PR #289, opened 2014-02-06
return to main menu from game menu by rclick

kyouryuukunn:
In main menu, a user can't return from game menu by r-click.
I allowed it.


nyaatrap:
Hi, will the new gui return to the main title by right click?

renpytom:
Now it will.

PR #498, opened 2014-11-20
Let choose to developer if strings need to be empty while generating translations

Neurone:
I created a specific option in the preference panel to handle this behavior and I subordinated it to the rot13 check.

I think it will be useful to change that check also and to create explicit request to the developer for rot13 option, maybe with just before the export and after the choose of the language name, so it will not be tied to the language name.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 saves
 tmp
 cache
+env
 
 log.txt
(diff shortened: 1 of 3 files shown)


renpytom:
Merged, but I'm going to change this so generating empty strings is the default.

PR #531, opened 2015-01-14
Add clipbord functions

kyouryuukunn:
def put_clipboard(string):
		    """
		    :doc: other

		    Put a string to clipboard and return True. If this fails, return None.
		    """

		def get_clipboard():
		    """
		    :doc: other

		    Get a string from clipboard and return it. If this fails, return None.
		    """


renpytom:
I'm not sure this adds much over using pygame.scrap directly. Is there a reason why we want this?

kyouryuukunn:
These simplify using clipboard features. This may be useful for people who don't know python and pygame well.

renpytom:
I don't think it simplifies it enough to be worth adding. You're turning one function call into one, less flexible, function call. I don't think it's worth doing unless we can improve over the pygame code - especially because I don't think these are in demand.

(Higher-level functions, like being able to paste text into an input, might be more useful. But I don't think they need this, as compared to scrap.)

kyouryuukunn:
OK.

PR #539, opened 2015-01-28
Update copyright notices and scripts for 2015

Nolanlemahn:
Just re-noticed that it was 2015; updated and ran the scripts.


renpytom:
Beat me to it. :smile:

PR #540, opened 2015-01-29
Allow rpy files to fully support indentation with spaces and tabs.

wjl:
This is 100% backwards compatible with all existing rpy files that use
spaces. Therefore, this does not require any change to anybody's
existing code. It only adds a new feature, allowing those who wish to
use tabs for indentation in rpy files to do so.

Previously, rpy files completely rejected indentation with tabs, and
forced all users to use spaces for indentation. Now, future rpy files
can be created using either tabs or spaces for indentation, at the
user's discretion.

To be sure that tabs and spaces are never mixed in a confusing manner,
this code forces all indentation to have the same prefix. Tabs are never
considered to be equivalent to any number of spaces. This matches what
we'd expect from files indented using best practices, with either spaces
only or tabs only, and eliminates any accidental "compatibility" between
the two in the same block.

### DIFF
--- a/renpy/parser.py
+++ b/renpy/parser.py
@@ -216,9 +216,6 @@ def list_logical_lines(filename, filedata=None):
 
             c = data[pos]
 
-            if c == '\t':
-                raise ParseError(filename, number, "Tab characters are not allowed in Ren'Py scripts.")
-
             if c == '\n':
                 number += 1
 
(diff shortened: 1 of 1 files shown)


renpytom:
I don't think it's a good idea to merge this request.

Experience from Python is that mixing tab and space indentation is a bad idea. Python 3 goes so far as to prevent the use of tabs and spaces for indentation in the same file; PEP8 says "Spaces are the preferred indentation method."

The problem with mixing tabs and spaces is there's no easy way, when looking at other people's code, to determine what was used. What's more, even when you determine that, you'd have to reconfigure your editor to match their preference. 

By limiting .rpy files to spaces only, we avoid these problems. I don't think there's a good reason to introduce them.

LuqueDaniel:
I agree with @renpytom

wjl:
This isn't about mixing tabs and spaces, but about supporting either one, as Python does.

We should make things easiy for our users, and prevent themselves from shooting themselves in the foot, but it is a shame if we would want not give users the freedom of using tabs for indentation, as this is an extremely popular method of indenting with many technical advantages.

But regardless of personal perferences (and spaces vs tabs flamewars), it is well agreed in almost all language communities that indenting with spaces is fine, and indenting with tabs is fine. What is universally reviled by everyone is indenting with some sort of weird mix of the two that confuses everybody.

I completely agree. In fact, this patch goes out of its way to prevent mixing in the same block, where Python 2 does a strange tabs-are-spaces modulo 8 computation, which is what screws a lot of people up.

Python 3 rejects tabs and spaces in the same indent. What if we did that?

Or, what if this patch went further, and allowed indenting with ONLY tabs or ONLY spaces in a single file. If any mixed indent was used, it would trigger a parse error with a message advising the user to either use spaces or tabs, but not both? This is a little heavy handed, but much better than the current "tab characters are not allowed, period". That would address any concern over people being confused over mixing tabs and spaces, while still letting those who wish to use ONLY TABS (never spaces) for indenting the freedom to do so, as they can in pretty much any other programming language (including Python, where this style is extremely popular and works very well, despite PEP-8 recommendations).

I would be happy to make this change if it would make it more agreeable to you folks. What do you think? =)

Either way, I will maintain this patch for my own development. Personally, I only ever use tabs for indentation and only use spaces for alignment. Along with an editor that lets you visualize the tab character, this technique is widely becoming a best practice in many projects and language communities.

renpytom:
I honestly think not having to deal with mixing spaces and tabs is an advantage. Even if we require a single indentation method per file, that advantage goes out the window when you copy and paste code from one file to another, or from a web page to a file.

I just don't see merging this as being worth the problems it will cause.

renpytom:
Actually, I'd allow this on one condition - it be made optional on a file-by-file basis. Like if there was a magic comment - something like #allow-unsupported-tabs# as the first line of the file, we'd allow it on a non-mixed basis.

PR #542, opened 2015-01-31
3D camera motion functions and Position Viewer

kyouryuukunn:
I added Ren'Py 3D camera functions and a gui viewer.
sample  https://www.youtube.com/watch?v=XYaVrwpcKkE

With some limitations, those functions allow Ren'Py to simulate 3D camera motion. This will easily
let you achieve a parallax effect by placing sprites and assets on a 3D field.
Those applies transforms to each layers which are registered as 3D layer by
positions of a camera and 3D layers on a 3D field to simulate 3D camera motion.

    def register_3d_layer(*layers):
        """
         :doc: camera

         Register layers as 3D layers. 3D layers is applied transforms to by
         positions of a camera and 3D layers. You can use this in game again
         and again.

         `layers`
              This should be the string or strings of a layer name.
         """

    def camera_reset():
        """
         :doc: camera

         Reset a camera and 3D layers positions.
         """

    def camera_move(x, y, z, rotate=0, duration=0, warper='linear', subpixel=True, loop=False):
        """
         :doc: camera

         Move the coordinate and rotate of a camera and apply transforms to all 3D layers.

         `x`
              the x coordinate of a camera
         `y`
              the y coordinate of a camera
         `z`
              the z coordinate of a camera
         `rotate`
              Defaul 0, the rotate of a camera
         `duration`
              Default 0, this is the second times taken to move a camera.
         `warper`
              Default 'linear', this should be string and the name of a warper
              registered with ATL.
         `subpixel`
              Default True, if True, causes things to be drawn on the screen
              using subpixel positioning
         `loop`
              Default False, if True, this motion repeats.
         """

    def layer_move(layer, z, duration=0, warper='linear', subpixel=True, loop=False):
        """
         :doc: camera

         Move the z coordinate of a layer and apply transform to the layer.

         `layer`
              the string of a layer name to be moved
         `z`
              the z coordinate of a layer
         `duration`
              Default 0, this is the second times taken to move a camera.
         `warper`
              Default 'linear', this should be the string of the name of a
              warper registered with ATL.
         `subpixel`
              Default True, if True, causes things to be drawn on the screen
              using subpixel positioning
         `loop`
              Default False, if True, this motion repeats.
         """

    def camera_moves(check_points, warper='linear', loop=False, subpixel=True):
        """
         :doc: camera

         Move a camera through check points and apply transforms to all 3D
         layers.

         `check_points`
              tuples of x, y, z, rotate, duration
         `loop`
              Default False, if True, this sequence of motions repeats.
         `subpixel`
              Default True, if True, causes things to be drawn on the screen
              using subpixel positioning
         """
    def layer_moves(layer, check_points, warper='linear', loop=False, subpixel=True):
        """
         :doc: camera

         Move a layer through check points and apply transform to the layer.

         `layer`
              the string of a layer name to be moved
         `check_points`
              tuples of z, duration
         `loop`
              Default False, if True, this sequence of motions repeats.
         `subpixel`
              Default True, if True, causes things to be drawn on the screen
              using subpixel positioning
         """
If you use this feature, in the first, set additional layers to be
registered as 3D layer.

For example, Write a code like below in options.rpy and add 'background', 'middle', 'forward'  layers.::

        config.layers = ['master', 'background', 'middle', 'forward', 'transient', 'screens', 'overlay']

Second, register layers which participate to a 3D motion as 3d layers by :func:`register_3d_layer`::

        init python:
            register_3d_layer('background', 'middle', 'forward')

Then, a camera and layers can move. ::

         label start:
             # reset a camera and layers positions
             $ camera_reset()
             scene bg onlayer background
             show A onlayer middle
             show B onlayer forward
             # It takes 0 second to move layers
             $ layer_move("background", 2000)
             $ layer_move("middle", 1500)
             $ layer_move("forward", 1000)
             with dissolve
             'It takes 1 second to move a camera to (1800, 0, 0)'
             $ camera_move(1800, 0, 0, 0, 1)
             'It takes 5 seconds to move a camera to (0, 0, 1600)'
             $ camera_move(0, 0, 1600, 0, 5)
             'A camera moves to (0, 0, 0) at the moment'
             $ camera_move(0, 0, 0)
             'It takes 1 second to rotate a camera to 180 degrees'
             $ camera_move(0, 0, 0, 180, 1)
             'It takes 1 second to rotate a camera to -180 degrees and 0.5 second to move a camera to (-1800, 0, 500)'
             $ camera_moves( ( (0, 0, 0, 0, 1), (-1800, 0, 500, 0, 1.5) ) )
             'a camera shuttles between (-1800, 0, 500) and (0, 0, 0)'
             $ camera_moves( ( (0, 0, 0, 0, .5), (-1800, 0, 500, 0, 1) ), loop=True)

When :var:`config.developer` is True, pressing position_viewer (by default,
"shift+P"), will open Position Viewer. This allow you to adjustment a camera
,3D layers and images positions by bars and manual inputs and see the result at
the moment.

Notice that 3D camera motion has some limits.:

* 3D camera motion applies transforms to 3D layers, so the show layer statement or
  :func:`renpy.show_layer_at` to 3D layers can't be usable.

* z coordinates of 3D layers don't affect the stacking order of 3d layers.

* A camera and a layer can't move at the same time.

### DIFF
--- a/renpy/common/00camera.rpy
+++ b/renpy/common/00camera.rpy
@@ -1,0 +1,231 @@
+# Copyright 2004-2014 Tom Rothamel <pytom@bishoujo.us>
+#
+# Permission is hereby granted, free of charge, to any person
+# obtaining a copy of this software and associated documentation files
+# (the "Software"), to deal in the Software without restriction,
+# including without limitation the rights to use, copy, modify, merge,
+# publish, distribute, sublicense, and/or sell copies of the Software,
+# and to permit persons to whom the Software is furnished to do so,
+# subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+# LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+# OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+# WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+init -1600 python:
+
+    ##########################################################################
+    # Camera functions
+
+    # value of _3d_layers isn't included in rollback.
+    # so I copy it many times.
+    _3d_layers = {}
+    _camera_x = 0
+    _camera_y = 0
+    _camera_z = 0
+    _camera_r = 0
+    _focal_length = 147.40
+    _layer_z = 1848.9
+
+
+    def register_3d_layer(*layers):
+        """
+         :doc: camera
+
+         Register layers as 3D layers. 3D layers is applied transforms to by a
+         camera motion. You can use this in game again and again.
+
+         `layers`
+              These should be the string or strings of a layer name.
+         """
+        global _3d_layers
+        af_3d_layers = {}
+        for layer in layers:
+            af_3d_layers[layer] = _layer_z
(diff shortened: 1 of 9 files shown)


renpytom:
I think this is too specific a technique to be something we want to support in core Ren'Py forever. I think real 3d support is more general, and I want to add it in sometime later this year.

(I think it's really nice - I just don't think it needs to or should be merged.)

kyouryuukunn:
OK. I send this code to CookBook.

PR #580, opened 2015-03-22
SpriteManager: update_args to allow universal update functions

caryoscelus:
Add update_args to SpriteManager and pass it as
kwargs to update function, so that it's possible
to have update function universal among multiple
SpriteManagers.

### DIFF
--- a/renpy/display/particle.py
+++ b/renpy/display/particle.py
@@ -127,15 +127,22 @@ class SpriteManager(renpy.display.core.Displayable):
     them at the fastest speed possible.
     """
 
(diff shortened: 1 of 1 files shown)


renpytom:
We don't actually need this. The renpy.curry function turns a function into a version that partially evaluates its arguments. I think partial evaluation is a better approach that trying to find every place that calls a function, and pass in arguments as well.

caryoscelus:
Yeah, that makes sense. The only problem is that curry function is never described in official docs (http://www.renpy.org/doc/html/)

PR #590, opened 2015-03-23
Remove old .rpyc files on "Force Recompile"

caryoscelus:
Otherwise, compiling may fail in case there some orphaned (without .rpy) .rpyc with same labels as in other scripts (usually happens on script rename).

The only problem with this i can see is that it may affect .rpyc files in renpy/common. Not sure if that is a real problem and whether or not those files are touched without this change.

### DIFF
--- a/renpy/main.py
+++ b/renpy/main.py
@@ -265,6 +265,18 @@ def main():
     renpy.style.build_styles() # @UndefinedVariable
     renpy.display.screen.prepare_screens()
 
+    # If recompiling everything, remove old .rpyc files.
+    # Otherwise, may fail in case there some orphaned (without .rpy) .rpyc
+    # with same labels as in other scripts (usually happens on script rename).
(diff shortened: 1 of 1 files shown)


renpytom:
This won't work, since we use .rpyc files to propagate statement names from the old version of the script to the new version of the script.

caryoscelus:
I don't quite understand what are those statement names, but since it won't work when .rpy file was renamed, it won't hurt if only orphan .rpyc files are removed. I'll fix that.

renpytom:
Do me a favor, and have it rename the files to .rpyc.bak? I really don't like deleting something that might contain user data. (And I've recovered games from .rpyc files before.)

caryoscelus:
done

renpytom:
This looks good. I'll merge it into 6.99.1, as I think 6.99.0 is done with the exception of some very minor distribution work.

PR #608, opened 2015-04-11
Update 00start.rpy comments

Nolanlemahn:
renpy.game.call_memory is deprecated/no longer in source.

### DIFF
--- a/renpy/common/00start.rpy
+++ b/renpy/common/00start.rpy
@@ -110,8 +110,8 @@ label _start_store:
     return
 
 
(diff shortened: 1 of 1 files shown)


Nolanlemahn:
Oh - "_start_memory" should be "_start_replay" on ( https://github.com/renpy/renpy/blob/master/renpy/common/00start.rpy#L99 ) as well too :P ... Just being a little nitpicky with the comments.

renpytom:
'tis a fair cop.

PR #616, opened 2015-04-21
Copy Android distributions to distributions directory.

d3nzil:
It would be more user friendly to copy Android distribution files to the same directory, where other distributions are.

This is my first attempt to do so. I'm still finding my way around the Launcher, so let me know if I should change something or if you have some questions.

### DIFF
--- a/launcher/game/android.rpy
+++ b/launcher/game/android.rpy
@@ -53,6 +53,7 @@ init python:
     import re
     import os
     import json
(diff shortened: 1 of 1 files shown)


renpytom:
Thanks. I'll change this a bit so that it uses information from rapt, rather than trying to scan the directories itself. But I like the idea.

PR #635, opened 2015-05-13
Multi character 2

vbkaisetsu:
Hi,

// Long time no see!

I created a new branch for MultiCharacter #144 with small changes.
Basically, the interface is not changed.

I think renpy 7.0 will be released soon. Will it contain statement changes around Character?
If not, I think the current implementation is good for parallel speaking.


renpytom:
This has been superseded by the multiple character support that will be in 6.99.14.

PR #652, opened 2015-06-02
Use dill for saving

caryoscelus:
This allows to save things which are impossible with regular pickle/cPickle, including cases which are pickled fine in py3, but can't be pickled by py2 pickle. [Dill](https://pypi.python.org/pypi/dill) seems to have no effect with cPickle, but as long as there's long-term plan to migrate to py3 this shouldn't be a problem.

I don't know what is correct way to add dependency (i just installed dill using 'pip install dill' in virtual env, but perhaps there's a way to automatize this?), but if you're going to accept this, i can either put that into README or do proper dependency if you tell me where to look.

### DIFF
--- a/renpy/loadsave.py
+++ b/renpy/loadsave.py
@@ -23,6 +23,7 @@
 
 import pickle
 import cPickle
+import dill
 
 from cStringIO import StringIO
 


renpytom:
I guess my question is - what does this add? 

Offhand, it looks like dill is focused on pickling various code types. I'm not sure we want to do this - my feeling is we generally want to refer to code types by name, so if the game changes, the instances are automatically upgraded. (That is, I think it's generally a misfeature to send code from version 1 of a game into version 2 of a game via a pickle.)

caryoscelus:
Yeah, that is right in most cases and i'm not trying to pickle code directly. The problem without dill was that py2 pickle have problems with callable objects, even such that only wrap normal function call (the reason for such objects was the problem with py2 instance method issues in the first place).

So for my case dill is a bad solution, but still better than nothing. If there's easy way to properly distribute third-party packages along with the game, perhaps there's not much sense in having it by default and i'll just use that.

renpytom:
Can you give an example of something you're having trouble with? There's likely a simple fix that works better. (For example, having a decorator return an object with an __call__ method rather than an inner function.)

caryoscelus:
I'm exactly using such decorator, which in fact does nothing except for unbinding method from its class.

    class unbound(object):
        def __init__(self, f):
            self.f = f
            functools.update_wrapper(self, f)
        def __call__(self, *args, **kwargs):
            return self.f(*args, **kwargs)

And then i'm getting following error:

    PicklingError: Can't pickle <function _init at 0x7f77ae244aa0>: it's not found as entity._init

(_init is just an @ unbound decorated function)

PR #658, opened 2015-06-12
Changes audio.py to work around an indefinite lockup that occurs if you try to quit while PulseAudio is locked.

SpikyCaterpillar:
To replicate the lockup:

startx as two different users.
Do something that locks the sound card as user1 (say, playing a pile of mp3s
through mpg321)
As the other user, start and then quit a Ren'Py game.

Result: strace will show the Ren'Py game sitting there waiting indefinitely
for a futex until such time as mpg321 releases the sound card.

### DIFF
--- a/renpy/audio/audio.py
+++ b/renpy/audio/audio.py
@@ -41,7 +41,18 @@ if 'pss' not in disable:
     try:
         import pysdlsound as pss
         pss.check_version(4)  # @UndefinedVariable
-        atexit.register(pss.quit)  # @UndefinedVariable
+        import platform
+        if platform.system()=='Linux':
+            def audio_atexit():
+                import signal
+                handler = signal.signal(signal.SIGALRM,signal.SIG_DFL)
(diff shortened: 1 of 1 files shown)


renpytom:
This strikes me as a bug in pulseaudio (or perhaps an ALSA misconfiguration) rather than a bug in Ren'Py. 

That being said, having atexit call pss.quit isn't really all that meaningful, as the sound will shut down in a small fraction of a second once the process quits. So I'm wondering if you remove line 45 of audio.py:

```
        atexit.register(pss.quit)  # @UndefinedVariable
```

entirely, does that adequately solve the problem. I think something like this is preferable to adding platform-specific bug work-around code (that I'm not sure I can ever remove) to Ren'Py.

SpikyCaterpillar:
Chopping out the atexit.register(pss.quit) line seems to work fine on the offending machine so far.  Haven't tested to see if it introduces problems anywhere else.

PR #672, opened 2015-06-27
Fix Nightly's translation

bbs3223474:
It should be "每夜版", not "暗黑版".
Old translation means "darkness version", not "nightly".


renpytom:
"Darkness version" sounds cooler. But I guess a more accurate translation is better. Merging, and thanks.

PR #692, opened 2015-07-22
fix (py_)eval: allow evaling code objects

caryoscelus:
The reason for this is that otherwise eval function available in
.rpy files is not fully compatible with regular python eval.


renpytom:
Merging, thanks.

We probably want to replace compile in minstore.py with something based on py_compile, to ensure that any code created gets the transformations required to create Revertable{Object,List,Dict,Set}.

PR #702, opened 2015-07-28
Typofix on docstring

Rikxz:
"The programmatic equivalent of the show screen statement." became "The programmatic equivalent of the call screen statement." for def call_screen


renpytom:
Merged. Thanks.

PR #717, opened 2015-08-14
Fixed common.rpy

RangHo:
Sorry that I messed up with common.rpy translation...
Here's a fix for that.
~~I hope it's not late yet~~


renpytom:
it will go in the next release.

PR #736, opened 2015-09-13
Update Spanish translation of the template

javimat:


templates/spanish/game/tl/None/common.rpy
```diff
@@ -512,3 +512,26 @@
...
+    old "Copies the traceback.txt file to the clipboard."
+    new "Copia el archivo traceback.txt al portapapeles."
+
+    # renpy/common/_errorhandling.rpym:613
+    old "Copies the errors.txt file to the clipboard."
+    new "Copia el archivo errors.txt al portapapeles."
+
+    # renpy/common/00gltest.rpy:91
+    old "Gamepad"
+    new "Gamepad"
```
LuqueDaniel: Maybe can be replaced for "Mando". It's only an idea :)

javimat: Thanks, Daniel, I did have a doubt here. I don't use that and don't know how it's called usually in Spanish. I'll put "mando".

PR #741, opened 2015-09-18
Updated the Finnish translation

Diapolo10:
I have updated the Finnish translation to the most recent version of
Ren'Py. Or, at least I sincerely hope so... things didn't go well last time.


javimat:
Please, remove the word TODO in comment lines.

renpytom:
I have a script that takes care of removing TODO lines automatically, so there's no need to do it by hand. It's a pretty common issue.

PR #764, opened 2015-10-19
Custom text replacement hook

emmeken:
This allows a custom function to for example replace '--' with em-dash and '...' with an actual ellipsis character. While it is possible to instead put those Unicode characters directly in the .rpy file, it is easier to type them as ASCII sequences.

### DIFF
--- a/renpy/config.py
+++ b/renpy/config.py
@@ -563,6 +563,9 @@ log_to_stdout = False
 # new-style custom text tags.
 custom_text_tags = { }
 
+# A function that given the text from a TEXT token, returns a replacement text.
+replace_text = None
+
 # A function that is called when a label is missing.
 missing_label_callback = None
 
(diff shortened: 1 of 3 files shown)


renpytom:
I like this.

Could you rename the config variable to replace_text, please? I think the "custom" is a bit much, as nearly everything is custom in Ren'Py.

It would also be nice if you could add documentation to sphinx/source/config.rst. This should probably go into the "occasionally used" section. (If you're not comfortable with sphinx, let me know and I'll take care of this.)

Once you do that,  I'll merge this.

emmeken:
I renamed the variable to `replace_text` and added documentation, as requested. I tried to write the documentation in such a way that it's both exact and accessible at the same time, but I'm not sure I succeeded, so feel free to adapt the text if necessary.

I also added a conversion to `unicode` for the returned value, since that is the required return type and it seemed more user friendly to perform the conversion if needed than to document this requirement.

renpytom:
Looks good. Merged.

PR #768, opened 2015-10-23
Add "arrowkeys" child to Viewport

Nolanlemahn:
I figure if "mousewheel" is a child, it might be nice to be able to have "arrowkeys" as a child... but I'm not entirely certain as to the correctness of the script changes.

It works on my end, but it may not be stylistically correct.


renpytom:
I like the idea of this, but not how it sort of conflicts with the keys Ren'Py uses for keyboard focus. Before I merge this, I need to check to see that it's usable in practice.

Nolanlemahn:
On the test case (a level editor for another project) I have set up, the arrow keys only affects a viewport if the mouse cursor is inside said viewport; arrow keys otherwise change the focused UI element (as I think is expected).

If there's any checking I can help with (or any changes I should make to my commit, or if you'd like to see the test case, or...), please let me know.

renpytom:
Sorry this took forever, but I've finally merged it into Ren'Py. Thank you!

PR #772, opened 2015-10-25
SW rendering bugfix and minor changes

SpikyCaterpillar:
This fixes a bug where exposure events didn't properly redraw the window on X if a player's using software rendering.  (Observed on two separate Debian machines, one with a Radeon and an OLPC with a Geode; neither have proprietary drivers installed.)

To reproduce the bug:
RENPY_RENDER=sw ./renpy.sh
Move another window partially over the launcher
move the window back off

The contents of the other window will be stuck atop the Ren'Py window until Ren'Py actually redraws that section of the screen.

This pull req also adds support for home and end in text input fields, and widens the cursor to two pixels to work around a bug where one-pixel-wide input field cursors don't show up on software rendering.  (This was observed on both Debian and El Capitan)


renpytom:
I merged this, but backed out the change in the width of the caret. 

I wasn't able to reproduce the missing caret on Ubuntu 15, but even if so, it seems likely that the bug is in the software renderer - and hence it doesn't make sense to change Input to work around it. So I backed out that change, and also fixed a typo.

I don't want to make changes to Ren'Py just to support the obsolete software rendering path, which will be going away in the next few versions of Ren'Py. I don't know how new the Radeon is, but the OLPC is a 2007-era design, and was IIRC low end even when it was first released. I think it's fair to consider it obsolete a decade later.

PR #800, opened 2015-12-09
Add a PushMove transition.

BlackDragonHunt:
This adds a new transition type, which causes the new scene to "push" the old scene off the screen in one of four directions. It's basically just a special case of CropMove--a combination of a slide in and a slide out applied simultaneously. I couldn't think of a nice way to do it with the tools available, so I made it its own transition, which I thought other people might find useful.

As a side note, because this is code I've written for a MangaGamer project, I'd appreciate it if you could credit them alongside me wherever you credit contributors, should you choose to accept this.


renpytom:
Thanks for both PRs. Both have been merged, and I'll credit you and MangaGamer.

PR #803, opened 2015-12-09
Configuration variables for audio volumes

vollschauer:
I would like to have something like this, not sure if the "set_volume" code syntax is correct ;)

### DIFF
--- a/renpy/common/00defaults.rpy
+++ b/renpy/common/00defaults.rpy
@@ -53,6 +53,14 @@ init -1500 python:
     # If not None, the default value of emphasize_audio.
     config.default_emphasize_audio = False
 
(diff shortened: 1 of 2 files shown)


renpytom:
Please add documentation to sphinx/source/config.rst, and then I'll merge this.

vollschauer:
I have added it at the end of the "Preference Defaults" section. Better proof read it ;) I didn't know what documentation is before...

renpytom:
I've merged this. I also renamed the variables to config.default_music_volume (etc), since I don't like abbreviations in the public API.

vollschauer:
Nice! Thank you ;)

PR #819, opened 2016-01-06
Fixed Drag.snap() animation to use delay

dvarin:
Passing a delay argument to Drag.snap() doesn't work, with the snap being instantaneous.  The cause of this seems to be the condition for deciding the snap is over: `if at >= (self.target_at - self.at)` , which is going to be true almost all the time.  So I changed it to what it looked like it should have been, `at >= self.target_at` .

That exposed another problem, which is that snap() doesn't account for the last render time stored in `self.at` being potentially out of date.  E.g., if the drag is moved and then held still without being dropped for 3 seconds, then dropped, any snap called in the drop handlers will see `self.at` as three seconds prior because there have been no calls to render() since the drag stopped moving.
So I added another property, `target_at_delay`, that tells the next render (which has an up-to-date `at` passed to it) to calculate the correct `target_at` and start the animation.  snap() already calls redraw(self,0) so that should be immediate.

### DIFF
--- a/renpy/display/dragdrop.py
+++ b/renpy/display/dragdrop.py
@@ -241,8 +241,12 @@ class Drag(renpy.display.core.Displayable, renpy.python.RevertableObject):
         self.at = 0
 
         # The (animation timebase) time at which we should reach
-        # the target coordinates.
+        # the target coordinates for the currently executing snap animation.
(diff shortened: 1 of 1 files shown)


renpytom:
Merged, thank you.

PR #820, opened 2016-01-17
Confirm EndReplay()

vollschauer:
An option to confirm the EndReplay(), I edit the "sphinx/source/inc/replay" cause I didn't know how
it get automatically generated.


renpytom:
This is going in soon, but I'm too tired tonging. 

Memo to self: Use this in the new gui nav when _in_replay is True.

renpytom:
Mereged, thank you. In 48690da I changed the wording and the layout variable name, so you may need to update to match that.

PR #830, opened 2016-01-26
Fix FTBFS on Big-endian architectures.

apoleon:
Ren'py fails to build from source on Big-endian architectures due to a missing semicolon in module/IMG_savepng.c.


renpytom:
Are there any big endian architectures anymore?

apoleon:
On Tue, 26. Jan 17:18 Tom Rothamel <notifications@github.com> wrote:
>    Are there any big endian architectures anymore?

oh yes there are still quite a few of them. It's another question how
many people have installed Ren'Py on them but it is possible. Here is a
list of all Debian architectures where it can be built successfully.

https://buildd.debian.org/status/package.php?p=renpy&suite=unstable

PR #867, opened 2016-03-01
Use Markdown to improve the GitHub issues

williamd1k0:
I'm suggesting the use of *templates* to improve the creation of issues.

`ISSUE_TEMPLATE.MD` redirects to `CONTRIBUTING.MD` that contains a basic guide about syntax highlighting with Markdown.

I don't revised English yet. First I would like to know what you think about this suggestion.

You can see how this works by my fork issue tracker.
https://github.com/williamd1k0/renpy/issues/new

### DIFF
--- a/.github/CONTRIBUTING.MD
+++ b/.github/CONTRIBUTING.MD
@@ -1,0 +1,69 @@
+# Markdown Issue Guide
+Basic guide to ease issues solution by using *GitHub Flavored Markdown*.
+
+## What is Markdown?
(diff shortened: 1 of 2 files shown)


renpytom:
I'm not sure we need this, at the moment. Generally, when people report issues on github, they're fairly complete. And they're also easier to work with - so I'm not sure I want to do something that would make reporting issues on Github more difficult, and that hence would lead to more issues being reported out of process.

williamd1k0:
Oh ok, I see. It was basically just an idea.

renpytom:
Ideas are great, so keep them coming. I just don't think it's time for this yet.

PR #872, opened 2016-03-10
Document the change to the default image init priority

jackmcbarn:
When init offset support was added in commit 679f9e31, the default image init
priority was changed from 500 to 990. This changes it back.

### DIFF
--- a/renpy/parser.py
+++ b/renpy/parser.py
@@ -1801,7 +1801,7 @@ def image_statement(l, loc):
     rv = ast.Image(loc, name, expr, atl)
 
     if not l.init:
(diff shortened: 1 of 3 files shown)


renpytom:
I wanted to lower it, so that an offset of 100 doesn't wind up creating numbers that are out of the user range. Did this cause a problem?

jackmcbarn:
I assumed this was unintentional since it wasn't mentioned in the commit message or in the changed documentation, and since it could theoretically be a breaking change (though probably not in most games), it would be wise to change it back. Since it was intentional, it should probably at least be documented.

PR #912, opened 2016-04-17
Updated Finnish translation of Ren'Py

Diapolo10:
The translation has now been updated to reflect the changes made between the previous translation and Ren'Py version 6.99.10.1227. (Unfortunately I don't use development versions often.)

### DIFF
--- a/launcher/game/tl/finnish/about.rpy
+++ b/launcher/game/tl/finnish/about.rpy
@@ -13,3 +13,17 @@ translate finnish strings:
     old "Back"
     new "Takaisin"
 
+translate Finnish strings:
+
+    # game/about.rpy:39
+    old "[version!q]"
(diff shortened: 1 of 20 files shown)


renpytom:
That works, but please sync up to master before translating - merging was a pain. Also, the translation name should be finnish (all lower-case), to match the existing stuff and the other languages.

Diapolo10:
Apologies. But before I sent the pull request, I checked and it was synced. Or... at least I understood it was.

I will be more careful from now on, promise.

renpytom:
Basically, I made a lot of changes (well, really, the big change was that I changed "Finnish" to "finnish". 

So when you update the translations in the future, start with a fresh clone of Ren'Py rather than this branch.

PR #915, opened 2016-04-19
use config.name & config.version for build info

caryoscelus:


renpytom:
This is too aggressive. config.name can be human readable, which won't work as the name of a directory. 

"PROJECTNAME-" + config.version

might be a worthful change.

caryoscelus:
I always assumed `config.name` should be filename safe. But if it's not, i see two other options:

* add `config.sys_name` (or something like that) which should be safe and add `config.name = config.sys_name` line to templates. Build info then should look for `config.sys_name` but fall back to something like UNNAMED_PROJECT in case it's not found so that users would rename `build.*` variables themselves.

* wrap `config.name` in conversion function, though this will not work well for non-latin names.

renpytom:
Or just do what we have now. I don't see a pressing reason to change this, except to add config.version automatically.

caryoscelus:
Ok, it's not pressing, just a bit annoying.

Btw, is there a reason why build info is added separately and isn't part of template?

caryoscelus:
Changed it to have directory name depend on executable name and version by default.

renpytom:
I did my own implementation of something like this as 49820ee . The idea is that there's a build.name that is then used to set build.directory_name and build.executable name.

PR #916, opened 2016-04-19
add local/nested labels support

caryoscelus:
Originally proposed here: https://lemmasoft.renai.us/forums/viewtopic.php?f=32&t=38122

> When the script is heavy-choice, amount of labels can soon get unmanageable. I think the usual way to avoid name collisions is to use something like "chapter1_episode1_choice1", but it is ugly, verbose and still doesn't guarantee avoidance of collisions, especially when using third-party code.
> 
> To resolve this problem i propose introducing 'local' labels, which can have identical names inside different 'global' labels. Here's how it works:
> 
> ```renpy
> label global_label:             # normal label, nothing changed for its behaviour
>   "..."
> label .a:                       # local label fully identified as global_label.a
>   "..."
>   jump .a                       # local jump, resolved to global_label.a
> 
> label another_global_label:
>   "..."
> label .a:                       # another_global_label.a
>   jump global_label.a
> ```
> 
> Syntax-wise, it's fully backwards compatible since currently label names cannot contain dot and thus all existing labels are considered global with this change.
> 
> It is also possible to expand this idea to have no nesting limit (i.e. making possible to have label a.b.c.d and jump to it inside a.b.c scope by simply "jump ...d")
> 
> I also found out that it is currently possible to 'nest' labels using indentation, but it doesn't protect from name collisions. It is possible to implement the same mechanism with such nested labels, but since it uses old syntax, it can cause problems with existing code. Also, it is more visual which can sometimes be good, but can also be bad when you prefer all statements to be on same indentation level.

While waiting for comments, i finished implementing it anyway and so making this pr for more specific discussion. If this is going to be accepted, i'll mention this in docs as well.

### DIFF
--- a/renpy/ast.py
+++ b/renpy/ast.py
@@ -1904,7 +1904,7 @@ class Screen(Node):
 class Translate(Node):
     """
     A translation block, produced either by explicit translation statements
-    or implicit translation blocs.
+    or implicit translation blocks.
 
     If language is None, when executed this transfers control to the translate
     statement in the current language, if any, and otherwise runs the block.
--- a/renpy/parser.py
+++ b/renpy/parser.py
@@ -491,7 +491,7 @@ class Lexer(object):
     sub-lexers to lex sub-blocks.
     """
 
-    def __init__(self, block, init=False, init_offset=0):
+    def __init__(self, block, init=False, init_offset=0, global_label=None):
 
         # Are we underneath an init block?
         self.init = init
@@ -509,6 +509,7 @@ class Lexer(object):
         self.text = ""
         self.number = 0
         self.subblock = [ ]
+        self.global_label = global_label
         self.pos = 0
         self.word_cache_pos = -1
         self.word_cache_newpos = -1
@@ -653,7 +654,7 @@ class Lexer(object):
 
         init = self.init or init
 
-        return Lexer(self.subblock, init=init, init_offset=self.init_offset)
+        return Lexer(self.subblock, init=init, init_offset=self.init_offset, global_label=self.global_label)
 
     def string(self):
         """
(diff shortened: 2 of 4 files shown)


renpytom:
I like this - I think it solves a problem and should be relatively easy to explain. 

I think my one concern is that local labels could cause problems with the translation system, which is probably not expecting a . in the label syntax. This can likely be solved by making all local labels hidden labels. 

Really, my main concerns with this are that I was too sick to answer you on the forum, and that I didn't come up with this myself. It looks good, and after the local labels are hidden by default, I think it's ready to merge once it it documented.

caryoscelus:
Hmm, translation does indeed fail with this. I think the easiest way is simply to fix `Lexer.hash` (doesn't seem like a proper name for translation id parsing, btw) to accept `.`. That seem to work. Or if that looks bad, another option would be to replace `.` for `_` in translation strings.

renpytom:
I think I like the idea of making local labels hidden. They feel like they should be as anonymous as possible - invisible to code outside the named label, at least for purposes like translations.

caryoscelus:
I think that depends on usage pattern, i.e. balance between global and local labels.

Just fixed ability to declare labels in form of `global.local` with `global` mismatching current scope. I wonder if `global.local` form should be forbidden altogether for declarations. The only time when it makes sense to use them is if you've got a lot of code between global label and local one and want to remind the scope.

Also, do you think it would make sense to allow more nesting levels?

renpytom:
I think one level is fine, and anything more would be excessive.

Shizmob:
I think leaving local labels visible might be nice, also to be able to create a sense of hierarchy.

renpytom:
Okay, then let's have the dot replaced with _ when generating translates.

caryoscelus:
Looks like that would require a bit more changes than i suspected.

caryoscelus:
For some reason, adding `identifier = identifier.replace('.', '_')` into `ScriptTranslator.lookup_translate` seems to work for one or two statements, but then fails with `Exception: Translation nodes cannot be run directly.`

caryoscelus:
Translation seems to be fixed, some documentation added.

vollschauer:
I'm really looking forward to this ....could sure use it right now ;)

caryoscelus:
Not sure why it wasn't merged yet, i think it was ready. Now there are some conflicts though. Will check them out soonish.

vollschauer:
Just curious, if the local labels will work also if there are indented inside in a "if else" statement?

caryoscelus:
Not sure what you mean. Can you give an example?

vollschauer:
This example is maybe silly but you get an idea what I mean:
`label test:
    if something == 0:
        if something1 == 0:
            if something2 == 0:
                if something3 == 0:
                $ something = "something"
                jump .local

        .local 

        if something4 == 0:
           if something5 == 0:
               $ something1 = "something1"`

caryoscelus:
Not sure if it's a good idea, but something like that (only syntax is a bit different) already works with normal labels, so should work with local as well.

```renpy
label test:
    if True:
        "True"
        jump false_as_well
    else:
        label false_as_well:
            pass                # used to avoid indentation increasing
        "False"
```

renpytom:
Merged this in. Thank you.

PR #922, opened 2016-04-22
Use proper API for determining if screen is showing in Show().

Shizmob:
`renpy.showing()` doesn't work as the `screens` layer is not provided, and providing it gives other weird behaviour. Use `renpy.get_screen()` instead, as this works properly.


renpytom:
Merged, thanks.

PR #944, opened 2016-05-16
Improve the buildsystem (setup.py)

susnux:
* Allow libav* libsw* headers in ffmpeg subdirectory
  (done by some distribution like openSUSE)
* Also search in lib64 and lib32 for libraries


renpytom:
Looks good. Thank you.

PR #947, opened 2016-05-22
change link to old wiki

xavi-mat:
The Quickstart had a link to the download page of the old wiki.
(Please check, I'm not sure if this is the correct place to change this.)


renpytom:
Merged, thanks.

PR #952, opened 2016-06-07
Add {alpha} to text_tags dict

kevinturner:
Prevents {alpha} from being incorrectly identified as an invalid text tag during Lint scans.


renpytom:
Merged, thank you.

PR #983, opened 2016-07-26
gallery: Just one option to advance trough unlocked images by left-cl…

vollschauer:
Just one option to control how to advance trough unlocked images, 
by left-clicking or using the navigation screen buttons...
This is my idea how it should work, I know this is not perfect but gives me the functionality I'm looking for.


renpytom:
Thanks, this probably works best. Merged.

PR #999, opened 2016-07-30
adding  tutorial\game\tl\indonesian\gui.rpy

pratomoastan:
adding  tutorial\game\tl\indonesian\gui.rpy


pratomoastan:
Somethings is wrong with the translation file i think i miss the "" somewhere 
because when i create new project using indonesian language its returning no attributes error
ill try to translate from the scratch

PR #1013, opened 2016-08-21
New preference for forcing fallback images on video sprites

kevinturner:
Video was too taxing for some of our systems and we wanted to be able to fall back to the already defined backup images for movie sprites without deleting the original file paths.


pratomoastan:
hmm this is good

Issue #681, opened 2015-07-11
Music room improvements


renpytom:
Additionally, the imagebutton state for the Play button is selected only while the first song is playing, and changes to idle when the song changes. The imagebutton for the Stop button does not change to selected when the song stops. Is it possible to change these so that the Play button is selected when any song is playing and the Stop button is selected when no music is playing?

http://lemmasoft.renai.us/forums/viewtopic.php?f=8&t=33060

PR #1033, opened 2016-09-13
Buttons tied to MusicRoom's Stop action now stay selected when clicked.

jsfehler:
Related ticket: https://github.com/renpy/renpy/issues/681


renpy/common/00musicroom.rpy
```diff
@@ -35,7 +35,6 @@ init -1500 python:
         identity_fields = [ "mr" ]
         equality_fields = [ "filename" ]
 
-
```
jsfehler: 2 blank lines was inconsistent with classes below. Can revert this part if it was by design.

PR #1037, opened 2016-09-19
Use built-in set instead of deprecated import

jsfehler:
In warp.py and python.py, sets.Set was being used. This is deprecated as of python 2.6. I replaced the usage with the built-in set and set comprehension.

There were 3 outright removals in RevertableSet(): The wrappers for **copy**, **deepcopy**, and union_update(), but I don't believe anything in ren'py hinges on that particular implementation. I'm mildly concerned about removing the copies, but none of the other Revertables use them.

The number of changes exploded a bit due to my IDE's linting. I only applied style changes that seemed inoffensive, but any can be reverted if they break overall style.

Tested the following scenarios:

1) On the command line, entered: renpy.exe "my_project" --warp "rpy_script":"line_number"
Result: Game opens at that line, effects occurring earlier are not applied, "return" statement at end of label returns player to the specified line number.

2) Created a set using both x = set([1, 2, 3, 4, 5, 1, 1]) and x = {1, 2, 3, 4, 5, 1, 1}, then verified each contained no duplicates. Verified that both create RevertableSet, not set.

### DIFF
--- a/renpy/python.py
+++ b/renpy/python.py
@@ -30,7 +30,6 @@ import marshal
 import random
 import weakref
 import re
-import sets
 import sys
 import time
 
@@ -42,11 +41,13 @@ import renpy.audio
 # Deleted is a singleton object that's used to represent an object that has
 # been deleted from the store.
 
+
 class StoreDeleted(object):
     def __reduce__(self):
         return "deleted"
 
 deleted = StoreDeleted()
+
 
 class StoreModule(object):
     """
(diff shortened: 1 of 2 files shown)


renpytom:
Have you tested loading a set saved with sets.Set with your new implementation?

jsfehler:
I didn't think of that. It breaks the save game, unfortunately.

cPickle: UnpicklingError: state is not a dictionary
pickle: AttributeError: 'tuple' object has no attribute 'iteritems'

In regular python, pickle doesn't seem to care about the difference, but it seems like the fact that both the objects are RevertableSet, but inherit from different objects is causing it to freak out.

renpytom:
Thanks for testing. I think I'll see if I can do something with __setstate__ before giving up - I'd really like to not break saves over this.

renpytom:
df88bdfd merges this in. It was rewritten to deal with some spacing issues, but based on your version, especially the special messages you gave. Thank you.

PR #1039, opened 2016-09-22
Replace print statement with print function

jsfehler:
Used py2to3 to convert everything.
Skipped the scripts and sphinx directories, wasn't sure if changing them was required for ren'py.

In launcher/game there were some files where it wasn't immediately clear if any of the code in there was from a 3rd party. I had to modify them to get renpy.exe to run.

Took the opportunity to run autopep8 with the following rules on the effected files:

E26  - Fix spacing after comment hash for inline comments.
E301 - Add missing blank line.
E302 - Add missing 2 blank lines.
E303 - Remove extra blank lines.

A very quick test had renpy running with all the changes, but it certainly needs to be heavily tested. If there's some test cases for features that use print, that would help.

### DIFF
--- a/renpy/lint.py
+++ b/renpy/lint.py
@@ -19,6 +19,7 @@
 # OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 # WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
+from __future__ import print_function
 import renpy.display
 import renpy.text
 import codecs
@@ -51,6 +52,8 @@ image_prefixes = None
 report_node = None
 
 # Reports a message to the user.
+
+
 def report(msg, *args):
     if report_node:
         out = u"%s:%d " % (renpy.parser.unicode_filename(report_node.filename), report_node.linenumber)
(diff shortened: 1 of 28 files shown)


renpy/lint.py
```diff
@@ -90,17 +95,19 @@ def try_eval(where, expr, additional=None):
...
 # Returns True of the expression can be compiled as python, False
 # otherwise.
+
+
 def try_compile(where, expr, additional=None):
 
     try:
         renpy.python.py_compile_eval_bytecode(expr)
-    except:
+    except SyntaxError:
```
jsfehler: Added this because my right eye twitches when try/except blocks don't specify what they're catching. However, I realized I'm here to fix print statements, not tweak exception handling. ...Then I forget to undo this before moving on.

I can either revert it, or if this function should only throw an exception on a SyntaxError, I guess we can keep it.

renpytom:
I'd really prefer it if this merge didn't include the PEP8 or exception handling changes - both seem out of scope for a focused pull request, and I think I'd rather come up with PEP 8 enforcement that I prefer, and apply it all at once. So if you can back this out, I'd merge it. 

When it comes to try_compile, I think we want to catch all errors, rather than stopping if we encounter an unexpected error.

renpytom:
On second thought - the pep8 stuff is a good idea. Good enough that I didn't want to wait to start using it. So I've merged this, then backed out the SyntaxError change.

Thanks!

jsfehler:
I don't disagree with not wanting style fixes in an otherwise unrelated set of changes. In hindsight I would have asked if you wanted style fixes included or if you wanted to apply a set of style changes to the whole codebase as a single commit.

In regards to the try/excepts in lint.py, personally I'd prefer to catch specific exception. If someone managed to trigger anything but a SyntaxError, it would be very unexpected and worth exploring. Stlll, now's not the time, and it's not a very pressing issue.

PR #1046, opened 2016-10-04
Fix cut off sentence in changelog

merumelu:
In the changelog there was a sentence that ended abruptly:
> Ren'Py can be made to return to the old behavior (in which only translate language style, translate language python, and callbacks are executed) by setti

I changed it back to how it was before 4b06e96d28e30234251271cc405573981c684fd9, where it got cut off for some reason.


renpytom:
Merged. Thank you.

PR #1060, opened 2016-10-19
Fix a few typos in docstrings

jsfehler:
+ PEP8 style fixes while I'm in there.
+ Replaced instances of "if not x in y" with "if x not in y"
+ Replaced multiple "or" check with a "in list" check.
(set or tuple check could theoretically be faster, but for such a tiny collection, I doubt there's a huge difference.)


renpytom:
Merged. I then changed the "in list" check with an "in string" check - I think it'll be faster, since it avoids allocating memory.

PR #1078, opened 2016-11-17
Draft of simplified Chinese translation

TakanashiTorimoinai:
Begin to translate the quickstart to simplified Chinese.


renpytom:
Thank you for translating this. 

I'm trying to think what I want to do with this. I'm leaning towards that we should probably make the Simplified Chinese reference manual into its own repo in the Ren'Py project, and then I'll have either a nightly job that builds it, or we can use readthedocs or something like that. 

My thinking behind this is that the English documentation is normative - it says what Ren'Py "should be", while translations are not. Translations should probably also go online immediately, rather than waiting for a full Ren'Py release.

renpytom:
Closing, as this was last updated in 2016.

PR #1083, opened 2016-12-03
Add Japanese translation for new gui

nyaatrap:
It also revised previous translation.


renpytom:
Merged, thank you. Does translating the date work?

nyaatrap:
No. If I use only one word ("{#weekday}%A") then it's translated correctly. But combination of words like ("{#weekday}%A {#month}%B") or "{#file_time}%A, %B %d %Y, %H:%M" will not be translated.

renpytom:
How are you using the format string? You really shouldn't be using the {#codes} directly - the FileTime function should be smart enough to pass the %a, %A, %b, and %B codes to the translation functions before displaying them to the player.

PR #1087, opened 2016-12-06
Small refactoring in ADVCharacter.__init__

caryoscelus:
I reduced some repeating. If you like this change, i can rewrite some more in the same fashion.

### DIFF
--- a/renpy/character.py
+++ b/renpy/character.py
@@ -570,7 +570,6 @@ class ADVCharacter(object):
 
     voice_tag = None
 
-    # When adding a new argument here, remember to add it to copy below.
     def __init__(
             self,
             name=NotSet,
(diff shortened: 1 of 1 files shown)


renpytom:
I'm not really a fan of this. It's more dynamic and less directly readable (with the fields being assigned inside a function call), and I'm not really sure there's any advantage given to making it so.

caryoscelus:
Readability is a matter of taste in this case - for me it's less readable with all the clutter of repeated actions and repeated names.

It's easier to add new properties and even more so to change method of initialization that way. It also makes possible to get rid of `v` function since it's called only once.

renpytom:
It's fair to call it a matter of taste. I find initializing a field directly to be more tasteful than calling setattr to do it.

PR #1100, opened 2016-12-24
Python3 port WIP

caryoscelus:
This is work in progress.

I try to keep compatibility with py2 as well for now.

### DIFF
--- a/launcher/game/EasyDialogsWin.py
+++ b/launcher/game/EasyDialogsWin.py
@@ -21,6 +21,7 @@ Based upon STDWIN dialogs with the same names and functions.
 
 from __future__ import division
 from __future__ import print_function
+from six import text_type
 
 import os
 
@@ -772,7 +773,7 @@ def AskFolder(
     def BrowseCallback(hwnd, uMsg, lParam, lpData):
         if uMsg == BFFM_INITIALIZED:
             if actionButtonLabel:
-                label = unicode(actionButtonLabel, errors='replace')
+                label = text_type(actionButtonLabel, errors='replace')
                 user32.SendMessageW(hwnd, BFFM_SETOKTEXT, 0, label)
             if cancelButtonLabel:
                 cancelButton = user32.GetDlgItem(hwnd, IDCANCEL)
--- a/launcher/game/archiver.rpy
+++ b/launcher/game/archiver.rpy
@@ -29,7 +29,11 @@ init python in archiver:
     import random
     import glob
 
-    from cPickle import dumps, HIGHEST_PROTOCOL
+    # TODO: get rid of cPickle when py3 migration is complete
+    try:
+        from cPickle import dumps, HIGHEST_PROTOCOL
+    except ImportError:
+        from pickle import dumps, HIGHEST_PROTOCOL
 
 
     class Archive(object):
--- a/launcher/game/change_icon.py
+++ b/launcher/game/change_icon.py
@@ -301,7 +301,7 @@ def change_icons(oldexe, icofn):
     physize = rsrc_section.SizeOfRawData
     virsize = rsrc_section.Misc_VirtualSize
 
-    f = file(oldexe, "rb")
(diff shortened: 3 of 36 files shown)


renpytom:
There's no way I can merge this in one go. I'm going to need lots of small PRS that are obviously correct. Python 2 support is still paramount, so we need to do that.

I'm also wondering if pasteurize is a better choice than six, since it lets us start writing Python 3 code.

caryoscelus:
I can obviously slice this pr into individual commits when they're ready, but what's the point? Just cherry-picking the good ones would have the same effect.

Also, why support py2 when py3 version will be ready? Such compatibility is going to provide problems with porting and i don't see much gains except for being able to continuously merge changes into master.

Python-future might be a better choice, yeah. I'll look into it later.

renpytom:
Ren'Py games can take up to five years to develop, and often the ones that take longest are the most Python heavy. We want to support python2 for at least a couple of years so the creators don't have to rewrite code that is working.

caryoscelus:
Ah, i see. Well, then i'm not sure my work will be directly usable for you.

If you don't have plans to work on porting in the nearest future, i might still continue this development and sustain a py3 fork until there's official py3 support.

jsfehler:
@caryoscelus You should pull out the commits for Exception catching and replacing file() with open() and resubmit them as 2 new PRs. At a glance they're both python 2.7 compatible.

@renpytom For migrating to python3, it might be helpful to differentiate between ren'py's guts and python code parsed by ren'py. You're right that we need to support python2.7 syntax, but perhaps the engine itself could use python3. 

I suspect that for all but the most intense projects, the amount of code change required to switch from python27 to 3x will be minor. Still, if there's a stable way to get the new features and performance of python3 without any risk of hurting game devs, that would be ideal.

kibertoad:
@jsfehler I've already extracted the Exception part as a separate PR, see #1101
file -> open : I think #1102 has that covered, if something is left, I'll do it after that one is merged.

caryoscelus:
@jsfehler , anybody can do it, commits are separate. I'm not interested in maintaining a lot of prs.

infirit:
[Python 2.7 will go eol in 2020](https://pythonclock.org/) so better move to 3.6 soon.

apoleon:
Hello,

could someone clarify the progress and timeline of the Python 3 port. As a matter of fact most Linux distributions will soon release without Python 2 thus software like renpy can no longer be part of these distributions. I believe other OS like Windows or MacOS will have a similar Policy because of the lack of security support for Python 2. I know that pygame-sdl2 works with Python 3 but renpy itself seems to be not ready yet.

Boundarybreaker:
Yeah, Python 2.7 is ***very*** dead now, so this is pretty crucial.

langbakk:
It's not "crucial", in any way, shape or form. RedHat has started 2.7 will be officially supported and provided till 2024, and the other big ones, like Ubuntu / Debian, has also stated that even if python 3 is default, 2.7 will be provided via official channels. There is no immediate threat of Renpy suddenly not working.

apoleon:
This is not correct. I am the Debian maintainer of Renpy and Debian plans to remove Python 2 completely before the next stable release. That means Renpy will also be removed if there is no Python 3 port.

Daniihh:
LGTM, `pip2` tells me that Python 2 has been deprecated since the start of this year.
If you don't want to solve the conflicting files just yet, I've made #2258 which just does the bare minimum syntax changes...
Something should be done soon, I know that many people would benefit from Python 3.

LuqueDaniel:
@caryoscelus Are you still interested in maintaining this PR?

caryoscelus:
No, unfortunately i don't have time to spend on ren'py

Jun 29, 2020 17:17:13 Daniel Luque <notifications@github.com>:

LuqueDaniel:
Then I will close this PR. You can continue the discussion #2003

Issue #1098, opened 2016-12-24
Python 3 porting


caryoscelus:
Is anybody currently working on this? I'll try it if nobody else does.

Do you plan to switch quickly or with a transition period? I.e. should the porting effort be compatible with both versions (e.g. using six) or only py3?

caryoscelus:
Closing this in favour of more specific code discussion in https://github.com/renpy/renpy/pull/1100

kibertoad:
I believe we still need an issue to track the progress, for Issues and PRs serve different purposes.

Zenadix:
+1, I think it's important to have Python 3 compatibility.

monospacedmagic:
This Ren'Py Python 3 port based on https://github.com/renpy/renpy/commit/01f536fc698ee5cf68f512e8125469573afac0be appears to work already: https://github.com/nx-python/renpy/tree/renpy-py3

kibertoad:
This is more relevant than ever, given https://pythonclock.org/
Are there any plans to drop Python 2 support at any point?

renpytom:
Closing, now that Ren'Py 8.0 has nightlies being build.

PR #1101, opened 2016-12-25
Fix exceptions to be compatible with Python 3.

kibertoad:
refs #1098

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,7 @@ styles.txt
 /.pydevproject.bak
 /.project
 /.cproject
+/.idea
 /.settings
 /LICENSE.txt
 /templates/english/README.html
--- a/renpy/execution.py
+++ b/renpy/execution.py
@@ -458,7 +458,7 @@ class Context(renpy.object.Object):
                     except renpy.game.CONTROL_EXCEPTIONS as ce:
                         raise ce
                     except Exception as ce:
-                        raise exc_info[0], exc_info[1], exc_info[2]
+                        raise exc_info[0](exc_info[1]).with_traceback(exc_info[2])
 
                 node = self.next_node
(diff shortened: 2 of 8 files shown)


caryoscelus:
I would suggest rewriting commit message to put issue reference into message body rather than title.

kibertoad:
Done.

caryoscelus:
Eh, i meant commit message, not pr.

Mirv:
You're very speedy kibertoad!  

I am not really involved in the coding yet, but I like to read what is going on!  I saw in the `execution.py` you had used literal numbers for the arrays.  Is there a way to loop over that which would make it so any additions or removals would not need to be hand coded?

caryoscelus:
@Mirv, it's just results of `sys.exc_info()` call. You aren't supposed to add or remove anything from it anyway.

renpy/execution.py
```diff
@@ -458,7 +458,7 @@ def run(self, node=None):
                     except renpy.game.CONTROL_EXCEPTIONS as ce:
                         raise ce
                     except Exception as ce:
-                        raise exc_info[0], exc_info[1], exc_info[2]
+                        raise exc_info[0](exc_info[1]).with_traceback(exc_info[2])
```
jsfehler: This only works in python3

.gitignore
```diff
@@ -38,6 +38,7 @@ styles.txt
 /.pydevproject.bak
 /.project
 /.cproject
+/.idea
```
jsfehler: This doesn't really have anything to do with exceptions or python3

tcyrus:
Shouldn't this be closed since #1596 was merged

renpytom:
Closing.

PR #1102, opened 2016-12-25
Replace 'open()' with 'with open()'

jsfehler:
I wish I could think of a commit message that didn't use the word "with" twice.


caryoscelus:
"Use 'with open()' instead of 'open()'". Or "Use 'with open()' for file handling"..

renpytom:
I generally like these, except for the change to dialogue.py. A pattern I use something is to use an object as a namespace in lieu of having to pass around a ton of parameters, and I think that makes sense in this case rather than adding a parameter. Please back that out so I can can merge the change.

jsfehler:
@renpytom Undid the changes to dialogue.py.

kibertoad:
Any follow-up on this?..

renpytom:
It will get merged once I'm sure 6.99.12 is fully done. I'm leaning towards putting out a 6.99.12.3, and stuff like this isn't necessary for that.

Uzume:
I notice this changes a number of instances of `file`. That still exists in Python 3 but is no longer globally available by default. That said, it might be better to consider moving from `open` to `io.open` (which uses key word args instead) and in a similar way move from `StringIO` and `cString` to `io.StringIO`. At some point in the PY3 port we will have to bite the bullet and move to unicode strings and using `io` will probably help there.

jsfehler:
@Uzume `file()` doesn't exist in Python 3 at all, as far as I know. The functionality is completely replaced by `open`. In Python 3, `open` is an alias of `io.open`.

In this PR, most of the reading/writing is binary so setting an encoding shouldn't matter. There's a few that read text and could have an encoding specified. `io.open` could be used, or `f.read().decode("UTF-8")` @renpytom would know if the text files are dangerous or not. Currently they don't specify encoding, so I imagine it hasn't been a problem so far.

Uzume:
@jsfehler Agreed. The canonical way to create a file object is by using the `open` function which as you said in CPython 3 is an alias of `io.open` and exactly why the most portable method handle such in CPython 2 is to use `io.open` with something like:
```python
from io import open
```
in method similar to but obviously different a `from __future__ import` statement.

In CPython 3 all `StringIO` implementations also end up being `io.StringIO` so we should probably use that in lieu of `StringIO.StringIO` and `cStringIO.StringIO`

I agree most of renpy's file manipulation is binary and thus any discussion of encodings are moot, however, it does support text for game content and fonts, etc. It should noted that in CPython 2, most platforms have broken Unicode support for things beyond the BMP  (and why `unichr(0x10000)` can throw a `ValueError`; this was basically fixed in CPython 3.3 with PEP 393).

Uzume:
@renpytom 
> I generally like these, except for the change to dialogue.py. A pattern I use something is to use an object as a namespace in lieu of having to pass around a ton of parameters, and I think that makes sense in this case rather than adding a parameter. Please back that out so I can can merge the change.

I cannot see what jsfehler's original code did before it was backed out but based on your above comment, this sounds much like Python 3's pathlib path objects. In any event, I do not see why the following in `renpy/translation/dialogue.py`:
```python
        self.f = open(output, "a")

        self.write_dialogue()

        self.f.close()
```
 cannot be replaced with:
```python
        with open(output, "a") as self.f:
            self.write_dialogue()
```

PR #1103, opened 2016-12-27
Fix typos in new GUI docs

morganwillcock:


sphinx/source/gui.rst
```diff
@@ -363,15 +363,15 @@ documented above, the following variables exist:
...
-    The color used by focused items in the gui, including the text of
+    The color used by focused items in the GUI, including the text of
     of buttons and the thumbs (movable areas) of sliders and scrollbars.
 
 .. var:: gui.selected_color = '#555555'
 
     The color used by the text of selected buttons. (This takes priority
     over the hover and idle colors.)
 
-.. var:: gui.insensitive_color = '#8888887f'
```
xavi-mat: Is this a typo? I thought this is gray with some level of transparency (~50%)

morganwillcock: Ah, could be. But I just followed the description in the same document:

`These variables should generally be set to hexadecimal color codes, which are strings of the form "#rrggbb"`

morganwillcock:
I've updated it, assuming that the alpha component is valid for use everywhere.

renpytom:
Merged, thank you.

PR #1106, opened 2017-01-04
Update about.rpy

doyousketch2:
Added Python version info to the About screen.

Image


renpytom:
Is there a reason for this? I'm not sure it adds any information beyond the Ren'Py version number, at least for official releases. We have to show the Ren'Py version number for support purposes, but it's not clear it makes to show the Python version and not the SDL, Freetype, FFMPEG, and other versions.

doyousketch2:
I've got at least 3 versions of Python on my computer.
As you can see, if I run it with the renpy.sh that came with it,
then Ren'Py uses an older version of Python.

Once I had this info, I wrote my own shell script to launch Ren'Py with Python 2.7.13 instead.

It also helps to know when trying to get renpygame to work, 
so I don't waste my time installing python3 dependencies.

doyousketch2:
I'll look into the other things you mentioned, 
I just don't know how I'd go about doing those, 
as they wouldn't be a simple sys call.

Might need to put that info into a global variable when they are initialized,
or write a function to return those values when you call renpy.gl from within the About.py script.

At first glance, neither of these methods do anything:
```
    renpy.gl.glGetString( GL_VERSION )
    renpy.gl.SDL_GetVersion()
```
So yeah, if anyone else has any idea how to implement the other ideas, I'm alright with that.

I just know that the info that I've added so far was helpful for me,
so I thought I'd pass it along to you,
to help others from struggling with the same things I did.

doyousketch2:
Ok, this looks like a promising solution:
```
    $ import pygame
    $ SDLmaj = pygame.display.gl_get_attribute(pygame.GL_CONTEXT_MAJOR_VERSION)
    $ SDLmin = pygame.display.gl_get_attribute(pygame.GL_CONTEXT_MINOR_VERSION)
```
I'll see if I can glean some of the other info as well, but that'll end up being later on today/tonight.

doyousketch2:
Here's how it looks so far.  I gotta crash tho, or brain is gonna crumple.
I'll upload stuff to GitHub when I get up again.

pic

renpytom:
I don't think the solution is to clutter the screen up with more information. Running Ren'Py with an alternative python implementation is something that has to be explicitly done, so I don't think it really adds much to display it, while I think it makes the screen more cluttered, and kind of a mess. 

I don't see a good reason to merge this that would affect more than a couple of users.

doyousketch2:
It'd be easy enough to put a button on there,
to only show the developer info when you click it,
like that "view license" link you've got there.

renpytom:
Sure, but we have to articulate why people would want developer info like this in the first place. I'm not seeing that.

renpytom:
I don't see why this is needed. What's more, it sure looks like much of the information will be wrong, since it sources things from pygame rather than Ren'Py. I'm going to close this PR so you don't spend time on something that's unlikely to be merged. 

I'm sorry, but I just don't see enough of a case for this.

doyousketch2:
Wow, that was rather rude and dismissive.
I thought we were having a discussion here.

The whole point is to get pygame info to get renpygame running.
So developers can take Ren'Py to the next level.

PR #1108, opened 2017-01-04
toggle_afm

vollschauer:
Simply bind 'a' to toggle afm ...

### DIFF
--- a/renpy/common/00keymap.rpy
+++ b/renpy/common/00keymap.rpy
@@ -27,7 +27,7 @@ init -1600 python:
         # disabled.
         rollback = [ 'K_PAGEUP', 'repeat_K_PAGEUP', 'K_AC_BACK', 'mousedown_4' ],
         screenshot = [ 's' ],
-        toggle_afm = [ 'a' ],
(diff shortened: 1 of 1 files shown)


renpytom:
I'm not a fan of this. The problem is that it makes it too easy for a single errant keypress to switch on afm mode, and then the player has no idea how to get out of it*. Is toggling AFM such a frequent activity that it needs a keybinding?

* This was a problem with self-voicing mode, and we had to fix it by adding the popup.

vollschauer:
I just thought it would be nice to have and how about merge it without binding it yet? 
So people could use it if they need a toggle afm.

renpytom:
I'd be fine merging it without the keybinding, and without the _toggle_afm or toggle_afm functions - renpy.run(Preference("auto-forward mode", "toggle")) does the same thing, as does just accessing the preference directly. We don't need a function for it. The _keymap_toggle_afm function should stay.

vollschauer:
Cool, I always learn something new :+1:  updated the file ... Thanks!

renpytom:
Merged. Thanks.

PR #1110, opened 2017-01-10
Include widget stack in traceback when relevant

SpikyCaterpillar:
When a game crashes because of an unclosed layer/widget stack, includes the widget stack in the traceback.


renpytom:
Merged, thanks.

I'm not sure how relevant this is, since I don't think this error is possible without calling UI functions rather than using screens. But that's no reason not to include it, and it doesn't look like it could hurt.

PR #1142, opened 2017-02-25
Arabic translation for 6.99 common.rpy

renoa:
common.rpy

### DIFF
--- a/launcher/game/tl/arabic/common.rpy
+++ b/launcher/game/tl/arabic/common.rpy
@@ -1,33 +1,33 @@
-﻿
+
 translate arabic strings:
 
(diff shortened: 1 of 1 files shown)


Ria-kon:
It must be named as "Arabic translation update for 6.99"
And… well, no need to create more than 1 pull request.

renoa:
i'm sorry about using the wrong name, i assumed it went into the exact folder where i edited it from...?? 
and also sorry about the multiple pull requests, it's been a while since i used github. I only really use it to translate Ren'Py, i couldn't find how to make a fork of only the translation file (cry)

PR #1148, opened 2017-03-10
Optimised some of the renpy PNG files with optipng -o7

mbrickn:
I'm pretty new so I'm not sure if I did this right. Sorry if I didn't!

It only saves a few kilobytes, but it's something right? ^_^

PS: Thank you for making Ren'Py!


mbrickn:
The Question got updated, so I think merging this might cause problems at this point? 

Sorry for any trouble! ^ _ ^

PR #1156, opened 2017-03-27
Updating russian tl for new version

Ria-kon:
Temp post for making russian renpy uptodate =D

So, adding legacy options…

### DIFF
--- a/launcher/game/tl/russian/launcher.rpy
+++ b/launcher/game/tl/russian/launcher.rpy
@@ -925,10 +925,6 @@ translate russian strings:
     old "Hardware rendering"
     new "Аппаратный рендеринг"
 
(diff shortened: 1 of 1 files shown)


renpytom:
Yep. I want to improve the default new game experience. 

Is there something that notifies you when I make a translatable change? You're very fast with these.

Ria-kon:
I just watching for your commits and also recreating new "fake translation" in prereleases and compare it (winmerge or anything like this staff) with existed one.

Sooo... I think, I notifies myself =D

renpytom:
Thanks. I feel like at some point, I should invest some time into developing a tool for making that easier on you and other translators (I was hoping you already had such a tool, so I wouldn't have to make one.)

PR #1162, opened 2017-04-08
Updated urls linking to renpy.org to use https

mbrickn:
Minor change. I just changed URLs to the Renpy website to use https instead of http.

This will make users go directly to the webpages linked, instead of automatically being redirected to the https version by the website. Should be (very slightly) better for user experience! 😊


renpytom:
Merged, thanks.

mbrickn:
Thank you! ^ _ ^

PR #1183, opened 2017-05-17
Russian Translation Ongoing

Ria-kon:
For nightly version of renpy. Right now we proofreading new tutorial

.### — notes (editing, problems, etc)
+++ — changes in our commits

For example:
- There are some http changes to https +++
- "Change the size and spacing of various things.." +++ 2 dots to 1
- Quick buttons on original was deleted. Is this ok? ###
- Theme test was deleted cause of commented line ###


Ria-kon:
Notes to original for now:

# e "Of course, there's a lot more to transforms than this. If you want to learn more, you can read the sections on Position Properties, Transforms, and Transform Properties." ### "Transforms and Animation", not "Transforms"
# e "Some games might prompt the player for input." ### "What's your name?" must be translated string!!!
# interactive director needs translation!!!

Ria-kon:
Notes №2:
# e5 "The ``what_outlines`` property puts an outline around the text." ### is this neccesary ``?
# e7 "The what_xalign property controls where all the text text itself is placed within the textbox, while what_textalign controls where rows of text are placed relative to each other." ### "text text" doubling
# e "The Text displayable lets Ren'Py treat text as if if it was an image." ### "as if if" and double space in original
# e "To break a line without pausing,\none can write \\n. \\' and \\\" include quotes in the text." ### comma, not dot
# e "For example, this displays the [variable]." ### variable not translating
# e "When the variable name is followed by !t, it is translated to [variable!t]. It could be something else in a different language." ### here must be translatable!t?
# e "Let's see them all in action." ### eileen up no sprite
# e "The AlphaDissolve transition lets you use one displayable to combine two others. For example..." ### example? where?
# e "The new screen is me just standing here." ### no alphadissolve too

Ria-kon:
Notes part 3:

# e "For example, in an RPG like visual novel, a screen can display the player's statistics." ### no translation in display for "Player", "Eileen", "HP"
# e "Complex screens can be the basis of whole game mechanics. A stats screen like this can be the basis of dating and life-sims." ### strange line in third bar. "Done" needs to be translated|
# e "I'll work from the inside out to describe the statements. But first, I'll show the screen so you can see it in action." ### "This is a screen" and "Okay" needs translation
# e "Here's an example of a screen that takes three parameters. The message parameter is a message to show, while the okay and cancel actions are run when the appropriate button is chosen." ### "Hello, world" and "Okay" not translatable, but not cancel =D
# e "Parameters let us change what a screen displays, simply by re-showing it with different arguments." ### "Shiro was here." not translatable
# e "The call screen statement can also take arguments, much like show screen does." ### "click either button to continue" not translatable
 # e "The python statement works just about the same way it does in the script. A single line of Python is introduced with a dollar sign. This line is run each time the screen updates." ### again with no translation "Hello, World."
# e "The for statement takes a list of values, and iterates through them, running the block inside the for loop with the variable bound to each list item." ### no translation for "Earth", "Moon", "Mars".
# old "How to show screens." ### "How to show screen?" or with dot?

Ria-kon:
Notes part 4. Main part of translation done. Now only grammar check, notes check, plus up-to-dating

# e "The HBox displayable is used to lay its children out horizontally. By default, there's no spacing between children, so they run together." ### HBox? HB?
# old "The road to the stars is steep and dangerous. But we are not afraid.\n–Yuri Gagarin" ### You'll be laughing, but there is no quote for this in russian lang. Only second part was founded in original russian =D
# e "Sorry, that's wrong. The xpos is .75, and the ypos is .25." ### double 1
# e "In other words, it's 75%% of the way from the left side, and 25%% of the way from the top." ### double 1
# translate russian tutorial_positions_6f926e18_1: ### double 2
# e "In other words, it's 75%% of the way from the left side, and 25%% of the way from the top." ### double 2
# e "Ren'Py uses transforms to animate, manipulate, andplace images. We've already seen the very simplest of transforms in use:" ### "andplace" to "and place"
# e "Animation is when the displayable being shown changes. For example, right now I am changing my expression." ### no animation, lol
# e "This tutorial game has only scratched the surface of what you can do with ATL. For example, we haven't even covered the on and event statements. For more information, you might want to check out the ATL chapter in the reference manual." ### link to documenation?
# "The subpixel property controls how things are lined up with the screen. When False, images can be pixel-perfect, but there can be pixel jumping." ### no original translatable text
# "When it's set to True, movement is smoother at the cost of blurring images a little." ### no original translatable text
# e "The text displayable can also interpolate values enclosed in square brackets" ### dot
# e "Without the background, text can be hard to read. While a frame isn't strictly required, many screens have one or more of them." ### "this is a screen" no translation
# e "But when I add a background, it's much easier. That's why there are two displayables that are intended to give backgrounds to user interface elements." ### "this is a screen" + "Okay" no translation
# e "A button takes another displayable as children. Since that child can be a layout, it can takes as many children as you want." ### again this strange line on bar… I don't know why
# e "Buttons take Window window properties, that are used to specify the background, margins, and padding. They also take Button-specific properties, like a sound to play on hover." ### "Window window" is "Window style"
# e "You chose to go home." ### error and Ren'Py crashes…

renpytom:
First off, thanks for the editing pass and the translation and the translation checking. 

From now on, can we not start each bullet point with a - rather than a #? The # character is causing your comments to be formatted in weird giant text, which made them hard to read and understand. I'm going in and editing that. 

And just to be clear, +++ are things you've dealt with, and ### means they need my attention, right?

Ria-kon:
Yes. Well, almost yes, i just doubting in example +++ and that's all

renpytom:
I've started to fix things in the russian-new-tutorial branch, which is based on yours.

renpytom:
I think I've addressed everything here. There was a regression that broke ATL that caused animations not to appear in some places, despite being mentioned in the script. Changes I didn't do:

* "Sorry, that's wrong. The xpos is .75, and the ypos is .25." - This is an intentional repeat.
* The subpixel stuff is commented out pending some work on the example.
* I'm not sure about the strange bar lines. I don't see them.

I need to research the Gagarin quote more. I seem to recall it being associated with Vladimir Komarov's death or funeral, and it's possible that I got it from either Slaton or more likely Stafford's book. If I can't find a good citation, I'll change it.

As for me, I'm going to merge and integrate the Russian translation now. Thanks for all this work - it makes Ren'Py a lot better for everyone.

renpytom:
17e2ed92 make the Russian translation public. It still needs to be updated, but I think what we have is quite good, and no reason not to show it off in the nightlies.

Thanks again.

PR #1186, opened 2017-05-20
translated error.rpy options.rpy script.rpym

renoa:
sorry if this isn't the right way to do a pull request ;-;


renpytom:
Merged. Sorry for the delay, I didn't see this. Thank you.

PR #1216, opened 2017-07-16
Improve compatibility support for games built with older renpy

bircoph:
The problem with Katawa Shoujo is quite known it at least two distributions: when people want to use the latest RenPy the game fails to start:
https://bugs.gentoo.org/show_bug.cgi?id=601200
https://aur.archlinux.org/packages/katawa-shoujo/

The second problem with ElvenRelations was found when looping through my local RenPy games collection.

The third change increases hardcoded values for the infinite loop check: on older hardware it triggers on valid games. Better solution will be to depend this values on CPU frequency/number of cores/generation or make it user configurable.

Commit messages describe both problems and solutions with greater detail. RenPy used for testing is the latest official release 6.99.12.4, though it looks like affected code was not changed in git HEAD.

### DIFF
--- a/renpy/ast.py
+++ b/renpy/ast.py
@@ -696,7 +696,7 @@ class Init(Node):
 
     def execute(self):
         next_node(self.next)
-        renpy.execution.not_infinite_loop(60)
+        renpy.execution.not_infinite_loop(300)
         statement_name("init")
 
     def restructure(self, callback):
@@ -858,7 +858,7 @@ class EarlyPython(Node):
 
     def execute(self):
         next_node(self.next)
-        renpy.execution.not_infinite_loop(60)
+        renpy.execution.not_infinite_loop(300)
         statement_name("python early")
 
     def early_execute(self):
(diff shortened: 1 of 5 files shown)


renpytom:
Thank you for sending these to me. It's good to know about problems with older games.

I've merged cf3f7fd, and fixes for the problems in the other two patches. 

1d350d9 removed compatibility code that's there for a reason, so I rewrote the compatibility code. Similarly, the infinite loop code is there for a reason. A 10 second pause is always a bug in a game during development, so the way ff30819 increased that to 50 seconds didn't seem like a good idea. Instead, I added some code to use config.developer to decide if we want to multiply the timeouts by 5 when they're set. 

In the future, please submit each change as a separate PR. It makes them easier to deal with.

Thank you for this - it was very valuable.

bircoph:
Hi, thank you for taking care of these patches. One comment, however:
> A 10 second pause is always a bug in a game during development
Usually, but not always. On very slow CPU it may take more time to start the game, e.g. on very old Intel Atom N270 which I use :) Modern phones are faster.

Issue #1113, opened 2017-01-16
Scaling window make text in Help goes out-of-frame (new GUI)


Ria-kon:
1280x720 (TAB) ![screenshot0001](https://cloud.githubusercontent.com/assets/24704640/21969489/6536fcd8-dba7-11e6-93fa-f33f521136d3.png)
1251x703 (ideal) ![screenshot0004](https://cloud.githubusercontent.com/assets/24704640/21969501/7f3b6c0e-dba7-11e6-88d6-e74c007c0ab3.png)
1176x662 (CTRL, H) ![screenshot0002](https://cloud.githubusercontent.com/assets/24704640/21969538/b733b10c-dba7-11e6-98fd-88cc1b57c777.png)
960x540 (TAB, H, g all over the text) ![screenshot0003](https://cloud.githubusercontent.com/assets/24704640/21969566/ebf7905c-dba7-11e6-9368-4b910e4d8f09.png)

renpytom:
Heh. I was working on this for a while when you filed the report. Thank you, though - this is a really good catch.

PR #1222, opened 2017-08-04
Updating russian translation and some more notes.

Ria-kon:
Well, I found some more issues and sorting them:

# Tutorial. 
# Example window not translating. 
I think, this is because of using variables and text tags.
- eslow "The nw tag causes Ren'Py to continue past slow text,{nw}" ### english example. next string is russian, lol
- e "To interpolate a variable, write it in square brackets. Isn't that right, [name]?" ### english example
- e "Variable interpolation also works with other variables. Here, the answer is [answer] and the flag is [flag]." ### example no translation
- old "The answer is [answer]." ### english example
- old "Text tags {color=#c8ffc8}work{/color} in screens." ### english example
- e "For example, we might want to have text that is {b}bold{/b}, {i}italic{/i}, {s}struckthrough{/s}, or {u}underlined{/u}." ### english example
- e "The a text tag {a=https://www.renpy.org}links to a website{/a}." ### english example
- e "The alpha text tag makes text {alpha=.5}translucent{/alpha}." ### english example
- e "The color text tag changes the {color=#0080c0}color{/color} of the text." ### english example
- e "The cps text tag {cps=25}makes text type itself out slowly{/cps}, even if slow text is off." ### english example
- e "The cps tag can also be relative to the default speed, {cps=*2}doubling{/cps} or {cps=*0.5}halving{/cps} it." ### english example
- e "The font tag changes the font, for example to {font=DejaVuSans-Bold.ttf}DejaVuSans-Bold.ttf{/font}." ### english example
- e "Sometimes, changing to a bold font looks better than using the {{b} tag." ### english example
- e "The k tag changes kerning. It can space the letters of a word {k=-.5}closer together{/k} or {k=.5}farther apart{/k}." ### english example
- e "The size tag changes the size of text. It can make text {size=+10}bigger{/size} or {size=-10}smaller{/size}, or set it to a {size=30}fixed size{/size}." ### english example
- e "The space tag {space=30} adds horizontal space in text.{vspace=30}The vspace tag adds vertical space between lines." ### english example
- e "The p tag breaks a paragraph,{p}and waits for the player to click." ### english example
- e "If it is given a number as an argument,{p=1.5}it waits that many seconds." ### english example
- e "The w tag also waits for a click,{w} except it doesn't break lines,{w=.5} the way p does." ### english example
- e "When the variable name is followed by !t, it is translated to [variable!t]. It could be something else in a different language." ### здесь должно быть translatable!t ### english example
- e "When the variable name is followed by !t, it is translated to [variable!t]. It could be something else in a different language." ### здесь должно быть translatable!t ### english example
- e "Finally, certain characters are special. [[, {{, and \\ need to be doubled if included in text. The %% character should be doubled if used in dialogue." ### english example
- e "For example, this displays the [variable!t]." ### english example
- e "When the variable name is followed by !q, special characters are quoted. This displays the raw [variable!q], including the italics tags." ### english example 

# variable!q!t
- e "When the variable name is followed by !q, special characters are quoted. This displays the raw [variable!q], including the italics tags." ### btw, set it to [variable!q!t] in original for translation measures. Otherwise it can be english raw text only.

# Where are displayable statements?
I just don't understand where to send tutorial player =D
- e "The screen language has a few statements that do things other than show displayables. If you haven't seen the displayable statements yet, you might want to check them out and come back here."

# Interactive Director. 
# String ID.
String ID take too much space. For example, in director part of the turorial string ID (left side) pushed strings themself right, so now string ID in the center of separating line. Watch it on screenshot.
- e "Let's add a background. Click the '+' to pick where to add it, then the 'scene' statement and 'washington' for the image. Finally, click 'Add' to add it." ### string ID take too much space

# Launcher 
- I still watching this (https://github.com/renpy/renpy/issues/1113) bug. Less now, but some symbols just keep go out-of-frame. "Пробел" after "Enter" on screenshot.

pythonw_2017-08-04_20-14-39
pythonw_2017-08-04_20-15-12

### DIFF
--- a/launcher/game/tl/russian/common.rpy
+++ b/launcher/game/tl/russian/common.rpy
@@ -165,6 +165,86 @@ translate russian strings:
     old "The interactive director is not enabled here."
     new "Интерактивный директор недоступен."
 
+    # 00director.rpy:1461
+    old "Done"
+    new "Принять"
+
+    # 00director.rpy:1469
+    old "(statement)"
+    new "(функция)"
+
+    # 00director.rpy:1470
+    old "(tag)"
+    new "(тег)"
+
+    # 00director.rpy:1471
+    old "(attributes)"
+    new "(аттрибут)"
+
+    # 00director.rpy:1472
+    old "(transform)"
+    new "(трансформация)"
+
+    # 00director.rpy:1497
+    old "(transition)"
+    new "(переход)"
+
+    # 00director.rpy:1509
+    old "(channel)"
+    new "(канал)"
+
+    # 00director.rpy:1510
+    old "(filename)"
+    new "(имя файла)"
+
+    # 00director.rpy:1535
+    old "Change"
+    new "Изменить"
+
+    # 00director.rpy:1537
+    old "Add"
+    new "Добавить"
+
+    # 00director.rpy:1540
+    old "Cancel"
+    new "Отмена"
+
+    # 00director.rpy:1543
+    old "Remove"
+    new "Убрать"
+
+    # 00director.rpy:1576
+    old "Statement:"
+    new "Функции:"
+
+    # 00director.rpy:1597
+    old "Tag:"
+    new "Теги:"
+
+    # 00director.rpy:1613
+    old "Attributes:"
+    new "Аттрибут:"
+
+    # 00director.rpy:1631
+    old "Transforms:"
+    new "Трансформации:"
+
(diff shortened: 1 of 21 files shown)


renpytom:
Just FYI, the Interactive Director section is blank.

Ria-kon:
I mean top part, not visualization zone. Where tutorial_director:0072 is.

PR #1230, opened 2017-08-27
Updating russian translation

Ria-kon:
Beside some changes in translation and updating translation.rst, there is one little change in tutorial itself:

Original:
"To break a line without pausing,\none can write \\n, \\' and \\\" include quotes in the text."

New:
"To break a line without pausing,\none can write \\n. \\' and \\\" include quotes in the text."

Otherwise, I understand this as "you can break a line with \\'  and \\\"…" and didn't understand "include quotes in the text".


renpytom:
Merged, but see c6f8a5d0. We don't need editorials in the TRANSLATORS file.

PR #1244, opened 2017-09-07
Translation update for custom resolution and russian

Ria-kon:
Main feature: Custom resolution now translatable in launcher.

Other is just update.


renpytom:
Thank you.

PR #1258, opened 2017-09-29
Update russian to prerelease

Ria-kon:


renpytom:
Thank you.

renpytom:
Would you mind watching #1262, please? I've opened that so I can point out changes to the translation to the translators.

PR #1260, opened 2017-10-01
fix unlimited save slots count

lgb:
By default, user can scroll to the right and see a lot of pages for save/load screen. Some users report this like a bug in game.

Otherwise, I think is may be configurable by another value - like config.max_save_pages or some, for split number of visible pages from all pages count.

### DIFF
--- a/renpy/common/_layout/classic_load_save.rpym
+++ b/renpy/common/_layout/classic_load_save.rpym
@@ -289,7 +289,7 @@ init python:
                     tb(True, name, ui.returns(("fppset", i)), fpp == i)
 
                 # Next
-                tb(True, u'Next', ui.returns(("fppdelta", +1)), False)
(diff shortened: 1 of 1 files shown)


Ria-kon:
This is a feature, not a bug~
And if you really need this bad user experience feature with limited save slots — FilePageNext() can take argument "max=n", where "n" by default is None. You can change it to any number you want.
https://www.renpy.org/doc/html/screen_actions.html#FilePageNext

renpytom:
I concur with Ria on this - thanks for wanting to contribute, but I think it's a good idea to let players have as many pages as they require. This is one of those things where we have a workaround for people who demand to limit the number of pages, but it should be hard to use, since it doesn't improve things. 

Just because users see it as a bug, doesn't mean it is. :)

Issue #1262, opened 2017-10-03
Translations need to be updated.


renpytom:
This is a bug that is intended to never be closed. It's here so translators can subscribe. I'll add comments to this issue when translations change after the initial prerelease of Ren'Py. The idea here is that translators can subscribe to this issue, so that github can email them when there's something to change.

renpytom:
I've recently added two major translation updates. 

The first is an error message in the launcher that displays when Ren'Py is run from a mac disk image. This is every important, since it can be the first thing a new creator sees. 

The second is a message in the common code that is displayed when a screenshot fails to save.

renpytom:
I had to add a new translation at the last minute. I'm planning on releasing 6.99.13 today or tomorrow.

renpytom:
Just to let everyone know, 6.99.14.1 is done. There are a few new strings, one on the front page of the launcher. If people can let me know you're working on it, I'll hold up the release until you're finished.

xavi-mat:
I'll try to translate most of Spanish missing strings today or tomorrow.

renpytom:
Everybody: 6.99.14.2 is almost done, so please get your translations in soon. (And let me know if you're planning on working on this in the next few days.)

renpytom:
Landed a number of new changes to 6.99.14.2, allowing the self-voicing mode to be translated. Please let me know ASAP if I should hold up the final release so this can be translated.

fblampe:
I added some new German translations for the launcher main menu in this PR: https://github.com/renpy/renpy/pull/1599

renpytom:
Closing, as this was never used for what it was intended.

PR #1263, opened 2017-10-04
Schinese update for launcher.

zedraxlo:


renpytom:
Merged, thank you. If you haven't already, please subscribe to  #1262 so you can be updated when the translation needs to be changed.

PR #1270, opened 2017-10-08
Fixed issue with Directory build option.

ianLeslie2:
Previously, as the files got added to the 'archive', if adding a file required more than one parent folder to be created then the build would fail.


renpytom:
Merged, thank you.

PR #1271, opened 2017-10-08
Added feature to renpy launcher allowing you to specify additional di…

ianLeslie2:
…rectories that Renpy should scan for projects.

These directories are specified in a file called "projectsPaths.txt" that is placed in the main projects directory.
---------

The need for this feature came up when I tried working with Renpy projects hosted in a google drive.
Getting the google-drive system to only share a specific renpy project folder is awkward. It's much easier to just point Renpy at the project folder you want it to scan.

### DIFF
--- a/launcher/game/project.rpy
+++ b/launcher/game/project.rpy
@@ -435,17 +435,15 @@ init python in project:
 
             if self.projects_directory is not None:
                 self.scan_directory(self.projects_directory)
-                
-                # If a file called "projectPaths.txt" exists in the main project directory,
-                # also scan any directories listed there
-                extraPathsFName = os.path.join(self.projects_directory,"projectPaths.txt")
-                if os.path.isfile(extraPathsFName):
-                    extraPathsF = open(extraPathsFName,"r")
-                    for path in extraPathsF:
+
+                # If a file called "projects.txt" exists in the main project directory,
+                # include any projects listed in it.
+                extra_projects_fn = os.path.join(self.projects_directory,"projects.txt")
+                with open(extra_projects_fn,"r") as f:
+                    for path in f:
                         path = path.strip()
-                        if len(path) > 0 and os.path.isdir(path):
-                            self.scan_directory(path)
-                    extraPathsF.close()
+                        if len(path) > 0:
+                            self.scan_directory_direct(path)
 
             self.scan_directory(config.renpy_base)
             self.scan_directory(os.path.join(config.renpy_base, "templates"))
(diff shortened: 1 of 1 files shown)


launcher/game/project.rpy
```diff
@@ -435,6 +435,17 @@ init python in project:
 
             if self.projects_directory is not None:
                 self.scan_directory(self.projects_directory)
+                
+                # If a file called "projectPaths.txt" exists in the main project directory,
+                # also scan any directories listed there
+                extraPathsFName = os.path.join(self.projects_directory,"projectPaths.txt")
+                if os.path.isfile(extraPathsFName):
+                    extraPathsF = open(extraPathsFName,"r")
```
jsfehler: Use "with open()" instead of open:
https://stackoverflow.com/questions/31334061/file-read-using-open-vs-with-open

launcher/game/project.rpy
```diff
@@ -435,6 +435,17 @@ init python in project:
 
             if self.projects_directory is not None:
                 self.scan_directory(self.projects_directory)
+                
+                # If a file called "projectPaths.txt" exists in the main project directory,
+                # also scan any directories listed there
+                extraPathsFName = os.path.join(self.projects_directory,"projectPaths.txt")
```
jsfehler: I'd recommend using underscores instead of CamelCase to conform to the rest of project.

jsfehler:
Nice idea, but why use a hard-coded .txt file? That doesn't conform to the rest of the Launcher's UX. Seems like an unnecessary vulnerability, too.

Personally, this option would be better served via a user input in the Launcher's preferences.
Store the result from the input in a persistent variable and let Ren'py handle saving it.
Then, display the additional directories as a list with "Remove" buttons next to each of them.

pratomoastan:
i agree with joshua, it would better if user could input it via launcher ,
so it will be much more simpler

ianLeslie2:
Ya I don't disagree with you, I was just trying to save myself some work by keeping it simple > _ >

I'm new to doing pull requests. Should I just go ahead and make the requested changes or should I wait for more input first?

pratomoastan:
so normal people (non advance) people could do it with ease

pratomoastan:
maybe you should wait for Tom feedback
because he's the project leader

renpytom:
I like this, at least as a first version of this feature. It seems something like this, in which projects are listed in a file, is something that it makes sense to add. I wouldn't want to add this as a new feature in the launcher, as we're getting very close to release, and a new interface would add a bunch of new translations. 

What I'd really like is a file projects.txt, in the projects directory, where each non-blank line is a path to a project that can be added to the launcher as if it was in the projects directory. That feels like the right level of granularity to have. 

A GUI feature is nice, but it can be added later if this proves popular. And the GUI would have to store its data somewhere - might as well use projects.txt. 

The comments on under_scores vs camelCase are relevant. I'd also suggest using filename or fn rather than FName. It probably also makes sense to just open a file, rather than using isfile - the file could be a symlink.

ianLeslie2:
The latest commit should incorporate the suggested changes,
I did some minor refactoring which resulted in a lot of changed lines, but the actual changes are quite minor.

renpytom:
Looks good. Merging now. Thank you.

PR #1277, opened 2017-10-16
Add OS Functions to Other Functions docs

kevinturner:
I have seen some newer visual novels break the 4th wall in interesting ways using these functions, but they aren't documented anywhere for a newer developer to find them.

### DIFF
--- a/sphinx/source/other.rst
+++ b/sphinx/source/other.rst
@@ -34,6 +34,30 @@ Ren'Py Version
 
     A string giving license text that should be included in a game's
     about screen.
+    
+Operating System
+-----------------
+Ren'Py includes functions that return which operating system is currently running the game. While most built-in Ren'Py functions that need OS-specific functionality (like :func:`CommandConsole`) already handle OS-specific behavior for you, these functions can be used to help create your own OS-specific functionality, especially for things like unusual file handling.
+
+.. var:: renpy.windows
+
+    Returns "True" if running on Windows, "False" otherwise.
+    
+.. var:: renpy.macintosh
+
+    Returns "True" if running on macOS, "False" otherwise.
+    
+.. var:: renpy.linux
+
+    Returns "True" if running on any Linux distribution, "False" otherwise.
+    
+.. var:: renpy.android
+
+    Returns "True" if running on an Android platform, "False" otherwise. Be careful not to equate this as a shorthand for mobile devices; not all Android devices will neccesarily use a portable touch-screen interface.
+    
+.. var:: renpy.ios
+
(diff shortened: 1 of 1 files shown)


pratomoastan:
Really ?
Is it really worth it ?
What do you mean by 4th wall?


On Oct 16, 2017 08:10, "Kevin Turner" <notifications@github.com> wrote:

I have seen some newer visual novels break the 4th wall in interesting ways
using these functions, but they aren't documented anywhere for a newer
developer to find them.
------------------------------
You can view, comment on, or merge this pull request online at:

  https://github.com/renpy/renpy/pull/1277
Commit Summary

   - Update "Other Functions" with OS functions
   - Revert "Update "Other Functions" with OS functions"
   - Update "Other Functions" with OS functions

File Changes

   - *M* sphinx/source/other.rst
   <https://github.com/renpy/renpy/pull/1277/files#diff-0> (24)

Patch Links:

   - https://github.com/renpy/renpy/pull/1277.patch
   - https://github.com/renpy/renpy/pull/1277.diff

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/renpy/renpy/pull/1277>, or mute the thread

kevinturner:
"4th wall" refers to the idea that the screen that the player is looking through is the 4th wall of the stage/game, with the other three walls being inside the game. In this case, I mean visual novels that explicitly reference the fact that they are running on a computer or are a video game.

To avoid spoiling too much about the game, the main inspiration for this commit was Doki Doki Literature Club, which uses this to tell the user about its file system in platform-specific ways, which is needed to progress further into the story. 

Additionally, you could use this in a more advanced way than even Doki Doki Literature Club uses it. For instance, at one point, it may display a Blue Screen of Death to fake out the user into thinking their computer crashed, although it's always the Windows version as far as I know. Using `renpy.macintosh` would allow you to use the macOS kernel panic screen when running on a Mac to keep it consistent.

I don't expect it to be very useful in a lot of contexts, but it's one of those things where even knowing that it exists can get you thinking outside of the box.

pratomoastan:
ooh i seee thats interesting ...

Pada tanggal 16 Okt 2017 08.40, "Kevin Turner" <notifications@github.com>
menulis:

renpytom:
I'm going to merge this, then rewrite it. I like the idea of it, but they're variables and not functions. Also, what's CommandConsole - either I'm brain-farting right now, or it's not actually a Ren'Py thing.

kevinturner:
CommandConsole is what happens when I accidentally flip "ConsoleCommand" in my mind. It also happens to be another undocumented feature that I have only seen in change logs, but I haven't figured out how it works so I can't really write about it.

[This is the source code for it](https://github.com/renpy/renpy/blob/b192a9faf0582b21603cad66bd6827053b19ae81/launcher/game/consolecommand.rpy)

PR #1281, opened 2017-10-28
Let's translate gui.nvl_list_length to Russian

Ria-kon:
https://github.com/renpy/renpy/issues/1262


renpytom:
Let's put a fork in 6.99.13 - it's done. Thanks for the translation.

PR #1282, opened 2017-10-29
[Py3] Make exceptions more compatible with Python3

shaform:
I am trying to speed up the progress of Python 3 porting by introducing small PRs starting from this one.
See #1100 #1098 #1101.


shaform:
I guess this could be closed due to #1596.

PR #1285, opened 2017-11-03
[Py3] fix filter & map

shaform:
I am trying to speed up the progress of Python 3 porting by introducing small PRs.
See #1100 #1098.

### DIFF
--- a/launcher/game/EasyDialogsWin.py
+++ b/launcher/game/EasyDialogsWin.py
@@ -673,10 +673,10 @@ def AskFileForOpen(
     ofn.lpfnHook = LPOFNHOOKPROC(hookProc)
 
     if fn(ctypes.byref(ofn)):
-        filenames = [fn for fn in filename.split('\0') if fn]
+        filenames = [pn for pn in filename.split('\0') if pn]
         if len(filenames) > 1:
             dir, filenames = filenames[0], filenames[1:]
-            return [os.path.join(dir, fn) for fn in filenames]
+            return [os.path.join(dir, pn) for pn in filenames]
         elif multiple:
             return filenames
         else:
--- a/launcher/game/EasyDialogsWin.py
+++ b/launcher/game/EasyDialogsWin.py
@@ -607,7 +607,7 @@ def AskFileForOpen(
     ofn.lpstrTitle = windowTitle
     ofn.lpstrInitialDir = defaultLocation
 
-    if typeList and filter(None, typeList):
+    if typeList and any(typeList):
         lpstrFilter = ''
         for typeSpec in typeList:
             try:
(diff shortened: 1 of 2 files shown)


launcher/game/EasyDialogsWin.py
```diff
@@ -607,7 +607,7 @@ def AskFileForOpen(
     ofn.lpstrTitle = windowTitle
     ofn.lpstrInitialDir = defaultLocation
 
-    if typeList and filter(None, typeList):
+    if typeList and [t for t in typeList if t]:
```
Uzume: This would probably be more straightforward and easier to understand as:
```python
    if typeList and any(typeList):
```

shaform: @Uzume Thanks! Fixed.

launcher/game/pefile.py
```diff
@@ -824,7 +824,7 @@ def dump(self, indentation=0):
                         except exceptions.ValueError, e:
                             val_str += ' [INVALID TIME]'
                 else:
-                    val_str = ''.join(filter(lambda c:c != '\0', str(val)))
+                    val_str = ''.join([c for c in str(val) if c != '\0'])
```
Uzume: This seems to just be converting to a string and then stripping NULs.
It would probably be more straightforward and easier to understand as:
```python
                    val_str = str(val).replace('\0', '')
```
It might also be faster as it does not have to convert to lists of individual character substrings.

launcher/game/EasyDialogsWin.py
```diff
@@ -673,10 +673,10 @@ def hookProc(hdlg, uiMsg, wParam, lParam):
     ofn.lpfnHook = LPOFNHOOKPROC(hookProc)
 
     if fn(ctypes.byref(ofn)):
-        filenames = filter(None, filename.split('\0'))
+        filenames = [fn for fn in filename.split('\0') if fn]
```
Uzume: I think reusing the identifier `fn` when it is used as a function earlier is confusing.

launcher/game/EasyDialogsWin.py
```diff
@@ -673,10 +673,10 @@ def hookProc(hdlg, uiMsg, wParam, lParam):
     ofn.lpfnHook = LPOFNHOOKPROC(hookProc)
 
     if fn(ctypes.byref(ofn)):
-        filenames = filter(None, filename.split('\0'))
+        filenames = [fn for fn in filename.split('\0') if fn]
         if len(filenames) > 1:
             dir, filenames = filenames[0], filenames[1:]
-            return map(lambda fn: os.path.join(dir, fn), filenames)
+            return [os.path.join(dir, fn) for fn in filenames]
```
Uzume: The same thing here. Reusing the identifier `fn` when it is used as a function earlier is confusing.

Uzume: I realize the original code also reused the same `fn` name but it seems to me renaming it something else like `pn` (for "path name") would be better.

shaform: Thanks. Updated. Not sure if we should change that in this PR though.

launcher/game/EasyDialogsWin.py
shaform: Updated.

PR #1286, opened 2017-11-03
[Py3] Use print function

shaform:
I am trying to speed up the progress of Python 3 porting by introducing small PRs.
See #1100 #1098.

### DIFF
--- a/scripts/style_to_statement.py
+++ b/scripts/style_to_statement.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 # Scans a file for python style assignments, and turns such lines into
 # style statements.
 #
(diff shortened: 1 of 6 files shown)


renpytom:
My plan is to audit these and start accepting them once 6.99.13.1 is out, maybe as early as this weekend.

shaform:
@renpytom Thanks! I've rebased the changes against the current master.

shaform:
@renpytom By the way, which Python 2 versions (2.5?, 2.6?, 2.7) must be supported? I can't seem to find this in the document.
It would be more difficult to make it compatible with both 2 and 3 when earlier versions (2.6, 2.5, etc) must be supported as well.

renpytom:
2.7, and then something towards the latest Python 3.

Uzume:
This seems like an update to #1039 which makes me wonder why these were missed back then.

shaform:
The PR said `Skipped the scripts and sphinx directories, wasn't sure if changing them was required for ren'py.` But considering that Python 2 would be out of support soon, I think that it might be better to update everything.

PR #1302, opened 2017-11-21
Russian The Question (Ren'Py 7 edition)

Ria-kon:
> Ria: So, in new The Question we learned that %Me% and Sylvie made that boy's love game first...

We're used in our translation russian old The Question (https://github.com/renpy/renpy/pull/1212)

Also, is the config.version (Ren'Py 7 Edition) shown by gui.show_name in Main menu not translatable at all?

### DIFF
--- a/the_question/game/screens.rpy
+++ b/the_question/game/screens.rpy
@@ -550,44 +550,44 @@ screen about():
 
             hbox:
                 spacing 15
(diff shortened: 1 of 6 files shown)


renpytom:
Merged, thank you. I'll go ahead and add the translation picker from the tutorial to TQ.

PR #1312, opened 2017-12-03
Fix major text segment layout regression.

elizagamedev:
0777826d introduces some major regressions, including breaking the
behavior of `{w}` and `{p}` tags, hyperlinks, and ruby text. This commit
correctly copies over the necessary information from `splits_from` in
the `Layout` constructor (as far as I can tell), and preserves the value
of `start_segment` in `Layout` objects so that it can be copied
properly.


renpytom:
Nice catch. Thank you so much for the fix.

Issue #1329, opened 2017-12-28
Extending "menu" statement to use custom screens


jsfehler:
I've been using the menu statement in a nested fashion, and I'd like to have the top level and nested menus use different screens.

This can be done with https://www.renpy.org/doc/html/statement_equivalents.html#renpy.display_menu
but it would require a fair of rewriting and force me not to use the menu statement.

Since renpy.display_menu is used internally anyways while parsing, I was thinking of the following as a way to send custom screen names down to it: 
```
menu:
    "oranges":
        "High in Vitamin C."
    "bananas":
        "High in Potassium."
    "crab":
        "Is not a fruit."
         menu:
             use response_screen             
             "Yes it is.":
                 "It's sea fruit, honest."
              "Oh, okay.":
                  "I'll stick to fruit."
```

`use response_screen` would cause `response_screen` to be sent to display_menu as the screen argument. If no argument is provided, the default would be used.

I was thinking of attempting this on my own, I'm just wondering if:
a) Something like this doesn't already exist and I've missed it
b) There's an inherent problem in attempting this, or a better syntax/pattern I should try.

Andykl:
I will add from myself that it is still necessary to give the opportunity to pass arguments to this screen
```
menu:
    use location_menu(pos = _return[1])
    "!":
       pass

```

jsfehler:
Added a proof of concept over at #1331

renpytom:
This has been taken over by menu arguments in 7.1.4.

PR #1331, opened 2017-12-30
Add 'use' keyword to the 'menu' statement

jsfehler:
- Allows screens other than 'choice' to be used for menus

Implements #1329 

It seems to work, but I have no idea if this is the best way, or if it causes bugs somewhere else.

### DIFF
--- a/renpy/ast.py
+++ b/renpy/ast.py
@@ -1439,14 +1439,16 @@ class Menu(Node):
         'items',
         'set',
         'with_',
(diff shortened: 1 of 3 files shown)


renpytom:
Menu arguments in 7.1.4 are a more flexible alternative to this.

lunarcloud:
Can we get a link to the documentation or commit with this "menu arguments" feature?

PR #1369, opened 2018-01-29
Russian 6.99.14.1

Ria-kon:
Not so much, huh?


renpytom:
Merging this in now, but I added one more translation, right on the first page of the launcher. I'll hold off on release until you get it in.

PR #1373, opened 2018-02-01
drag & drop change

Enerccio:
Added drag&drop functionality that uses mouse exact position instead of overlap when deciding drop drags.

Example script.rpy for test included
[script.rpy.txt](https://github.com/renpy/renpy/files/1687064/script.rpy.txt)


renpytom:
On dragdrop.py, line 846, shouldn't that be 

x < (c.x + c.w) 

rather than <=? Same thing for y and h.

Enerccio:
Updated with new keyword name and fixed `<=` to `<`

New test:
[script.rpy.txt](https://github.com/renpy/renpy/files/1687244/script.rpy.txt)

renpytom:
Looks good. Thank you.

PR #1388, opened 2018-02-19
Updated URLs to use https when possible. (Readme file only)

mbrickn:
I updated urls to use https when possible to avoid redirects.


renpytom:
Merged, thank you.

PR #1389, opened 2018-02-20
Update 00console.rpy

wap3:


vollschauer:
I don't see any sense to enable the console as default, why you wanna do this?

renpytom:
No chance of this getting merged.

PR #1395, opened 2018-03-04
Allow specifying the baseline for an SFont

eevee:
A `self.baseline` was already implemented for `ImageFont` and even loaded in `BMFont`, but never used.  Now it's used and exposed to `register_sfont`, which is useful for mixing an sfont with descenders with some other font.

Actual motivating example: I wanted to inject a few colorful custom glyphs, so I put them in a quick sfont and used it alongside my (truetype) dialogue font.  Unfortunately, they sit right on the baseline, which shoves the ascent higher and changes the height of the whole line:

before: emoji-style glyph changes line height

I could shrink them a bit, of course, but they were really designed to fill the entire line height.  With this patch and `baseline=-2`:

after: glyphs dip below baseline as intended

### DIFF
--- a/renpy/text/font.py
+++ b/renpy/text/font.py
@@ -64,7 +64,7 @@ class ImageFont(object):
             g = textsupport.Glyph()  # @UndefinedVariable
 
             g.character = ord(c)
-            g.ascent = self.height
+            g.ascent = self.baseline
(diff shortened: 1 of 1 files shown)


renpytom:
Merged, thank you.

PR #1396, opened 2018-03-04
call_screen() shouldn't pass along _with_none

eevee:
Previously, `_with_none` was still in `kwargs` when the screen was shown, so the call would explode unless your screen also happened to have a `_with_none` argument.

Incidentally, [the documentation suggests](https://www.renpy.org/doc/html/screens.html#call-screen) that a `with` after a `call screen` should _just work_, but it doesn't if `config.implicit_with_none` is true — which it is by default!  Not sure if the docs ought to point this out, or if it's planned for `_with_none` to be added to the call automatically someday...?

### DIFF
--- a/renpy/exports.py
+++ b/renpy/exports.py
@@ -2647,6 +2647,11 @@ def call_screen(_screen_name, *args, **kwargs):
 
     renpy.exports.mode('screen')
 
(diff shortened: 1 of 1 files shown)


renpytom:
Merged, thank you.

PR #1406, opened 2018-03-12
Allow interpolating to any prop-only ATL transform

eevee:
It's possible to use an ATL transform as an interpolation target, but only if all the properties are on the same line:

```
transform truecenter:
    xcenter 0.5 ycenter 0.5

transform tocenter:
    linear 1.0 truecenter
```

Now you can do this too:

```
transform truecenter:
    xcenter 0.5
    ycenter 0.5

transform tocenter:
    linear 1.0 truecenter
```


renpytom:
I like this change. Merged.

PR #1407, opened 2018-03-12
Fix typo in "renpy.get_attributes" doc

kevinturner:
Saying that this returns "None" when an image is showing instead of when it isn't makes this function incredibly confusing at first sight.


renpytom:
Yup. Thanks, merged.

PR #1416, opened 2018-03-27
Fix wrong predict in DynamicDisplayable

Andykl:


renpytom:
Yep. Good fix. Thanks for the catch.

PR #1423, opened 2018-04-02
Translate comments from templates/portuguese/options.rpy from spanish/english to portuguese

el-unicorn:


renpytom:
Merged, but note that this will only translate the old options.rpy, which is all but obsolete now.

PR #1443, opened 2018-04-28
LiveComposite replaced by Composite in tutorial and its translations.

Alexandre-T:
In this small pull request, I replaced Live Composite by Composite in tutorial and its translation, according to your [changelog](https://www.renpy.org/doc/html/changelog.html?highlight=changelog#renpy-6-99-14-3) in version 6-99-14-3.


renpytom:
Nice catch, thank you.

PR #1463, opened 2018-05-18
Additional documentation corrections

AXYPB:
The majority of documentation pages have been edited. Please verify that all names for functions, variables, properties, etc. are to be marked with two backticks rather than one. In this case, further editing passes will be required.


renpytom:
Merged, thank you so much.

PR #1495, opened 2018-06-30
Compress launcher's images

jsfehler:
Used https://tinypng.com/ to compress the images.


renpytom:
I used pngcrush on the images in 574c0d9c  . I'm not planning on merging anything that involves passing images through a website that quantize the image to reduce quality. If I wanted to do that, I'd just use webp, but I don't want to do that.

PR #1520, opened 2018-07-19
Typo corrections in documentation

AXYPB:


sphinx/source/text.rst
```diff
@@ -685,4 +685,4 @@ necessary.
 3. Run the game.
 
 Whenever text is displayed that overflows the available area, Ren'Py
-will log an error to the ``text_overflow.txt`` file.
+will log an error to the text_overflow.txt file.
```
Uzume: What is wrong with this? Inline literals are denoted by double backtick. They are supported Sphinx too.

AXYPB: Other file names are not formatted this way throughout the documentation. This is a change for consistency.

Uzume: It seems to me, the better solution to consistency is to format all filenames (and any other literal values) this way rather than remove this.

renpytom:
Thanks again.

PR #1524, opened 2018-07-24
game_menu() no longer ignores screen parameter.

Saltome:
The screen argument needs to be passed as a keyword.


renpytom:
Merged, thank you.

PR #1550, opened 2018-08-23
Update Korean translations for Ren'py 7

sbrn10:
Old translations are missing a bunch of new content around the new GUI.


renpytom:
Merged, thank you. There are a few new translations that are in 7.1, related to the Android support.

PR #1597, opened 2018-10-19
[Py3] Use print function in pure python files

wchill:
Use the print function in pure Python files.

`launcher/game/distribute.rpy` and `renpy/common/_developer/developer.rpym` both also use the old python2 print statements, but I'm not sure how they get turned into Python code and `from __future__ import print_function` has to be declared at the top of a Python file for it to work properly. So some guidance would be helpful for those two files.

### DIFF
--- a/launcher/game/change_icon.py
+++ b/launcher/game/change_icon.py
@@ -317,7 +317,7 @@ def change_icons(oldexe, icofn):
 
     alignment = pe.OPTIONAL_HEADER.SectionAlignment
 
-    # print "Alignment is", alignment
+    # print("Alignment is", alignment)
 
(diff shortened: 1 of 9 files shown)


wchill:
I also edited print statements I found in comments to prevent confusion if someone decides to uncomment them for whatever reason.

renpytom:
You should be able to just convert them to use print(). Ren'Py tries compiling each print block both ways. First by compiling it with print_function, and then doing it again without if that fails. So you should be able to just convert the .rpy and .rpym files, so long as you do it a file at a time.

wchill:
Sounds good. I've fixed those two files to use the new syntax.

PR #1599, opened 2018-10-21
translation: Translated more strings to German, fixed ellipses.

fblampe:
I noticed that a few strings are not translated to German, especially in the main launcher menu screens, so I added those that I found.

I also replaced the ellipsis characters with three regular dots as in the original english strings.

I'm a native German speaker, so I think this should be good enough :)


renpytom:
Merged, thank you.

Issue #1598, opened 2018-10-20
Report text tag errors more properly.


renpytom:
At  least to the level of the  tag in question,.

fblampe:
Hi, I'd be interested to work on this, but I'm not quite sure what you're referring to.

The messages seem already quite detailed to me (see below). The file and line number as well as the problematic tag are all given, so I'm not sure what could be added..?

```
While running game code:
  File "game/script.rpy", line 21, in script
    "Professor {cps/=2}Eileen's{/cps} lectures are usually interesting, but today I just couldn't concentrate on it."
Exception: Unknown text tag u'cps/=2'
```

PR #1601, opened 2018-10-23
Raise exception if tag value is empty

ShehriyarQureshi:
Concerning #1598 

Using empty value for tag caused the game to exit displaying
Python's exception.

e.g. {size=} resulted in IndexError.

Added a check to see if value isn't empty. If so, give a general
exception telling the user to see if they supplied appropriate
value.

I don't know if all the tags checked below this addition require a value. I appreciate modifications to it.

### DIFF
--- a/renpy/text/text.py
+++ b/renpy/text/text.py
@@ -875,212 +875,218 @@ class Layout(object):
 
         for type, text in tokens:  # @ReservedAssignment
 
-            if type == PARAGRAPH:
+            try:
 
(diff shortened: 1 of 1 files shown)


renpytom:
Rejected. There are plenty of text tags that take empty values ({i}, for example.)

ShehriyarQureshi:
My bad. Should've thoroughly looked all of them. I'll correct it.

PR #1602, opened 2018-10-24
Add checks for tags that require value

ShehriyarQureshi:
Some tags that require value, when supplied none, raise python's
exceptions which might be hard to interpret.

Added simple checks to those tags to see if the supplied value is
empty. Raises exception that is easier to read than the ones
Python raises.

To a person who doesn't know programming, working on a large project might get them confused if they leave a tag empty and see Python's exceptions which might make this simple problem hard to understand.

It's for #1598

### DIFF
--- a/renpy/text/text.py
+++ b/renpy/text/text.py
@@ -924,10 +924,18 @@ class Layout(object):
                     line = [ ]
 
                 elif tag == "space":
+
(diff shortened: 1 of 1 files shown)


renpytom:
Merged, thanks.

PR #1611, opened 2018-10-29
Add missing history_allow_tags set to tutorial and The question.

KosMos-pv:
This will fix crash when trying to open history in these two example projects.


renpytom:
Thank you, even if I have some egg on my face for this.

PR #1616, opened 2018-11-02
Add first test using pytest

ShehriyarQureshi:
The test checks a small function in main.py called
path_to_renpy_base. It is not a perfect test but I guess it's
better than having none.


ShehriyarQureshi:
This test needs to be improved. There are functions that check for folders inside the renpy directory. We should see if we can get the test to mimic as if renpy is being run.

ShehriyarQureshi:
It's still wrong. I'll work on it when I get home.

ShehriyarQureshi:
Does this look good for a start?

tests/test_main.py
```diff
@@ -0,0 +1,31 @@
...
+class TestRenpyPath:
+    def test_renpy_path_to_renpy_base(self):
+
+        # We're just doing what the function does and comparing the values.
+        # This test should fail if the function is changed and returns
+        # something else.
+
+        # we change sys.argv to mimimc as if renpy is being run
+        sys.argv.pop(1)
+        sys.argv[0] = 'renpy.py'
```
jsfehler: Modifying sys.argv doesn't seem relevant here. If you modify sys.argv[0] to equal what you want, then you're not actually testing that main.path_to_renpy_base equals the result of sys.argv[0]. You're only testing that main.path_to_renpy_base equals what you think it should be.

tests/test_main.py
```diff
@@ -0,0 +1,31 @@
...
+
+        renpy_base = main.path_to_renpy_base()
+
+        # we use sys.path[0] as it is the expected result of function
+        assert renpy_base == sys.path[0]
+
+    def test_path_to_common(self):
+        assert main.path_to_common(sys.path[0]) == "{}/renpy/common".format(
```
jsfehler: You should use yoda conditions in this case.

`assert immutable == mutable` is safer than `assert mutable == immutable`, because it prevents typos along the lines of
`assert mutable = immutable`

jsfehler:
I don't know what the  "tags" file is for, but it clearly doesn't belong in this PR.

tests/test_main.py
jsfehler: Is this test relying on the previous testing having changed sys.path[0]? If so:
- Tests must be able to run in isolation.
- Like the previous test, it's using a modified sys.argv for no benefit.

tests/test_main.py
```diff
@@ -0,0 +1,31 @@
...
+
+# add renpy directory to path else main module isn't found
+current_directory = os.getcwd()
+sys.path.insert(0, current_directory)
+
+# thou shalt not move this artifact unless thou can without errors
+import main
+
+
+class TestRenpyPath:
```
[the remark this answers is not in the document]
ShehriyarQureshi: I think I followed this style from python3. pytest needed a class in version 3 if I'm not wrong. I'll remove the class and just use functions.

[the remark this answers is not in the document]
jsfehler: pytest doesn't need classes in any version.

Using classes is a holdover from the unittest style of writing setup/teardown methods. This is largely unnecessary in pytest due to the fixture system.

tests/test_main.py
ShehriyarQureshi: I don't think I get you. The second function is using `sys.path[0]`, not `sys.argv`. Is there a relation b/w `sys.argv` and `sys.path[0]` because `sys.path[0]` returns path to directory where the interpreter was called, while `sys.argv` is a list of command line arguments.

Upon testing `sys.path[0]` I realized the pytest command needs to be run in the renpy folder and not anywhere else because then `import main` throws `ImportError: no module named main` exception.

jsfehler: I might have blinked and seen argv instead of path by mistake. I'll give this another look in a bit.

tests/test_main.py
ShehriyarQureshi: There are two things.

- We can leave `sys.argv` and check what the function does, which should be something similar i.e. it should give us a path to (folder where pytest executable is, in my case) but there is a problem with this approach:

    - There are functions that check for other folders inside of `renpy_base`. Creating tests for those functions will fail as we wouldn't be looking into the renpy directory.

The logical thing to do here in my opinion is to replicate normal renpy execution. That way other functions should work as intended and can be tested properly.

What I did here is not the best but it should do the trick for now. `sys.argv[0]` is replicating that `renpy.py` is being called from renpy directory.

There should be a better way to do this.

tests/test_main.py
ShehriyarQureshi: Reversed parts of assert statement.

tests/test_main.py
ShehriyarQureshi: The other functions aren't really **checking** if the folders were inside renpy directory. They are just appending the folders string to the path where script resides.

i.e. running renpy from renpy directory we get
`sys.argv = ['renpy.py']`
suppose path to `renpy.py` is `a/b/c/renpy.py`
the function returns path where script is placed, so it returns `a/b/c`

I updated the implementation to use whatever sys.argv we get (_which will be path to pytest_). This way:

`sys.argv[0] = 'x/y/z/pytest`
so the function should return `'x/y/z'`. We use `os.path.dirname` to get this path and use it
as our known output value.

This way we can just append the folders to `'x/y/z'.

i.e. `path_to_common` when supplied above path should return `'x/y/z/renpy/common'` and we can just append `/renpy/common` to above path and it will serve as our known output value. The test is going to fail if the function is changed in a way that it returns a different value.

tests/test_main.py
ShehriyarQureshi: Removed the use of sys.path[0] as it is not constant.

ShehriyarQureshi:
@renpytom what's your input on this?

renpytom:
I'd like to see tests of non-trivial functions before committing to merging a testing framework. Right now, this just seems like something super-trivial. I'd like to know how something like, for example, renpy.exports.display_menu could be tested.

ShehriyarQureshi:
Alright. I'll see what I can do about it.

ShehriyarQureshi:
I'll create a new pull request if I can get tests for `renpy.exports.display_menu` or similar functions working.

PR #1621, opened 2018-11-04
Add min argument to FilePagePrevious

jsfehler:
Adds a min argument to FilePagePrevious. This is to support scenarios where Auto and Quick Save are enabled, but the UI wants to organize them separately from manual saves.


renpytom:
I like the idea of it, but I think the min argument isn't the way to do it. I replaced it with a pair of flags, one for auto and one for quick, and added them to FilePageNext as well.

Thanks for the PR.

PR #1625, opened 2018-11-07
Fixes to French translation

Beuc:
I couldn't find the central point for the translations though. These strings also appear in "Tutorial" and to some extent in "Launcher". Should I replicate the changes there as well?

### DIFF
--- a/the_question/game/tl/french/screens.rpy
+++ b/the_question/game/tl/french/screens.rpy
@@ -187,7 +187,7 @@ translate french strings:
 
     # screens.rpy:782
     old "Rollback Side"
-    new "Rollback"
+    new "Rembobinage côté"
 
     # screens.rpy:783
     #old "Disable"
@@ -215,7 +215,7 @@ translate french strings:
 
     # screens.rpy:800
     old "Language"
-    new "Langage"
+    new "Langue"
 
     # screens.rpy:815
     old "Text Speed"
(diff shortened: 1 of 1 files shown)


renpytom:
Merged. Yes, the changes have to be replicated 3 times.

Alexandre-T:
Le mot rembobinage est un bien meilleur choix que ce que j'avais proposé. Je pense que tu peux le répliquer. Je pense préférable d'utiliser des mots-clés comme : Sens, Rembobinage et Avance rapide. L'analogie avec les lecteurs vidéo est excellente. En revanche, «Rembobinage côté» n'est pas français. «Rembobinage» suffirait. À la rigueur «Sens du rembobinage». Dans tous les cas, c'est un meilleur choix que "Rollback"

The "Rembobinage" word is a better word than "Rollback". Some word like "Sens, Rembobinage, Avance rapide" could be used. But, "Rembobinage côté" isn't french. "Remboninage" is enough. "Sens du rembobinage" could be accepted. But "Rembobinage" is still a better choice than "Rollback".

Beuc:
Hi Alexandre and thanks for your input,

"Sens" (way/direction) wouldn't be accurate because we're talking about which side ("côté") of the screen becomes active to perform a rollback - useful on mobile devices with no mousewheel.
In both cases this is a rollback, not a fast-forward (no "avance rapide").

"Rembobinage côté" is a compromise between accuracy and space constraints. It's used in the preferences screen that shows this block:
```
Rembobinage côté:
- Désactivé
- Gauche
- Droit
```

The full sentence (prompt + chosen value) becomes e.g. "Rembobinage coté droit" which is French.
This is not perfect but this manages to convey that "side" part, which is important to understand what this option is about.
What do you think?

renpytom:
Please make sure I find out (probably via a new Pull Request) if this needs to be changed.

Beuc:
Sure.
If there's no objection I'll replicate the patch for tutorial/launcher in a new pull request this week-end.

Alexandre-T:
Your solution seems to be the best one.

Beuc:
Done, see #1630 :)

PR #1631, opened 2018-11-10
emscripten: signal when data is written (so it persists)

Beuc:
The virtual filesystem is transient until an asynchronous sync message is signaled.
This patch does it when the user saves their game.

### DIFF
--- a/renpy/savelocation.py
+++ b/renpy/savelocation.py
@@ -36,6 +36,9 @@ import threading
 from renpy.loadsave import clear_slot, safe_rename
 import shutil
 
+if renpy.emscripten:
+    import emscripten
+
 disk_lock = threading.RLock()
 
 # A suffix used to disambguate temporary files being written by multiple
@@ -147,6 +150,9 @@ class FileLocation(object):
 
         with disk_lock:
             record.write_file(filename)
+
+        if renpy.emscripten:
+            emscripten.syncfs()
 
         self.scan()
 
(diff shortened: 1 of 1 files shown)


renpytom:
I'm holding off until 7.1.2 is out to merge this. 

That being said, I'm wondering if it might make sense to deal with this at some sort of lower level - like through a patch to Python. It just feels really wrong to annotate Ren'Py with code that just brings things into compliance with the way Python is documented to work, you know? 

Can you call the emscripten stuff from C? It should be easy to check to see if a Python file object is read-write, and call if from there.

Beuc:
Hi,

This is necessary to synchronize the `~/.renpy/` mount point only. The root filesystem is in transient.
Reference for that function is at https://kripken.github.io/emscripten-site/docs/api_reference/Filesystem-API.html
It would not make sense to call syncfs e.g. everytime a line is written to log.txt.
Also architecturally, doing syncs purely at the Python level means we can miss syncs in C modules.

The `emscripten` module is my C wrapper:
https://www.beuc.net/python-emscripten/python/artifact/a308c76692f5c513
https://github.com/python-emscripten/python/blob/trunk/emscripten.pyx
so everything is called from C.

renpytom:
I rewrote this in 372216e, to route all of this through a single method.

Beuc:
I had made a few changes since then to avoid doing many syncs during autosave (which does around 10 renames in a row) - i.e. no sync on rename, and sync at the end of autosave_thread.
This doesn't fit defining sync in FileLocation; I'll probably write another patch unless you want to proceed differently.

PR #1639, opened 2018-11-15
Explicitely close audio on quit

Beuc:
With Emscripten, this tells the browser to stop calling the audio callback.

PR #1653, opened 2018-11-25
Explicitely close audio on quit

Beuc:
With Emscripten, this tells the browser to stop calling the audio callback.
(resubmission of #1639 since it was reverted and since #1642 fixed the issue)


renpytom:
And merged again. Thanks.

PR #1655, opened 2018-11-26
When providing a custom function to the transform, make it sure that …

Meithal:
…the default function is not called.

That allows us to keep the same instance of a Transform during the lifetime of a CDD.


Meithal:
If you want to write the example user defined displayable like that 

```renpy
init python:

    import math

    class Appearing(renpy.Displayable):

        def __init__(self, child, opaque_distance, transparent_distance, **kwargs):

            super(Appearing, self).__init__(**kwargs)

            self.child = renpy.displayable(child)

            self.opaque_distance = opaque_distance
            self.transparent_distance = transparent_distance

            self.alpha = 0.0

            self.width = 0
            self.height = 0

            self.tr = Transform(child=self.child, alpha=self.alpha, function=self.control)

        def control(self, tr, st, at):
            tr.alpha = self.alpha

            child_render = renpy.render(self.tr, width, height, st, at)

            self.width, self.height = child_render.get_size()

            render = renpy.Render(self.width, self.height)

            render.blit(child_render, (0, 0))

            return render

        def event(self, ev, x, y, st):

            distance = math.hypot(x - (self.width / 2), y - (self.height / 2))

            if distance <= self.opaque_distance:
                alpha = 1.0
            elif distance >= self.transparent_distance:
                alpha = 0.0
            else:
                alpha = 1.0 - 1.0 * (distance - self.opaque_distance) / (self.transparent_distance - self.opaque_distance)

            if alpha != self.alpha:
                self.alpha = alpha
                renpy.redraw(self, 0)

            return self.child.event(ev, x, y, st)

        def visit(self):
            return [ self.child ]
```

It is currently not possible because the default function will be called after our own function is, thus we have to reinstantiate a new transform at every render pass, that pr makes the code above possible, maybe it can be done better because the default function seems to do more that just clamp the transform to its initial values so maybe that will break things,

renpytom:
I'm going to close this, per above. I don't think that having the ability to set the a property in both the default and the function is something we want, and I can't see a downside apart from that.

PR #1695, opened 2019-01-22
shortcut for change_language

pionere:
change_language is called often even if there was no real language change -> better to shortcut it as early as possible
also block_rollback was never called due to the setting of old_language earlier


renpytom:
Merged, finally. Thanks.

PR #1711, opened 2019-02-09
Option to set fadein for main menu music.

KosMos-pv:
Also added this option to template for new projects and edited documentation. I think it need to be edited as my English is not good enough.

### DIFF
--- a/gui/game/options.rpy
+++ b/gui/game/options.rpy
@@ -67,6 +67,12 @@ define config.has_voice = True
 # define config.main_menu_music = "main-menu-theme.ogg"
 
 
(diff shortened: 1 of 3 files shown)


KosMos-pv:
Looks like I really hurried with the creation of this pull request. I tried to add fadein for music in the main menu, but I got an error when I used the method suggested on the forum. But, as it turned out, I was just inattentive and could not understand the reason for my mistake. In the end, I decided to modify the engine files and created a new option to customize the game. If you think that adding this setting is optional, you can reject my pull request. Sorry for disturbing.

KosMos-pv:
Well, it was a very stupid idea to push this. Closing this request...

Issue #1761, opened 2019-03-07
Unexpected temporary attribute transition behaviour


mal:
Using `with dissolve` on a say line with a temporary attribute only affects the initial change, not the restoration. Conversely using `with dissolve` alone on a line following a say line with a temporary attribute only affects the restorative change.

In addition to this being strange, it also differs from behaviour of normal say lines which are un-impacted by `with dissolve` statements on separate lines.

```renpy
image r = Solid('f773', xysize=(320, 240))
image b = Solid('7ff3', xysize=(320, 240))
image y = Solid('ff73', xysize=(320, 240))

define foo = Character('Foo', image='square')

layeredimage square:
  group colour:
    attribute red default "r"
    attribute blue "b"
    attribute yellow "y"


label main_menu:
    $ _confirm_quit = False
    return


label start:
    scene expression '#111'
    show square at truecenter
    foo "start as red"
    foo @ blue "dissolve temporarily into blue" with dissolve
    foo "restore red (should dissolve back, but doesn't)"
    foo @ yellow "dissolve temporarily into yellow" with dissolve
    with dissolve
    foo "restore red (this time it {b}did{/b} dissolve back)"
    foo @ blue "temporarily switch to blue without dissolve"
    with dissolve
    foo "restore red (with dissolve once again)"

    foo yellow "this will dissolve to yellow" with dissolve
    with dissolve # this will act as a weird delay
    foo blue "this won't dissolve to blue"
    with dissolve # this will act as a weird delay

    "end"
    return
```

renpytom:
I don't believe there's a problem here. The ``with`` clause for a say statement isn't meant as a shorthand for setting config.say_attribute_transition, which it kind of looks like you might think it is. Instead, it's shorthand for 

```
$ renpy.transition(dissolve)
"Say statement."
```

And as far as I can tell, it's doing that job flawlessly. Right now, there isn't syntax for setting a say_attribute_transition on a say statement basis, though say arguments and an appropriate function could work together to make this work. (A dedicated callback would be cleaner, though - I'd be happy to add that if you want.) 

I don't see this as a bug, however.

mal:
Ah, OK that makes sense. The confusion came from seeing the transition apply when images were changing as a result of temporary attributes being added, but not they were changing due to them being reset.

[...]

PR #1774, opened 2019-03-17
Extend say_attribute_transition_callback with canonical tag list

mal:
Follows from discussion and concept in https://github.com/renpy/renpy/issues/1761#issuecomment-473678693

This provides context to the callback about the incoming image it's affecting with the transition. This can be used in isolation, or with the current state to intelligently utilise a transition based upon the tags/attributes of the image being displayed.

---

For example, this would allow transitions to be selected based on which attributes are/will be changing:

```renpy
init python hide:
    dissolve_attrs = {'running', 'walking', 'jumping', 'skipping'}

    def example(tag, attrs, mode):
        current_attrs = _set(renpy.get_attributes(tag))
        future_attrs = _set(attrs)
        future_attrs.remove(tag)
        # calc attrs that are actually changing, not just those passed
        changed_attrs = current_attrs ^ future_attrs 
        if dissolve_attrs & changed_attrs:
            return dissolve, config.say_attribute_transition_layer
        return config.say_attribute_transition, config.say_attribute_transition_layer

    config.say_attribute_transition_callback = example
```

### DIFF
--- a/renpy/character.py
+++ b/renpy/character.py
@@ -896,7 +896,7 @@ class ADVCharacter(object):
                 images.predict_show(new_image)
 
             else:
-                trans, layer = renpy.config.say_attribute_transition_callback(self.image_tag, mode)
+                trans, layer = renpy.config.say_attribute_transition_callback(self.image_tag, new_image, mode)
 
                 if not skip_trans:
                     if (trans is not None) and (layer is not None):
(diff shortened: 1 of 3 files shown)


mal:
@renpytom any thoughts or feedback on this?

renpytom:
Ah, shoot - missed this one for 7.2.2. It'll go into the next release, in some form.

renpytom:
Just merged it, and then added some compatibility for existing games.

mal:
Sweet! Thanks Tom! 🎉

PR #1781, opened 2019-03-21
Perform re-create Lexer.eob on revert

Andykl:
This allows to do revert correctly even if the lexer has been exhausted before.


renpytom:
Merged, thanks.

PR #1785, opened 2019-03-29
Add ability to parse kerning messages of BMFont fonts

kobaltcore:
This allows the BMFont object to understand "kerning" lines embedded within BMFont definitions whereas previously they would be ignored.
The kerning information is added to the existing "self.kerns" property.

I suspect this should have originally been the case as well, given that the entire kerning support system is already in place and kerning support is mentioned in the documentation for BMFonts.


renpytom:
Merged. I don't think image fonts are all that useful anymore, but if we're going to support a format, might as well support it as best we can.

Issue #1800, opened 2019-04-22
Reload doesn't reload py modules


mal:
Seems that they're not reloaded when changed even when triggered manually.

**Reproduce**

- Start game
  - see `foo` printed ✔️ 
- Shift+R
  - autoreload on
  - see `foo` printed again ✔️ 
- Edit `game/mylib/__init__.rpy` to `test = 'bar'`
  - _No reload triggered_ ❌ 
- Shift+R twice (toggle autoreload off and back on, to force a reload)
  - see `foo` printed again ❌ _should have been `bar`_
  - `mylib` not reloaded

**Test Case**
<details><summary>Tiny example game</summary>

```
game
├── mylib
│   └── __init__.py
└── script.rpy
```

**game/mylib/__init__.rpy**
```
test = 'foo'
```

**game/script.rpy**
```
init python:
    from mylib import test
    print(test)

label start:
    $ renpy.pause()
```
</details>

**Proposal**

A new config var for a list of module names that should participate in reloads. I'm thinking the condition for `del`ing them from `sys.modules` should be something like `any(i == m or i.startswith(m + '.') for m in config_var)` such that the reload will apply to the listed modules and any submodules. I'll try and PoC it tomorrow and submit a PR if it seems sane and solves the problem.

PR #1801, opened 2019-04-23
Python module reloading support

mal:
Fixes https://github.com/renpy/renpy/issues/1800

### DIFF
--- a/renpy/__init__.py
+++ b/renpy/__init__.py
@@ -551,6 +551,10 @@ def post_import():
         vars(renpy.exports).setdefault(k, v)
 
 
+def issubmodule(sub, module):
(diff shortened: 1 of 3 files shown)


renpytom:
This looks quite good. Would you mind adding documentation to sphinx/source/config.rst, in the "Occasionally Used" section? That's pretty much all I need to merge this.

mal:
I tried to keep it informative enough for those looking for it and terse enough to avoid confusing anyone else, but I can elaborate on the expected use case being the development of first-party pure python modules if you feel that would be helpful?

renpytom:
Yep. Works for me. I'll edit the config.rst to clarify it's a list of strings giving modules names - having the basic types is helpful.

PR #1802, opened 2019-04-24
Keep vanilla python's late import binding

mal:
In python it's possible for a module to meddle with `sys.modules` during it's initialisation, this allows it to dynamically redefine itself on-the-fly, it's not common, but it happens. In vanilla python assignments of imported modules take place after the module is exec'd, however in renpy the opposite is true. This early assignment can lead to a case where the imported module does not match the value in `sys.modules` and subsequent `import`s yield a different result. This is shown in the simple (albeit contrived) test case provided, along with output samples pre- and post-patch.

#### Test Case

**game/lib.py**
```py
from sys import modules
from types import ModuleType
mod = modules[__name__] = ModuleType(__name__)
del ModuleType, modules

mod.a = 'hello'
```

**game/script.rpy**
```renpy
init python:
    import sys
    import lib
    print('1 ', lib is sys.modules['lib'])
    try:
        print('1a', lib.a)
    except AttributeError as e:
        print('1a', e)
    import lib
    print('2 ', lib is sys.modules['lib'])
    print('2a', lib.a)

label start:
    $ _confirm_quit = False
    $ renpy.pause()
```

**before patch**
```
1  False
1a 'module' object has no attribute 'a'
2  True
2a hello
```

**after patch**
```
1  True
1a hello
2  True
2a hello
```

### DIFF
--- a/renpy/loader.py
+++ b/renpy/loader.py
@@ -744,7 +744,7 @@ class RenpyImporter(object):
         code = compile(source, filename, 'exec', renpy.python.old_compile_flags, 1)
         exec code in mod.__dict__
 
-        return mod
+        return sys.modules[fullname]
 
     def get_data(self, filename):
         return load(filename).read()


renpytom:
I guess it doesn't hurt anything, but I have to ask - how did you find this?

mal:
A train of curiosity really. I found myself wishing that it was possible to augment the `module` API, specifically to be able to provide it a `__getitem__` method. It's possible in the newer versions of py3 via `__class__` assignment (it's readonly in py2) and that eventually lead me to [`metamodule`](https://pypi.org/project/metamodule/), which uses some shady `ctypes` pointer magic to function in cpython2. Both approaches use the dynamic redefinition within the module pattern.

Seeing the pointer stuff got me curious as to whether it'd be compatible with `renpy` builds, which is where I ran into the situation of having imported something, but not what I'd been expecting compared to the same code run in py2. Checking `sys.modules` showed all as expected, so it was just the local assignment that was off. It was at that point I started poking around `loader.py`, and found the module ref bridging over the execution. Then a PoC to ensure it was the dynamic redefinition pattern rather than metamodule causing the issue and here we are.&lt;/life-story&gt; 😉

renpytom:
Interesting. I didn't know about metamodule. I'd be very careful playing about here - Ren'Py is also fairly intimately involved with modules, so changing something Ren'Py uses may break things. 

That being said, merged.

PR #1804, opened 2019-04-29
Python module bytecode

mal:
This adds support for compiling `.py` files in `game/` and `game/site-packages/` (when on disk) to their `.pyo` equivalents to accelerate module loading, and to use those `.pyo` files when present in archives. Unfortunately it's not the most straight forward process. This sort of thing looks much more accessible in py3, but since py2 is well defined, and unlikely to change, I'm hoping this is acceptable (or at least on the road to being so) until py3 rolls around.

For this change it's necessary to know the origin of code (disk, rpa, apk). In order to do this `renpy.loadable` (and the cache that backs it) has been tweaked to return (truthy) `int` constants rather than `True`. An appropriate compat config has been added to maintain the public API for older script versions.

The search order for loadable files is defined thus (`__debug__` is `False` when running with `-O`):
https://github.com/renpy/renpy/blob/3165b2a08c45e105ff55a130e219cad4aea23d6b/renpy/loader.py#L699-L701

If a compiled file is selected, the corresponding `.py` file is looked for on disk, and if found, used to check the compiled file timestamp and if needed recompile it (if missing or mismatched timestamps). If the `.py` is _not_ found on disk, the compiled file found originally is used without further conditioning.

This results in a loading preference hierarchy of:
- `.pyo` files on disk that are in sync with their corresponding `.py` file (also on disk)
- `.py` files on disk (which will compile a `.pyo` for next load)
- `.pyo` files on disk (with no corresponding `.py` file)
- `.pyo` files in rpa or apks
- `.py` files in rpa or apks

I appreciate this isn't a small change and will likely require some time to mull over, so thanks for looking! 🙂

### DIFF
--- a/renpy/common/00compat.rpy
+++ b/renpy/common/00compat.rpy
@@ -184,6 +184,7 @@ init -1900 python:
             config.menu_actions = False
 
         if version <= (7, 2, 2):
+            config.loadable_boolean = True
             config.say_attribute_transition_callback_attrs = False
 
     # The version of Ren'Py this script is intended for, or
--- a/renpy/config.py
+++ b/renpy/config.py
@@ -825,6 +825,9 @@ preload_fonts = [ ]
 # Should Ren'Py process multiple ATL events in a single update?
 atl_multiple_events = True
 
+# Should renpy.loadable return a boolean?
+loadable_boolean = False
+
 # A callback that's called when checking to see if a file is loadable.
 loadable_callback = None
 
--- a/renpy/exports.py
+++ b/renpy/exports.py
@@ -1953,12 +1953,13 @@ def loadable(filename):
     """
     :doc: file
 
-    Returns True if the given filename is loadable, meaning that it
-    can be loaded from the disk or from inside an archive. Returns
-    False if this is not the case.
+    Returns an int representing the origin from which the given filename
+    is loadable, meaning that it can be loaded from the disk, or from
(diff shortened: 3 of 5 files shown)


renpytom:
I'm going to reject this patch, not over anything like code quality, but because I simply don't want to include this in Ren'Py. 

I've been very careful throughout Ren'Py to always include source code for any Python creators give us. (Source for all Python blocks are included in .rpyc files.) This is to make it likely that when we upgrade Python versions, older games have at least a chance of running. 

This goes against that - if we support loading .pyo files, then we'll also support people distributing .pyo files without the corresponding .py source code. That's going to bite a lot of people in the ass from a game preservation and engine evolution perspective. 

I realize this is kind of fast, but this is something I thought about while initially implementing Python loading. These concerns are more important than startup time to me.

renpytom:
By  the way, if you want to discuss a feature like this before putting the work in, I'm very available on Discord (renpytom#9751).

mal:
Ahh ok, makes sense it's already been considered. I was thinking that because the version of python is bundled in the distribution phase, there wouldn't be a chance for the `pyo` files to be incompatible, but in light of the other goals set out, perhaps that doesn't bare out.

Will definitely seek you out on Discord in future if I have crazy ideas, although when I started this I expected it to be relatively small change (classic case of underestimation!) Regardless, learnt a lot about the python import and bytecode systems and the py2/py3 differences so it's far from a bust!

PR #1837, opened 2019-05-24
Update common french translations

Beuc:
Some new common strings (the accessibility menu that pops-up with key 'a' ; some strings from the error screen) are not available in French.
Since these strings can be imported in new games, let's translate them :)
(some other strings are still translated, but that's a more exhaustive job for later)

@maxlefou I think you're the usual translator, do you want to proof-read?

### DIFF
--- a/launcher/game/tl/french/common.rpy
+++ b/launcher/game/tl/french/common.rpy
@@ -207,15 +207,15 @@ translate french strings:
 
     # 00library.rpy:142
     old "Self-voicing disabled."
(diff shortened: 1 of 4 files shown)


renpytom:
Oh, sorry - clicked merge by accident. Please do a second commit if we need proofreading.

maxlefou:
I checked the commit, i didn't see any errors in french. Everything is valid to me, no problem. ^^

PR #1839, opened 2019-05-27
Fix Curry hash exception

### LINKED ISSUE #1840: renpy.curry.Curry.__hash__ exception
```
    def __hash__(self):
        return hash(self.callable) ^ hash(self.args) ^ hash(self.kwargs)
```
kwargs unhashable

Andykl:
This is fix #1840


renpytom:
Already fixed.

PR #1870, opened 2019-06-10
Do not redraw at per_frame, it's already being redraw.

### LINKED ISSUE #1869: Redraw recursion
from https://lemmasoft.renai.us/forums/viewtopic.php?p=513004#p513004
per_frame screens give recursion error.

```
def redraw(d, when):
    if per_frame:
        invalidate(d)
        return

def invalidate(d):
    if not rendering:
        redraw(d, 0)
        return
```

Andykl:
This is fix #1869

### DIFF
--- a/renpy/display/render.pyx
+++ b/renpy/display/render.pyx
@@ -279,17 +279,12 @@ def invalidate(d):
     a redraw to start.
     """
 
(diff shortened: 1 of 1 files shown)


Andykl:
Too hard

PR #1884, opened 2019-06-15
accessibility message changed to translatable string

Moshibit:


renpytom:
Merged. Are you going to follow this up with the translation soon? I'd like to get 7.3 out the door today.

Moshibit:
I just added that last string, and for now it would be everything, I already reviewed the translation three times.

renpytom:
Sure, but we don't actually have the translation yet, as far as I can see. I want to make sure that's on the way, so I know when I can release.

Moshibit:
The translation is ready

PR #1899, opened 2019-06-22
prevent crash when get_selected is called on a non-existent field

pionere:
A field (especially a global Variable) might not exists at the beginning of the interaction. This causes the application to crash e.g. if the variable is set on the 'action' of a button.
The solution is that the get_selected returns False if the given field does not exists at the moment.

### DIFF
--- a/renpy/common/00action_data.rpy
+++ b/renpy/common/00action_data.rpy
@@ -24,19 +24,22 @@ init -1600 python:
    ##########################################################################
     # Functions that set variables or fields.
 
-    __FieldNotFound = object()
+    __FieldNotFound, __FieldValueUndefined = object(), object()
 
-    def __get_field(obj, name, kind):
+    def __get_field(obj, name, kind, value=__FieldValueUndefined):
 
         if not name:
             return obj
 
         rv = obj
 
-        for i in name.split("."):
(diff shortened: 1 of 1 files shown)


renpytom:
I don't think I want to merge this. 

My feeling is that this isn't a crash - it's Ren'Py reporting an error in the game, namely the use of a field or variable that doesn't exist. I don't think it makes sense to cover up something that's a logic error. (And probably punt the problem down the road until the next time the field is used.)

pionere:
Why would it be an error something like this in the code:
   `textbutton "jumper":`
   `   action [SetVariable("button_pressed"), Jump(somewhere)]`
where the button_pressed variable does not exist yet?

Andykl:
Because the error clearly indicates the miss of a variable.
Python is a strong typing language, and I don't think we need depart from it in this case.
You can always write `default button_pressed = None` to prevent exception.

pionere:
What do you mean a miss of a variable?
setattr(xxx, "var", value) is perfectly valid in python even if there is no var variable in the xxx entity at the moment.
the exception can be prevented by setting an explicit "selected" parameter on the button or doing exactly the same thing in a control loop or a function.

PR #1906, opened 2019-06-25
Fix spurious spaces

Moshibit:


renpytom:
Merged this in by hand, since there were too many merge conflicts to do in manually. (And also, one of the translations was missing a close quote.) Could you go ahead and resynchronize your branch with Ren'Py, so the next time we merge, we're on the same page?

As always, thank you!

PR #1932, opened 2019-07-11
CDSL displayables should allow 'if' statements inside self.

Andykl:


renpytom:
Reviewed and merged, thanks.

PR #1942, opened 2019-07-15
Make automatic_images support webp

methanoliver:
I am not sure just why is this feature in 00obsolete.rpy, but not supporting webp when RenPy itself does otherwise sounds like an oversight.

### DIFF
--- a/renpy/common/00obsolete.rpy
+++ b/renpy/common/00obsolete.rpy
@@ -86,12 +86,13 @@ init 1900 python hide:
             if fn.startswith("_"):
                 continue
 
-            # Only .png and .jpg
-            if not fn.lower().endswith(".png") and not fn.lower().endswith(".jpg"):
+            # Only .png, .jpg, .webp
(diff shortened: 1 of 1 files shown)


renpytom:
Since it's obsolete. It's been replaced by the images/ directory, which supports webp. If you want more control, you can use renpy.list_files() to get a list of files, and define them yourself. 

Since this is an obsolete feature, I'm not going to accept a pull request to improve it.

methanoliver:
Images directory doesn't support subdirectories, though. :(

StudioErrilhl:
The images directory does indeed support subdirectories...

methanoliver:
It does, however, only parse individual file names for tags, and directory names are not taken into account.

renpytom:
That's correct. If you'd like to change the rules that Ren'Py uses, you can use renpy.list_files() to iterate over all files, and then renpy.image to define them. This is more flexible than the previous system, which is why the old one has been marked obsolete.

PR #1953, opened 2019-07-20
web: set variants according to what the browser runs on

Beuc:
This proposal detects additional variants by querying the browser.

This allows the same web game to e.g. plays with a mobile UI on mobile, and with a full-blown UI on desktop/"pc". This can be tested with real hardware as well as "Responsive Design Mode" in browsers.

I'm not entirely sure about setting "android" and "ios", maybe some games use it to target OS-specific features not available through the browser.

(tested on 7.3.2.320, master is broken with the on-going GL work AFAICS)

### DIFF
--- a/gui/game/screens.rpy
+++ b/gui/game/screens.rpy
@@ -323,15 +323,15 @@ screen navigation():
 
         textbutton _("About") action ShowMenu("about")
 
-        if renpy.variant("pc"):
+        if renpy.variant("pc") or (renpy.variant("web") and not renpy.variant("mobile")):
 
             ## Help isn't necessary or relevant to mobile devices.
             textbutton _("Help") action ShowMenu("help")
 
-            if not renpy.variant("web"):
+        if renpy.variant("pc"):
 
-                ## The quit button is banned on iOS and unnecessary on Android and Web.
-                textbutton _("Quit") action Quit(confirm=not main_menu)
+            ## The quit button is banned on iOS and unnecessary on Android and Web.
+            textbutton _("Quit") action Quit(confirm=not main_menu)
 
 
 style navigation_button is gui_button
@@ -727,7 +727,7 @@ screen preferences():
             hbox:
                 box_wrap True
 
-                if renpy.variant("pc"):
+                if renpy.variant("pc") or renpy.variant("web"):
 
                     vbox:
                         style_prefix "radio"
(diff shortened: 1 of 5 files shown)


renpytom:
My feeling is that web should be exclusive of the other platform-specific things. I like the idea of detecting touch and small/medium/large, but I don't think a web platform game should set mobile, pc, android, or ios. (Or linux/windows/macintosh, for that matters.) 

Basically, if renpy.variant("android") was used to implement some requirement for Android, I don't think that should affect the Android value.

Master shouldn't be broken, though you might have to delete renpy/module/gen to build it.

Beuc:
- One issue is whether to display the (keyboard/mouse-related) help menu; that's mainly why I set 'pc' vs. 'mobile'&'tv'. What would you suggest?

- Looking again at 'touch', my heuristic seems error prone due to hybrids like microsoft surface, and Edge's way to declare touchpads as touch devices. I'm thinking of only defining it when we detect a mobile browser (even if we don't define the 'mobile' variant) to mitigate, there doesn't seem to be a clean way to address this, all the doc I found recommends moving away from touch/not-touch binary logic.

> Master shouldn't be broken, though you might have to delete renpy/module/gen to build it.
I (obviously) did that, as well as registered the new matrix module. Does it work for you??

Beuc:
(added a note about missing window/full-screen menu as well)

renpytom:
Hm... is it possible to detect the presence of a keyboard or mouse on the device? It might be a good idea to proactively set keyboard and mouse variants, rather than touch. 

I haven't tried building the web version of the gl2 codebase, though I'll hopefully get to that soon.

Beuc:
AFAICS browsers push devs to handle all kind of events as they are received, rather than attempt to double-guess how the user will use the webpage. Consequently they do not provide devs with device capabilities ahead of time (i.e. the code doesn't know there's a keyboard until it receives a keyboard event).
I found a [possible way](https://stackoverflow.com/a/52854585) to detect a mouse but nothing for keyboards.

Beuc:
I reworked the proposal.
- 'android'/'ios'/'pc' dropped
- mitigate pc-tablet hybrids by defining "touch" only on mobile
- keep 'mobile' as it's the basis for many decisions
- display Help only if pc or web&!mobile
- display Fullscreen option for pc&web

PR #1961, opened 2019-07-27
web: detect when fullscreen status changed externally

Beuc:
When Ren'Py "caches" the fullscreen status, this lead to issues in RenPyWeb, as exiting fullscreen can be triggered directly by the user pressing "Esc." (handled by the browser, rather than requested by Ren'Py/SDL2). This makes Ren'Py keep requesting fullscreen right after the user exits it.

Getting the low-level fullscreen status is a bit convoluted but I could do it without modifying the API.


Beuc:
Ping? :)

Issue #1963, opened 2019-07-28
Shader error: 'highp' : precision is not supported in fragment shader


Beuc:
I just got a report from a RenPyWeb user getting this on a tablet:
```
  File "glenviron_shader.pyx", line 315, in renpy.gl.glenviron_shader.check_status
Exception: Shader error: ERROR: 0:3: 'highp' : precision is not supported in fragment shader
```
I requested more info on the hardware, but looking at the code, it sounds like Ren'Py assumes 'highp' is available everywhere.

In the OpenGL wikibooks I wrote a few years back, I used this kind of boilerplate:
https://gitlab.com/wikibooks-opengl/modern-tutorials/blob/master/common-sdl2/shader_utils.cpp#L90
```
		"#ifdef GL_ES                        \n"
		"#  ifdef GL_FRAGMENT_PRECISION_HIGH \n"
		"     precision highp float;         \n"
		"#  else                             \n"
		"     precision mediump float;       \n"
		"#  endif                            \n"
		"#else                               \n"
		// Ignore unsupported precision specifiers
		"#  define lowp                      \n"
		"#  define mediump                   \n"
		"#  define highp                     \n"
		"#endif                              \n";
```
Not sure how current this is in 2019.

Beuc:
The hardware is a Huawei Mediapad 10 Link running Android 4.4.
According to the specs it carries a Vivante GC4000 GPU:
https://www.gsmarena.com/huawei_mediapad_10_link-5507.php
which should fully supported GLES 1-3:
https://bensontao.wordpress.com/2013/12/04/vivante-vega-gpu-geometry-and-tessellation-shader-overview/

The lack of 'highp' is not detected by Ren'Py and RenPyWeb quickly crashes with a JavaScript error ("cannot read property 'createImageData' of null").

I asked the user if they can test with a RAPT version for comparison just to make sure.

Beuc:
The user (who decided not to comment here, so I'll proxy) couldn't compare with RAPT since RAPT now requires Android 5.0 (since 7.2 or 7.3 AFAICS, I didn't find the info in the Ren'Py changelog).
I unearth my first smartphone but it already supported 'highp' so I cannot test personally (however I could check the funny results that a native 'lowp' can produce ;)).

Beuc:
I send a build following #1965 to the user to see if this fixes the issue.

renpytom:
This has been merged in, and I handled in in gl2, so I think this can be closed.

PR #1965, opened 2019-07-30
gles: handle GPUs that do not support highp

Beuc:
POC for #1963

### DIFF
--- a/renpy/gl/glenviron_shader.pyx
+++ b/renpy/gl/glenviron_shader.pyx
@@ -31,11 +31,17 @@ cdef int round(double d):
     return <int> (d + .5)
 
 
-VERTEX_SHADER1 = """\
+GLES_PORTABILITY = """\
 #ifdef GL_ES
-precision highp float;
+  #ifdef GL_FRAGMENT_PRECISION_HIGH
+    precision highp float;
+  # else
+    precision mediump float;
+  # endif
 #endif
+"""
 
+VERTEX_SHADER1 = GLES_PORTABILITY + """\
 uniform mat4 Projection;
 
 attribute vec4 Vertex;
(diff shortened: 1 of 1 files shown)


renpytom:
Can you get rid of the gl2environ_shader changes, since it went away.

Beuc:
This is food for thought at this point, which may come in handy since you're rewriting these parts.

Beuc:
AFAICS we'd need to replicate the change in gl2shadercache.py and gl2ftl.pyx.
Do you want me to do that?

renpytom:
I'll take care of the gl2 stuff - right now, it's what I'm concentrating on, and I'd like to focus my efforts on it without having to worry about pull requests to it. 

Were you able to get master running? I've seen a few graphical glitches on the old gl renderer (probably due to the change to us the new 4x4 Matrix implementation), but I'm able to get it to run and render most games. So if you're not able to run master, that's a problem.

Beuc:
I didn't try renpyweb/master again, but given the kind of on-going changes it's best you check directly IMHO.
In addition, it would be worth considering switching web to the gl2 backend since WebGL 1 is based on OpenGL ES 2.

renpytom:
Both make sense. The plan is to let the creator opt in to gl2 for a bit.

Note that the Ren'Py gl backend already uses OpenGL ES 2 - the point of Ren'Py's gl2 renderer is to take better advantage of features that weren't in the gl1 fixed-function stack. (Also, to move Ren'Py from quads to meshes.)

Beuc:
Re-applied patch to master, gl/-only, as discussed.

Beuc:
Is it OK?

Beuc:
Ping? :)

renpytom:
Sorry, merged this in.

PR #1967, opened 2019-08-03
Added is_init_phase in exports.

Andykl:


renpytom:
Not sure where this is needed, but it's not a bad addition. Merged.

PR #1972, opened 2019-08-06
Adjustment force_step

Andykl:
A test script that demonstrates the changes
[script.zip](https://github.com/renpy/renpy/files/3469953/script.zip)


renpytom:
In round_value, the last line will force the value to be a float, even when it wasn't before. 

We should use something like type(self.value)(new_value) to force it to whatever the current type of value is.

Andykl:
Done

PR #1983, opened 2019-08-25
Fix an incorrect determiner

tdh8316:
The determiner `this` seems to incorrect in its context.


renpytom:
I think you found a valid problem, but I think your fix was wrong. I fixed it myself in 3441e0e. Thank you!

PR #1997, opened 2019-09-07
fixed sound-queue call_back

pionere:
call queue_empty_callback only if the queue is really empty


renpytom:
Looks good, merged. Thanks.

PR #1999, opened 2019-09-09
remove callback on stop

pionere:
Remove callback when a channel is stopped, otherwise the empty queue triggers the the callback, which then fills the queue and filling the queue starts the playback again.


renpytom:
I don't think we want this - the callback is for a very specific type of game, where the callback is always called to fill the queue.

PR #2002, opened 2019-09-13
Fixed bug with edgescroll in viewport

Herpior:
Fixed bug where viewport stops edgescrolling for a fraction of a second after every mousemotion event when mouse in the edgescroll area. It's caused by the event function setting edge_last_st = st when it calls check_edge_redraw. 

Fixed by adding a boolean parameter to check_edge_redraw that keeps the method exactly as before by default, but allows it to skip resetting edge_last_st when called from the event function.

### DIFF
--- a/renpy/display/viewport.py
+++ b/renpy/display/viewport.py
@@ -278,7 +278,7 @@ class Viewport(renpy.display.layout.Container):
 
         return rv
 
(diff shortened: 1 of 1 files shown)


renpytom:
Merged, thanks.

PR #2004, opened 2019-09-24
Join previous autosave thread before spawning a new one

uyjulian:
On some platforms, the amount of thread resources are limited.

To avoid reaching the limit quickly, join the previous autosave thread before spawning a new one.


renpytom:
Looks good, merged. What platform had this problem?

uyjulian:
Horizon / Nintendo Switch.

PR #2005, opened 2019-09-27
Clear edgescroll timer when mouse moved out of bounds

Herpior:
Fixed a bug I caused in #2002, where the viewport would jump to a different position when mouse re-enters the viewport.


renpytom:
Merging in. Sorry I missed this for the second pre-release, it'll be here for the next one.

PR #2006, opened 2019-09-27
config.cache_surfaces is False by default

Beuc:
AFAICS the only place where it's set to True is in a `version <= (6, 99, 14)` compatibility block.
I only read and test it to False otherwise.


renpytom:
Good catch, I think this changed without the documentation being changed.

Issue #2003, opened 2019-09-15
Python 3 Support


renpytom:
Since I've been getting a lot of people asking about this. Yes, Ren'Py is going to be ported to Python 3. That will very likely take place later this year, or early next year, once the OpenGL Rewrite that I'm working on is mature enough that I can spend some time on something else. 

My goal is to keep releasing a Python 2.7-based Ren'Py 7 for several years after that, to allow older games to keep working on newer computer. At the same time, I'll to release Ren'Py 8, based on a Python 3.x version, for newer development. 

It's important to realize that Python 2.7 will not stop working in 2020 or anytime soon. The games that work today will still work today, and new games created with Ren'Py 7 will still continue to work. Many of the advantages of Python 3, like unicode strings, have been adopted by Ren'Py. So while this is something that I plan to do soon, right now I'd like to focus on work that more directly improves what Ren'Py is capable of. (And to be honest, to focus on something I'd enjoy way more than getting the Python 3 stack up on a bunch of unsupported platforms.)

thirdtruck:
Hi Tom! Looking for help with this?

I'm a Ren'Py user (http://www.anthrotari.com), fresh off of helping with a Python 2->3 migration for work, who wants to contribute something for Hacktoberfest (https://hacktoberfest.digitalocean.com). I'd be glad to get the ball rolling on this.

danballance:
Hi guys, how is the python 3 migration looking? I only ask because my brother is learning python in order to develop games with Renpy (great project btw!) and I've advised him to get his machine set up for python 3. Then we've realised your actually running python 2 still. I'm a bit hesitant to advise him to start learning python 2 in 2019. Is the migration definitely going to happen or are you struggling to get the time and resources? (Completely understandable if you are of course!)

LuqueDaniel:
@danballance To develop games with Ren'Py it isn't necessary to setup a specific environment. In addition there isn't much difference between Python 2 and 3 in terms of syntax. Regarding the current status of Python 3 support, some contributors have been [working on it](https://github.com/renpy/renpy/pulls?q=is%3Apr+Py3+is%3Aclosed) recently.

danballance:
Thanks for the pointer and update. Much appreciated :)

darrickyee:
[...]

PR #2017, opened 2019-10-02
[Py3] Make Python 3 integer related changes

wchill:
Continuing the work from #1098, #2003 and related PRs.

1. Remove the L suffix from integers.
2. Fix integer division in Python 3.
3. Use the new octal syntax to prevent breakage in Python 3.

### DIFF
--- a/launcher/game/package_formats.rpy
+++ b/launcher/game/package_formats.rpy
@@ -30,6 +30,7 @@ init python in distribute:
     import shutil
     import threading
 
+    from builtins import int
     from zipfile import crc32
 
     zlib.Z_DEFAULT_COMPRESSION = 5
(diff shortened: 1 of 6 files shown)


renpytom:
I'd like to move us onto a dual-mode system, so we run on Python 3 and Python 2. Does this support that?

wchill:
Hi, yes, I believe so:

1. all constants with the L suffix, except for one, are 32-bit which should be fine as normal ints in Python 2. There's actually one thing I need to change here which I'll push out in a new commit (I was under the impression that `from builtins import int` would work without extra dependencies, oops).
2. Yes, integer division still works the same under Py2
3. The new octal syntax is supported since 2.6

infirit:
> Yes, integer division still works the same under Py2

Integer division is actually different where py2 always returns an int. Py3 however uses "true division" and may return a float. So ``1 / 3`` will return ``0.3333333333333333``

Using floor division "//" like this PR does is the right thing and it work the same between 2 and 3. You can get the py3 behaviour by import division from ``__future__``.

PR #2018, opened 2019-10-03
[Py3] Replace exec, execfile, reload with Py3 compatible versions

wchill:
1. Detect if reload is built-in (Py2.7 only) and import it from importlib if not (Py3 only).
2. Replace exec built-in with exec function call.
3. Replace execfile built-in with file read + compile + exec calls.

### DIFF
--- a/distribute.py
+++ b/distribute.py
@@ -10,6 +10,7 @@ import shutil
 import subprocess
 import argparse
 import time
+import importlib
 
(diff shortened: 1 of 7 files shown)


wchill:
This PR should also work under Python 2.7. exec in Python 2.7 can accept a tuple as the first argument, which means you can use it with the same syntax as in Python 3 even though it's not actually a function.

PR #2020, opened 2019-10-03
[Py3] Use open() instead of file()

wchill:
file() doesn't exist anymore in Py3.


renpytom:
I've merged all of your PRs into Ren'Py.

Issue #2022, opened 2019-10-03
Only show quick_menu while window is present (request)


lunalucid:
It's becoming a popular desire to only show the quick menu while the dialogue box is being shown. Some people have been asking how to hide the quick menu when the dialogue is hidden via showing screens or hiding the textbox (without setting quick_menu to true or false every time for the ones who do know of this option)

Not everyone wants the quick menu attached to the textbox but I'm assuming that those in that category would have a higher-level knowledge on GUI customization in general. 

So maybe we could possibly consider making this the default? At least for PC
Pull request will be made to reflect a solution if interested

PR #2023, opened 2019-10-03
Quick menu request per issue #2022

lunalucid:
"It's becoming a popular desire to only show the quick menu while the dialogue box is being shown. Some people have been asking how to hide the quick menu when the dialogue is hidden via showing screens or hiding the textbox (without setting quick_menu to true or false every time for the ones who do know of this option)

Not everyone wants the quick menu attached to the textbox but I'm assuming that those in that category would have a higher-level knowledge on GUI customization in general.

So maybe we could possibly consider making this the default? At least for PC"
This is the proposed solution if interested.

### DIFF
--- a/gui/game/screens.rpy
+++ b/gui/game/screens.rpy
@@ -244,23 +244,23 @@ screen quick_menu():
 
     ## Ensure this appears on top of other screens.
     zorder 100
+    if renpy.get_screen("say"):
+        if quick_menu:
 
-    if quick_menu:
+            hbox:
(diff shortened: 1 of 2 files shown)


renpytom:
Rejected.

The most useful time for saving is at menus, and also at things like imagemaps and choice screens. Having the quick menu go away during these most useful times for saving isn't a good idea, so a patch like this would be counterproductive.

lunalucid:
Good point, thanks.

PR #2033, opened 2019-10-11
Config exception handler

Andykl:
Some details:
1. I have placed it in this place because handle before the screen can be initialized (e.g. an error in early) have no sense. I'm not sure if I need to docs it.
2. This handles CONTROL_EXCEPTIONS, so technically it is possible to use `config.exception_handler = Jump(label)` or `config.exception_handler = renpy.full_restart` and it will work, but only during the execution of the game. But maybe it's a good side effect if the developer is sure what he's doing.
3. It doesn't pass `(short, full, traceback_fn)` because these data are very raw and I don't think that it should be processed outside the Ren'Py.

### DIFF
--- a/renpy/config.py
+++ b/renpy/config.py
@@ -1006,6 +1006,10 @@ depth_size = None
 # A list of screens to remove when the context is copied.
 context_copy_remove_screens = [ "notify" ]
 
+# A function which will be called in new context to handle excpetion or
+# None to use default handler.
+exception_handler = None
+
 del os
 del collections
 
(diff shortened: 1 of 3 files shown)


renpytom:
No, this isn't something I'll merge. This is too radical of a change for something that has to run in an exception context, where Ren'Py is already potentially having problems.  I also think that without the traceback information, you've made the exception screen completely useless - a player will see this, and not be able to report anything worthwhile about the problem.

I'll add in what I think is the right way to handle this.

Andykl:
Okay, maybe I didn't get the what "error handling" is in your point of view.

PR #2036, opened 2019-10-14
[Py3] Updated behavior of RevertableList

Andykl:
1. Fixed bug when `RevertableList[::1]` returns Python list.
2. Fixed bug when `RevertableList[:] = []` did not participate in the rollback.
3. Since slice methods are not available in Python 3 list, their use is wrapped in six.
4. Added useful `clear` and `copy` methods of Python 3 list.

### DIFF
--- a/renpy/python.py
+++ b/renpy/python.py
@@ -38,6 +38,8 @@ import re
 import sys
 import time
 
+from renpy import six
+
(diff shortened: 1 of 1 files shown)


renpytom:
Merged.

Issue #1926, opened 2019-07-08
Renpy presplash will cause Windows to want to kill the process if you click on the presplash.


sdn3rd:
Upon clicking the pre-splash in Windows 10:
PS C:\Users\nelsont>  [System.Environment]::OSVersion.Version

Major  Minor  Build  Revision
-----  -----  -----  --------
10     0      17763  0

It thinks Python has froze and asks to kill the process.

![presplash](https://user-images.githubusercontent.com/5367766/60783355-40cd6380-a118-11e9-8b9f-73ed609fde17.png)

uyjulian:
Windows does that automatically whenever a long operation is taking place and no events have been responded to.

Andykl:
Yes, but maybe there is a way to pass some event to the presplash so as not to cause freezing.

renpytom:
Yeah, we probably have to pump the event loop occasionally so this doesn't happen.

Keyon77:
I seem to be having a similar issue, the splash screen shows but the game fails to load.  Earlier versions of the game worked fine, as well as my other renpy games.  I have directx 11 installed and play games that require much higher graphics, i cleared persistent files and also the registry, the game is also on allowed lists for firewalls/anti virus etc.  currently running windows 8.1

uyjulian:
Another solution to the problem: call `DisableProcessWindowsGhosting`

PR #2039, opened 2019-10-15
Rework presplash processing

Andykl:
Since pygame_sdl cannot handle events by threads, we pump the presplash window after each file loading. Also checks that the user asked to close the game without waiting end of loading.


Andykl:
This is fix #1926

PR #2044, opened 2019-10-24
Input parameters "allow" and "exclude" can take regex

Lezalith:


renpytom:
This can't be merged as given, as it changes the behavior of Ren'Py.

````
$ test = renpy.input("Morse code only!", allow=".- ") 
```

goes from matching very little to matching everything. If you want to make a version that checks to see if allow and exclude are regexes created with re.compile, I'd consider that pull request.

Issue #2032, opened 2019-10-10
Ren'Py 7.3.4 Unable to make a ZIP Distribution of the SDK


GanstaKingofSA:
**Description**
Currently in the latest version of Ren'Py when the user selects `Open launcher project` and tries to build the Ren'Py Launcher with `sdk` selected, it will give the user a traceback error like the one below.

```
I'm sorry, but an uncaught exception occurred.

While running game code:
  File "game/distribute.rpy", line 1512, in <module>
  File "game/distribute.rpy", line 1519, in _execute_python_hide
  File "game/distribute.rpy", line 572, in __init__
  File "game/package_formats.rpy", line 398, in wait_parallel_threads
  File "game/distribute.rpy", line 1299, in done
  File "game/distribute.rpy", line 122, in hash_file
IOError: [Errno 2] No such file or directory: 'E:\\renpy-7.3.4-sdk\\renpy-7.3.4-dists\\renpy-7.3.4-sdk.zip'

-- Full Traceback ------------------------------------------------------------

Full traceback:
  File "launcher/game/distribute.rpyc", line 1512, in script
  File "E:\renpy-7.3.4-sdk\renpy\ast.py", line 914, in execute
    renpy.python.py_exec_bytecode(self.code.bytecode, self.hide, store=self.store)
  File "E:\renpy-7.3.4-sdk\renpy\python.py", line 2008, in py_exec_bytecode
    exec bytecode in globals, locals
  File "game/distribute.rpy", line 1512, in <module>
  File "game/distribute.rpy", line 1519, in _execute_python_hide
  File "game/distribute.rpy", line 572, in __init__
  File "game/package_formats.rpy", line 398, in wait_parallel_threads
  File "game/distribute.rpy", line 1299, in done
  File "game/distribute.rpy", line 122, in hash_file
IOError: [Errno 2] No such file or directory: 'E:\\renpy-7.3.4-sdk\\renpy-7.3.4-dists\\renpy-7.3.4-sdk.zip'

Windows-8-6.2.9200
Ren'Py 7.3.4.596
Ren'Py Launcher 7.3.4.596
Thu Oct 10 18:09:15 2019
```

I know this is mainly a issue that only Tom can see when he builds Ren'Py updates for release but wanted to point that out as I have encountered it on it along with my own modification of the SDK.

renpytom:
[...]

PR #2050, opened 2019-10-29
Fix for Building Ren'Py in Ren'Py

GanstaKingofSA:
This PR does the following

- Fixes `package_formats.rpy` in Ren'Py Launcher to build Ren'Py itself with `Build Distributions` in Windows

As discussed in Discord, the issue surrounds Lines 1311-1314 of `distribute.rpy` where during building, Ren'Py throws an exception (reference Issue #2032 for error) of building .zip's of the SDK and of IDE that come with Ren'Py, updates and other packages except `.tar.bz2`.

The issue stems on these lines of code
```python
elif format == "zip" or format == "app-zip":
    if self.build['renpy']:
        pkg = ExternalZipPackage(path)
    else:
        pkg = ZipPackage(path)
```
Ren'Py from after 6.99.12.4 (6.99.13 -> 7.3.5) will be unable to build the SDK version of itself due to a issue with `ExternalZipPackage(path)` and of it executing the result in Windows due to ExternalZipPackage using Linux/Unix commands that Windows does not recognize when subprocess.Popen engages the code.

Refer to this [comment](https://github.com/renpy/renpy/pull/2050#issuecomment-547919162) and Commit 83bd509 for new fix.

When this is executed in Windows, all builds of Ren'Py, Updates, and DLC build properly and output with no errors as it was building a regular traditional Ren'Py game. Hopefully this fixes the headache beforehand and the need of `distribute.py` to now distribute in the launcher itself for Tom.


GanstaKingofSA:
Refer to Commit https://github.com/renpy/renpy/commit/83bd509b589d09055e90a3b6e20612d6b3d20463 for the PowerShell Script used. Originally I wanted to use 7-Zip but since not many will have that installed, I decided to use the Global PowerShell ZIP code in Windows instead

uyjulian:
I would avoid PowerShell, since it won't create archives with the correct directory separator: `/` instead of  `\`

GanstaKingofSA:
Just tested it again using the correct separator and it looks like PowerShell is fine with it. As this affects Windows, the few options are 7-Zip or VBS/.NET apart from PowerShell at the end of `ExternalZipPackage` in order to compile the SDK to ZIP. This doesn't mainly affect regular projects but only if building the Ren'Py ones. @uyjulian 
Do you have a suggested method?
Will update the commit to use `/*` instead as it works fine.

PR #2061, opened 2019-11-17
Fix unreachable code on handling exception

tdh8316:
Statements after `raise` seems to be unreachable.

### DIFF
--- a/renpy.py
+++ b/renpy.py
@@ -152,8 +152,8 @@ def path_to_renpy_base():
 try:
     import ast; ast
 except:
-    raise
     print("Ren'Py requires at least python 2.6.")
+    raise
     sys.exit(0)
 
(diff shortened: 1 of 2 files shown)


uyjulian:
`sys.exit(0)` is still unreachable.

tdh8316:
Thanks! I think `sys.exit` is not required as the `raise` statement performs that action.

renpytom:
I need to look at these more closely, since I think this might be spurious debug code.

tdh8316:
Yes, there were other unreachable codes, which look like spurious debug code.
For instance:
```python
if False:
    ...
```
And I didn't edit these codes.

renpytom:
This is a good catch, but it was more just debug code that didn't get removed.

tdh8316:
Well, thanks for following up anyway.

PR #2063, opened 2019-11-23
Add RPi4 Raspbian Support.

lee-yunseok:
Just linked aarch64 to current armv7l RPi libs. At least until now, I haven't found any problems.


uyjulian:
I thought RPi4 is armv8 not armv7l

lee-yunseok:
RPi4 is armv8 but armv7l libs work for RPi4. No need make armv8 libs on Ren'Py.

uyjulian:
This may break Linux 4 Tegra and other native arm64 bit distributions

lee-yunseok:
Is this break switch homebrews?

uyjulian:
No, since it is not using Linux.

However, other native aarch64 platforms may be broken due to not having 32 bit libraries.

uyjulian:
Looks good. May I suggest squash

lee-yunseok:
@uyjulian I have considered the expected situation. Now it looks safe.

uyjulian:
You should squash your commits.

lee-yunseok:
The only one issue is lacking performance in Raspbian 64bit kernel with libegl error. I believe this will be solved with next mesa.

lee-yunseok:
libEGL problem has solved in recent RPi update.

lee-yunseok:
libEGL error has solved in recent RPi update.

PR #2096, opened 2020-01-21
Encode and decode the text during copypaste

### LINKED ISSUE #2055: Input copypaste have problems with non-ASCII characters
Look at if it is possible to read the unicode or filter the input and output lines if this is not possible.

Andykl:
This is fix #2055

### DIFF
--- a/renpy/display/behavior.py
+++ b/renpy/display/behavior.py
@@ -1371,11 +1371,13 @@ class Input(renpy.text.text.Text):  # @UndefinedVariable
             raise renpy.display.core.IgnoreEvent()
 
         elif self.copypaste and map_event(ev, "input_copy"):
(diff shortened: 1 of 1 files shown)


infirit:
I fear you are triggering python2's implicit decode and encode which is not available in python3. The ``encode`` method on strings is an operation on codepoints (unicode type in python2). If ``self.content`` is of type unicode you are good but if it is of type str then this change will cause problems in python3.

jsfehler:
@infirit ren'py is using six under the hood: https://github.com/renpy/renpy/blob/c2206245ed5e4a83946b8f7133f5521a5654e35b/renpy/compat/__init__.py#L97

So this change is probably safe, but good catch.

Issue #2101, opened 2020-01-22
Potentially performance leak.


Andykl:
Discribed in https://github.com/renpy/renpy/commit/0b450d9aceac8a71bae887210710bf07be7b6e2d#r36886826

PR #2105, opened 2020-01-25
Add participation of pure funcs and constants in ATL analysis

### LINKED ISSUE #2100: ATL images doesn't sees pure functions
```
image test:
    Solid("#fff")
```
This image will be marked as not const. This happens because for ATL based images we do not pass parameters to the pyanalysis.

Andykl:
First commit fixes #2100.
It doesn't help with ATL for show, scene and show layer statements.


Andykl:
Second commit fixes issue with
```
default xpos = None
transform use_default:  # this is not constant
    xpos xpos

transform use_use_default:  # this is constant because the name use_default is constant
    use_default
```
It also enabled analysis for show scene and show layer statements.

Andykl:
This commit fixes #2101.

renpytom:
Merged, thank you.

PR #2106, opened 2020-01-28
Revert "web: Avoid the use of float arguments."

Beuc:
This work-around was specific to Emterpreter, which is obsoleted by ASYNCIFY.

This reverts commit dce59141a7f940da9f5b2820ee4b61016b9113b3.


Beuc:
Hi. Let me know if you have any question :)

Beuc:
If you're reworking the GL handling, I'd suggest merging this patch now, rather than after everything is tested :)

PR #2114, opened 2020-02-13
doc: reference im.AlphaMask in AlphaMask

Beuc:
It took me forever to spot that there are 2 functions with the same name, and the first one in the documentation (which we find through Ctrl+F) did not mention that.


Beuc:
Committing this small doc fix to ease the backlog -- feel free to revert.

PR #2121, opened 2020-02-19
Add equality comparison to HistoryEntry

jsfehler:
I'm currently working on a game where dialogue can be repeated by revisiting the same label.  Using an equality check, I can use a callback to prevent the history from showing the same line multiple times in a row.

### DIFF
--- a/renpy/character.py
+++ b/renpy/character.py
@@ -648,6 +648,9 @@ class HistoryEntry(renpy.object.Object):
 
     multiple = None
 
+    def __eq__(self, other):
+        return self.who == other.who and self.what == other.what
+    
     def __repr__(self):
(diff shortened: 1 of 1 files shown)


infirit:
For now renpy supports python2 so you will also need to implement ``__ne__``, see https://docs.python.org/2.7/reference/datamodel.html#object.__ne__.

Python3 by default will invert the result ``__eq__`` so there you can skip ``__ne__``.

jsfehler:
@infirit Good catch, but I suspect @renpytom will release the python3 update to Ren'py before he considers merging this.

GiovanH:
No; the plan is currently to support [both platforms](https://github.com/renpy/renpy/pull/2017#issuecomment-537714917), so this patch needs to implement `__ne__`.

PR #2123, opened 2020-02-22
Add ability to show progress bar instead of static presplash

kobaltcore:
## What
Recently I was toying around with adding a presplash screen to a game I'm involved in because it's gotten quite large and takes a while to start up on older computers.

I experimented a bit and eventually implemented what's in this pull request, which is a way of defining a themable progress bar as the presplash which shows the actual progress towards main window creation.

## Why
A presplash is a great way of showing that the game has actually launched before the main window appears (which happens between 5 and 30 seconds after launching our game based on user reports). However, I found a static image quite uninspiring. Generally it is often times helpful to visually convey to the user that actual work is being done, hence this proposal. It also helps with estimating startup duration on the user's side.

## Code Talk
This change is backwards compatible with the current way of adding a static splash screen as the static one takes precedence over the progress bar one if it is given. As such, all current behavior remains exactly the same, making this an entirely optional feature.

The process of triggering this progress bar presplash is very similar to how the regular system works: Instead of creating the file `presplash.png` in the game directory, one instead creates two files: `presplash_background.png` and `presplash_foreground.png` (can also be `JPG`, mirroring the static presplash).

The progress bar code will then render the background as the full presplash screen and after that render the foreground on top, incpororating alpha. The foreground is cropped to `full_width * current_progress` and is thus revealed over the startup phase.

To update the screen I am using the `pump_window` method as this updates whenever the event loop needs to be pumped anyway and thus mirrors progress throughout the startup phase.

Now, there is one thing I'd consider a "code smell" in here, which is that while I am able to dynamically determine the amount of script files (and thus the amount of `pump_window` calls while script is being loaded), I can not easily determine the `pump_window` calls that happen outside of that, namely during the actual initialization phase.

Currently there is a fixed number of additional calls, which is why there is the magic number `23` in there. I don't like having magic numbers like this in there, however in this case I deemed it as begrudingly acceptable given that the amount of init phase calls to `pump_window` are unlikely to change. If there is a better way to figure this out, that would definitely be a good improvement.

## Samples
I have attached a sample video of how this may look using the `Bar` GUI example from the documentation. Additionally I've attached the two files required for triggering this progress bar so you can take a peek at how those could look like.

Let me know whether this is a feature worthwhile adding or whether it's an abomination of hell that should be incinerated.

### Shoddy GIF Demo
sample

### Background
presplash_background

### Foreground
presplash_foreground

### DIFF
--- a/renpy/display/presplash.py
+++ b/renpy/display/presplash.py
@@ -40,6 +40,40 @@ window = None
 start_time = time.time()
 
 
+class ProgressBar(pygame_sdl2.sprite.Sprite):
+
+    def __init__(self, foreground, background):
+        super(ProgressBar, self).__init__()
+        self.foreground = pygame_sdl2.image.load(foreground)
+        self.background = pygame_sdl2.image.load(background)
+        self.width, self.height = self.background.get_size()
+        self.image = pygame_sdl2.Surface((self.width, self.height))
+        self.progress = 0
+        self.counter = 0.0
+
+    def convert_alpha(self, surface=None):
+        self.foreground = self.foreground.convert_alpha(surface)
+        self.background = self.background.convert_alpha(surface)
+
+    def get_size(self):
+        return (self.width, self.height)
+
+    def update(self, total):
+        self.counter += 1
+        self.progress = min(self.counter / total, 1)
+        self.image.blit(self.background, (0, 0))
+        self.image.blit(self.foreground.subsurface(0, 0, self.width * self.progress, self.height), (0, 0))
+
+
+def find_file(base_name, root):
+    allowed_exts = [ ".png", ".jpg" ]
+    for ext in allowed_exts:
+        fn = os.path.join(root, base_name + ext)
+        if os.path.exists(fn):
+            return fn
+    return None
+
+
 def start(basedir, gamedir):
     """
     Called to display the presplash when necessary.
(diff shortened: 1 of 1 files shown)


Andykl:
This addition looks nice to me.
1. I find it strange to add 2 new special files for this. Perhaps it is better to make one file that will be interpreted as foreground. This change will still be backwards compatible.
2. There is a problem with your code not having minimum_presplash_time processing

kobaltcore:
In regards to 1.) my reasoning behind adding 2 files was customizability. By splitting background and foreground, people could theme both the background as well as the foreground separately. For example they could replace the rather uninspiring example background I provided by one with a more elaborate border.
If one wants to have a border, it would be necessary to have 2 distinct files because otherwise we'd be revealing the entire picture, including border, which looks strange.

As far as 2.) goes, the current solution does obey `minimum_presplash_time`, though maybe not in the best way. It progresses to 100% until the game has loaded completely, after which it enters the `sleep()` loop where it simply waits (with full loading bar) until the time has elapsed.

I'm not fully certain on how one would handle waiting on a progress bar that has already fully loaded and has thus served its purpose. Additionally, the `minimum_presplash_time` is only known after most of the init stages have already completed (given that it's loaded from `options.rpy` inside the actual game), so we can not adjust the progress without jumping backwards quite a bit.

PR #2124, opened 2020-02-23
Use or operator instead of if statement

jsfehler:
Faster, technically, but I think this makes it easier to read

### DIFF
--- a/renpy/common/00start.rpy
+++ b/renpy/common/00start.rpy
@@ -72,12 +72,7 @@ init -1600 python:
 
         import os
 
-        if "RENPY_LANGUAGE" in os.environ:
-            language = os.environ["RENPY_LANGUAGE"]
-        elif config.language is not None:
-            language = config.language
-        else:
-            language = _preferences.language
+        language = os.environ.get("RENPY_LANGUAGE") or config.language or _preferences.language
 
         renpy.change_language(language)
 


infirit:
For ``config.language`` it explicitly checks for ``None``. With this change any "falsy" value will used.

jsfehler:
@infirit config.language's type is None or str. Given an empty string isn't a valid language, the difference in behaviour is that if someone does use a blank string, they drop down to _preferences.language instead of using that blank string. In most cases this will not cause any differences.

You can't reasonably create a language that's mapped to an empty string, but `renpy.change_language` could crash out if you don't feed it None or a non-empty string. Any other type will crash, but later with a more opaque message. A blank string won't crash, but probably isn't what the user really wanted to do.

infirit:
If your intention is to fix a bug when language is an empty string then put that in the commit message for the reviewer. Right now it just "maybe performance" and your definition of "readability" which many would disagree with.

And think about it, the original author chose to check for None. People don't do that unless they want to check of a variable has been set or reset not to check validity. If the original author meant to check for a valid language then it would have been much more thorough than what you did now.

jsfehler:
I specifically said "I think" so yes, this is my definition of readability. If pytom thinks my opinion is incompatible with his direction, he can ignore/close the PR and I've learned more about the style he wants to use in the future. He can also tell me to buzz off with code changes in the name of aesthetic beauty.

Since the Falsey-but-intended use case was obviously not valid, I didn't concern myself with it. I'm more than happy to discuss unintended side-effects of this change, and/or ways of preventing users from creating bugs by entering an invalid language. (I'm sure the edge case of using a control character in a language string would be a fun experiment)

I never said the purpose of the original code was to check validity, but it is implicitly looking for a non-empty string, not just checking for not-None. The real question is if a Falsey value should drop to _preferences.language,  or a Falsey value should crash immediately, since anything but a non-empty string will crash anyways.

renpytom:
I think this is reasonable, so I'll merge it.

PR #2127, opened 2020-02-29
Add linting for redefined define/default statements

jsfehler:
Given the assumption that define/default statements shouldn't be duplicated, eg:

```
define pineapple = "pineapple"

default has_banana = False
default has_orange = False

define pineapple = "banana"
```

This PR adds a lint rule to inform the user when it happens:
```
game/script.rpy:9 define pineapple already defined at game/script.rpy:5
```

However, ren'py itself breaks this rule:
```
game/screens.rpy:567 define gui.about already defined at game/options.rpy:32
```

@renpytom If you like this rule but want to whitelist certain statements, we can do that. Or should gui.about not be duplicated?

### DIFF
--- a/renpy/lint.py
+++ b/renpy/lint.py
@@ -53,6 +53,11 @@ image_prefixes = None
 # The node the report will be about:
 report_node = None
 
+# Collect define/default statements to check for duplication
+all_define_statments = {}
+all_default_statements = {}
+
+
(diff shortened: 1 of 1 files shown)


renpytom:
Merged in, thank you. I think the duplication needs to be removed for new games, but I'll handle gui.about separately.

PR #2128, opened 2020-03-03
Lint: Warn when text displayables in a screen use string formatting

jsfehler:
Based on the documentation at: https://www.renpy.org/doc/html/screen_optimization.html#const-text

Text displayables in a screen should favour ren'py text substitution over using the % operator for string formatting.

I'm not sure if this is the best implementation of the check, and I'm not married to the specific message.

Example Scenario:

```
screen my_screen():
    $ potato_amount = 5
    text "I have % potatoes" % potato_amount
```

Should be written as:
```
screen my_screen():
    $ potato_amount = 5
    text "I have [potato_amount] potatoes"
```

To create a const displayable for the screen.

### DIFF
--- a/renpy/lint.py
+++ b/renpy/lint.py
@@ -618,10 +618,24 @@ def check_label(node):
 
 
 def check_screen(node):
-
     if (node.screen.parameters is None) and renpy.config.lint_screens_without_parameters:
         report("The screen %s has not been given a parameter list.", node.screen.name)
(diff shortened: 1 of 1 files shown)


renpytom:
I don't believe this is an important enough performance loss to be worth a lint check. It also gets too deep into the rest of Ren'Py's internals, dealing with things like the way class names are formatted. I just don't see this as worth the added complexity.

PR #2129, opened 2020-03-06
Improved renpy.pure function.

Andykl:
If this function is applied to a function in a module, it will be marked correctly.


Andykl:
Also fixed a bug with define statement with namespace. Example:
```
define audio.some = 'audio/some.mp3'

screen test():
    textbutton "Text" activate_sound audio.some action Return()
```
This screen was treated as non-constant.

renpytom:
Merged, thank you.

PR #2160, opened 2020-04-11
Fix parse error on `queue music [expr] channel [expr]`

hsandt:
Added missing `continue` in parse_queue_music for `channel` parameter parsing
Continues a series of fixes/refactoring done on music statement parsing in https://github.com/renpy/renpy/pull/2159


hsandt:
Until PR is approved, you can use the workaround:
`queue [channel_expr] [music_expr]`

PR #2163, opened 2020-04-13
Fix menu_kwargs restore assignment from old value

hsandt:
It's mostly a "code typo" fix. I don't have a specific case where it fixes a bug, since the restore didn't seem to matter in my project, but it should now have the intended behavior.


renpytom:
Thank you!

PR #2169, opened 2020-04-20
gl2draw: Explicitly specify profile and version when using desktop GL

uyjulian:


renpytom:
I like the idea of this patch, but I'm not sure the version is right. I think gl2 now requires OpenGL 2.1

uyjulian:
Rebased and fixed

renpytom:
Looks good, thank you.

PR #2170, opened 2020-04-20
Set SDL2 hint to convert touch events to mouse events

uyjulian:


uyjulian:
This doesn't actually work…

uyjulian:
Okay, this should actually work now.  
Some platforms may have `SDL_TOUCH_MOUSE_EVENTS` hint disabled by default, so enable it.  
Since Ren'Py does not use touch events directly, disable `SDL_MOUSE_TOUCH_EVENTS` to avoid unneeded processing.

renpytom:
Makes sense, I think. I'll merge this once Live2D is done.

PR #2174, opened 2020-04-24
Translations for Chinese update, and new fonts.

zedraxlo:
This time I uploaded 2 fonts for Chinese and Japanese. Be sure only use one of them.


renpytom:
I'm a little confused by this - how is one font covering both Chinese styles and Japanese as well? Is this some sort of OpenType magic?

zedraxlo:
> I'm a little confused by this - how is one font covering both Chinese styles and Japanese as well? Is this some sort of OpenType magic?

Some of them are not same on unicode, like 图 圖 図 .
Some of them are same, and using Japanese style, but their writing is not much different and does not affect the use.
In fact, Chinese deep users still need to change the font to their customs, because this is a lite version font, it is even not contain the whole 2nd-table.
But after testing, I ensured that this font can meet the normal use of the engine, and the release of lightweight games.
The current font may be slightly flawed, and I will re-adjust it later.

renpytom:
Could you re-open this? I apologize for being behind in merging things.

PR #2177, opened 2020-04-27
Make config.use_cpickle work again in Python 2.x

### LINKED ISSUE #2175: master no longer respects `config.use_cpickle`, but still documents it
52e447f8b7 has removed all the code which respects the `config.use_cpickle` flag. Either it should be reinstated, or the docs should be updated to remove the mention of that config flag (or, even better, mention that there *used* to be a flag, but it's been removed), and the release notes should point that out as an incompatible change.

mathrick:
Also document that it has no effect under Python 3

Resolves #2175 

*Note*: it doesn't touch code that has never made the distinction between `pickle`  and `cPickle`, which means in effect it restores the behaviour prior to c2206245e in those parts.

Tested manually under Python 2.7 with the following script:

```renpy
init python:
    renpy.config.use_cpickle = False
    
    class PickleSurprise(object):
        pass
        def __init__(self):
            self.surprise = None

label start:
    # Cause a pickle error to test config.use_cpickle
    $ surprise = PickleSurprise()

    python hide:
        def foo():
            def bar():
                return 42
            return bar

        surprise.surprise = foo()

label breakme:
    "Try saving now for a pickle surprise"
```

### DIFF
--- a/renpy/__init__.py
+++ b/renpy/__init__.py
@@ -59,7 +59,6 @@ from renpy.compat import *
 
 import sys
 import os
-import pickle
 import copy
 import types
 
--- a/renpy/compat/__init__.py
+++ b/renpy/compat/__init__.py
@@ -34,6 +34,9 @@ Right now, it does the following things:
 
 * Defines PY2 in the current context, to make Python 2 conditional.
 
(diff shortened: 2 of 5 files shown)


renpytom:
This is good work. Thank you! Merged.

PR #2184, opened 2020-05-02
Create splash_screen.rpy

YasPHP:
Good to go?

### DIFF
--- a/tutorial/game/splash_screen.rpy
+++ b/tutorial/game/splash_screen.rpy
@@ -1,0 +1,21 @@
+# JEDIT code to make a splash screen
+
+# splash screen: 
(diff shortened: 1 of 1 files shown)


renpytom:
Why do we want this? Ren'Py supports them, since people like them - but since it strictly slows the game down, I don't think it needs to be encouraged like this.

YasPHP:
Ah okay, thought it could be a good move to promote general game logistics like copyright notice, version, etc.

renpytom:
Oh, those go in the about screen, which is part of the default gui.

YasPHP:
Awesome, thanks for clarifying.

PR #2189, opened 2020-05-08
add "mobile" screen variant to iOS builds

capntrips:
According to [here](https://github.com/renpy/renpy/commit/118d39d83b40620f15033a707680b51c04ba19ce#diff-3deaadb5853b1b2467152b00910bbfe9L240-L246) and [here](https://www.renpy.org/doc/html/screens.html#screen-variants), iOS builds should define the "mobile" screen variant.


renpytom:
Merged, thank you.

PR #2191, opened 2020-05-10
Allow specification of per-play/per-queue volume

kobaltcore:
## What
This change enables users to specify `volume` in `play` and `queue` statements as well as via the Python API `renpy.music.play`, `renpy.sound.play` and the `queue` variants respectively.  
Specifically it adds a tertiary volume which in turn affects the secondary volume such that it still works fine with mixer controls and with minimal changes to the codebase.

## Why
It is highly likely that a lot of users utilize music from multiple sources, which is almost never loudness-normalized. Few users do this themselves.  
Thus it can be the case that some music or sound effects are louder or more quiet than others, which can be a hassle to deal with as a player, because one would have to constantly change the mixer controls around depending on what's playing.

While this change does not resolve the loudness issue itself, it gives game authors an easy way out of this by enabling them to specify per-play/per-queue volumes, circumventing these jumps in loudness.

In addition, it is sometimes desirable to play an existing sound effect at a higher or lower volume to fit a specific scene, for example contrasting regular knocking on a door with forceful knocking (the latter of which should be louder).

This was previously possible as well, using constructs like the following:
```renpy
# First
$ renpy.music.set_volume(0.75)
play music my_music_var

# Later
$ renpy.music.set_volume(1.0)
```

The proposed change is a more convenient solution for this issue as it has the advantage of not having to remember to re-set the volume back to its original level after tweaking it. Additionally it does not mess with the secondary volume, which is already modified by the audio emphasis system.

## Code Talk
I changed the affected statements (namely `play` and `queue` and their sub-statements where necessary) as well as the matching Python API points.  
Regarding naming conventions, I named the parameter in the Python API `relative_volume`, however the tag used for statements is simply `volume` for convenience.
This can of course be changed if a different naming scheme would be more clear.

I added a tertiary volume which is multiplied with the secondary volume. This way it does not interfere with the audio emphasis system, which may lift or reduce secondary volume at times. Using this method, we remain relative to the actual volume, even when audio emphasis is active.

I added basic docs to the affected Python methods, which can be expanded upon if required.

## Samples
```renpy
play music my_music_var volume 0.75
queue music "audio/music.mp3" fadein 1.5 volume 0.3 fadeout 2.3
play sound "audio/some_file.opus" volume 1.2
queue sound some_sfx_var volume 1.2
```

This also plays well with emphasized audio:
```renpy
# Configuration
define config.emphasize_audio_channels = ["sound"]  # reduce music when sound effects are playing
init python:
    renpy.game.preferences.emphasize_audio = True

# In the script
play music my_music_var  # full loudness
"dialogue"
play music my_music_var volume 0.5  # half loudness
"dialogue"
play sound sfx  # reduces "music" channel further, plays sound, ramps back up to 0.5
"dialogue"
play music my_music_var  # full loudness again
```

Let me know your thoughts.

### DIFF
--- a/renpy/audio/audio.py
+++ b/renpy/audio/audio.py
@@ -142,6 +142,7 @@ class MusicContext(renpy.python.RevertableObject):
     __version__ = 0
 
     pause = False
+    tertiary_volume = 1.0
 
     def __init__(self):
 
--- a/renpy/audio/audio.py
+++ b/renpy/audio/audio.py
@@ -124,11 +124,12 @@ class QueueEntry(object):
     A queue entry object.
     """
 
-    def __init__(self, filename, fadein, tight, loop):
+    def __init__(self, filename, fadein, tight, loop, relative_volume):
         self.filename = filename
         self.fadein = fadein
         self.tight = tight
         self.loop = loop
+        self.relative_volume = relative_volume
 
 
 class MusicContext(renpy.python.RevertableObject):
@@ -157,6 +158,9 @@ class MusicContext(renpy.python.RevertableObject):
 
         # The secondary volume.
         self.secondary_volume = 1.0
+
+        # The tertiary volume.
+        self.tertiary_volume = 1.0
 
         # The time the channel was ordered last changed.
         self.last_changed = 0
@@ -476,6 +480,8 @@ class Channel(object):
             try:
                 filename, start, end = self.split_filename(topq.filename, topq.loop)
 
+                self.set_tertiary_volume(topq.relative_volume)
+
                 if (end >= 0) and ((end - start) <= 0) and self.queue:
                     continue
 
@@ -513,9 +519,9 @@ class Channel(object):
             if self.loop:
                 for i in self.loop:
                     if topq is not None:
-                        newq = QueueEntry(i, 0, topq.tight, True)
+                        newq = QueueEntry(i, 0, topq.tight, True, topq.relative_volume)
                     else:
-                        newq = QueueEntry(i, 0, False, True)
+                        newq = QueueEntry(i, 0, False, True, topq.relative_volume)
 
                     self.queue.append(newq)
             # Try callback:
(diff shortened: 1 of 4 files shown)


renpytom:
I like the idea of this. 

I think that this would break save compatibility, but it's a an easy fix. What you want to do is to add a class-level tertiary volume field to MusicContext, the same as was done with pause. So:

```
class MusicContext(renpy.python.RevertableObject):
    """
    This stores information about the music in a game. This object
    participates in rollback, so when the user goes back in time, all
    the values get reverted as well.
    """

    __version__ = 0

    pause = False
    tertiary_volume = 1.0

    def __init__(self):

        super(MusicContext, self).__init__()

        # The time this channel was last ordered panned.
        self.pan_time = None
```

That will fix the backwards compatibility issue. Other than that I think this is a really good patch.

kobaltcore:
I've added the class-level-field as requested, thanks for being super-quick on the draw!

Andykl:
Also, to make it backward compaliable with old rpyc's data in statements execute code it's nessesary to change `relative_volume=eval(p["volume"])` to `relative_volume=eval(p.get("volume", "1.0"))`.
Otherwise it will nessesary to Force Recompile the game.

kobaltcore:
I've made the requested tweaks, which should hopefully make this a backwards-compatible change now. Is there anything else that may need revising?

Main points would be:
- The naming of `relative_volume` in the Python API (keep or change)
- The doc strings (keep, expand or modify)
- Ren'Py's documentation proper (add something about this in the actual docs?)

Andykl:
You forgot about the last one ;)
Everything else looks good.
I think Tom will merge it as soon as he sorts out the main tree for release.

kobaltcore:
Drast, I must've accidentally un-staged that particular hunk, whoops.

renpytom:
Merged, thank you.

Issue #2187, opened 2020-05-05
The `at transform` syntax doesn't get recognized inside a custom screen statement.


hell-oh-world:
I'm trying to implement a custom screen statement using the `renpy.register_sl_statement` class. The code is working well, except that any anonymous `atl` doesn't get passed. So this is somehow my code looks like:
```py
# in a file something like 01early.rpy
python early:
    renpy.register_sl_statement("my_statement", positional=0, children="many").add_property_group("ui").add_property_group("position")
```
```py
# the screen statement screen
screen my_statement(**properties):
    frame:
        properties properties
```
```py
# implementation in other screens
screen sample:
    my_statement:
        background "#f00"
        at transform:
            alpha 0.5
```
Pytom said that the `at transform` or the anonymous atl is kinda new in Ren'Py. This thing works if an ordinarily defined transform is used rather than an anonymous one, unfortunately, even though it's passed by printing the contents of `properties`, the atl doesn't map itself using the `properties` UI statement. In a short note, anonymous atl doesn't get passed, defined atl, on the other hand, gets passed, but it just won't be mapped by the `properties` statement unless you explicitly tell it to by using conditionals like this...
```py
# the screen statement screen
screen my_statement(**properties):
    frame:
        properties properties

        if "at" in properties:
            at properties["at"]
```
This one works, as long as the atl is defined and not an anonymous one.
Another thing to note is that I observed that you can't do a direct anonymous atl inside the screen statement, so...
```py
# the screen statement screen
screen my_statement(**properties):
    frame:
        properties properties

        at transform:
            alpha 0.5
```
This one won't work, but if the transform specified is defined, then the transform will be recognized.

Andykl:
Thank you for your response. I'll check it when I get a chance.

Andykl:
This issue should be fixed in next nightly build.

Issue #2154, opened 2020-04-06
 unexpected behavior when combining "use" with "if" as part of button in screen


Kiaazad:
Here's the topic about this issue on lemmasoft: https://lemmasoft.renai.us/forums/viewtopic.php?f=8&t=58541&p=527320#p527320
and here's how to replicate the bug
```python
screen scr222:
    frame:
	background "#fff4"
        text "123"
screen scr122:
    default tsts = 1
    button:
        background None hover_background "#333"
        use scr222
        if tsts:
            action NullAction()

label start:
    "..."
    show screen scr122
    "...."
```
the button action will be disabled whether the condition is True or False

Andykl:
I have found that this is not happening with
```
screen scr122:
    default tsts = 1
    button:
        background None hover_background "#333"
        if tsts:
            action NullAction()
        use scr222
```
So I'm guessing that execution of `SLUse`  somehow corrupts the parent screen scope.

Andykl:
For screen
```
screen scr122():
    default tsts = 1
    hbox:
        align (0.5, 0.5)
        button:
            background None
            hover_background "#333"
            if tsts:
                action Return()
            use scr222
        button:
            background None
            hover_background "#333"
            use scr222
            if tsts:
                action Return()
```
I traced it back to:
`SLDisplayable.__dict__` for these buttons differ only in the `children` and `keyword_children` fields:
```
first button:
    children = [<renpy.sl2.slast.SLIf object>, <renpy.sl2.slast.SLUse object>]
    keyword_children = [<renpy.sl2.slast.SLIf object>]
second button:
    children = [<renpy.sl2.slast.SLUse object>, <renpy.sl2.slast.SLIf object>]
    keyword_children = []
```
This happens because SLUse is always last_keyword:
https://github.com/renpy/renpy/blob/5b492825b0df713697c97da3cef389723ac8477a/renpy/sl2/slast.py#L1684

This is not only true for SLUse, so in this screen all buttons also ignore the If block.
```
screen scr122():
    hbox:
        align (0.5, 0.5)
        button:
            background None
            hover_background "#333"
            if False:  # SLIf with last_keyword child
                use scr222
            if True:
                action Return()
            use scr222
        button:
            background None
            hover_background "#333"
            showif False:  # SLShowIf is last_keyword
                pass
            else:
                use scr222
            if True:
                action Return()
        button:
            background None
            hover_background "#333"
            default var = True  # SLDefault is last_keyword
            if True:
[...]

PR #2210, opened 2020-05-25
Sl fixes

Andykl:
1. Fixed a performance leak for SLUse when use without a block was referring to `not_const_ast`.
2. Fixed lost copying `SLBlock.atl_transform`
3. Fixed a bug when `at` property inside `properties` has not updated the parsed keyword. Per #2187
```
screen some():
    fixed:
        at transform:
            alpha 0.5
        properties dict(at=Transform(alpha=1.0))
```
4. Slightly redesigned `SLIf.execute_predicting` so that the true block will perform the prediction exactly first. I.e. `false_screen` will predict its images first for:
```
define SOME_SCREEN_COND = False
screen some():
    if SOME_SCREEN_COND:
        use true_screen
    else:
        use false_screen
```
5. Stop add expressions in cycle finder for SLUse
```
screen expr():
    use false_cycle

default expr = "some"
screen false_cycle():
    use expression expr
```
6. Added a `SLBlock.keyword_exist` method to check if some keyword was parsed and use of it for `id` property and `arguments` property.
7. Added an error message if keyword children follow after last_keyword children. Per #2154 
8. Added separate ast class for CD SLS. It works the same as SLUse but creates arguments on each execute with processing if statements. Per #2187 
This change is backwards compatible with the compiled code. 
But to fix the bug with `add_positional` this requires an update `renpy.register_sl_statement("name", positional=1)` to `renpy.register_sl_statement("name").add_positional("name")`

### DIFF
--- a/renpy/display/screen.py
+++ b/renpy/display/screen.py
@@ -786,13 +786,15 @@ def get_all_screen_variants(name):
     order.
     """
 
-    rv = [ ]
+    if isinstance(name, basestring):
+        name = tuple(name.split())
 
-    for k, v in screens.items():
-        if k[0] == name:
-            rv.append((k[1], v))
+    name = name[0]
 
-    return rv
+    if name not in screens_by_name:
+        return [ ]
+
+    return list(screens_by_name[name].items())
 
 
 # Have all screens been analyzed?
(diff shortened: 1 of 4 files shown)


renpytom:
Merged this in, finally. Thank you.

Issue #1987, opened 2019-08-28
maxsize will upscale smaller images


mal:
I'm not sure if this is a bug or expected behaviour. The original changelog entry for `maxsize` suggests that images will only be scaled down to fit within the defined bounds, a description that appears to match up with the name. Actual behaviour differs from this in that smaller images will scaled up until one of their dimensions touches the bounds. The documentation proper doesn't preclude this, but neither does it appear to acknowledge it, hence the uncertain as to classifying it as a bug.

When using it, I was expecting images that were already sized within the bounds would be left untouched. The use case being to define an area in which a arbitrary displayable could be placed while being certain that larger displayables would be suitably scaled down to fit the area on an as needed basis. Unfortunately the upscaling negatively impacts the situation when using a smaller displayable.

If intended, `maxsize` seems like an odd name, and perhaps `minsize` (upscale only), `maxsize` (downscale only) and `fitsize`(? there are probably better names) (current `maxsize` behaviour), might be worth considering?

Test case follows:

```renpy
label main_menu:
    $ _confirm_quit = False
    return


image box = Solid('f777', xysize=(150, 150))


label start:
    scene
    show box at top
    'normal box 150x150'
    show expression Transform('box', maxsize=(100, 250)) at topleft
    'maxsize 100x250 - {color=7f7}shrinks{/color} to 100x100'
    show expression Transform('box', maxsize=(250, 100)) at topright
    'maxsize 250x100 - {color=7f7}shrinks{/color} to 100x100'
    show expression Transform('box', maxsize=(250, 250)) at truecenter
    'maxsize 250x250 - {color=f77}grows{/color} to 250x250'
    'fin'
    return
```

Potential fix if no upscaling is desired (**untested**):

https://github.com/renpy/renpy/blob/6754d50c12e58914bb6d5760946b074545c25b00/renpy/display/accelerator.pyx#L208
```diff
-        mul = min(maxsizex / width, maxsizey / height)
+        mul = min(maxsizex / width, maxsizey / height, 1)
```

mal:
This now also applies to `minsize`, which will present the same problem in the inverse, shrinking any images that already exceed the minimum size specified.

renpytom:
Since we have the ``fit`` property now, this is resolved. Right?

mal:
Yup 👍

PR #2230, opened 2020-06-11
add minsize transform property

Gouvernathor:
I added the `minsize` transform property.
It's useful when trying to fit a displayable to the background without changing its proportions, for example.
If maxsize is specified, minsize is ignored (better to have an underfull box than to have it overflow).

To paraphrase the words of the [doc](https://www.renpy.org/doc/html/atl.html#transform-property-maxsize) :
If not None, causes the displayable to be scaled so that it fills a box of this size, while preserving aspect ratio (Note that this means that one of the dimensions may be greater than the size of this box.). Maxsize takes priority over minsize.

### DIFF
--- a/renpy/atl.py
+++ b/renpy/atl.py
@@ -119,6 +119,7 @@ PROPERTIES = {
     "crop_relative" : bool,
     "size" : (int, int),
     "maxsize" : (int, int),
+    "minsize" : (int, int),
     "corner1" : (float, float),
     "corner2" : (float, float),
     "subpixel" : bool,
--- a/renpy/display/accelerator.pyx
+++ b/renpy/display/accelerator.pyx
@@ -202,11 +202,18 @@ def transform_render(self, widtho, heighto, st, at):
     # Size.
     size = state.size
     maxsize = state.maxsize
+    minsize = state.minsize
 
-    if (maxsize is not None) and (width != 0) and (height != 0):
-        maxsizex, maxsizey = maxsize
-        mul = min(maxsizex / width, maxsizey / height)
-        size = (width * mul, height * mul)
+    if (width != 0) and (height != 0):
+        mul = False
+        if (maxsize is not None):
+            maxsizex, maxsizey = maxsize
+            mul = min(maxsizex / width, maxsizey / height)
+        elif (minsize is not None):
+            minsizex, minsizey = minsize
+            mul = max(minsizex / width, minsizey / height)
+        if mul:
+            size = (width * mul, height * mul)
 
     if (size is not None) and (size != (width, height)) and (width != 0) and (height != 0):
         nw, nh = size
--- a/renpy/display/accelerator.pyx
+++ b/renpy/display/accelerator.pyx
@@ -205,14 +205,14 @@ def transform_render(self, widtho, heighto, st, at):
     minsize = state.minsize
 
     if (width != 0) and (height != 0):
-        mul = False
+        mul = None
         if (maxsize is not None):
             maxsizex, maxsizey = maxsize
             mul = min(maxsizex / width, maxsizey / height)
         elif (minsize is not None):
             minsizex, minsizey = minsize
             mul = max(minsizex / width, minsizey / height)
-        if mul:
+        if mul is not None:
             size = (width * mul, height * mul)
 
     if (size is not None) and (size != (width, height)) and (width != 0) and (height != 0):
--- a/renpy/display/transform.py
+++ b/renpy/display/transform.py
@@ -112,6 +112,7 @@ class TransformState(renpy.object.Object):
     ytile = 1
     last_angle = None
     maxsize = None
+    minsize = None
 
     def __init__(self):
         self.alpha = 1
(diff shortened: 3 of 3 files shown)


Gouvernathor:
I was always struggling to have backgrounds fill the entire screen while being zoomed in as least as possible.
For images larger than 1280*720, I used either `maxsize (1280, 7000)` ore `maxsize (7000, 720)`, depending on whether the ratio of the image was wider than 16/9 or not. But it was quite inefficient, as I needed to manually check the ratio of the image, and also it would cause bugs if I happened to have an image wider or taller than the specified 7000 pixels.
So I wondered why there was no `minsize` transform property, and there it is.

Also if you merge it, the doc would need to be updated. I wrote a description for it that you can use.

renpytom:
I like this. Would you mind changing the initialization of mull to None, and then the test to ``mul is not None`` ? I don't like using False mixed with a number, since False in Python has the integer value of 0.

renpytom:
Merged, thank you.

Gouvernathor:
Well, thank you :)

mal:
I believe this will suffer from the same problem as `maxsize` reported in #1987 only inverted.

`maxsize` counter intuitively scales up images that are smaller than the provided bounds.
`minsize` as implemented will counter intuitively scale down images that are larger than the provided bounds.

I believe the fix for these are
```py
mul = min(maxsizex / width, maxsizey / height, 1) # maxsize - do not scale up (can't exceed 1x)
```
and
```py
mul = max(minsizex / width, minsizey / height, 1) # minsize - do not scale down (can't fall below 1x)
```

respectively.

This would prevent `maxsize` from enlarging images that already fit within its bounds, and prevent `minsize` from shrinking images that already exceed its bounds.

Gouvernathor:
That woud be changing the purpose of the properties in themselves, and it would ruin the compatiibility for almost anyone using it.
More clarity on the behavior of `minsize` and `maxsize` would be a good thing, but simply changing it would not.

With that being said (and after some thought given to it), turning the 2 properties into 4 would be over-complicated, since the limited maxsize and the full maxsize would have no business being used in parallel.
What I've thought about is to create a boolean "flag" property, kinda like `crop_relative` for example, which would affect the behavior of `minsize` and `maxsize`, switching it between its current, "full" behavior and the limited one you propose. And for sake of compatibility, its default value would be the one matching it current behavior.

I have no idea how to call that flag, however. Tom, would you be interested in a solution like that ?

mal:
The point about compatibility is well taken, which is partly why I wanted to raise this prior to it making it into a release, but I do realise creators are already using `maxsize`. With that said, there are feature flags in Ren'Py to facilitate this kind of change between versions while maintaining compatibility for existing games.

Therefore my suggestion would be to add a `config.allow_maxsize_scaleup` compatibility variable which would default to `False` but be set to `True` for scripts built in versions prior to `7.3.5` (I believe that's how such flags are meant to work, but not 100%), to govern whether `maxsize` would result in upscaling, and to alter the implementation of minsize to never support shrinking in the first place.

My preference for altering the default behaviour comes from a standpoint of least confusion; something named `minsize` shouldn't be scaling anything down that already exceeds the defined minimum, in the same way we wouldn't expect `xmaximum` properties to force something smaller to be expanded to that size, but I do appreciate that it's not a decision to be taken lightly.

It definitely feels like there's room for a separate transform property to best-effort scale a displayable to fit a bounding box beyond the existing `maxsize` and `minsize` constraints. I spent a bit of time trying to think out the various possibilities and came up with this by way of example:

fit

If that makes sense, I'd be happy to convert it into an issue/feature request?

renpytom:
My current thinking is that we should have a size_limit property, that can be None or a number, but usually 1.0. If the number is not None, the mul becomes min(mul, size_limit).

Gouvernathor:
So that would be an upper limit for the zoom applied to the displayable by the minsize ? I find that quite complex, but that's just my impression.

The more I think about it, the more I think mal's solution using `config.allow_maxsize_scaleup` is the best, given we can make it work.
The "alternate" 4 proposed properties are cool, but I think passing it as a single property (since they can't be added simultaneously), called `fit`, defaulting to None of course, accepting a str, either "x", "y", "in" or "over".

The "zoom limit" of some sorts proposed by Tom could be implemented by zooming the displayable with the specified factor using a first transform, then applying it the `Transform(fix="in")` transform. Yes it would be quite complex but a `size_limit` of 1.0 would already be built in the `fit="in"` property.

What do you think ?

renpytom:
I'm warming up to the fit property, perhaps as a replacement for minsize, and perhaps deprecating maxsize in the process. It probably would make sense to have fit interact with size. So could have an fix "scale" to make  the displayable the size given, (so scale would have to be the default), and then you could change it to x, y, inside, over, etc.

mal:
😄
I was writing up a very similar suggestion as you posted - with regard to the `size` property being useful in this context. I'll create a new issue for it and link back here. 👍

mal:
There we go! I've tried to capture the ideas shared here and think about how we might present it such that a sane guess could be made as to it's meaning/intent if stumbling into a transform definition without being completely familiar with docs. Hopefully I've not missed or butchered anything, but please have a look and see what you think. 😁

Issue #2235, opened 2020-06-11
Proposed feature to help "fit" displayables


mal:
Originates from discussion in https://github.com/renpy/renpy/pull/2230.

tl;dr: Current `maxsize` behaviour leaves the door open for confusion, and there are many related use-cases it would be nice to support. Below is a table of identified use-cases and a proposal based on the ideas shared in the original discussion.

<details>
<summary>Identified use-cases (click to expand)</summary>

![fit](https://user-images.githubusercontent.com/591257/84430861-3b2ea680-ac22-11ea-8de3-91c8b4a7db60.png)
</details>

---

The original discussion raises the prospect of a new `fit` property for transforms, that same concept is shown here but named `size_type` to more strongly link it with the `size` property but this is only a suggestion:

_Comments refer back to the identified use-cases._
```rpy
# size_type := (exact|fit|over|min|max) default: exact

# minsize - scales up until the smaller dimension (proportionally) matches or exceeds it's bound
size (300, 200)
size_type 'min'

# maxsize - scales down until the larger dimension (proportionally) matches or lies below it's bound
size (300, 200)
size_type 'max'

# fit x - scales proportionally until the width matches xsize
xsize 300
size_type 'fit'

# fit y - scales proportionally until the height matches ysize
ysize 200
size_type 'fit'

# fit inside - scales proportionally until the displayable fits inside the bounding box
size (300, 200)
size_type 'fit'

# fit over - scales proportionally until the displayable is able to cover the bounding box
size (300, 200)
size_type 'over'
```

The default would be current behaviour when setting `size`, with the introduction of `xsize` and `ysize` to facilitate only specifying a single dimension, and `size_type` to govern how the sizing is done. This would then allow the suggested deprecation of `minsize`/`maxsize` as both old and new cases should be covered. Additionally, if only one of `xsize`/`ysize` is set, the default `size_type` should probably become fit.

One limitation with this approach is that it wouldn't be possible to set separate min and max dimensions on the same transform, but I don't believe this is supported at the moment either. It could also be achieved relatively painlessly with one additional nested transform to allow for what is hopefully a relatively rare edge-case.

Gouvernathor:
OK, so the basic behavior of `xsize` and `ysize` : I think the best would be to store `xsize` and `ysize` in `size`, overriding the preceding values if necessary. For example, something like :
`if size or xsize or ysize:`
    `size = (xsize or size[0], ysize or size[1])`
[...]

PR #2236, opened 2020-06-12
size_type, and all that stuff around maxsize

Gouvernathor:
Implementation of #2235 
At the moment the code is untested, and it completely removes maxsize and minsize.
And also it contains some unnecessary comments that I will clean up before merging.
Also I'm unaware of other places the `size` property might be used, since it's a pretty common keyword, so there could be places where adding `size=(xsize, ysize)` may be necessary.

Minsize can be removed with no consequence, but I don't really know what to do with maxsize.
Should I make `maxsize=(a, b)` an alias of `size=(a, b), size_type="over"` to support retro-compatibility ? If so, with what priority over the `size`, `xsize` and `ysize` properties ?


Gouvernathor:
Also I've thought of another possibility, without adding xsize nor ysize : since specifying both is useless (since putting the two as `size` is equivalent), when we want to x-fit, why not passing "xfit" as `size_type` then passing the sole int as the `size` property ? And the same for yfit of course.
It would require to check more closely if the `size` property isn't used elsewhere. And it would mean the `size` property not necessarily being an iterable. But I think it would solve a lot of questions about priority of one property above another, and it would make everything simpler to understand.

It's also possible to keep `size` as an `(int, int)` tuple, but just considering the relevant part of `size` when the user asks for "xfit" or "yfit". It would require the user to put an irrelevant value alongside the relevant one, but all problems of simplicity, priority and of other uses of the size property would be solved.

For example, instead of `xsize=400, size_type="fit"` or `xsize=400`, you would do `size=400, size_type="xfit"` (1st possibility), or `size=(400, 678), size_type="xfit"` (2nd possibility).

I'm not 100% convinced about these new solutions, but anyway here they are.
They are not in the code I commited so far.

renpytom:
I merged in #2241, which I think is a more complete and expandable version of this. Thank you for pushing in this direction.

PR #2242, opened 2020-06-13
Remove `\n` replacement in log files, because `io.open` is dealing with it

Andykl:
Without it, the raw line is written to the file as `LINE\r\r\n` and some text editors treat it as Mac line endings.

### DIFF
--- a/renpy/log.py
+++ b/renpy/log.py
@@ -159,8 +159,6 @@ class LogFile(object):
             if not isinstance(s, str):
                 s = s.decode("latin-1")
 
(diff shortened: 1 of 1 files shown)


uyjulian:
Instead of deleting that line, I would probably do
```py
s = s.replace("\r\n", "\n").replace("\r", "").replace("\r\n", "\n")
```
to normalize line endings.

renpytom:
I think Andy's version is simpler and suitable. The logging system should never generate a \r on its own, so I think it makes sense for io.open to handle this as part of text mode processing.

PR #2252, opened 2020-06-22
Fix typo (draw -> gl2draw)

dogtopus:
Fixes `AttributeError` while trying to use `renpy.gl2.gl2draw.GL2DrawingContext.draw` with a Surface.


renpytom:
Merged, thank you.

PR #2258, opened 2020-06-26
Python3 Readiness

Daniihh:
I have made a few adjustments to project code to allow for compiling and running with Python 3.

## Changes
- Replaced all instances of `except <type>, <var>` with `except <type> as <var>` in Python and RenPy code. The ladder format is valid in both Python 3.X and 2.7.
- Replaced all instances of `<dict>.has_key(<key>)` with `<key> in <dict>` in Python and RenPy code. The ladder format is also valid in both Python 3.X and 2.7.

## Goals
- Do the bare minimum to let RenPy use Python 3.X.
- To be a stepping stone in getting RenPy completely Python 3.X compliant.

### Non-Goals
- Change how RenPy projects are packaged or provide configuration options to change the Python runtime's version.
- Refactor code that uses deprecated methods / functions. (Except of those deprecated methods / functions that have been *removed*.)

## Notes
Before, you would need to use the `python2` command to build and run RenPy projects. This pull request allows the `python` / `python3` command to be able to do the same things, without errors. Using Python 3 will run RenPy code using Python 3, this can be seen by adding this snippet to anywhere in a RenPy script:
```renpy
python:
  import sys
  print(sys.version_info)
```

### DIFF
--- a/renpy/audio/audio.py
+++ b/renpy/audio/audio.py
@@ -72,7 +72,7 @@ def load(fn):
 
     try:
         rv = renpy.loader.load(fn)
-    except renpy.webloader.DownloadNeeded, exception:
+    except renpy.webloader.DownloadNeeded as exception:
         if exception.rtype == 'music':
             renpy.webloader.enqueue(exception.relpath, 'music', None)
         elif exception.rtype == 'voice':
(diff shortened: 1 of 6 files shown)


daihaminkey:
👍

Boundarybreaker (approved):

renpytom:
Merged, thank you.

PR #2263, opened 2020-07-01
Delete file

DaniilPetuhov:
Test


renpytom:
I don't think this is an intended PR, so I'm closing it.

PR #2278, opened 2020-07-10
Handle variation sequences (VS1-256 and FVS1-3)

### LINKED ISSUE #2272: Feature request: Proper emoji rendering support
Currently attempting to render emojis result in a lot of weird behaviors.

- Noto Color Emoji font fails to load properly with an error saying "loca" table not found.
- Noto Emoji loads, but rendering is broken (`heart emoji` (`U+2764 U+FE0F`) shows as `heart character` (`U+2764`) and a placeholder, `skull` shows as 2 placeholders. Both cases the characters are clearly present in the set but aren't rendered properly)

Since modern emojis provide a lot of useful symbols in a standard and self-documenting way (no more inserting some pictures as characters, what you see in script is what you get on-screen if the fonts are the same) and AFAIK newer freetype does handle emojis (even the colored ones) it makes sense to support them properly.

dogtopus:
This partially fixes #2272, specifically the heart case where the variation selector character is ignored and rendered as a placeholder instead of getting respected properly.

### DIFF
--- a/module/include/freetype.pxd
+++ b/module/include/freetype.pxd
@@ -423,6 +423,7 @@ cdef extern from "pyfreetype.h":
 
     FT_Int FT_Get_Charmap_Index(FT_CharMap charmap)
     FT_UInt FT_Get_Char_Index(FT_Face face, FT_ULong charcode)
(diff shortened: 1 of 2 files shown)


renpytom:
Merged, finally. Thank you.

PR #2280, opened 2020-07-12
Set start_interact before call start_interact_callbacks

Andykl:
This change allows to get the correct value of `start_interact` (via renpy.is_start_interact) in `config.start_interact_callbacks`.
It allows to use the same function in also `config.interact_callbacks` and reduce code duplication.


renpytom:
Thanks!

PR #2287, opened 2020-07-18
web: mobile: browser virtual keyboard support

Beuc:
Currently players on mobile web are stuck whenever the game asks text input (typically "what's your name").

As usual web support is tricky, and there's no API for the mobile browser virtual keyboard. My leads are:

(1) Map a HTML text input with SDL2 (character-based): cf. https://github.com/emscripten-ports/SDL2/issues/80 -- APIs/events don't match, security policies are in the way, little external interest, probably not feasible

(2) Implement a virtual keyboard within Ren'Py: sounds OK; it won't support many characters or input methods but integration would be best

(3) Map a HTML text input with Ren'Py (line-based): sounds OK; supports the characters / input methods that the browser supports, integration is a bit rough (translucent text input at the top) but works

(4) Give up and tell devs/porters to do their [own keyboard in Screen Language](https://lemmasoft.renai.us/forums/viewtopic.php?f=51&t=25972): hopefully not, I think we can provide a default something that works in the common case 

This PR is a working attempt at (3).

I just see now that there's some issue with Firefox who simulates _some_ keys events (space, /... but not letters) and messes things up, so this isn't ready yet, but I thought I'd be good to start the conversation.

virtual-keyboard

### DIFF
--- a/renpy/display/behavior.py
+++ b/renpy/display/behavior.py
@@ -152,6 +152,8 @@ def compile_event(key, keydown):
 
     return rv
 
+if renpy.emscripten:
+    import emscripten, json
 
 # These store a lambda for each compiled key in the system.
 event_cache = { }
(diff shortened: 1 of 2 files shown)


renpytom:
It's no worse that what I've seen in billion dollar android games. Merging.

Beuc:
> > I just see now that there's some issue with Firefox who simulates _some_ keys events (space, /... but not letters) and messes things up, so this isn't ready yet, but I thought I'd be good to start the conversation.

> Merging.

Thanks, but as I explained I didn't intend for this to be merged as-is right away :/

I was about to revert the merge but it seems this is a [tricky case](https://git.github.io/htmldocs/howto/revert-a-faulty-merge.html).
It's probably best to revert the merge commit and recreate this branch from scratch (and a new issue/non-PR for continuing the conversation). What do you think?

PR #2304, opened 2020-08-02
Refactor of renpy.loader

uyjulian:
`renpy.loader` is refactored so it is easier to add new loader callbacks and new archive formats.


renpytom:
This and #2305 look like they want to add support for python-only modules that load files. 

I'm not completely convinced this is a good idea. I'd think any loader callback would need to make sure to handle SubFile properly (it's key to performance, since it means that various modules do not need the gil when reading files).

uyjulian:
This would only support uncompressed files, since it would return an index (like the current RPA implementation)

PR #2305, opened 2020-08-02
Add support for loading pyo and pyc using RenpyImporter

uyjulian:
pyo and pyc can now be loaded with `RenpyImporter`.

### DIFF
--- a/renpy/loader.py
+++ b/renpy/loader.py
@@ -26,6 +26,9 @@ import renpy
 import os.path
 from pickle import loads
 import sys
+import marshal
+if sys.version_info[0] == 3:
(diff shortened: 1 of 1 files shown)


renpytom:
I don't like this as something to be included in Ren'Py proper. I think it's possible to write a Python file to load these files in, if you want to - but I think encouraging .pyo/.pyc files will have the effect of making games more fragile - if Ren'Py is updated to an incompatible version of Python, then the .pyo/.pyc files would break. 

This isn't really an issue on the 2.7 side of things, but Python 3 support is reasonably close (more or less), so I don't really want to add this.

PR #2307, opened 2020-08-02
Disable automatic garbage collection during loading of pyanalysis/script/slast data

uyjulian:
Improves startup performance.

### DIFF
--- a/renpy/pyanalysis.py
+++ b/renpy/pyanalysis.py
@@ -30,6 +30,7 @@ import ast
 
 import zlib
 from pickle import loads, dumps
+import gc
 
 # The set of names that should be treated as constants.
 always_constants = { 'True', 'False', 'None' }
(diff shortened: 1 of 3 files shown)


renpytom:
I don't like the idea of entirely disabling garbage collection. What do you think about tweaking the settings, so it happens less often (or not at all). That's what we do after loading is finished, and I don't think we should be less aggressive about garbage collection when performance is less important (init) then when trying to keep up a framerate.

uyjulian:
How about disabling automatic garbage collection at pickle loads?

renpytom:
I think that makes more sense, though I think setting the GC threshold higher is better than a full disable.

uyjulian:
I moved the disabled portions of automatic garbage collection to localized areas where pickle takes a while with automatic garbage collection enabled.

renpytom:
I wrote  ab8c1bb4 as an alternative to this, since I think it makes sense to up the gc threshold, rather than disabling it entirely.

uyjulian:
I think that's good.

PR #2318, opened 2020-08-12
Added the new Japanese linebreaking algorithm using Janome

kyouryuukunn:
Hello, pytom.

I allowed a user to use the new Japanese linebreaking algorithm "japanese-janome" using janome.

"japanese-janome" breaks line by `janome <https://github.com/mocobeta/janome>`_, 
which is Japanese Morphological Analysis library. This breaks the line
after a postpositional particle and below characters::

    )]）｝〕〉》」』】〙〗｠?!？！‼⁇⁈⁉、。

but, doesn't break line before below characters::

    ,)]）｝、〕〉》」』】〙〗〟’”｠»ゝゞァィゥェォッャュョヮヵヶぁぃぅぇぉっゃゅょゎゕゖㇰㇱㇲㇳㇴㇵㇶㇷㇸㇹㇷ゚ㇺㇻㇼㇽㇾㇿ々〻"ー‐゠–〜～?!‼⁇⁈⁉・:;/。.？！

A user should install janome and pkgutil into his game directory if he use this.::

    pip install --target game/python-packages janome

Then, copy pkgutil.py into the same python-packages directory from https://github.com/python/cpython/blob/2.7/Lib/pkgutil.py.

This is very natural but fairly slow line-breaking algorithm, so isn't recommended for huge amounts of text, such as in history.

A user should import janome in init block and place presplash.png into his game directory because it takes dozens of seconds.::

    init python:
        from janome.tokenizer import Tokenizer
        renpy.text.textsupport._janome_tokenize = Tokenizer().tokenize

janome is under Apache License 2.0 and uses the MeCab-IPADIC dictionary/statistical model, so I suggest using the below wording in his project's README file or App Store description to use janome:

    This program contains Ren'Py and below software.

    Ren'Py contains free software licensed under a number of licenses,
    including the GNU Lesser General Public License. A complete list of software
    is available at http://www.renpy.org/doc/html/license.html.

    janome (https://github.com/mocobeta/janome) Apache License 2.0 (http://www.apache.org/licenses/LICENSE-2.0)

    This program also includes MeCab-IPADIC (Nara Institute of Science and Technology)

And he should include a readable copy of
`NOTICE.txt <https://github.com/mocobeta/janome/blob/master/NOTICE.txt>`_ with
every copy of his Ren'Py distribute or display below wording in his application:

    This program includes MeCab-IPADIC (Nara Institute of Science and Technology)

### DIFF
--- a/launcher/game/tl/korean/launcher.rpy
+++ b/launcher/game/tl/korean/launcher.rpy
@@ -30,11 +30,11 @@
 
     # android.rpy:30
     old "To build Android packages, please download RAPT, unzip it, and place it into the Ren'Py directory. Then restart the Ren'Py launcher."
-    new "안드로이드 패키지를 만드려면, RAPT 파일을 내려받은 뒤에 렌파이 디렉토리에 압축 해제하세요. 그 다음 렌파이 런처를 재시작하세요."
+    new "안드로이드 패키지를 만드려면, RAPT 파일을 내려받고 렌파이 디렉터리 안에 압축 해제하세요. 그다음 렌파이를 재시작하세요."
 
     # android.rpy:31
     old "A 64-bit/x64 Java 8 Development Kit is required to build Android packages on Windows. The JDK is different from the JRE, so it's possible you have Java without having the JDK.\n\nPlease {a=http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html}download and install the JDK{/a}, then restart the Ren'Py launcher."
-    new "윈도우에서 안드로이드 패키지를 만드려면 64비트 JDK가 필요합니다. JDK는 JRE와 다르므로 PC에 JDK가 없는 자바가 설치되어 있을 수 있습니다.\n\n{a=http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html}JDK를 내려받아 설치한 뒤{/a}, 렌파이 런처를 재시작해주세요."
+    new "윈도우에서 안드로이드 패키지를 만드려면 64비트 JDK가 필요합니다. JDK는 JRE와 다르므로 PC에 JDK가 없는 자바가 설치되어 있을 수 있습니다.\n\n{a=http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html}JDK를 내려받아 설치한 뒤{/a}, 렌파이를 재시작해주세요."
 
     # android.rpy:32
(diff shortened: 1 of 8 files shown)


renpytom:
Hm... the talk of adding multiple seconds to the startup time scares me. What's causing that? 

I'm sort of wondering if it's the right idea to include this with Ren'Py, or have some sort of tool that can annotate the source or translations with zero-width space characters that provide breaking points, and then have a new annotator that tells Ren'Py to only break at such points.

kyouryuukunn:
Japanese Morphological Analysis needs a big dictionary file about a part of speech.
That file is loaded when Tokenizer class is initialized and it takes time.
That class isn't called if a user doesn't set language to "japanese-janome".

It might be better to post this code on a forum if annotate_functions like ``annotate_unicode`` wasn't in a pyx file, but it needs to be compiled.

kyouryuukunn:
Janome doesn't support python2 after v3.1.10, so I fixed document. 

pip install --target game/python-packages janome==0.3.10

And  I fixed some sample code in document.
I recommended to use janome for only say(nvl)_dialogue so that we can proceed efficiently 

    init python:
        from janome.tokenizer import Tokenizer
        renpy.text.textsupport._janome_tokenize = Tokenizer().tokenize
        style.say_dialogue.language = "japanese-janome"
        style.nvl_dialogue.language = "japanese-janome"

kyouryuukunn:
Hello, Pytom.

I measured the load time. The long load time may be due to what Ren'py does.
I added below code and launched game.
```
init python:
    import time
    start = time.time()
    from janome.tokenizer import Tokenizer
    t = Tokenizer()
    test_time = time.time()-start
```
test_time was 20.25s when I installed janome into the game directory, but test_time was 1.77s when I installed it into lib/python2.7. 

I assume Ren'Py do something when a library is placed in a game directory, but I don't know the solution for it.

I also mesured the load times by python3 and janome v0.4.1 whose load time was improved from v0.4.0.
test_time was 1.94s by python2 and janome==0.3.10
test_time was 1.15s by python3 and janome==0.3.10
test_time was 0.34s by python3 and janome==0.4.1

The load will be faster when Ren'Py supports python3. I think the load time of Janome will not be problem If above problem was solved.

kyouryuukunn:
Sorry, I didn't take non pc platform into acount.
In Android smartphone, it took 4,5 seconds to show text on say screen. In web, Ren'Py couldn't launch.
Caching linebreak is required before ditributing a game.

kyouryuukunn:
This is resolved by #3733.

PR #2324, opened 2020-08-16
Fixes #2292: Use utf-8 encoding when writing to lint file output

### LINKED ISSUE #2292: Check Script (Lint) is not working
Checking the script through Renpy SDK is not working. It's throwing an error screen saying "Launching the project failed.", even though the project launches without any problems when clicking "Launch Project" button. 
Renpy SDK version: 7.4.0.602

kapilgain:
Using codecs.open here to do implicit conversion of any str to unicode while writing to the lint output file. If a filename is not provided as an argument to the lint command, then output is sent to sys.stdout for which the str to unicode conversion is already in place (using future.utils.text_type) in https://github.com/renpy/renpy/blob/master/renpy/log.py#L228-L229

### DIFF
--- a/renpy/lint.py
+++ b/renpy/lint.py
@@ -708,7 +708,7 @@ def lint():
     args = ap.parse_args()
 
     if args.filename:
-        f = open(args.filename, "w")
+        f = codecs.open(args.filename, "w", encoding="utf-8")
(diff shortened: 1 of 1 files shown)


renpytom:
Thank you!

PR #2327, opened 2020-08-18
Normalize unicode to NFC

uyjulian:
Some environments (especially macOS on HFS+ filesystem) will automatically normalize filenames to a customized version of NFD (see also https://gist.github.com/JamesChevalier/8448512 ).

This will allow those files to be found, even if the filesystem is normalization sensitive.

### DIFF
--- a/renpy/loader.py
+++ b/renpy/loader.py
@@ -30,6 +30,7 @@ import threading
 import zlib
 import re
 import io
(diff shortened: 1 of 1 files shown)


renpytom:
Thank you, Apple. /s

Thanks! Merged.

PR #2334, opened 2020-08-26
Add K_key keysyms to default Ren'Py keymap

Andykl:
For "director" and "developer" keysyms are replaced so that CapsLock does not swap them.
For "input_copy" and "input_paste" added noshift modifiers.
In all other cases K_keys are added, so it doesn't break backward compatibility.
Also updated the table in the documentation.


renpytom:
Merged, thank you.

PR #2335, opened 2020-08-26
Fix incorrect docs for pushup/down transitions

jsfehler:
pushup and pushdown were called pushtop and pushbottom.


renpytom:
Merged. IIRC, this was a late change because I didn't want wipetop and wipebottom - more the latter was explicitly not wanted.

PR #2336, opened 2020-08-28
web: mobile: add built-in simple virtual keyboard

Beuc:
This adds a basic, translucent on-screen keyboard programmed in screen language.
It should work in landscape and portrait mode.

It synthesizes SDL keyboard events. I believe I dodged possible conflicts (like sending K_ENTER which triggers the current (enter) button again), targeted the right layer, and restarted interaction appropriately. New screen elements are prefixed with '_'.

It doesn't depend on browser-specific feature as it's 100% Ren'Py, so we can expect browser compatibility.

It's not perfect but it does the job: prevent players from getting stuck when playing with Android/iOS browsers when e.g. the game asks the MC name ;)

Gamedevs may want to customize it, so maybe this needs to be reorganized at a point.

keyb3-2

### DIFF
--- a/renpy/common/00touchwebkeyboard.rpy
+++ b/renpy/common/00touchwebkeyboard.rpy
@@ -1,0 +1,221 @@
+# Copyright 2020 Sylvain Beucler
+#
+# Permission is hereby granted, free of charge, to any person
+# obtaining a copy of this software and associated documentation files
+# (the "Software"), to deal in the Software without restriction,
+# including without limitation the rights to use, copy, modify, merge,
+# publish, distribute, sublicense, and/or sell copies of the Software,
+# and to permit persons to whom the Software is furnished to do so,
+# subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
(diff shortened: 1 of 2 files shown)


Beuc:
Hi, let me know if you agree with this direction :)
Unlike the prior attempt, this PR works.

renpytom:
Honestly,  it seems really ugly, as it is now, which makes me hesitate to want to merge it.  I think if we have a keyboard like this, it should follow a qwerty layout. I'd be fine with including the hooks in core.py (provided they check to see if _touchwebkeyboard exists, but I'd rather people have to find a keyboard, as opposed to standardizing on this design.

Beuc:
Hi,

I picked this layout for the following reasons:
- It's a common RPGMaker* layout, already used in web games
- It works in both portrait and landscape mode
- It's not a US-centric (qwerty) keyboard, nor a anything-centric keyboard for that matter
- It can easily be extended with more pages of characters

I went for a default, out-of-the-box behavior, because:
- That's one fewer thing to worry about when building a web version (there are a lot of [other issues](https://renpy.beuc.net/#faq) to deal with)
- Gamedevs don't realize they need to manually add a keyboard until they play on mobile+web specifically
- There are a lot of web games distributed/hosted by mass web porters, who usually do a quick porting (decompile+rebuild) job and I don't expect them to think about including a keyboard; the consequence is players bashing not the careless porter, but Ren'Py/RenPyWeb in general for apparently lacking keyboard support on mobile.

With that being said, I'm fine with dropping either.
Let me know.

renpytom:
Hm... let me have a go of it, later today. I think my big problem with this is that the keyboard is overlapping the text being typed, and that feels kind of bad to me. What I'm thinking is that if we made each side one or two letters wider, and perhaps put the keyboard inside a frame, we'd get something that looked a lot better and provided the same functionality.

I think the justifications make sense, it's just a visceral reaction to the design, but I think it's possible to make a better design.

renpytom:
I merged this in, with a different layout for the keyboard, and some renaming of things - I don't think there's anything particularly web-specific here, so I've made it an option on non-web platforms and renamed things accordingly.

image

Beuc:
Hi and thanks for the merge!

Last time I saw a "Enter your name" prompt in a Ren'Py game it was in the middle of the screen, hence why I went for an overlapping keyboard.

There was one more bit to rename, which broke mobile browsers only. I pushed a fix.

PR #2351, opened 2020-09-10
Add Sublime Text and VSCode support

### LINKED ISSUE #2348: [Feature/Q] Add VSCode (or VSCodium) to the Suggested Text Editors
When editing a file, some text editors (such as Atom, Editra and jEdit) are suggested to install or open files in.
It'd be awesome to have VSCode (or VSCodium) as a choice here.
VSCode is quite similar to Atom, it brings capabilities to edit and execute python code as well.

Is there some way I can add VSCode as a choice by myself as a workaround?

manuth:
This PR adds support for `Sublime Text` and `VSCode`.
Adding a plugin for Sublime Text or VSCode won't be necessary as the description of the editor provides a link for downloading the editors alongside with the remark to add the editor to `PATH`.

Thanks to @kobaltcore for providing me the code-snippets which are necessary for integrating said editors into `Ren'Py`.

Merging this PR fixes #2348

### DIFF
--- a/launcher/Sublime Text.edit.py
+++ b/launcher/Sublime Text.edit.py
@@ -1,0 +1,10 @@
+import subprocess
+
+import renpy.editor
+
+class Editor(renpy.editor.Editor):
+
+    def open(self, filename, line=None, **kwargs):
+        filename = renpy.exports.fsencode(filename)
+        address = filename if line is None else "%s:%d" % (filename, line)
+        subprocess.call(["subl", arg], shell=True)
--- a/launcher/VSCode.py
+++ b/launcher/VSCode.py
@@ -1,0 +1,10 @@
+import subprocess
+
+import renpy.editor
+
+class Editor(renpy.editor.Editor):
+
+    def open(self, filename, line=None, **kwargs):
+        filename = renpy.exports.fsencode(filename)
+        address = filename if line is None else "%s:%d" % (filename, line)
+        subprocess.call(["code", "-g", address], shell=True)
--- a/launcher/game/editor.rpy
+++ b/launcher/game/editor.rpy
@@ -1,4 +1,4 @@
-﻿# Copyright 2004-2020 Tom Rothamel <pytom@bishoujo.us>
+# Copyright 2004-2020 Tom Rothamel <pytom@bishoujo.us>
 #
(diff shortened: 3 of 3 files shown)


renpytom:
I don't think this is something I want to merge. 

One of the big benefits of Ren'Py is that if you don't pick "System Editor" or "None", it sets the editor up for you and gets you going. This removes that - it expects the user to have these editors set up for you, and for a user that's new to programming, downloading an editor and setting up PATH is likely a bridge too far. 

(Experienced users can just drop in the file.) 

I also don't like the remove of System Editor and None from the translations system - I don't know why we'd want that.

manuth:
Alright, I get you
I wouldn't want you to have to create and maintain another new plugin so we'd best leave it as it is for now.

The second change wasn't about removing "System Editor" and "None" from the translation system.

Yet Ren'Py checks whether editor-files intersect with an existing fancy editor by comparing it against a static list as seen here:
https://github.com/renpy/renpy/blob/cf6409bf719d6ba495e213fb033f325f02e0574e/launcher/game/editor.rpy#L192-L194

My change stores the untranslated version of the "System Editor" and "None" editor as well as all the other fancy editors into a variable in order to prevent intersections:
https://github.com/renpy/renpy/blob/f98c7213e25743a83e410dc7d7bdc69e57faeb51/launcher/game/editor.rpy#L149-L151

And translates the "System Editor" and "None" Editor at a later point:
https://github.com/renpy/renpy/blob/f98c7213e25743a83e410dc7d7bdc69e57faeb51/launcher/game/editor.rpy#L202-L206
https://github.com/renpy/renpy/blob/f98c7213e25743a83e410dc7d7bdc69e57faeb51/launcher/game/editor.rpy#L221-L225

If you want to merge this change I can create a separate PR for that matter

Thanks for taking time to have a look at this one ^^

renpytom:
I think you might be confusing what _ does, which is to just mark the string for translation. It doesn't actually translate anything on its own, that's done when the string reaches a Text displayable. 

In Ren'Py, _("string") == "string" is always true. 

so I don't think these changes are needed.

manuth:
Nice, thanks for the elaboration
I'll close the PR then

PR #2356, opened 2020-09-12
doc: _window_show and _window_hide don't disable auto

jackmcbarn:
Since `_window_show` and `_window_hide` don't do `store._window_auto = False`, they're equivalent to `window auto show` and `window auto hide`, not `window show` and `window hide`.

### DIFF
--- a/renpy/common/000window.rpy
+++ b/renpy/common/000window.rpy
@@ -40,7 +40,7 @@ init -1200 python:
         """
         :doc: window
 
-        The Python equivalent of the "window show" statement.
+        The Python equivalent of the "window auto show" statement.
 
         `trans`
             If False, the default window show transition is used. If None,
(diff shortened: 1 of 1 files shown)


jackmcbarn:
Wait, this still isn't quite right, because `window auto show` and `window auto hide` **enable** store._window_auto, which this doesn't do. These functions leave it alone, which doesn't match how any of the statements work.

renpytom:
b66cf5d6239b54799cce774b32b8b24f490557bc should address this.

jackmcbarn:
Well, it does address this. I'm surprised that you did that though, since it's a breaking change.

renpytom:
Did it actually  break things for you? There's a bit of a judgement call, as it doesn't seem like mixing window auto and _window_show/_window_hide was something that worked before this.

jackmcbarn:
Not for me personally, no, but it certainly seems plausible that there's games that it will break.

renpytom:
A reasonable point. Probably worth a compat variable.

PR #2372, opened 2020-09-27
Optimization for checking archive extension

uyjulian:
Instead of calling `.endswith` for each extension, use `os.path.splitext` then use `in` operator with the extension on the list.


Andykl:
Looks good, merged.

PR #2376, opened 2020-10-02
Option to specify transition in ShowMenu (for #2194)

### LINKED ISSUE #2194: Request: Add transition variable to ShowMenu action
Currently, when the `ShowMenu` action is called, it performs `renpy.transition(config.intra_transition)` to transition to the new menu. It would be useful if `ShowMenu` could take an optional keyword argument that would override this with a different transition for special cases, without changing the default intra-menu transition in the configuration.

I have a custom action defined that handles this behavior, but this seems like something that would be useful to have in-engine.
```python
@renpy.pure
class ShowMenuPlus(ShowMenu):
    """
    A menu action.
    Like ShowMenu, but allows overriding the intra-menu transition
    """

    def __init__(self, screen, _transition=None, *args, **kwargs):
    # note: should use super here, copied for clarity
        self.screen = screen
        self.transition = _transition
        self.args = args
        self.kwargs = kwargs

    def __call__(self):
        if not self.get_sensitive():
            return

        orig_screen = screen = self.screen or store._game_menu_screen
        if not (renpy.has_screen(screen) or renpy.has_label(screen)):
            screen = screen + "_screen"

        # Ugly. We have different code depending on if we're in the
        # game menu or not.
        if renpy.context()._menu:
            if renpy.has_screen(screen):
                renpy.transition(self.transition or config.intra_transition)
                renpy.show_screen(screen, _transient=True, *self.args, **self.kwargs)
                renpy.restart_interaction()

            elif renpy.has_label(screen):
                renpy.transition(self.transition)
                ui.layer("screens")
                ui.remove_above(None)
                ui.close()
                renpy.jump(screen)
            else:
                raise Exception("%r is not a screen or a label." % orig_screen)

        else:
            renpy.call_in_new_context("_game_menu", *self.args, _game_menu_screen=screen, **self.kwargs)
```

GiovanH:
Draft to implement #2194 (Add transition variable to ShowMenu action), please review.

Please also do note that this is technically a breaking change, since the `transition` kwarg was previously consumed by `**kwargs` and passed to the screen. `transition` could be renamed, but any new kwarg to `ShowMenu` has this issue.

Resolves #2194 

~~this isn't hacktober spam I promise~~

### DIFF
--- a/renpy/common/00action_menu.rpy
+++ b/renpy/common/00action_menu.rpy
@@ -58,8 +58,9 @@ init -1500 python:
          Extra arguments and keyword arguments are passed on to the screen
          """
 
-        def __init__(self, screen=None, *args, **kwargs):
+        def __init__(self, screen=None, transition=None, *args, **kwargs):
             self.screen = screen
+            self.transition = transition
             self.args = args
             self.kwargs = kwargs
 
@@ -83,7 +84,7 @@ init -1500 python:
 
                 if renpy.has_screen(screen):
 
-                    renpy.transition(config.intra_transition)
+                    renpy.transition(self.transition or config.intra_transition)
                     renpy.show_screen(screen, _transient=True, *self.args, **self.kwargs)
                     renpy.restart_interaction()
 
--- a/renpy/common/00action_menu.rpy
+++ b/renpy/common/00action_menu.rpy
@@ -58,9 +58,9 @@ init -1500 python:
          Extra arguments and keyword arguments are passed on to the screen
          """
 
-        def __init__(self, screen=None, transition=None, *args, **kwargs):
+        def __init__(self, screen=None, _transition=None, *args, **kwargs):
             self.screen = screen
-            self.transition = transition
+            self.transition = _transition
             self.args = args
             self.kwargs = kwargs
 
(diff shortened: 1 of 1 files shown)


renpytom:
Yeah, this is a breaking change, so can't be accepted. What might work would be to look for _transition in kwargs, and if it's present use that as the transition. Variables beginning with _ are reserved for Ren'Py.

GiovanH:
It can definitely be called `_transition`, although that's a little less clear and less parallel with how `screen` works. It's still technically a breaking change, since screens might try to consume `_transition`, but if `_transition` better fits the specification, that should be okay.

renpytom:
This still won't work when it's a positional parameter like that - it will change what the first element of args is. 

What you need to do is something like:

self.transition = kwargs.pop("_transition", None)

You also need to put transition = None as part of the class, to keep save compatibility. If that's not present, then trying to use a ShowMenu found in a save file won't work.

GiovanH:
Oh, good catch, I missed that, thanks.

renpytom:
I'd suggest a phrasing like "If the keyword argument _transition is given...", and then this is probably good for merging.

renpytom:
Looks good.

GiovanH:
Okay, marked.

PR #2384, opened 2020-10-04
Fix a bug that would cause the secondary volume to reset in some cases

### LINKED ISSUE #2383: set_volume before loading music does not seem to apply properly in 7.4 nightly
Hello,

When attempting to set the volume of music before playing the music, the setting of the volume seems to be ignored.

Take a look at the following code snippet.

```py
$renpy.music.set_volume(0.1, delay=0.0,  channel='music');
play music "audio/backgroundSound/whitenoise.wav"
"This sound should be quiet."
$renpy.music.set_volume(1.0, delay=0.5, channel='music');
"This sound should be loud."
```

I would expect it to play a quiet sound on the first text line, then become louder on the second text line.

In 7.3.5, this works as expected.

In the latest nightly (nightly-2020-10-04-58c02c9a3), however, it does not seem to work, and instead plays the sound at full volume.

kobaltcore:
Fixes #2383 and stems from a discussion with @Andykl 

When calling `set_tertiary_volume` the secondary volume would be reset due to a magic number remaining in the code.
This patch fixes this by introducing a storage field for the raw value of the secondary volume, which it subsequently uses when setting the tertiary volume, thus keeping the propagation intact and correct.

### DIFF
--- a/renpy/audio/audio.py
+++ b/renpy/audio/audio.py
@@ -143,6 +143,7 @@ class MusicContext(renpy.python.RevertableObject):
 
     pause = False
     tertiary_volume = 1.0
+    raw_secondary_volume = 1.0
 
     def __init__(self):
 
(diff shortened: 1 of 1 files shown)


kobaltcore:
Made a small change in presentation at the behest of Andy to avoid introducing new fields.
This now computes the actual volume on demand and leaves the two fields for secondary and tertiary volume unchanged, making for an overall more stable interface.

PR #2397, opened 2020-10-11
Garbage collection related changes

uyjulian:


renpytom:
I like the changes that fix the cleaning of gc.garbage, but I don't believe that we should ever fully disable garbage collection. So I don't want to include a context manager that makes it easy to do so.

uyjulian:
Sure. I trashed the last two commits.

renpytom:
Thank you, I've gone ahead and merged this in.

PR #2398, opened 2020-10-11
When OOM error occurs during image loading, clear memory and try again

uyjulian:


uyjulian:
I don't think this is needed, so I'm closing

PR #2400, opened 2020-10-11
Add ffmedia MediaState resource manager

uyjulian:
The resource manager thread will handle thread joining and `MediaState` deallocation.  

This will avoid usage of `SDL_DetachThread`

### DIFF
--- a/module/ffmedia.c
+++ b/module/ffmedia.c
@@ -141,6 +141,8 @@ typedef struct SurfaceQueueEntry {
 
 typedef struct MediaState {
 
+	/* The thread. */
+	SDL_Thread* thread;
 
 	/* The condition and lock. */
 	SDL_cond* cond;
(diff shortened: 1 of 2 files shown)


renpytom:
What are you trying to accomplish with this?

renpytom:
In more detail - I'm not sure why we'd want to make big changes to avoid the use of SDL_DetachThread - does it not work somewhere?

uyjulian:
Switch. Thread detaching is not implemented on that platform, so it will eventually run out of thread resources due to thread handle cleanup not being performed.

renpytom:
4b0849723bb6e95d8f69924e12245670c3b8ee80 is my go at this, which just puts the MediaState onto a linked list, and walks the linked list to wait for threads and deallocate the MediaState objects. 

I think this is a bit simpler than your approach to it, and pretty clearly can't deadlock, as the code inside the new lock is guaranteed to terminate in small amounts of time. I wasn't completely sure about how ms_manager_cond was being used in this PR.

uyjulian:
That is a much better and cleaner solution.

PR #2409, opened 2020-10-18
Add Gtk support for project directory chooser

joshua-stone:
This PR hopes to make RenPy's project directory chooser more robust by supporting multiple backends. In theory Tkinter should be readily-available on Linux distributions, but in practice I find it surprisingly difficult to set up when trying to build and run RenPy when python3-tkinter is an optional package that doesn't come preinstalled on many distributions.

Neither the [Freedesktop SDK](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/blob/master/NEWS) nor the [Gnome SDK](https://gitlab.gnome.org/GNOME/gnome-build-meta/-/tree/gnome-3-38/elements/sdk) include Tkinter, even excluding it from their Python builds, and even more problematic that there doesn't appear to be a complete tkinter package available on PyPi. Gtk on the other hand is much better supported and is straightforward to set up on Linux.

I've only added dialog messages referencing PyGObject in the English translation so far. More languages can be added if there's interest in upstreaming this support.

### DIFF
--- a/launcher/game/choose_directory.rpy
+++ b/launcher/game/choose_directory.rpy
@@ -90,10 +90,10 @@ init python:
                 choice = ""
                 path = None
 
-                interface.error(_("Ren'Py was unable to run python with tkinter or PyGObject to choose the directory. Please install the python3-tk, tkinter, or python3-gobject package."), label=None)
+                interface.error(_("Ren'Py was unable to run python with tkinter to choose the directory. Please install the python3-tk or tkinter package."), label=None)
 
             if code:
-                interface.error(_("Ren'Py was unable to run python with tkinter or PyGobject to choose the directory. Please install the python3-tk, tkinter, or python3-gobject package."), label=None)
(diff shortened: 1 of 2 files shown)


renpytom:
I am leaning towards accepting 9709bec without 5a11ff8. I think it makes sense to try both gtk and tkinter, but only mention the tkinter failure in the error messages. My thinking is that installing tkinter will fix the problem, and not having to retranslate this high-probability error message into a bunch of languages is will benefit the end user. 

What do you think?

joshua-stone:
I think accepting just 9709bec should work given the circumstances. Seeing as RenPy has used Tkinter for a long time it might be a bit premature to add complexity to error messages when it has to be consistent on all major platforms.

Shall I revert 5a11ff8?

renpytom:
Please, and then I'll merge it.

joshua-stone:
5a11ff8 has been reverted!

renpytom:
Merged, thank you.

PR #2411, opened 2020-10-19
Allow grids to be underfull

shawna-p:
This update adds a config variable, `allow_underfull`, which is `False` by default. If set to `True`, Ren'Py will display grids which are not full.

### DIFF
--- a/renpy/config.py
+++ b/renpy/config.py
@@ -459,6 +459,9 @@ log_enable = True
 # Should we log text overflows?
 debug_text_overflow = False
 
+# Should underfull grids raise an exception?
+allow_underfull = False
+
 # Should we save the window size in the preferences?
 save_physical_size = True
 
--- a/renpy/display/layout.py
+++ b/renpy/display/layout.py
@@ -434,7 +434,7 @@ class Grid(Container):
         rows = self.rows
 
         if len(self.children) != cols * rows:
-            if not config.allow_underfull and len(self.children) < cols * rows:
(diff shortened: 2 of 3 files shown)


renpytom:
Did you intend to close this?

shawna-p:
This is my first pass at allowing grids to display without being full. If it suits your needs, you're welcome to merge it. I was thinking I'd try working on making this idea into a style property instead, so it can be set on a per-grid basis or for all grids by changing the default style. The two ideas could be compatible, or I could submit a pull request for the second method later which removes this method in favour of the style property method.

renpytom:
I don't think it should be a style. I think it probably could be a config variable - config.allow_underfull_grids, with allow_underfull being a True/None/False choice, where None means "use config.allow_underfull". I'd want to avoid adding it to a style, as that takes up a lot of ram, and winds up with questions like "what does it mean if idle_allow_underfull is True, and hover_allow_undefull is False?"

renpytom:
... and then I actually re-read your code. I actually think this is something I can merge right now.

shawna-p:
Great! You're welcome to rename the variable itself to something more suitable if you prefer, or let me know if there's something I can modify to improve it.

Andykl:
In this code, it seems wrong to me that the list of children changes in `render`. I think it would be better to move this check to `per_interact` because render can be run multiple times in one iteration.

Andykl:
Also, `Container` have special method to add children, so it should be used instead of direct children list changing

renpytom:
Merged, thanks to everyone.

PR #2413, opened 2020-10-19
edited the about page to look more professional and detailed.

RukshanJS:
I added some details to the about page. Also made the image a little smaller.


renpytom:
This reads as more promotional than I think the about page should read.

PR #2418, opened 2020-10-21
Make arm64 builds work properly

jackmcbarn:
The Raspberry Pi is currently beta-testing an arm64 OS, and arm64 Chromebooks can also use this via crostini.

### DIFF
--- a/launcher/game/options.rpy
+++ b/launcher/game/options.rpy
@@ -376,6 +376,8 @@ init python:
     build.classify_renpy("lib/**/*steam_api*", None)
     build.classify_renpy("lib/linux-armv7l/", "raspi")
     build.classify_renpy("lib/linux-armv7l/**", "raspi")
+    build.classify_renpy("lib/linux-aarch64/", "raspi")
+    build.classify_renpy("lib/linux-aarch64/**", "raspi")
     build.classify_renpy("lib/**", "binary")
     build.classify_renpy("renpy.sh", "binary")
     # renpy.app is now built from scratch from distribute.rpy.
--- a/launcher/game/distribute.rpy
+++ b/launcher/game/distribute.rpy
@@ -903,6 +903,16 @@ init python in distribute:
                     armfn,
                     True)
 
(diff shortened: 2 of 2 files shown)


launcher/game/options.rpy
```diff
@@ -376,6 +376,8 @@ init python:
     build.classify_renpy("lib/**/*steam_api*", None)
     build.classify_renpy("lib/linux-armv7l/", "raspi")
     build.classify_renpy("lib/linux-armv7l/**", "raspi")
+    build.classify_renpy("lib/linux-aarch64/", "raspi")
+    build.classify_renpy("lib/linux-aarch64/**", "raspi")
```
jackmcbarn, line 380: As I think about this some more, why do we even have "raspi" separated out? Why not just always bundle it in the regular Ren'Py distribution? (Although that'd be something to change in a different PR for sure.)

renpytom: Fundamentally, it's because the audience for Raspberry pi is small enough, that it's not worth increasing the size of the default Ren'Py download. I think it's worth supporting as a way of, say, deploying a dozen demos of a game, but I think the number of people that develop on raspi is kind of tiny.

jackmcbarn: But you need this to play on Raspberry Pi, not just to develop on it. Also, is an extra 20MB that much when so many Ren'Py games these days are 5GB?

renpytom: How many people actually play on a raspberry pi? I don't think there are all that many at the moment, though if you have data that proves that wrong, I'd love to hear it.

I'd tend to think the raspi 4 is the first one that can run desktop-class software without major compromises, and I think I'd rather wait for a supported arm64 OS than focus on a 32-bit distro that's on its way out.

renpytom:
This is out of date, so I'm closing it. I'll be having a go at adding aarch64 support to renpy-build soon. Probably as a way of keeping myself sane during the leadup to the 7.5/8.0 release.

PR #2423, opened 2020-10-26
Add type comments to atl.py

jsfehler:
This is a proposal for adding type comments to the project. The PR itself isn't a complete job on atl.py and is just to show what they would look like.

I don't think I need to evangelize type comments, but adding these across the board increases the probability that major work can continue if Tom gets hit by a bus.

This also opens up the possibility of using mypy or another tool to catch nasty bugs when building the project.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,3 @@
+[mypy]
+ignore_missing_imports = True
+allow_redefinition = True
--- a/renpy/atl.py
+++ b/renpy/atl.py
@@ -29,13 +29,13 @@ import random
 
 
 def compiling(loc):
-    file, number = loc # @ReservedAssignment
+    file, number = loc  # @ReservedAssignment
 
     renpy.game.exception_info = "Compiling ATL code at %s:%d" % (file, number)
 
 
 def executing(loc):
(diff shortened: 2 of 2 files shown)


renpytom:
I like the idea of this, at the very least. Thank you for doing the work. 

I do wonder if the right way would be to write Python 3, and then turn it into Python 2 as part of the build process - but that can probably wait until after 7.4. And in any case, these comments could be automatically converted.

jsfehler:
I've amended the PR to make mypy in particular scream less. I stuck some Any types where I gave up on figuring it out. Sorry.

Ultimately, you're going to have to run mypy locally and make decisions about if it's correct about some errors, they don't apply, or maybe I goofed somewhere. If we can make it okay with atl.py, then follow the errors for every line of code touched by atl.py, we'll end up at a point where merging in type comments is minimally useful to an IDE.

I wouldn't merge this with the ImportError try/catch there. You'd have to add the typing library backport to the project to get it to run normally.

renpytom:
I think this has been done, at least to the poing where VSCode's type analysis isn't complaining. So, I'm closing this.

PR #2427, opened 2020-10-28
Porting ImageDissolve to the new matrixcolor system

Gouvernathor:
Allowed any displayable to take the role of the image controlling the transition.
Formerly only proper images were allowed, due to the use of the im. matrixes.

### DIFF
--- a/renpy/display/transition.py
+++ b/renpy/display/transition.py
@@ -408,9 +408,9 @@ class ImageDissolve(Transition):
     dissolve in first, and black pixels will dissolve in last.
 
     `image`
-        A control image to use. This must be either an image file or
-        image manipulator. The control image should be the size of
-        the scenes being dissolved.
+        A control image to use. This can be any displayable. The control image
+        should be the size of the scenes being dissolved, and if `reverse=True`,
+        it should be fully opaque.
 
     `time`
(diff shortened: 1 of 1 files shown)


Gouvernathor:
Maybe it's not the best place to suggest this (but I don't think it deserves raising an issue), how would you feel about giving (in the doc') a workaround for each documented "Image Manipulators" and `im.MatrixColor` functionality, since they are now all replaced with `Transform` or other ways to do it ?

I could write it in a pull request, but does it look interesting to you, giving such examples in the doc' ?

renpytom:
I think it's too early to do this, unless config.gl2 is set. It's still possible to use the gl renderer - it's still the default. I'd like to get a little more experience with how well gl2 works before people switch over to it. 

I do think documentation about how to replace image manipulators would be good.

Gouvernathor:
About the Matrix business - I can make an if-else switch, like `if config.gl2` the new 4*4 matrix is set then used, `else` the former one. Then, you can merge the pull request if you feel confident enough about gl2's reliability to perform that task.

renpytom:
That seems like a good idea.

Gouvernathor:
I'll get to work with the doc examples.

PR #2428, opened 2020-10-29
Handle common directory in renpy.zip

uyjulian:
Handle common directory in `renpy.zip`.

When the `renpy` folder is added into a zip archive, around 612 files can be combined into 1 file, increasing performance on slow filesystems.

### DIFF
--- a/renpy/loader.py
+++ b/renpy/loader.py
@@ -31,6 +31,8 @@ import zlib
 import re
 import io
 import unicodedata
+import zipfile
+import struct
 
 from renpy.compat.pickle import loads
 from renpy.webloader import DownloadNeeded
@@ -95,6 +97,9 @@ else:
 # Files on disk should be checked before archives. Otherwise, among
 # other things, using a new version of bytecode.rpyb will break.
 archives = [ ]
+
+# A list of paths that should be handled as the Ren'Py common folder.
+common_paths = [ ]
 
 # The value of renpy.config.archives the last time index_archives was
 # run.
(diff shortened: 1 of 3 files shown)


renpy/loader.py
```diff
@@ -96,6 +98,9 @@ def get_path(fn):
 # other things, using a new version of bytecode.rpyb will break.
 archives = [ ]
 
+# A list of paths that should be handled as the Ren'Py common folder.
+paths_renpy_common = [ ]
```
renpytom, line 102: I think this would be better named ``common_paths``.

renpy/loader.py
```diff
@@ -243,6 +248,40 @@ def index_archives():
                     except:
                         raise
 
+    global paths_renpy_common
```
renpytom, line 251: I don't like this function. There are too many constants and byte strings - I don't know what  b"\x50\x4b\x03\x04" means here. 

I'd suggest the following changes:

* Let's not do a search for renpy.zip in sys.path. I think it's better to simply check for it in config.renpy_base, and if it doesn't exist, that's it. 
* https://github.com/renpy/renpy-build/blob/master/runtime/android/apk.py#L134 does something very similar, without knowing as much about how a zip file works. I think this could be based off that.

renpy/loader.py
```diff
@@ -400,10 +439,13 @@ def scandirfiles_from_archives(add, seen):
     Scans archives and fills out game_files.
     """
 
-    files = game_files
-
     for _prefix, index in archives:
         for j in index:
+            # Check if the file should be treated as a common file.
+            if len(index[j]) > 3 and index[j][3] is True:
```
renpytom, line 445: I don't like the len() here. It might make sense to store the common vs not-common distinction in archives, and then do:

```
for _prefix, index, common in archives:
    if common:
         files = common_files 
    else:
         files = game_files 
```

(And the same thing down around line 722.)

renpytom (changes requested):
I think I get what your'e going for here, but there a number of things I dislike about the implementation. I've done a review, which hopefully should work.

uyjulian:
I've made changes.

For the search behavior, it is done in case the contents of `renpy/common` is in another archive (for example `python27.zip`) or if it is extracted in a directory.

I moved the check for common file from per-file to per-archive.

renpytom:
I don't see the use case for putting it in python27.zip, and I can see cases where doing a long search like this will lead to problems. (That might accidentally have renpy/common in them.)  I'll be honest, when I'm evaluating changes like this, I'm caring more about risk to every other platform, than reward on game consoles.

I'd be fine using renpy.zip if it exists in config.renpy_base. I really don't want this to turn into a search all over the user's system, such that it's possible to find a renpy/common that doesn't exist.

uyjulian:
I made it so it will only consider `sys.path` entries inside `renpy.config.renpy_base`.

renpytom:
Is there any use case where we'd be so out of control of the layout of the files that searching like this would be appropriate?

uyjulian:
There are `os.path.exists` calls guarding inserts to `sys.path` inside top-level `renpy.py`.  
This will prevent code duplication and hardcoded `renpy.zip` string in multiple files.

renpytom:
That doesn't really answer my question, though. I think a relative path to renpy.zip would be a good idea, and a search like this less good. So I'm trying to see if there is any reason for having the search path.

uyjulian:
If, in the future, there are multiple `renpy.zip` (like `renpy27.zip` and `renpy39.zip`) (when Python 3 support gets added), only one file (`renpy.py`) needs to be changed to handle this case instead of multiple files (`renpy.py` and `loader.py`).  
Also, if `renpy.zip` does not get added to `sys.path` due to it not existing, it won't be checked later on in the search path. This means the file will only get checked once (inside `renpy.py`) if it does not exist.  

The search path code is for connivence for any further changes to where the `renpy/common` folder is located.

renpytom:
Wouldn't it be way simpler in this case to just allow path_to_common to return a path to the zip file, whatever it is?

uyjulian:
I removed the sys.path search and instead changed the behavior around `path_to_common` to return a tuple with the path to `renpy.zip`.

renpytom:
I've decided to reject this. I don't really see it getting a lot of use, and the zipfile support in renpy.loader doesn't look clean to me. (And I'm trying to go through old PRs and close ones I don't see getting in.)

PR #2435, opened 2020-10-30
Adding im and im.MatrixColor workarounds in documentation

Gouvernathor:
As discussed in #2427.

I didn't include `im.Data` because I don't think the same can be done without it (I have no idea how to do it, and it's a kinda obsolete functionality anyway).
I didn't include `im.AlphaMask` either because `Alphamask` is already linked in `im.AlphaMask`'s docline. Furthermore, using `Alphamask` to do `im.AlphaMask`'s work would require passing the control displayable through a Matrix to replace its alpha layer with its red layer (and even then it wouldn't override the base displayable's alpha channel, it would multiply it), resulting in a more complex workaround than the other examples I gave.

### DIFF
--- a/renpy/display/im.py
+++ b/renpy/display/im.py
@@ -872,8 +872,8 @@ class FactorScale(ImageBase):
 
         image logo doubled = im.FactorScale("logo.png", 1.5)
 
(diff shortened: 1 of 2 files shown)


renpytom:
Merged, thanks.

PR #2437, opened 2020-10-30
Reuse dict and list objects

uyjulian:
Small optimization by reusing `dict` and `list` objects whenever possible.  

Instead of assigning a new empty `list` or `dict`, or slicing with a new empty `list`, clear the existing one using `del x[:]` or `x.clear()`.

### DIFF
--- a/renpy/atl.py
+++ b/renpy/atl.py
@@ -291,13 +291,11 @@ def compile_all():
     to compile all transforms.
     """
 
-    global compile_queue
-
     for i in compile_queue:
         if i.atl.constant == GLOBAL_CONST:
             i.compile()
 
-    compile_queue = [ ]
+    del compile_queue[:]
 
 
 # This is the context used when compiling an ATL statement. It stores the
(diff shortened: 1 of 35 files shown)


renpytom:
Running some tests:

$ python -m timeit -s 'a = {}' 'a = {}'
100000000 loops, best of 3: 0.015 usec per loop

$ python -m timeit -s 'a = {}' 'a.clear()'
10000000 loops, best of 3: 0.0299 usec per loop

$ python -m timeit -s 'a = []' 'a = []'
100000000 loops, best of 3: 0.0167 usec per loop

$ python -m timeit -s 'a = []' 'del a[:]'
100000000 loops, best of 3: 0.0171 usec per loop

I'm not really seeing a justification for this change, which would need a really good justifcation given the chance of introducing aliasing problems.

I've noticed you've been doing a lot of changes to try and minimize garbage collection, but I'm not sure why. Python generally uses reference counting for it's garbage collector, which is good at collecting small amounts of garbage without long GC pauses. (I believe that the mark-sweep collector rarely if ever runs in Ren'Py, after initial startup, unless I've introduced a regression.)

uyjulian:
Looks like my assumptions of speed based on another environment were incorrect.

PR #2438, opened 2020-10-30
When OOM error occurs during image loading, clean some memory and try again

uyjulian:
When OOM error occurs during image loading, clean some memory and try again.  

This shouldn't happen very often, but when it does, it will clean up some memory, hopefully having enough contingent space for the image to be loaded.

### DIFF
--- a/renpy/display/im.py
+++ b/renpy/display/im.py
@@ -258,19 +258,30 @@ class Cache(object):
         # Otherwise, we load the image ourselves.
         if ce is None:
 
-            try:
-                if image in self.pin_cache:
-                    surf = self.pin_cache[image]
-                else:
-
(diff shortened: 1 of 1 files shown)


renpytom:
I don't like this, as renpy.free_memory isn't something that's meant to be called during an interaction. You'd be removing something from the image cache that Ren'Py expects to be there, very likely turning the error that's actually happening (running out of memory due to not being able to fit anything more into memory) into a random error.

I think we need to step back and figure out why you're getting OOMs in the first case, as simply adding handling for them isn't really useful.

uyjulian:
That is probably correct. I managed to get a kernel panic sometimes while calling `renpy.free_memory` inside of that function.

PR #2446, opened 2020-11-02
Android additional archive loading on startup

mal:
**Problem:** Mod and patch `.rpa` files are ignored during startup on Android despite being in the `config.searchpath`.

---

At present archives are loadable when in `config.searchpath`, but their names are not collected for indexing during startup as `renpy.main.main()` only scans `config.gamedir`. In order to facilitate distributing mods and patches as `.rpa` files this change adds `config.gamedir` to Android's `config.searchpath` and then scans `config.searchpath` to populate `config.archives`.

It also makes use of sets and iterators to keep the process quick and prevent any duplicate processing later during the init sequence.

The end result is a `config.archives` containing a reverse sorted list of archive names as before, only now including names found in `config.searchpath` directories. As the list contains no path information, and `config.searchpath` precedence is handled by file loading code elsewhere, the thinking is that only a simple reverse sort over all members is required to ensure expected and predictable indexing and load order later on in the process.

### DIFF
--- a/renpy/main.py
+++ b/renpy/main.py
@@ -374,7 +374,7 @@ def main():
         renpy.config.searchpath.extend(os.environ["RENPY_SEARCHPATH"].split("::"))
 
     if renpy.android:
-        renpy.config.searchpath = [ ]
+        renpy.config.searchpath = [ renpy.config.gamedir ]
         renpy.config.commondir = None
 
         android_searchpath()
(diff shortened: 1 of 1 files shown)


mal:
@renpytom please could this feature be considered for inclusion (in some form) in 7.4.9? It's not quite clear to me if the approach here or the paths involved would still the best way forward in light of the increase in file path restrictions in newer versions of Android, but in general a way for mobile users to be able to consume `.rpa` packs to update and mod their games would be most welcome.

renpytom:
I'll consider it, though modding games is still not going to be supported. Patching kind of makes sense, though.

mal:
FWIW It's also a convenient way to distribute content expansions and language packs which can contain fairly large files (translated media/assets; images, video, audio etc) where it can be impractical to distribute them bundled them with the base-game.

mal:
Rebased in light of #2927.

Despite discussion in the thread on the subject, that change looks to have had the unintended consequence of removing `gamedir` from consideration when loading `.rpa` files on Android. Prior to that change only `gamedir` was examined, now, due to only looking at `searchpath`, and that list still being reset on Android (as before, [main.py#L377](https://github.com/renpy/renpy/blob/969cf58a38a0216140c52af6e21da517beb1d339/renpy/main.py#L377), I believe to clear the `commondir`) that location is no longer considered.

Additional changes are just optimisations to the archive gathering, as discussed in the original PR body above.

renpytom:
renpy.config.gamedir is basically a dead place on Android - since it's not writable without root, it's only useful for cache data. Given that nothing can be put there, and Ren'Py doesn't put things there, I don't see the point of searching it.

I dislike the rewrite of the search logic to use sets - it removes ordering, which is bad. 

I don't think there's anything left in this PR to merge, as far as I can see.

renpytom:
Thinking about it for a few more seconds, downloading data to config.gamedir is not a good idea, but should probably be supported. So we could add it for that. I will not accept the changes to archive scanning, as they seem to change the archive search order.

renpytom:
fe796bbc9 is my go at fixing this - I don't think there's anything else here.

PR #2471, opened 2020-11-19
Translate tutorial game into simplified Chinese

Neoteus:
Though [Chinese documentation](https://renpy.cn/doc/) of Ren'py has already been complete, I think the tutorial game is more intuitive. Starters can see how it works in action. So I spent some time translating it into Chinese.

I name the translation as "schinese". And I add an appropriate option in Language preference screen. I also try to keep technical terms in the tutorial consistent with the Chinese documentation while translating. I hope these would be helpful.

---

By the way, have you thought about changing the font for Chinese, Japanese and Korean? In my view, SourceHanSansLite.tff is too limited, and sometimes not beautiful in Chinese. I suggest [Noto Sans CJK](http://www.google.com/get/noto/help/cjk/) from Google, covering Simplified Chinese, Traditional Chinese, Japanese, and Korean in a unified font family. All Noto fonts are published under the SIL Open Font License, Version 1.1. It may be as large as 23 MB, but that's not a problem for modern PC.

### DIFF
--- a/tutorial/game/screens.rpy
+++ b/tutorial/game/screens.rpy
@@ -778,7 +778,8 @@ screen preferences():
                     textbutton "Español" text_font "DejaVuSans.ttf" action Language("spanish")
                     textbutton "한국어" text_font "../../launcher/game/fonts/SourceHanSansLite.ttf" action Language("korean")
                     textbutton "日本語" text_font "../../launcher/game/fonts/SourceHanSansLite.ttf" action Language("japanese")
-                    
+                    textbutton "简体中文" text_font "../../launcher/game/fonts/SourceHanSansLite.ttf" action Language("schinese")
+
                     # This should be last.
                     textbutton "Pig Latin" text_font "DejaVuSans.ttf" action Language("piglatin")
 
--- a/tutorial/game/tl/schinese/01example.rpy
+++ b/tutorial/game/tl/schinese/01example.rpy
@@ -1,0 +1,11 @@
+﻿# TODO: Translation updated at 2020-04-12 22:59
+
+translate schinese strings:
+
+    # game/01example.rpy:473
+    old "Copied the example to the clipboard."
+    new "复制示例到剪贴板。"
+
+    # game/01example.rpy:546
+    old "copy"
+    new "复制"
--- a/tutorial/game/tl/schinese/common.rpy
+++ b/tutorial/game/tl/schinese/common.rpy
@@ -1,0 +1,1178 @@
+﻿# TODO: Translation updated at 2020-04-12 22:59
+
+translate schinese strings:
+
+    # renpy/common/00accessibility.rpy:28
+    old "Self-voicing disabled."
+    new "自动语音关闭。"
+
+    # renpy/common/00accessibility.rpy:29
+    old "Clipboard voicing enabled. "
+    new "剪贴板语音启用。"
+
+    # renpy/common/00accessibility.rpy:30
+    old "Self-voicing enabled. "
+    new "自动语音启用。"
+
+    # renpy/common/00accessibility.rpy:32
+    old "bar"
+    new "条"
+
+    # renpy/common/00accessibility.rpy:33
+    old "selected"
+    new "选中"
+
+    # renpy/common/00accessibility.rpy:34
+    old "viewport"
+    new "视口"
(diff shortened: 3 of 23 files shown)


renpytom:
@zedraxlo , can you take a look at this?

I'm thinking that this is probably fine, though I'd like it if schinese became simplified_chinese, to match the launcher. That's something I can easily write a script to take care of, though. 

I'm not sure about Noto vs SourceHanSans. I think we might want to take the .ttc version, but that might require some changes to Ren'Py's font support to make work.

zedraxlo:
It's fine to make "simplified_chinese" became "schinese", because "simplified_chinese" is too long, also Steam is using "schinese". This is also apply to "tchinese".
In fact, Noto is SourceHanSans, SourceHanSans is Noto. This font is developed by Google and Adobe, Google called it as Noto, Adobe called it as SourceHanSans. SourceHanSans also have a large ttc version, but I remembered that 2 years ago Tom is against use this one because it's too large to increase size of SDK and server expenses. In fact I also against use this one because dl.renpy.org have no CDN in China, the download speed is almost 20KB/s, add more 20+ MB will wait 16+ more minutes.
But the final decision is right in you, Tom.

renpytom:
I think I'm fine with schinese if that's what Steam is using. I don't have a good idea as to what sort of download speeds people are experiencing, but my general thinking is that if they're the same font, we should keep using the smaller variant. Is the difference between the lettershapes, or because we're cutting SourceHanSans down, and so the character selection is limited?

Neoteus:
I notice several characters a little strange, and I show them below. It is inconvenient to check which character needs to be added to the cut-down font, though adding no more than 10 characters may fix that.

Snipaste_2020-11-20_09-31-18

I use the cut-down font in the second line, and full-size in the third line. You see, there are some differences. There are many similar characters in Chinese, but they are replaced with the same one in cut-down. Some characters can be separated into two parts, and each is used as an individual character, which is half-width of the standard ones.

I think using lite fonts makes sdk smaller and download fast, while full-size fonts provide convenience if the developer use the default without character missing. Anyway, that's not a big problem. It is acceptable either it takes 20 minutes more to download, or people spend time to find and get their own font.

zedraxlo:
In SorceHanSans, CJK character shapes are not exactly the same, especially in japanese, schinese and tchinese. The .ttc has certainly solved this problem, but obviously Tom would like to use a smaller version. There was a logic problem in previous  versions (especially 1st pre-release) of Lite that caused some characters to display in a style that was not customary for use, but I've fixed that in the October 29th pull request. It is recommended to use the latest nightly version to view the display of the above characters.

With respect to the current logic and character range, it goes like this:
Chinese characters, also known as kanji, are hereafter collectively referred to as kanji.
For kanji used only in japanese, the japanese writing style is displayed.
For kanji used only in schinese, the schinese writing style is displayed.
For kanji used only in tchinese, the tchinese writing style is displayed.
For kanji used in both japanese and schinese, the japanese writing style is displayed.
For kanji used in both japanese and tchinese, the japanese writing style is displayed.
For kanji used in both schinese and tchinese, the schinese writing style is displayed.
For kanji used in both japanese, schinese and tchinese, the japanese writing style is displayed.
This logic gives priority to japanese. If we want to change the priority to schinese, we should ask RenPy's Japanese maintainer for advice.

Character List:
- Basic English numbers and symbols, including half and full angles.
- Japanese hiragana and katakana.
- Korean characters, based on KS X 1001.
- Japanese kanji, based on JIS X 0208 + joyo kanji.
- Schinese common kanji, based on GB2312, but with about 1.4k characters removed, these characters are not often used, for example they are only used in rare names of places, mountains and rivers.
- Tchinese commonly used Chinese characters, based on Big5 + Taiwan Ministry of Education's character list.
- The above characters are partially de-emphasized.

renpytom:
Let me at least merge the translation in as it is now. I'm thinking it might be good to have a proper CJK font out of the box, perhaps improving TTC support to make that happen. Not living in the affected region, and knowing how good/bad the current fontset is, I'm not sure I understand if the problem is worth the increased download. 

It might make sense to simply offer users the choice of a stripped-down Ren'Py, or one with the additional fonts. 

I think it might be best to continue this discussion, so I created #2479.

PR #2478, opened 2020-11-25
Add basic support for parameter annotations

jsfehler:
This PR adds annotation support to parameters. These are arbitrary strings, similar to PEP 3107.

```
screen inventory(apples_qty: "int"):
    text "You have [apples_qty] apples"

screen inventory(apples_qty: "How many apples you have" = 10):
    text "You have [apples_qty] apples"
```

Return values weren't added since screens and labels don't return anything.

Checking for strings was used here as a test, but should probably be changed to accept any object type.


renpytom:
This might work, but it probably should match what Python supports better, needs some backwards compatibility, and probably should wait until after 7.4 is out.

jsfehler:
Updated to use l.simple_expression instead of l.string.

I hesitate to add a return annotation in this PR, like:

`label talk_with_shopkeeper(current_money: int) -> "jump back to downtown"`

It's not really a "return" value, but more of a "end of node" action. Since it's different enough from regular python it might need more thought.

Something like a typing library:

`label talk_with_shopkeeper(current_money: int) -> renpy.typing.Jump['downtown']`

might be interesting? It could be paired with validation in the future. Unlike parameters, it's not something where python's typing library could just be used.

jsfehler:
@renpytom Thinking about an effective way to unit test this. Does something like the following make sense:

```
def test_parameter_annotations_no_default_value():
    block = "screen test_screen(qty: int):"
    lex = Lexer(block)

    result = parse_parameters(lex)

    assert result.annotations = {'qty': int}


def test_parameter_annotations_with_default_value():
    block = "screen test_screen(qty: int = 10):"
    lex = Lexer(block)

    result = parse_parameters(lex)

    assert result.annotations = {'qty': int}
   
```

Andykl:
I like the idea. I will update #2391 today. Could you make your PR on top of it after?

Andykl:
Hello, could you rebase your pull request on base of #3310?

Gouvernathor:
I redirected the PR to the py3-arguments branch so that it includes its edits. I also did a merge commit to cleanup the conflicts.

PR #2480, opened 2020-11-26
Update show statement documentation

pauljmartinez:
I'm a new Ren'Py user and thought I'd see if this suggestion might help make the documentation a bit clearer.

I'd be happy to update this PR with translations if my suggestion is helpful.


renpytom:
I think this is a good contribution - the original version of the game was wrong. What do you plan to do with translations?

PR #2481, opened 2020-11-26
Update menu documentation

pauljmartinez:
I'm a new Ren'Py user and thought I'd see if this suggestion might help make the documentation a bit clearer.

I'd be happy to update this PR with translations if my suggestion is helpful.

### DIFF
--- a/tutorial/game/tutorial_quickstart.rpy
+++ b/tutorial/game/tutorial_quickstart.rpy
@@ -539,9 +539,9 @@ label tutorial_menus:
 
     e "Similarly, picking the second choice jumps us to choice1_no, which also runs two lines of script."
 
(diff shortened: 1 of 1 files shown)


Lord-V15 (approved):

renpytom:
Honestly, I don't think this is worth it, as it would mean updating all the translations. The second change is wrong, as we're talking about Ren'Py's if statement, not the Python if statement. The first one is fine, but I'm not sure it's worth making if it means updating all the translations.

Lord-V15:
Oh my I'm so sorry, I didn't think it through about the translations. 
Yeah I agree with you. Absolutely no need to add these to be honest.

Lord-V15 (changes requested):
"The if statement" was meant to be a reference to the checking of flag variable and not the Python's if statement in general.

PR #2507, opened 2020-12-19
Random fixes in the doc

Gouvernathor:
When `matrixcolor` started being documented, the im.matrixcolor functions' doc anchors were renamed from im_matrixcolor to im_matrix (in [this commit](https://github.com/renpy/renpy/commit/7468b5645943d0d4f88056ce995f8471dd19793a)), but the matching anchor in displayables.rst wasn't updated, so I assume this was a mistake.

I'm not that experienced in translations in Ren'py so plz check this, but I believe putting a variable before a $ results in an error even in that context.

Also since the section where the placeholder contacts the web service is inactivated, I removed references to that in the docstring.


renpytom:
Ah, sorry - I missed this. I think I saw it and thought it was a chat comment, and not a PR. Could you back out 
01a0365 and b435b86 so I can merge the rest?

PR #2510, opened 2020-12-19
Add events keyword for Dissolve

Andykl:
This is useful in case the transition is a screen layout change, e.g. in the game menu, so `config.intra_transition = {"screens": Dissolve(0.3, events=True)}`.


renpytom:
Hm... I think this should be a change that happens either systematically, or not at all, rather than something limited to just Dissolve.

PR #2512, opened 2020-12-21
Random fixes in the doc - 2

Gouvernathor:
Another version of #2507 without the messy merge commit and the matrixcolor fix

### DIFF
--- a/renpy/common/00layeredimage.rpy
+++ b/renpy/common/00layeredimage.rpy
@@ -26,7 +26,7 @@ python early in layeredimage:
             which is used to create better error messages.
 
         `name`
-            The name of the attribute image.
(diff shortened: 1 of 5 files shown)


Gouvernathor:
So the removal of the im. matrices is intentional ?

Gouvernathor:
Also I've just noticed that Image-like Displayables are missing from the same doc page, is it also intentional ?

renpytom:
Yes, in 7.4 there isn't much need for image manipulators, so I want to deemphasize them. 

It looks like image_like displayables are still there, unless I'm missing something.

Gouvernathor:
Here is what I have on both versions of /displayables.html
Screenshot_121
Screenshot_122

PR #2516, opened 2020-12-22
Add a dynamic condition to layeredimage layers

Gouvernathor:
The idea is to be able to take into account both what attributes were requested by the user and the context of the game to make the choice of which attributes to display, while having these attributes declard automatically.
Currently, if I'm not mistaken, only the `if` blocks and the attributes having square brackets are switched dynamically, but neither can be defined automatically in an auto block.
It would be done by adding an `if_dyn` (dyn for dynamic) clause to the layeredimage layers, whose value would be a string which once evaluated yields a boolean. That boolean value is then taken into account in the same way as `if_all`, `if_any` or `if_not`.

The only thing left to be done is to trigger a redraw on the layeredimages having an `if_dyn` clause on any layer, on each interaction. These layeredimage objects now have the (arguably renamable) `is_dynamic` flag set to true, but I don't know (yet) how to tell ren'py to redraw each such layeredimage on each interaction.
I witnessed such a per-interaction redraw for attributes whose manually-set displayables were strings with square brackets in them. But if I'm not mistaken, it's because they become DynamicImages inside a Fixed (potentially within a Transform), and the willingness of the DynamicImage to get a redraw is transmitted back to the Fixed. So, unhelpful for this purpose. But I still have hope.

### DIFF
--- a/renpy/common/00layeredimage.rpy
+++ b/renpy/common/00layeredimage.rpy
@@ -8,7 +8,7 @@ python early in layeredimage:
     ATL_PROPERTIES = [ i for i in renpy.atl.PROPERTIES ]
 
     # The properties for attribute layers.
-    LAYER_PROPERTIES = [ "if_all", "if_any", "if_not", "at" ] + ATL_PROPERTIES
+    LAYER_PROPERTIES = [ "if_all", "if_any", "if_not", "at", "if_dyn" ] + ATL_PROPERTIES
 
     # This is the default value for predict_all given to conditions.
     predict_all = False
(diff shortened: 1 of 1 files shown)


Gouvernathor:
By the way, this question of refreshing dynamic layeredimages when the underlying data changes also apply to the already implemented layeredimageproxy which have square-brackets substitution, which I referred to in #2513 .
Perhaps the LayeredImageProxies should also have that `is_dynamic` property.

Gouvernathor:
Rendered mainly obsolete by the config.adjust_attribute

PR #2524, opened 2020-12-24
Fix for choose_image ambiguity

Gouvernathor:
Now when several matches are found (given they all have every required argument), only the ones with the largest amount of optional arguments are kept. If there are still several, the exception is raised as it would have been, otherwise the one having the max amount of optionals is returned.


renpytom:
I think this is correct, so I merged it.

PR #2529, opened 2020-12-25
Favor optional attributes over unrequested ones

### LINKED ISSUE #2528: Possibly unintended change in layeredimage conflict managing
040ec21 changed stuff in the choice of which displayable to choose, when one or several of those are layeredimage or live2D.
In renpy/display/image.py/choose_image(), displayables are polled and the ones with the most attributes, when they include all required ones, are stored in `matches`.
However, the attributes already being in a layeredimage's (or live2D's) name ("sitting" if it was declared as `layeredimage eileen sitting`) are counted, and do not bar it from being part of `matches`, when it would be the case were it not a layeredimage (nor a live2D).
This happens _before_ my last commit, which only matters when several matches are found with the same number of attributes.
I'll show an example where it has an influence on the final result.
```
layeredimage eileen one:
    size (400, 400)
    xalign .0 # shows on the left
    attribute aa default:
        Solid('#c60')
    attribute ab:
        Solid('#f00')
    attribute base:
        Solid('#0f0')

layeredimage eileen:
    size (400, 400)
    xalign 1.0 # shows on the right
    attribute ba default:
        Solid('#60c')
    attribute bb:
        Solid('#00f')
    attribute base:
        Solid('#0f0')
```
Say I call `show eileen base`. The polling finds `('one', 'base')` for eileen one, and `('base',)` for eileen.
With the old behavior, 'one' being neither required nor optional, `eileen one` would be barred from the `matches` list, leaving `eileen` as the chosen one. (a green square would appear on the right)
With the new behavior, since `eileen one` is a layeredimage, the fact that it has a non-optional attribute is waived, and it is kept in the tally with 2 attributes. `eileen` doesn't make it, having only 1. `eileen one` is chosen and displayed. (the green square appears on the left)

TL;DR now when several poses for the same layeredimage can respond to the requested attributes, the ones with the largest amount of attributes in their name are given an unfair advantage, whereas they used to be banned.

You might think it's stupid to have an eileen with a pose and eileen with none, and it's probably true, but the same thing would be happening would we have an `eileen one`, an `eileen two sitting` and `eileen two standing`.

Gouvernathor:
fixes #2528
It basically switches from a first tally for the _**most**_ number of attributes with a runoff for the most optional ones,
to a first tally for the most number of optional attributes with a runoff for the _**least**_ total number of attributes.

Only ever matters when dealing with layeredimages or live2D, since it's the only case where unrequested (= not required nor optional) attributes make it into the matches list.

Checked with layeredimage, it works, but it's not tested with live2D.

### DIFF
--- a/renpy/display/image.py
+++ b/renpy/display/image.py
@@ -945,7 +945,7 @@ class ShownImageInfo(renpy.object.Object):
     def choose_image(self, tag, required, optional, exception_name):
 
         # The longest length of an image that matches.
-        max_len = -1
+        max_opt = -1
 
         # The list of matching images.
         matches = None
@@ -965,13 +965,16 @@ class ShownImageInfo(renpy.object.Object):
                 attrs = attrs + newattrs
 
             num_required = 0
+            num_optional = 0
 
             for i in attrs:
                 if i in required:
                     num_required += 1
                     continue
-
-                elif (i not in optional) and (ca is None):
+                if i in optional:
+                    num_optional += 1
+                    continue
+                elif ca is None:
                     break
 
             else:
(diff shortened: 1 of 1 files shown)


Gouvernathor:
This implementation does not check for doubles when counting optional (or required) attributes. It could if I use `len(set(stuff))` like in the runoff. Is there a reason not to do that when counting required attributes ?

PR #2532, opened 2020-12-25
Refresh LayeredImageProxy when interpolation is performed on its name

Gouvernathor:
When the string given to the LayeredImageProxy has square brackets in it, the data it accesses can change while the image is displayed, but the changes won't be taken into account until another `show` instruction is performed on the LayeredImageProxy.

Now, only when there's interpolation (not to waste processing time), the displayable which would be returned by the `_duplicate` method to be displayed (which is basically a Fixed wrapped in Transforms) is wrapped in a Fixed whose `per_interact` method is overriden.
The new method, having a link to the LayeredImageProxy, recomputes the displayable, and if updated, replaces its child with it. (Actually it always replaces it, since Transforms and Fixed don't support equality.)

Therefore, when the underlying data updates, the image is refreshed.

### DIFF
--- a/renpy/common/00layeredimage.rpy
+++ b/renpy/common/00layeredimage.rpy
@@ -1143,7 +1143,7 @@ python early in layeredimage:
 
                 def new_per_interact():
                     newchild = self._base_duplicate(args)
-                    if rv.child is not newchild:
+                    if rv.child != newchild:
                         rv.remove(rv.child)
                         rv.add(newchild)
                         renpy.display.render.redraw(rv, 0)
--- a/renpy/common/00layeredimage.rpy
+++ b/renpy/common/00layeredimage.rpy
@@ -1122,13 +1122,35 @@ python early in layeredimage:
             return image
 
 
-        def _duplicate(self, args):
+        def _base_duplicate(self, args):
 
             rv = self.image._duplicate(args)
 
 
             for i in self.transform:
                 rv = i(rv)
+
+            return rv
+
+        def _duplicate(self, args):
+
+            rv = self._base_duplicate(args)
+
+            if "[" in self.name:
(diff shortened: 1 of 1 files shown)


renpytom:
I'm a little confused by what you're trying to accomplish here, but I don't think this is the solution to it - changing a method at runtime doesn't seem like a good idea. 

What's the underlying problem you're trying to address?

Gouvernathor:
Take this code for example :
```
default phoenixgender = 'm'

layeredimage mphoenix:
    group base auto:
        attribute default default
    group face auto:
        attribute smile default
    attribute blush
    attribute sweat

layeredimage fphoenix:
    group base auto:
        attribute default default
    group face auto:
        attribute smile default
    attribute blush
    attribute sweat

image phoenix = LayeredImageProxy("[phoenixgender]phoenix")
```
If the `phoenixgender` variable changes while `phoenix` is displayed, for example from 'm' to 'f', the masculine version of phoenix will still be displayed. If the game refreshes, it will reload the feminine version. That state of limbo can be problematic if (unlike gender in this example) the variable reflects a change affecting the display of several different characters, since it requires to call the `show phoenix` instruction for each one.

Concerning the implementation, I thought about another version where a subclass of Fixed would be defined inside LayeredImageProxy, with an override on the `per_interact` method, and wrap the displayable inside that one instead of a normal fixed.
There are several issues with that, one being that Fixed is not a real class, it's something like _layout_class and I can't really get my mind around it. Another being that the LayeredImageProxy needs to be stored in that subclass, and the `args` parameter to `_duplicate` too, since these variables are used in `new_per_interact`.

Gouvernathor:
Would cause pickling problems since a nested function is sneaked out into a returned value.

Issue #2314, opened 2020-08-08
Script reloading & replay mode


DinakiS:
When script reloading in replay mode it disables replay mode and continue replay like it's a regular game.

Example:
```
label start:
    menu:
        "Simple jump":
            jump replay
        "Start replay":
            $ renpy.call_replay("replay")
            jump start

label replay:
    if _in_replay:
        "Replay. If you press Shift + R here, you'll get the text 'Not replay'."
        "Press Shift + R here to unlock unreachable text."

        $ renpy.end_replay()

        "In replay this text can't be reached"
    else:
        "Not replay"

    jump start
```

renpytom:
I don't consider this a bug.

PR #2587, opened 2021-01-09
Document #2314

jackmcbarn:
#2314 is really unexpected behavior. Since it isn't considered a bug, document it.


renpytom:
Do you think the right answer to this would be to disable reload while in replay?

jackmcbarn:
If it were just disabling manual reload then that would be fine, but for it to be effective, it would have to disable the auto-reload from editing a file too, and that would also be really surprising behavior. If we also disabled starting a replay while auto-reload was on, then I think that could work. By the way, is there a reason the underlying behavior can't just be changed?

PR #2595, opened 2021-01-14
Add a getter for the achievement progress

Gouvernathor:
In addition to the setter already in place (the `achievement.progress` function), there's now a getter to display to the user inside ren'py, giving a use to the non-steam achievement progress tracking.

I directly used the persistent's implementation, since the steam's implementation of the `progress` function uses it directly, but if in the future several ways to store progress emerge, a `get_progress` method could be added to all subclasses of Backend, and the function could be a foreach loop like the others.

### DIFF
--- a/renpy/common/00achievement.rpy
+++ b/renpy/common/00achievement.rpy
@@ -320,6 +320,17 @@ init -1500 python in achievement:
         for i in backends:
             i.clear_all()
 
+    def get_progress(name):
+        """
+        :doc: achievement
+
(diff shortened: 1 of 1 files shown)


Gouvernathor:
Come to think of it, there could also be an equivalent of the way steam grants the achievement once the required number of points is reached, but on the persistent storage.
Let me know if you're interested.

renpytom:
Merged, and yes, I'd be interested in an independent achievement system.

PR #2597, opened 2021-01-14
Fix typo in ATL tutorial

ajkhan96:


renpytom:
A good catch. Thank you for the first contribution.

PR #2616, opened 2021-01-22
Update focus.py

midgethetree:
Made it so the focus_extreme function (used when arrow key is pressed if self voicing is off and no displayable on screen is currently focused) ignores displayables with keyboard_focus set to False.


renpytom:
Merged, thank you.

PR #2627, opened 2021-01-26
Improvements to default achievement backend

Gouvernathor:
Added the `register` method, so the `stat_max` value is given.

Added achievement granting when the progress reaches the `stat_max` value.


Gouvernathor:
I didn't implement `stat_modulo` because I don't see the point of giving a moduoed value when the user requests it (whereas steam needs it). A modulo can always be applied afterwards, in an achievements screen for example.

If it should be implemented, it would be in the `get_progress` function.

PR #2643, opened 2021-02-02
Add rgba property to Color

RenpyRemix:
Add attribute _rgba and property rgba to Color class and update TintMatrix to use that property


renpytom:
That works. Merged. Thank you!

PR #2652, opened 2021-02-06
Live2D: Added ability to specifiy user function to blend parameters

Andykl:
This is meant to be used in the way:
```
init python:
    def update_parallax(model, st):
        xpos, ypos = renpy.get_mouse_pos()
        # Parallax parameters takes on a value from -100 to 100
        model.blend_parameter("parallax_x", "add", (xpos / 9.6) - 100)
        model.blend_parameter("parallax_y", "add", (ypos / 5.4) - 100)
        return 0

image main_menu_l2d = Live2D("l2d_menu_prlx", loop=True, update_function=update_parallax)
```
If the implementation looks good, I will document it.


renpytom:
I think this is good, save that I think it would be better to have "Add" instead of "add", to match Live2D.

Andykl:
I have updated this and added the documentation.

renpytom:
Merged, thank you.

PR #2657, opened 2021-02-08
relative volume management independant of context

AndriWilford:
Possibly, take this commit as a guide or idea to approach this. 

This may address Issue 2654#

Adds the ability for channels to manage their "tertiary" volume independently of context.
In addition, set this tertiary volume for a context switch (loading).

Tertiary volume is set per-track, such as:
renpy.music.play("music.opus", relative_volume=0.7)  
  or
play music "music.opus" volume 0.7

As opposed to "secondary" volume which is set by channel controls, like
$ renpy.music.set_volume(0.4, delay=0.5, channel="music")

The problem is that things like loading a game can cause sudden differences in 'track' volumes. Since these volumes were stored with the context, and immediately applied via mixing with the 'secondary' volume, there was no way (I could see) to allow smooth volume transitions when  fade-in/fade-out occurred with the context switch,

For instance, if one had a track playing at 0.3 tertiary volume, then loaded a game with 1.0 volume,  the fades would occur, but the channel tertiary volume would immediately jump to 1.0, making the old track play very loudly for some part of the fadeout duration.

I attempted to do this in a way that minimized change to behavior...
Some decisions were made:
If a new list of loop tracks are queued, and the currently playing track is among them, then immediately alter the volume.

Also, added a 'dequeue' to the context-switch interact, when the channel is currently playing a loop. Previously, the channel would 'fadeout' (and so clear the queue and playing track) if the current playing track wasn't in the upcoming filename list,  but if it *was*,  it would NOT dequeue, but still queue the whole list, potentially adding many copies of the filename list to the queue if several loads were done in a row.

### DIFF
--- a/renpy/audio/audio.py
+++ b/renpy/audio/audio.py
@@ -249,6 +249,15 @@ class Channel(object):
 
         # The time the secondary volume of this channel was last set.
         self.secondary_volume_time = None
+        
+        #The next filename to play on this channel
+        self.next_channel_filename = None
+
+        #Queue a volume adjustment for the relative channel volume.
+        self.queued_tertiary_volume = None
+
+        #The current relative channel volume that the player is hearing. separating this from the "context" volume allows for smooth volume transitions on loads and such
+        self.current_tertiary_volume = 1.0
 
         # Should we stop playing on mute?
         self.stop_on_mute = stop_on_mute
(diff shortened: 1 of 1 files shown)


renpytom:
649b8469a was based on this, but accomplishes things in a different way, since I was able to edit the C and Cython source that does audio playback. Thank you, and I can close this now.

PR #2664, opened 2021-02-13
Pass kwargs of input() to the shown screen

Lezalith:
Right now, passing a keyword argument to the renpy.input() will throw:
TypeError: input() got an unexpected keyword argument '{}'

With this, renpy.input() takes keyword arguments and passes them to the screen shown with the function.


Lezalith:
Update with current functionality:

Calling input() with kwargs that have the prefix "show_" will strip the prefix and pass the kwargs to the screen.
kwargs without the prefix and undefined by input() still throw an error.

renpytom:
Merged. Thank you.

PR #2668, opened 2021-02-16
Make screen main_menu() easier to customize

xavi-mat:
`screen main_menu()` is probably the most customized of all built-in screens.

Currently styles in `screen main_menu` are applied to one `frame` and one `vbox` from `style_prefix`. This way, any time a developer adds a `frame` or a `vbox` to customize their `main_menu` they will inherit some unexpected properties (like `xsize`, `yfill`, `xalign`, `yalign`, a very specific `background`...).

Actually: `style_prefix` is useful to reuse the same properties in different children of the same type. But current properties defined in `style main_menu_frame` and `style main_menu_vbox` will hardly be reused.

My proposal is not to use `style_prefix` in the main menu, but apply directly those two styles to the two unique children that will use them.

This way, if developers add more `vbox` or `frame` to their `main_menu`, they will behave as expected with default properties.

### DIFF
--- a/gui/game/screens.rpy
+++ b/gui/game/screens.rpy
@@ -356,13 +356,11 @@ screen main_menu():
     ## This ensures that any other menu screen is replaced.
     tag menu
 
-    style_prefix "main_menu"
-
     add gui.main_menu_background
 
     ## This empty frame darkens the main menu.
     frame:
-        pass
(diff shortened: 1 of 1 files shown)


Gouvernathor (approved):

renpytom:
Merged, thank you.

PR #2671, opened 2021-02-17
Update behavior.py

hell-oh-world:
Allow the `keysym` and `alternate_keysym` properties to take a list of `keysym`s or single string `keysym`.

### DIFF
--- a/renpy/display/behavior.py
+++ b/renpy/display/behavior.py
@@ -920,14 +920,24 @@ class Button(renpy.display.layout.Window):
             rv = super(Button, self).event(ev, x, y, st)
             if rv is not None:
                 return rv
-
+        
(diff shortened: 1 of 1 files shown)


Andykl:
It's not nessesary, because `map_event` merge lists itself. However, there was a bug with lists and cache check. I fixed that in 79296865.

hell-oh-world:
> It's not nessesary, because `map_event` merge lists itself. However, there was a bug with lists and cache check. I fixed that in [7929686](https://github.com/renpy/renpy/commit/79296865dfc6f304e6d5ecb8582a2337a2f73dd8).

If that feature is already supported, then why come I cannot do something like...
```py
button:
  keysym ["K_SPACE", "K_a"]
```

PR #2677, opened 2021-02-19
Allow labels not to take colons

Gouvernathor:
I always thought it was weird that the label statement requires a colon, but indenting the following line (and giving it a block) is not mandatory. Imho, scripts would make a lot more sense imo if there was no indenting after labels, since labels have no scope like functions do.

So I disabled the requirement to have a colon ending a label's line.
It helps especially because most IDEs automatically indent a line following a colon.
It seems to work fine, after limited testing.

It could also be
```py
if l.match(':'):
    block = parse_block(l.subblock_lexer(init))
else:
    block = []
```
It would maybe save some execution time.

### DIFF
--- a/renpy/parser.py
+++ b/renpy/parser.py
@@ -2335,9 +2335,7 @@ def label_statement(l, loc, init=False):
     else:
         hide = False
 
-    l.require(':')
-    l.expect_eol()
-
+    l.match(':')
     # Optional block here. It's empty if no block is associated with
     # this statement.
     block = parse_block(l.subblock_lexer(init))


jsfehler:
It does seem weird that the following is valid:

```
label foo:
label bar:
    "Some text"
```

However it does serve a purpose, or at least seems to: https://www.renpy.org/doc/html/label.html

Variables in a label aren't scoped, but labels inside another label can be. Someone could have something like:

```
label global_label:
label .local_one:
    "Really?"
label .local_two:
    "Yes."
```

I don't think the solution should be to make the following valid:

```
label global_label
label .local_one
    "Really?"
label .local_two
    "Yes."
```

I think this is going to make reading complex label layouts a nightmare. It means multiple ways of writing the same script. More complexity to the DSL, and make documentation ever more difficult. Multiple ways to do the same thing is a red flag to me.

If anything, I would actually argue for making indentation after a label statement mandatory, but I shudder to think at how many older games that would break.

Lack of block scoping is just a python thing. Labels behave more like fancy goto statements than functions, so you probably shouldn't want label scoped variables.

Gouvernathor:
I never understood how sub-labels (or local labels) are supposed to work. In your second example, is `local_one` really a sub-label of `global_label`, even though it's not in its block ?

If so, then there's no reason to keep indenting after labels, since making sub-labels works without indenting and blocks (and I believe there should be more explaination about how all of this works), and if there's no indenting then there should be no colon.

If not, if you meant the following :
```rpy
label global_label:
    label .local_one:
        "Really?"
    label .local_two:
        "Yes."
```
then you could say that having a colon and a block is still mandatory if you want to have sub-labels, but you can omit it if there aren't.
It would lead to that kind of code, which seems a lot more understandable to me, in a control flow point of view :
```rpy
label global_label:
    label .local_one
    "Really?"
    label .local_two
    "Yes."
```
Also there could (and probably should) be a check so that there can't be indenting after a label without a colon.


You're right about calling labels `goto` statements, but goto statements in other languages don't have scopes at all, and it wouldn't make sense then to have a colon or an indented block beneath.

All in all, label statements in renpy as it is are weird. I used to believe that a label's block's script was unreachable unless you jumped or called the label directly, because the control flow would skip it like an `if False:` statement. Or that a `return` was implied at the end of a label's block.
It turns out neither of these are true, and since they aren't (and assuming you didn't typo the indentation in your example), labels' indented blocks serve no purpose at all and only lead to misunderstandings.
So, what I'm saying is that colons need to go as well. I don't get why you would keep the colons without the indentation, if that's what you're defending.
(And to be clear, I have no intention of banning indenting nor colons after a label.)

jsfehler:
I agree that the control flow here is misleading. The following:

```
namespace foo:
    (script, code, whatever)
```

really implies some level of encapsulation regarding when everything under the namespace executes. Currently:

```
label start:
    "Start"

label foo:
    "Foo"
```

will have everything in `foo` execute directly after `start`. Unless a jump or return is added at the end of `start`.

So in that sense, the label is more of a line marker. However, labels take arguments! Which means in the following scenario:

```
label start:
    "Start"

label foo(a='AAA'):
    "[a]"
```

What you see in `foo` will be different depending what you feed it. Looks more like a function than a goto. Return is optional, and the difference between jump and call matters for the return statement.

If we say something like "labels are reusable script blocks" then it's easy to imagine something like:

```
label my_script_1:
    "All I can say is..."

label string_formatter(a='Hello', b='Wisconsin'):
    "[a] [b]"
    return

label my_script_2:
    call string_formatter(a='Goodbye', b='New York')
    "and get the hell out of here."
```

Which might look fine, but it's a trap! A very hard to catch trap, regardless of indentation.

renpytom:
I don't plan to merge this.  Ren'Py is deliberately an unstructured programming language, as the domain it was meant for - visual novels - tends not to repeat blocks of code.  Similar usages are found in C and C++, for example, and inherited from the older languages that Ren'Py takes after.

You can also write:

```
label foo:
    pass
label bar:
    "..."
```

If you want.

Issue #2682, opened 2021-02-20
Character() constructor seemingly confusing name and kind


Gouvernathor:
```rpy
define sar = Character(_('Sardaukar'))
define hyr = Character((lambda : _("Hyrulan") if believe_hyrulan else _("Sardaukar Soldier")), dynamic=True)
default believe_hyrulan = false

label start:
    sar "hello" # works as intended
    hyr "hello" # wai
    Character(sar) "hello" # wai
    Character(hyr) "hello" # fails because "eval() takes a string or code object"
    Character(kind=hyr) "hello" # wai
```

AndriWilford:
I don't think it is "confusing name and kind".
The string type conversion of the class does not evaluate a function.  "Character(hyr)" attempts to convert the object 'hyr' to string, so crashes, since there is no calling of the function. Strictly speaking   "Character(hyr.name())("Hello")" works, since it is 'calling' hyr's name instead of string conversion..

I think string conversion can just have function evaluation as in pull request #2683

Gouvernathor:
Sure, but the system shouldn't try to evaluate `hyr` as a string, if `sar` is not. Whatever the specified behavior is, that's a discrepancy.

Furthermore, I removed it from my example but in my original code both `sar` and `hyr` have other custom settings other than its name, and they are carried along when calling `Character(sar)` or `Character(kind=sar)`, and they wouldn't if I called `Character(sar.name)`.

AndriWilford:
The behavior on line 8 and 9 is the same. They are both type-cast to string for the parameter to the Character constructor. It's just that it works for  8, since it successfully evaluates to a string, while line 9's does not.

Can you give an example of code where custom settings are carrying over for line 9 and 8? The lines 9 and 8 and 10 create new character objects that are used once and discarded (the character object is not saved to any variable). Line 10 should work to bring over custom settings, from the use of the  "kind"  parameter,  however lines 8 and 9 should not... they are just creating new character objects,  with the string created from evaluating 'sar' and 'hur',  but with no parameters.

Gouvernathor:
Indeed. I re-tried it and it turns out I had completely misunderstood how the constructor works.
I thought if you only give a Character instance as a name and no `kind` argument, the Character argument would be interpreted as a `kind`.
I believe that behavior would be more handy, but it would probably cause incompatible changes.

renpytom:
I don't think there's a bug here.

PR #2683, opened 2021-02-20
Character class callable for string type

AndriWilford:
This fixes issue #2682 .
Adds the ability for the string type conversion to also "call" a callable "who". Currently, only __call__ will attempt a callable who conversion.

### DIFF
--- a/renpy/character.py
+++ b/renpy/character.py
@@ -1047,7 +1047,10 @@ class ADVCharacter(object):
         who = self.name
 
         if self.dynamic:
-            who = renpy.python.py_eval(who)
+            if callable(who):
+                who = who()
+            else:
+                who = renpy.python.py_eval(who)
 
         return renpy.substitutions.substitute(who)[0]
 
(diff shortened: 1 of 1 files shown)


AndriWilford:
Apparently, using web interface to "update" a fork creates a dirty history.

Maybe just make the changes in the ..
   "Add function callable to string type"
 .. commit above oneself, rather than merge this request directly.

Gouvernathor:
Passing a callable as a name to a dynamic character already works.
Examples : `define hyr = Character((lambda : _("Hyrulan") if believe_hyrulan else _("Sardaukar Soldier")), dynamic=True)`
```py
init python:
    def hyrname():
        return _("Hyrulan") if believe_hyrulan else _("Sardaukar Soldier")
define hyr = Character(hyrname, dynamic=True)
```
Both of these work.

Gouvernathor:
You couls avoid some code duplication by replacing l. 1059 to 1067 with `rv = self.__unicode__()`.

renpytom:
It looks good, and I don't think overloading unicode for this would be a good idea.

PR #2695, opened 2021-02-25
Update common French translations

Beuc:
Update missing French strings for 7.4.
Drop a few obsolete strings.

Btw, why are internal strings like
https://github.com/renpy/renpy/blob/27bc9891060f5eee62983b2881906fc8ae1501c4/renpy/common/00preferences.rpy#L388
marked for translation?
(I see we consistently translated them with the english version)

Cc @maxlefou for proof-reading :)

### DIFF
--- a/the_question/game/tl/french/common.rpy
+++ b/the_question/game/tl/french/common.rpy
@@ -489,6 +489,14 @@ translate french strings:
     old "self voicing"
     new "self voicing"
 
+    # renpy/common/00preferences.rpy:388
+    old "self voicing volume drop"
+    new "self voicing volume drop"
+
+    # renpy/common/00preferences.rpy:430
+    old "font transform"
+    new "font transform"
+
     # 00preferences.rpy:355
     old "clipboard voicing"
     new "clipboard voicing"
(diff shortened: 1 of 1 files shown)


Gouvernathor:
In the launcher translations I turned "slef voicing volume drop" by something like "baisse de volume de la vocalisation automatique" or something like that. It should be translated.

Beuc:
Hi. I believe you are confusing my code quote above with:
https://github.com/renpy/renpy/blob/5b782f70fa7c506ddfcf019084bc7a9600fecda5/renpy/common/00accessibility.rpy#L193
which is the human-readable version, and which I translated.
I don't see a reason to translate the internal name of the preference.

Incidentally, your translation is unclear: this is about dropping the global volume of non-voice channels *in favor* of the voicing channel, as explained at:
https://github.com/renpy/renpy/blob/27bc9891060f5eee62983b2881906fc8ae1501c4/renpy/common/00preferences.rpy#L180
(I translated as "Baisse de volume pendant la lecture vocale")

renpytom:
I need to check, but I think it's pretty likely I marked the internal names for translation by mistake, when using an automatic tool to mark strings to be translated.

Gouvernathor:
I thought the low-caps version was the one to be passed to auto-voicing, hence why I translated it.
Good point @Beuc, I'll look into how I actually translated it, might need a correction.

PR #2698, opened 2021-02-27
Harmonizing and correcting french translation

Gouvernathor:
As pointed out in #2695, plus fixing an inconsistency


Gouvernathor:
I chose to translate "self-voicing" to "lecture vocale", because although being more precise and accurate, "vocalisation automatique" was too lengthy and risked overflowing the places it's put in.

Issue #2466, opened 2020-11-14
An In-built Dark Mode


peachiestking:
I understand you can do this via editing some files in the launcher, but I feel as though it would be better to have an option in the preferences screen or something to enable or disable dark mode or to use a custom skin, which can be added via file editing.

I made a concept for what this would look like in preferences and how this would look in general. I can provide my Theme.rpy also.

![DarkModeConcept](https://user-images.githubusercontent.com/45401754/99145781-61d98900-2669-11eb-8909-5168956fcebf.png)

The concept I made is quite quick and not too detailed, but I doubt it needs to be.

Here is the code in my Theme.rpy:

```
init python:
    # The color of non-interactive text.
    TEXT = "#F5F5F5" #545454

    # Colors for buttons in various states.
    IDLE = "#BDBDBD" #42637b
    HOVER = "#9E9E9E" #d86b45
    DISABLED = "#757575" #808080

    # Colors for reversed text buttons (selected list entries).
    REVERSE_IDLE = "#BDBDBD" #78a5c5
    REVERSE_HOVER = "#9E9E9E" #d86b45
    REVERSE_TEXT = "#757575" #ffffff

    # Colors for the scrollbar thumb.
    SCROLLBAR_IDLE = "#5c6e91" #dfdfdf
    SCROLLBAR_HOVER = "#31326f" #d86b45
    # An image used as a separator pattern.
    PATTERN = "images/pattern.png"

    # A displayable used for the background of everything.
    BACKGROUND = "theme/theme_background.jpg"

    # A displayable used for the background of windows
    # containing commands, preferences, and navigation info.
    WINDOW = "#393e46" # Frame("window.png", 0, 0, tile=True) #ffffff80

    # A displayable used for the background of the projects list.
    PROJECTS_WINDOW = Null()

    # A displayable used the background of information boxes.
    INFO_WINDOW = "#393e46" #f9f9f9c0

    # Colors for the titles of information boxes.
    ERROR_COLOR = "#f05454" #d15353
    INFO_COLOR = "#34626c" #545454
    INTERACTION_COLOR = "#9E9E9E" #d19753
    QUESTION_COLOR = "#f5b461" #d19753

    # The color of input text.
    INPUT_COLOR = "#FAFAFA" #d86b45
```

The stuff in the # next to the new colour is the original, but I'm sure you knew this already.

The background.png and the window.png can be replaced from white to `"#212121"`

[...]

PR #2703, opened 2021-03-03
Launcher Dark Theme

Gouvernathor:
Implements #2466 (with different colors)
This adds a dark theme and incidentally a clear theme to the launcher, with the related preference options.

Switching theme triggers a mandatory reload of the launcher from disk, because the variables and displayables' definitions are in init blocks. Why not put the theme redefinition in functions which could be called by both init blocks and the related buttons ?
Because of custom themes. They have to trigger at init 0, and there's no way to trigger their block on the click of a button without reloading the launcher from disk.
Regarding custom themes, I haven't tested all possibilities since they are quite endless, but the changes I made trigger in init -1 except for seemless changes on scrollbar_centers and checkboxes, which (both) check if an image already exists before replacing it.
Therefore, custom themes are applied over and are overriding whichever of these 3 themes is ticked in the options.

Of course, some custom themes may glitch if applied upon a non-default theme if they used default assets or colors, but switching back to default theme in the options will solve that.

The only other change I made was turning the `l_checkbox_box` style into a transform, because applying a style to a Fixed was glitchy. I believe that's the only way a custom theme could become incompatible (but that style wasn't documented anyway).

The colors I put in the clear theme are directly taken from the base assets, the only removal is the gradient at the top of the window and background images, which aren't reproducible programmatically.
The colors of the dark theme are not @peachiestking 's nor @centaurus-dj 's (whom helped me build the dark mode), they are partly custom and mainly taken from the default colors. The result is mainly darker, and with kinda flashier color accents. I'm not an UI expert, so feel free to express remarks or change proposals, as this is supposed to become the main and common dark theme of the ren'py launcher.


Gouvernathor:
Screenshots, the purple bars are Windows's.
Default (identical to vanilla, if I made no mistake) :
Screenshot_178
Screenshot_179
Clear (note the absence of gradients) :
Screenshot_180
Screenshot_181
Dark theme :
Screenshot_182
Screenshot_183

renpytom:
I remove access to the clear theme in c92151db7 - I think the difference is subtle enough that the end result was confusing.

PR #2705, opened 2021-03-03
doc: clarify video library used

Beuc:
Hi. I got contacted by a dev who wondered about poor video performances and blamed Ren'Py usage of a defunc library (libav.org).
I answered that there was probably confusion between the libav* libraries provided by FFmpeg and the old libav fork.
Nonetheless, the current documentation has an ambiguous usage of "libav" and I figured we could clarify.

### DIFF
--- a/sphinx/source/movie.rst
+++ b/sphinx/source/movie.rst
@@ -3,7 +3,7 @@
 Movie
 =====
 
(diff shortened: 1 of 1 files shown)


renpytom:
Thank you.

PR #2707, opened 2021-03-05
Add frame background to default notify screen

Gouvernathor:
Adds a grey transparent background to improve contrast and readability over any surface.
Comes from the fact that it appears over a white surface in the launcher, and could pose the same problem in games with white backdrops if the gui asset or the screens.rpy notify screen is missing.

### DIFF
--- a/renpy/common/00action_other.rpy
+++ b/renpy/common/00action_other.rpy
@@ -665,7 +665,9 @@ init -1500:
     screen notify:
         zorder 100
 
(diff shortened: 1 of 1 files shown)


renpytom:
I don't think this is the right thing to do for this. Any modern Ren'Py game should be customizing the notify screen. For older games that aren't, it doesn't make sense to change the behavior in an incompatible way. If this needs to be changed for launcher theming, it should be done as part of the launcher, not as part of the engine.

Issue #1891, opened 2019-06-19
OutOfMemory for RAPT on upgrade to 7.3


trajano:
I was able to fix this by just changing the gradle.settings file so instead of 3g, I use 4g.  But it would be nice if there was a setting to set that up from the UI.

Andykl:
Same problem with 500Mb+ projects

```
Ren'Py 7.3.0.271

Starting a Gradle Daemon (subsequent builds will be faster)
:app:preBuild UP-TO-DATE
:renpyandroid:preBuild UP-TO-DATE
:renpyandroid:preDebugBuild UP-TO-DATE
:renpyandroid:checkDebugManifest
:renpyandroid:processDebugManifest
:app:preDebugBuild
:renpyandroid:compileDebugAidl
:app:compileDebugAidl
:renpyandroid:packageDebugRenderscript NO-SOURCE
:app:compileDebugRenderscript
:app:checkDebugManifest
:app:generateDebugBuildConfig
:app:prepareLintJar UP-TO-DATE
:app:mainApkListPersistenceDebug
:app:generateDebugResValues
:app:generateDebugResources
:renpyandroid:compileDebugRenderscript
:renpyandroid:generateDebugResValues
:renpyandroid:generateDebugResources
:renpyandroid:packageDebugResources
:app:mergeDebugResources
:app:createDebugCompatibleScreenManifests
:app:processDebugManifest
:app:splitsDiscoveryTaskDebug
:renpyandroid:platformAttrExtractor UP-TO-DATE
:renpyandroid:generateDebugRFile
:app:processDebugResources
:app:generateDebugSources
:renpyandroid:generateDebugBuildConfig
:renpyandroid:prepareLintJar UP-TO-DATE
:renpyandroid:generateDebugSources
:renpyandroid:javaPreCompileDebug
:renpyandroid:compileDebugJavaWithJavacNote: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.

:renpyandroid:processDebugJavaRes NO-SOURCE
:renpyandroid:transformClassesAndResourcesWithPrepareIntermediateJarsForDebug
:app:javaPreCompileDebug
:app:compileDebugJavaWithJavac
:app:compileDebugNdk NO-SOURCE
:app:compileDebugSources
:app:mergeDebugShaders
:app:compileDebugShaders
:app:generateDebugAssets
:renpyandroid:mergeDebugShaders
:renpyandroid:compileDebugShaders
[...]

PR #2734, opened 2021-03-22
Adjust Java Heap Size Limit in Ren'Py

GanstaKingofSA:
## Issue
In some instances, when building a project on Android, you might get a `java.lang.OutOfMemory` error when Gradle attempts to package your project for release (Issue #1891). This is normally caused when a project is approximately >500 MB in size and will cause the current Java Allocation of 3 GB to be filled completely which causes the following issue.

## Versions Affected
Ren'Py Versions that use Gradle as of late with `-Xms3g` coded in *gradle.properties*

## Solution
This PR addresses this issue by making a button available to the user under **Other**, below _Logcat_ named __Change Java Heap Size__ with this description.
> Opens "gradle.properties" in a editor in order to change the Java heap size limit in RAPT. Use this if you received a "java.lang.OutOfMemory" error while building.

Button Code
```python
textbutton _("Change Java Heap Size"):
    action AndroidIfState(state, ANDROID_NO_CONFIG, editor.Edit(config.basedir + "/rapt/project/gradle.properties"))
    hovered tt.Action(HEAP_SIZE_TEXT)
```

This button will only be clickable once the Android SDK is installed to avoid any file does not exist errors.

### DIFF
--- a/launcher/game/android.rpy
+++ b/launcher/game/android.rpy
@@ -49,7 +49,7 @@ init python:
 
     DEBUG_TEXT = _("Selects the Debug build, which can be accessed through Android Studio. Changing between debug and release builds requires an uninstall from your device.")
     RELEASE_TEXT = _("Selects the Release build, which can be uploaded to stores. Changing between debug and release builds requires an uninstall from your device.")
-
+    HEAP_SIZE_TEXT = _("Opens {i}gradle.properties{/i} in a editor to change Java Heap Size in RAPT.\n\nUse this only if you recieve a {i}java.lang.OutOfMemory{/i} error while building.")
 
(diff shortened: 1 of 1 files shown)


renpytom:
Let me think about this. I'm wondering if it mike sense to add a gradle heap size field in the android configuration, and then generate gradle.properties on build.

GanstaKingofSA:
That can work too. I did do that method in another solution.

The only stuff I found that should be noted for thia thought is a failsafe to not be less than 3 GB but not greater than 8 GB or the max RAM size of the system. 8 GB seems reasonable but it may not be a full fix to some rare big sized projects.

PR #2741, opened 2021-03-27
Add example explaining label parameter scope

jsfehler:
This really threw me off so I hope an example in the docs will help someone else.


renpytom:
I've approached this in a different way with 349b20440. I think your example focuses on what happens in the problematic case, but that's not really the way documentation should work. I've added some documentation to explain how label parameters are intended to be used (with calls), and then point the reader at the call example.

PR #2790, opened 2021-05-03
Add a fallback value for ConditionSwitches

Gouvernathor:
When no terminal value is given to a ConditionSwitch, it shows Null().
Avoids triggering a "could not choose a displayable" exception.

Could be linked to a compatibility config variable.

### DIFF
--- a/renpy/display/layout.py
+++ b/renpy/display/layout.py
@@ -1475,7 +1475,7 @@ def ConditionSwitch(*args, **kwargs):
     if len(args) % 2 != 0:
         raise Exception('ConditionSwitch takes an even number of arguments')
 
-    for cond, d in zip(args[0::2], args[1::2]):
+    for cond, d in zip(args[0::2], args[1::2])+[(True, Null())]:
 
         if cond not in cond_cache:
             code = renpy.python.py_compile(cond, 'eval')
--- a/renpy/display/layout.py
+++ b/renpy/display/layout.py
@@ -1418,7 +1418,9 @@ def condition_switch_pick(switch):
         if renpy.python.py_eval_bytecode(code):
             return d
 
-    raise Exception("Switch could not choose a displayable.")
+    if renpy.config.developer:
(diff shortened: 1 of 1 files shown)


Andykl:
It won't work with python 3 because zip creates an iterator, not a list. And also, this is a backwards incompatible change, I think it's better to do it as keyword parameter instead of config variable, although maybe make a config that defines the value of this parameter if it is not specified.

renpytom:
I could see doing this if renpy.config.developer is False, and of course made to work with Python 3. When the game is under development, a failure like this should not be hidden. It should probably be done  just above the exception, so it's zero-cost in the non error case. 

I don't think this is a compatibility change, since it would just be making an error the player would see a non-error. 

Do you want to make the change, or should I do it?

Gouvernathor:
Allowing it for release safety is a good idea, but I think it would save a lot of copy-paste to allow devs to just add the fallback value to all their ConditionSwitch by default.
Maybe with a config variable which would keep the current behavior by default, and only add the fallback value if the dev opts-in the config variable (or the developer mode is disabled) ?

And maybe have an additional kwarg which would override the behavior in an instance-per-instance basis, but if the former is added I feel it would over-complicate it.

renpytom:
No on having a fallback value by default. Not covering all the cases with a ConditionSwitch is an error, and we shouldn't make that error go unseen. If any particular developer wants this, they could always write their own ConditionSwitch wrapper - 

```
init python:
    def ConditionSwitchWrapper(*args, **kwargs):
         args = args + ("True", Null())
         return ConditionSwitch(*args, **kwargs)
```

But getting things wrong is not something the engine should encourage. (Heck, at this point we should not be encouraging the use of ConditionSwitch. It's usually a sign of a poorly-programmed game, now that LayeredImage exists.)

Gouvernathor:
I believe this is what you want - no config variable, only avoiding the exception when the game is in non-dev mode.

PR #2803, opened 2021-05-10
Fix typo and vary showcases

Gouvernathor:


renpytom:
I don't like the rewrite - I had tried that, and found it ugly to read, and a bit confusing. (I know where you were trying to go, I went there to, but fundamentally though it a bit overwhelming.) 

I've taken the typo fixes - thank you.

PR #2804, opened 2021-05-10
Add quality-of-life editing abilities to the Input field

kobaltcore:
## What is this?
This PR adds common editing capabilities normally found in text editors like Atom, Sublime Text, Visual Studio Code etc to the input handler.

Namely, it contains the following features:
- Jump word boundaries forwards and backwards via pressing the ALT modifier key and the left or right arrow keys
- Jump to the start or end of the current line via pressing the CMD or Windows modifier key and the left or right arrow keys
- Backward delete from the start of the current line up to the cursor via pressing the CMD or Windows modifier key and backspace

## Why?
The goal of this is to make it much easier to navigate around long lines of text, especially in the in-game console, but likely also in any regular input field.

Specifically while debugging, one may have long strings in the console and changing characters at the beginning of one can be quite the hassle. The current model forces one to hold the arrow keys and move to the front character-by-character, which may take a while and is suboptimal.

The word boundary jumps can cope with additional whitespace, so the below line would still work properly when jumping between words:
```
this is      a  test
```

## Discussion

There is at least one thing that still requires feedback or some discussion at minimum, which is the keymap. I'm not sure how to create key bindings which are platform specific.

This chiefly concerns the full jumps, which are bound to CMD on macOS (which is correct).
However, using `meta_` binds the same keybind to the Windows key on Windows whereas it should be `ctrl_`.

If we could find a way to differentiate this keybind between the two platforms, we should be golden.
Using `ctrl_` on both is not possible because `ctrl_` + arrow keys switches Spaces on macOS, whereas Windows + arrow keys activates window management on Windows, so neither is universal.

I think this might also interfere with Skipping controls, so maybe choosing an entirely different modifier key would be best.
Let me know your thoughts.


kobaltcore:
In that case we would switch from `meta_` to `ctrl_`.
That would be compliant on Windows and Linux, but leaves macOS as the odd one out.

Is the conditional that @Gouvernathor proposed acceptable to differentiate this?
```py
['meta_K_BACKSPACE' if renpy.macintosh else 'ctrl_K_BACKSPACE']
```

Alternatively, having a keybind which maps to Ctrl on Windows and Linux and CMD on macOS may be the optimal way to handle this, but that would depend on updates to the pygame library, if I'm not mistaken.

renpytom:
I wouldn't want the conditional, but we could add the keybind in renpy/display/behavior.py

renpytom:
What do you think it should be called? I'm thinking 'shortcut', but I'm not wild about that name.

Gouvernathor:
I have a doubt about the keystrokes themselves. While Alt+Left/Right work on Atom, the Ctrl+Left/Right keystroke to jump from word to word is much more common and considered as a standard by Microsoft, at the same level as Shift+Left/Right to select a portion of a text for example. It also works on Atom, as well as a lot of other programs including Firefox for example.
Alt+Left/Right also have the inconvenient to be mapped to previous/next page on a lot of programs, including Firefox where it goes back and forward in page history, and Discord where it does the same with servers.
Therefore, I would recommand using Ctrl+Left/Right instead of Alt+Left/Right as a standard for words jumping, at least for Windows and Linux.

Concerning the current assignment of Ctrl+Left/Right, there are already specific keys to do that, called Begin (symbolized as a northwest arrow) and End, and both already work with renpy's console, so I don't consider it to be necessary to map a new keystroke to it, especially one that could confuse users (again, only regarding Windows/Linux systems).
Same about Ctrl+Backspace, which on Atom does the same as Alt+Backspace, which is deleting up to the previous word boundary, and which can be done in the console by swiching lines using the Up/Down keys then back to the bottom one. In that case a keystroke could be added to do it more cleanly and to add it to input fields, but Ctrl+Backspace is perhaps not the best choice.

kobaltcore:
Comparing to Atom may only be partially useful, since keymaps may differ per editor. For example, SublimeText handles it the way I implemented the first version here, and VSCode may do so as well (though my keymap there is customized, so I can't tell exactly).

The keybinds are certainly able to be changed still, if we can come up with better combinations.

renpytom:
I think the CUA bindings are fairly standard, so we should follow those.

Gouvernathor:
The [CUA](https://wptoolbox.com/library/CUA-DOS.pdf) apparently does advise Ctrl+Left/Right for word jumping, and Ctrl+Backspace to delete a word.
However, according to [Wikipedia](https://en.wikipedia.org/wiki/Table_of_keyboard_shortcuts), the `Opt` key (which is Alt I guess ?) is used instead of Ctrl in both cases, so I believe it would make sense to keep Mac in accord with its own standard.

kobaltcore:
Using Ctrl+Arrows to jump words would introduce a slight conflict, where macOS has two distinct keys (Alt and CMD) but on Windows and Linux, Ctrl would take on responsibilities from both, which is not possible.

For example, jumping between words with Ctrl+Arrows would require new keybinds for the full backward and forward movement (as well as delete). On Windows, the new key proposed for this appears to be "Fn", which I've seen on laptop keyboards, but not on regular desktop ones, which could be a point of confusion.

Of course we could drop the full line jumps entirely in favor of just having the Begin/End keys, but personally I'd find it rather unergonomic to press those instead of jumping around via keybinds where my hands are already positioned.

Imho I think it'd be better to remain consistent internally rather than to adopt a standard that introduces additional conflicts.
The following key bindings should be relatively intuitive, are close together and allow one to move around the text without having to reposition ones' hands:
- Jump Word Left: Alt+Left on all platforms
- Jump Word Right: Alt+Right on all platforms
- Delete Word Left: Alt+Backspace on all platforms
- Jump to End: Ctrl+Right on Windows/Linux and CMD+Right on macOS
- Jump to Start: Ctrl+Left on Windows/Linux and CMD+Left on macOS
- Delete to Start: Ctrl+Backspace on Windows/Linux and CMD+Backspace on macOS

This would also be congruent with the "merged" new keybind which aliases between CTRL and CMD on Windows/Linux and macOS respectively. If we were to swap the Ctrl/Alt responsibilities around, that setup would not work anymore.

Of course some of this may be down to personal preference, but I guess there's two things to decide on:
- Do we keep the Jump to Start/End keybinds or do we remove them entirely since Begin/End already does the same?
- Do we follow an internally consistent keybind scheme or do we adopt CUA?

If I were to build the keybinds according to the CUA table, they'd look like this:
- Jump Word Left: Ctrl+Left on Windows/Linux and Alt+Left on macOS
- Jump Word Right: Ctrl+Right on Windows/Linux and Alt+Right on macOS
- Delete Word Left: Ctrl+Backspace on Windows/Linux and Alt+Backspace on macOS
- Jump to End: End or Fn+Right on Windows/Linux and CMD+Right on macOS
- Jump to Start: Begin or Fn+Left on Windows/Linux and CMD+Left on macOS
- Delete to Start: Ctrl+Home on Windows and CMD+Backspace on macOS

Personally I think this makes things less portable and more confusing, but your guys' opinion may differ.
Since Ren'Py is portable across many different platforms, I'd argue that a consistent keybind scheme that ports with minimal changes across platforms would be a boon, not a drawback.
The CUA scheme would also make it impossible to have the `osctrl` keybind introduced above as there is no consistent way to use it in that case.

Gouvernathor:
If the keybindings have to be consistent between all versions of Renpy, imo the CUA ones should be chosen and not Mac ones.

The Fn key is not a standard even within windows systems, it actually depends on the keypad builder. In my case Fn+Right is mapped to the End key, but on other computers it can be volume+ or brightness+. Most of the time the key doesn't exist.
Afaik the substitution is done either by hardware or by the OS, so I doubt Ren'py would even register an Fn keystroke.

So my vote would go to the last table you proposed.
Either by adopting Windows/Linux ones as common to all renpy versions (which would make jump word undoable on Mac if I understood correctly).
Or by solving those key issues.
My solution would be to have Ctrl+Backspace *and* Alt+Backspace doing the same thing on all platforms (like it does in Atom afaik), which is deleting the previous word.
Then, Delete to Start could be mapped to Windows+Backspace on Windows/Linux, which is not already OS-mapped (unlike Windows+Left), so it would be CMD+Backspace on all platforms. We could leave it documented only for Mac, just as CMD+Left/Right, since it's not as useful in renpy as it can be in text programs : Up-Down can be done in the console or Rollback-Rollforward in an input to clear the line.

Sum up (the **bold** is already implemented) :
- Jump Word: Ctrl+Left/Right on Windows/Linux and Alt+Left/Right on Mac, `osctrl+Left/Right`
- Delete Word Left: Ctrl+Backspace and Alt+Backspace on all platforms, `Ctrl+Backspace or Alt+Backspace`
- Jump to begin/end: **Home/End on windows/Linux** and CMD+Left/Right on Mac, `Home/End or CMD+Left/Right`
- Delete to start: Windows+Backspace for Windows and CMD+Backspace on Mac, `CMD+Backspace`
That way, `osctrl` means `Alt if renpy.macintosh else Ctrl`, and nothing more.

uyjulian:
On mac, some additional functionality is mapped with the "Fn" key.

Fn+Left arrow: Home
Fn+Right arrow: end
Fn+Up: page up
Fn+Down: page down

Also, there are additional text navigation on mac:  
Ctrl+a: go to beginning of line
Ctrl+e: go to end of line

kobaltcore:
I guess we should stick with keymaps which are common across systems or have useful cross-platform keybind alternatives.

Either way, if we *are* going with the CUA keybinds, then @renpytom might have some input on how keybinds per platform should be handled.
With my original approach this was circumvented (by only adding the `osctrl` binding), but if we have the more complex setup as elaborated upon by @Gouvernathor, then we need to figure out a way of handling keybinds differently per platform.

Tom has already expressed he would prefer not to have `<key_a> if renpy.macintosh else <key_b>` in the keymap, so maybe there is another approach to handling this.

Short of adding support for layered and/or os-dependent keymaps to Ren'Py, are there other options of tackling this?

Gouvernathor:
My proposition does not need any new pseudo-key apart from `osctrl`. It's implementable as-is.
The only platform-dependent mappings are the Jump Word ones. All others are common (provided I was not wrong about CMD meaning the Windows key on windows).

The only other difficulty would be mapping different keystrokes to the same function, but that's already the case with `H` and `mouse3` both hiding the user interface, so I don't believe it would be hard to have Alt+Backspace and Ctrl+Backspace mapped to the same thing.

kobaltcore:
Ah, my bad. The only hang-up of sorts in this case is the CMD-Windows-key conundrum.
From the behavior of the key, CMD is actually much closer to "Ctrl" than the Windows key, as it largely takes over most of the functions that would normally be bound to Ctrl on Windows/Linux. It has very little comparable attributes to the Windows key.
As such, it wouldn't really be proper to alias them as that would not mirror what it's normally used for.
If we're okay with having that little bit of weirdness, then this is an acceptable solution that I can implement for this PR, though I would like Tom to sign off on that since it would kind of muddy the waters between Ctrl-Alt and CMD-Windows between the two platforms.

kobaltcore:
Alright, sorry for dropping off the face of the earth for a bit, but I'm back now.

I've implemented the keybinds as @Gouvernathor specified in our earlier conversation, which now means that `osctrl` is `Alt` on macOS and `Ctrl` everywhere else.

The only keybind that may be a bit iffy is this one:
* Jump to begin/end: **Home/End on Windows/Linux** and CMD+Left/Right on Mac

Which is implemented something like this:
```python
input_home = [ 'meta_K_LEFT' if renpy.macintosh else 'K_HOME' ],
input_end = [ 'meta_K_RIGHT' if renpy.macintosh else 'K_END' ],
```

Tom has previously expressed that he would like to avoid this kind of pattern in the keymap, but I don't see an easy way around this, save for creating a new pseudo-key which handles this internally. Not sure if making a new pseudo-key for this is the better alternative here or not.

Gouvernathor:
I think having `input_end = ['meta_K_RIGHT', 'K_END']` (and the same for input_home) would do the job.
It would mean that Home and End are still valid on mac, to begin with.
It would also bind "Windows+Right" to these actions, but we would leave it as an undocumented feature, so people in Windows won't use them and mess around with their window being sent to the side, and we can have a cleaner code without adding another pseudo-key.

The workaround would be to have something like `input_end = ['macmeta_K_RIGHT', 'K_END']`, with `macmeta` as a pseudo-key which would be`meta` on Mac and unmapped on Windows.
But the only advantage would be to disable an undocumented behavior, so I don't think it's worth it.

kobaltcore:
If we're fine with that undocumented behavior on Windows, then that works.
I've made the change, if we want to modify this later down the line we still can, I suppose.

renpytom:
I'm looking at merging this. Shouldn't we be using osctrl in the keymaps?

kobaltcore:
We should indeed. For most of the bindings this was already the case where applicable, but I missed the "delete" one, which has now been rectified.
I also rebased onto master so merging is easier, just in case.

renpytom:
Merged in. Thanks.

PR #2815, opened 2021-05-19
Document Character() not requiring a name

Gouvernathor:
Document that `define eright = Character(kind=e, namebox.xalign=1.0)` is vallid Ren'py.

Another phrasing not mentioning the NotSet sentinel may be better.

### DIFF
--- a/renpy/character.py
+++ b/renpy/character.py
@@ -1342,7 +1342,7 @@ class ADVCharacter(object):
 def Character(name=NotSet, kind=None, **properties):
     """
     :doc: character
-    :args: (name, kind=adv, **args)
+    :args: (name=NotSet, kind=adv, **args)
     :name: Character
 
     Creates and returns a Character object, which controls the look
@@ -1351,7 +1351,8 @@ def Character(name=NotSet, kind=None, **properties):
     `name`
         If a string, the name of the character for dialogue. When
         ``name`` is None, display of the name is omitted, as for
-        narration.
+        narration. If no name is given, the name is taken from
+        ``kind``.
 
     `kind`
         The Character to base this Character off of. When used, the


renpytom:
Character is documented the way I want it to be - explicit calls to it should be given a name. There are cases behind the scenes where that name isn't required, but that's an internal detail that end users shouldn't rely on.

Gouvernathor:
The current implementation makes is crystal-simple to have different versions of the same character, with one detail that's changing - the position of the namebox in my example but usually I'm using it for the sideimage position.

Trying to do this while always specifying the name requires to have `ch2 = Character(ch1, kind=ch1, **kwargs)`, which is fine, just less handy. It works with normal characters as well as with dynamic character taking an evaluable string as a name.

However, with dynamic characters taking a callable, it doesn't work. Having `ch3 = Character(ch1, kind=ch1)` with ch1 being a callable-driven dynamic character raises an exception when trying to make ch3 speak, because it tries to call ch1 with no arguments.

That means you have to write the callable again. If it's a function it just creates code duplication to copy/paste its name, but if it's a class instance it forces you to define it in a variable or a constant. If it's a lambda, you have to copy/paste it (or to define it as a fuction, polluting the namespace in the process).
And then the code becomes harder to maintain, especially if you have the first character defined at the beginning of the script, and the other one right in the middle, like :
```
ch1 "I'm on the right..."
Character(kind=ch1, sideimage_xalign=.0) "And then I'm on the left !"
ch1 "And I'll never be on the left again."
```

Actually it works when you do `Character(ch1.name, kind=ch1)` - at least as far as I tested, a.k.a with ADVCharacters.
But then again, the `name` attribute is not documented, and that code being less handy than `Character(kind=ch1)`, I'd choose to document the name argument rather than the name attribute.

Issue #2827, opened 2021-05-26
[New Feature] Generalized mesh Frame


Gouvernathor:
A more general 2D mesh system which Frame and Tile could be special cases of.
It would basically allow slicing an image horizontally and vertically into rectangular zones, and arbitrarily saying which and how much each is tiled and/or stretched.

renpytom:
Just tossing an idea out here - the px type/object, as part of specifying these. 

The idea would be something like a shorter name for absolute that could combine with object. For example, to specify the x expansion, one could do:

```
[ 0px, 10px, 0.5 - 10*px, 0.5 + 10*px, 1.0 - 10*px, 1.0 ]
```

Where 1.0 - 10 * px means "10 pixels to the left of the right side of the screen."

RenpyRemix:
Theoretically, now we have access to GPU based rendering, 9-patch (using opaque/transparent pixels along two sides to dictate the (possibly multiple) stretch parts) should be quite feasible.

renpytom:
Ren'Py's Frame is nine-patch.

PR #2832, opened 2021-05-29
Add px and the coordinate object

Gouvernathor:
As introduced in #2827

### DIFF
--- a/renpy/display/core.py
+++ b/renpy/display/core.py
@@ -239,14 +239,6 @@ class PX(coordinate):
         cls._store.append(self)
         return self
 
-    def __rmul__(self, other):
-        if isinstance(other, (int, absolute)):
-            return coordinate(offset=other)
-        return NotImplemented
-
-    def __mul__(self, other): # not documented
-        return other*self
-
 px = PX()
 
 
--- a/renpy/display/core.py
+++ b/renpy/display/core.py
@@ -194,6 +194,8 @@ class coordinate:
             return coordinate(offset=self.offset, relative=self.relative+other)
         if isinstance(other, coordinate) and isinstance(self.relative, int) and isinstance(other.relative, int):
             return coordinate(offset=self.offset+other.offset)
+        elif isinstance(other, coordinate):
+            raise AttributeError("Coordinates with a relative value don't support addition")
         return NotImplemented
 
     def __radd__(self, other):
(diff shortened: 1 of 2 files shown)


Gouvernathor:
Summing up all the operations availables :
- rmul is enabled with ints and absolutes - not with floats to avoid confusion and errors - and as long as there is no relative value
  - `10*px`, `absolute(12.5)*px`, `2*(10*px)`, `2*coordinate(offset=10)` are valid
  - `10*coordinate(offset=10, relative=.5)`, `12.5*px` are not
- mul is enabled and results in the same, but it's not documented
  - `2*(px*10)` is the same as `2*px*10` or `2*(10*px)`
- adding a (non-absolute) float as a relative value is enabled
  - `10*px+.5`, `.5+20*px`, `.2+10*px+.3` are valid
- adding two coordinate instances is enabled as long as neither have a relative value
  - `10*px+10*px` is the same as `(10+10)*px`
  - `.5+10*px+(.2+20*px)` is not a valid expression
- radd is enabled in the same conditions as add, and is documented
- sub is enabled to complete add
  - `10*px-.5` is the same as `10*px+(-.5)`
- rsub and unary neg are _**not enabled**_
  - `anything-10*px` and `-(10*px)` are errors
- equality is enabled, two coordinates will be equal if their offset and relative values are equal - if they correspond to the same place on the screen
  - with `a=coordinate(offset=10)` and `b=coordinate(offset=10, relative=.0)`, `a == b` even though `10*a` is valid and `10*b` is not

The coordinate constructor is `coordinate(offset, relative=0)`, which could be documented as `coordinate(*, offset, relative=None)`.
The `offset` and `relative` attributes are documented, the former always being an `absolute`, the latter being either a `float` or 0.

px is a singleton. As `type(None)() is None`, `type(px)() is px`.

Possible improvement : multiplication could be enabled only on px and not on coordinate, because it's kind of absurd to multiply an offset absolute number of pixels by another absolute. Maybe it should be disabled like float multiplication.

Gouvernathor:
Just noticed, to enable the `1.0-10*px` expression, shown in #2827, the unary neg operation and the rsub binary must be enabled. Maybe only when there is no relative value ?

Gouvernathor:
Replaced with #3214

PR #2881, opened 2021-07-02
autosave on input config variable

midgethetree:
Almost all places where Ren'Py autosaves have a configuration variable that can be set to False to prevent Ren'Py from autosaving at that time. The exception is the input function. This commit creates a new config variable autosave_on_input which can be set to False to disable autosaving when players input text.

### DIFF
--- a/renpy/config.py
+++ b/renpy/config.py
@@ -550,6 +550,9 @@ adjust_view_size = None
 # True if we should autosave when a choice occurs.
 autosave_on_choice = True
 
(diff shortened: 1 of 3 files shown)


renpytom:
Looks good, but needs documentation in sphinx/source/config.rst .

midgethetree:
Added some documentation just now, hope it's enough.

renpytom:
It is, thank you.

PR #2891, opened 2021-07-07
Upgrade input's caret

Gouvernathor:
From @mal's idea and code.

Makes the default input caret blinking, and stops it from offsetting characters by 1px when activating/deactivating like it did before (that's the `xsize=0` part, no I don't understand it either but it works).

The blinking effect has 3 opt-outs :
1. you can set `config.input_caret_blink` to a False value for the whole game
2. you can override it for a specific input by setting its new `caret_blink` property
3. you can override all of this by replacing the default caret using the already existing `caret` style property

An addition to `renpy/common/00compat.rpy` can easily be done to disable it for previous versions, although I don't think it's worth it.
Adding `caret_blink` as a keyword in `renpy/screenlang.py` and `renpy/sl2/sldisplayables.py` didn't seem necessary to make it work (as far as I tested) but it can be added.
The config variable and the property could be renamed to reflect that they only change the default caret, and don't apply when a custom one is supplied via the `caret` style property, but I feel it would make their names too long.

### DIFF
--- a/renpy/config.py
+++ b/renpy/config.py
@@ -1162,6 +1162,9 @@ audio_filename_callback = None
 # Should minimums be adjusted when x/yminimum and x/ymaximum are both floats?
 adjust_minimums = True
 
+# Should the default input caret blink ?
+input_caret_blink = True
+
 del os
 del collections
 
--- a/renpy/config.py
+++ b/renpy/config.py
@@ -1163,7 +1163,7 @@ audio_filename_callback = None
 adjust_minimums = True
 
 # Should the default input caret blink ?
-input_caret_blink = True
+input_caret_blink = 1.
 
 del os
 del collections
--- a/renpy/display/behavior.py
+++ b/renpy/display/behavior.py
@@ -1131,11 +1131,10 @@ def input_post_per_interact():
 
 
 def blink(trans, st, at, caret_blink):
-    if int(st*2/caret_blink)%2:
-        trans.alpha = .0
-    else:
-        trans.alpha = 1.
-    return (caret_blink/2)-(st%(caret_blink/2))
+    ttl = caret_blink - st % caret_blink
+    trans.alpha = ttl > caret_blink / 2.
+    return ttl % (caret_blink / 2.)
+
 
 class Input(renpy.text.text.Text): # @UndefinedVariable
     """
--- a/renpy/display/behavior.py
+++ b/renpy/display/behavior.py
@@ -1202,10 +1202,8 @@ class Input(renpy.text.text.Text): # @UndefinedVariable
 
         caret = renpy.display.image.Solid(xsize=1, style=style, **caretprops)
         if caret_blink:
-            self.caret = renpy.store.Fixed(renpy.display.transform.Transform(caret, function=renpy.curry.curry(blink)(caret_blink=caret_blink)),
-                                           xsize=0,)
-        else:
-            self.caret = renpy.store.Fixed(caret, xsize=0)
+            caret = renpy.display.transform.Transform(caret, function=renpy.curry.partial(blink, caret_blink=caret_blink))
+    self.caret = renpy.store.Fixed(caret, xsize=0)
         self.caret_pos = len(self.content)
         self.old_caret_pos = self.caret_pos
 
--- a/renpy/display/behavior.py
+++ b/renpy/display/behavior.py
@@ -1203,7 +1203,7 @@ class Input(renpy.text.text.Text): # @UndefinedVariable
         caret = renpy.display.image.Solid(xsize=1, style=style, **caretprops)
         if caret_blink:
             caret = renpy.display.transform.Transform(caret, function=renpy.curry.partial(blink, caret_blink=caret_blink))
-    self.caret = renpy.store.Fixed(caret, xsize=0)
+        self.caret = renpy.store.Fixed(caret, xsize=0)
         self.caret_pos = len(self.content)
         self.old_caret_pos = self.caret_pos
 
--- a/renpy/display/behavior.py
+++ b/renpy/display/behavior.py
@@ -1161,6 +1161,7 @@ class Input(renpy.text.text.Text): # @UndefinedVariable
                  pixel_width=None,
                  value=None,
                  copypaste=False,
+                 caret_blink=renpy.config.input_caret_blink,
                  **properties):
 
         super(Input, self).__init__("", style=style, replaces=replaces, substitute=False, **properties)
@@ -1192,7 +1193,13 @@ class Input(renpy.text.text.Text): # @UndefinedVariable
             if i.endswith("color"):
                 caretprops[i] = properties[i]
 
-        self.caret = renpy.display.image.Solid(xmaximum=1, style=style, **caretprops)
+        caret = renpy.display.image.Solid(xsize=1, style=style, **caretprops)
+        if caret_blink:
+            self.caret = renpy.store.Fixed(renpy.display.anim.Animation(caret, 0.5,
+                                                                        renpy.display.layout.Null(xsize=1), 0.5),
+                                           xsize=0)
+        else:
+            self.caret = renpy.store.Fixed(caret, xsize=0)
         self.caret_pos = len(self.content)
         self.old_caret_pos = self.caret_pos
 
--- a/renpy/display/behavior.py
+++ b/renpy/display/behavior.py
@@ -1130,6 +1130,13 @@ def input_post_per_interact():
             i.caret_pos = len(content)
 
 
+def blink(trans, st, at, caret_blink):
+    if int(st*2/caret_blink)%2:
+        trans.alpha = .0
+    else:
+        trans.alpha = 1.
+    return (caret_blink/2)-(st%(caret_blink/2))
+
 class Input(renpy.text.text.Text): # @UndefinedVariable
     """
     This is a Displayable that takes text as input.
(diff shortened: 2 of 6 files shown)


renpytom:
This is basically okay, but I don't like the use of Animation here. Let's use a Transform with a function that adjusts alpha instead.

Gouvernathor:
Done and tested.

renpytom:
That function transform isn't right. Two problems with it:

* Nested functions don't pickle, and so I think this might have a problem being loaded in from a save. It should be moved to an outer level, I think. 
* Returning .5 from the transform isn't right, since ``st`` can be anything. It makes sense to properly compute the amount of time next until the blink has to be update. 

I also think it would be better if the blink rate was controllable - it is now, and I suspect that's something creators will ask for.

Gouvernathor:
How about having `config.input_caret_blink` defaulting to 1.0, and having both it and the property actually being the looping period of the blinking ?

And for the function, something resembling this (as is it doesn't work, the blinking is erratic, I must have missed something in the math)
```renpy
def blink(trans, st, at, caret_blink):
    if int(st*2/caret_blink)%2:
        trans.alpha = .0
    else:
        trans.alpha = 1.
    return st%(caret_blink/2)
```
which I would define just before Input in behavior.py and currify in the Transform.

Would that be good ?

renpytom:
I think that works. 

As for blink, I think it should probably be:

```
def blink(trans, st, at, caret_blink):
   st = 2 * st / caret_blink
   trans.alpha = 0.0 if (st % 2.0 > 1.0) else 1.0
   return 1.0 - (st % 1.0)
```

Probably. I haven't tried it.

Gouvernathor:
I found something similar by drawing it on paper, it works fine. Adding it.

renpy/config.py
```diff
@@ -1162,6 +1162,9 @@ def say_attribute_transition_callback(*args):
 # Should minimums be adjusted when x/yminimum and x/ymaximum are both floats?
 adjust_minimums = True
 
+# Should the default input caret blink ?
+input_caret_blink = 1.
```
mal, line 1166: I think having it blink by default will result in considerable increase in frame rate on previously idle screens displaying an input, perhaps `False` ought to be the default?

renpy/display/behavior.py
```diff
@@ -1192,7 +1200,12 @@ def __init__(self,
             if i.endswith("color"):
                 caretprops[i] = properties[i]
 
-        self.caret = renpy.display.image.Solid(xmaximum=1, style=style, **caretprops)
+        caret = renpy.display.image.Solid(xsize=1, style=style, **caretprops)
+        if caret_blink:
+            self.caret = renpy.store.Fixed(renpy.display.transform.Transform(caret, function=renpy.curry.curry(blink)(caret_blink=caret_blink)),
+                                           xsize=0,)
+        else:
+            self.caret = renpy.store.Fixed(caret, xsize=0)
```
mal, line 1208: I think you can simply this section down to:
```suggestion
        if caret_blink:
            caret = renpy.display.transform.Transform(caret, function=renpy.curry.partial(blink, caret_blink=caret_blink))
    self.caret = renpy.store.Fixed(caret, xsize=0)
```

mal (commented):
Couple of suggestions. 🙂

renpy/display/behavior.py
```diff
@@ -1130,6 +1130,13 @@ def input_post_per_interact():
             i.caret_pos = len(content)
 
 
+def blink(trans, st, at, caret_blink):
+    if int(st*2/caret_blink)%2:
+        trans.alpha = .0
+    else:
+        trans.alpha = 1.
+    return (caret_blink/2)-(st%(caret_blink/2))
```
mal, line 1138: I think we can simplify this down too, to use fewer ops and even managing to do away with the conditional:
```suggestion
    def blink(trans, st, at, caret_blink):
        ttl = caret_blink - st % caret_blink
        trans.alpha = ttl > caret_blink / 2
        return ttl % caret_blink / 2

```

mal:
Test code to verify suggested `blink` function (will apply it to the console's caret):

```rpy
init python:
    def blink(trans, st, at, caret_blink):
        ttl = caret_blink - st % caret_blink
        trans.alpha = ttl > caret_blink / 2.
        return ttl % caret_blink / 2.

style _console_input_text:
    caret Transform(Solid('000', xsize=1), function=renpy.partial(blink, caret_blink=1.))
```

Gouvernathor:
The period is `1.`, not `.1`, so the required framerate is 2 frames per second... I think it's a fairly reasonable redraw rate for something rather small.

About the blink rewrite, I don't understand the math but if it works then I guess it's fine to add it.

mal:
On my machine enabling/disabling the blink accounts for around a 5-6% CPU bump when Ren'Py is idle, which isn't the end of the world but also isn't super nice just to blink a cursor on minimised of occluded window. Also consider that on mobile devices CPU usage equates fairly directly to battery consumption. It's definitely a nice addition for some primary input fields where a creator feels it make an impactful difference, I'm just not sure it should be on for all inputs by default. 🤷

That's my thought process on it anyway, I'll leave you and Tom to decide which way to go. 👍 

And a few notes on the `blink` suggestion which might help a bit:
First line works out how long in seconds is left in the current blink cycle.
Second line sets the alpha to 1 when in the first half of the cycle (more than half the cycle left), otherwise 0.
Third line returns in seconds the time until the next half cycle.

renpy/display/behavior.py
Gouvernathor: Shouldn't the last line be `ttl % (caret_blink/2.)` ?

mal: Yup, you're right /o\ I've updated it in the suggestion accordingly. 👍

Gouvernathor:
After comparatively testing the 1. and False values for the config variable, I come up with a difference of half a percent in GPU consumption.
Maybe your results were influenced by a higher refresh rate induced by the priority error in the blink function rewrite ?
I think that difference is negligible and the default should remain `1.`, but if Tom wants to disable it, globally or for older versions of renpy, then so be it.

renpytom:
I don't have a problem with this being on by default. I don't think the framerate is a big deal, and inputs aren't on the screen enough to mess with things like GPU prediction. And I think that blinking by default is a nice cue that says "you can type her".

One change we might want to make - and just because I noticed it here - is we might want to somehow reset the blink cycle each time the player types something. I'm noticing that at least here on Linux, the caret only goes away when I haven't typed in a half-second, and that's a good behavior, I think.

Gouvernathor:
Agreed, and it would also be good if it didn't disappear when the user presses right or left.
I've thought about this but I have no idea how to do it.
I thought using st instead of at in the blink function would suffice, but it doesn't

It could be done in the system that shows the caret in the middle of the text, which we didn't touch so far, but then it could mess up with carets already customized via the `caret` style property in other games. Not sure it's a big deal, but well...

renpytom:
I'll merge this in the next day or so, and do a pass myself.

mal:
Just a few passing thoughts I had on this:

1. Perhaps `caret_frequency` would better capture the nature of the `config` attribute and input property than `caret_blink`.
2. There's already an `anim.Blink` displayable which may be of use here to avoid having a separate `blink`, however it's not immediately clear to me if that would add a cost over and above the current functional Transform approach.

renpytom:
Reviewing this, a few small issues: 

* What's the deal with the Fixed at https://github.com/renpy/renpy/pull/2891/files#diff-132b26cd00a2f575d26dcb33a4d4ef8dcc53b6542e9ec100c271161585d15bd7R1205 
? I'm not sure why it's there, and it likely slows things down. I'm also concerned that setting an xsize of 0 will break more substantial carets.

* https://github.com/renpy/renpy/pull/2891/files#diff-2766cab532c3484eadcd2531156ed9dbecce29566e607f7467133506a03a2a85R601 I'm not sure why config.key_repeat came back, but it should go away. 

* This probably needs a line in 00compat.rpy, as it's a change in behavior.

Gouvernathor:
The xsize=0 thing is weird, but it works in making the width of the caret null when it comes to putting it in the middle of a text line. The white bar visually remains there, as I believe it would need a Flatten operation to make it disappear, or a `Transform(crop_relative=True, crop=(.0, .0, 1., 1.))`.
It keeps the characters to the right of the caret from slightly moving left to right by 1px when the caret moves in the line. You can try typing a long string of random characters in the console and moving the caret back and forth, and compare between versions of the code.

renpytom:
I think I'm convinced by the caret change, and I can't recall any game that used a customized caret, so I don't think it needs compat. 

Last thing:

https://github.com/renpy/renpy/pull/2891/files#diff-132b26cd00a2f575d26dcb33a4d4ef8dcc53b6542e9ec100c271161585d15bd7R1170

Since this is python, the default value of caret_blink gets set when Ren'Py is loading in, and renpy.config.caret_blink won't do anything. What you want to do is default it to None, and then do:

```
if caret_blink is None:
     caret_blink = renpy.config.caret_blink
```

Gouvernathor:
Nice catch.

mal:
One final thought on the `xsize=0`, does it have any impact on the cursor display when using a resized game window? Worth confirming it doesn't vanish away to nothing (or at least no more so than I imagine the existing cursor does currently).

renpytom:
Merged it in, thank you.  I don't think xsize=0 will affect much.

renpytom:
7708bf0cb tweaks it kind of bit. Having tried it, it's kind of tough to position the caret if it's not present after hitting one of the arrow keys.

PR #2894, opened 2021-07-10
Remove obsolete gl2 references in the doc

Gouvernathor:
Obsolete since config.gl2 is now enabled by default.
A message like "This requires model-based rendering, which is enabled by default." could be put in its place, but I don't feel it's really relevent.


renpytom:
I'd like to keep this in until 7.5 or so, and then remove it. We still get a couple of people a week on discord caught using older versions where config.gl2 isn't the default.

PR #2897, opened 2021-07-10
Update screens.rst with an example of a common button

henvu50:
An example of a common button that could be used for various games.

### DIFF
--- a/sphinx/source/screens.rst
+++ b/sphinx/source/screens.rst
@@ -402,7 +402,7 @@ Here's an example of a button with a hoverable arrow image centered to its outli
         idle_background Transform("arrow_idle.png", xalign=0.5, yalign=0.5)
         hover_background Transform("arrow_hover.png", xalign=0.5, yalign=0.5)
         text "north" outlines [ (2, "#333", 0, 0) ] size 44
-        action Jump("label_town_north")
+        action Jump("town_north") # a label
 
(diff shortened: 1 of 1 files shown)


renpytom:
I don't think that this is a good example of button, as it's a reimplementation of a textbutton. 

What would be a good example would be to have  a button that has, say, an hbox inside it that takes both text and an image. 

Stylistically, properties like ``action`` should be before displayables, like text.

henvu50:
> I don't think that this is a good example of button, as it's a reimplementation of a textbutton.

Oh, I see.  So I'm making the mistake of re-implementing a textbutton via button.  IMO, this is important information for new users to have. Would it be okay to warn new users who are reading about button on the wiki to avoid this mistake?  Example:

_"Tip: In most cases, if you need a button with image and text, it's more appropriate to use a textbutton with a background and/or foreground style containing an idle and hover image. However, depending on how much re-usability you need, for say an inventory list, then a button may be more appropriate."_  

I used the explanation you gave on the discord.

renpytom:
I can close it, once conversation stops. 

I'm not sure that we need the warning like this. At the end of the day, there are lots of things people can do with Ren'Py, and some are better than others. I'm not seeing a huge rash of people using buttons where imagebuttons or textbuttons would suffice, and so I don't really feel we need a warning like that in the docs.

PR #2912, opened 2021-07-18
Document access to the warper functions

Gouvernathor:
Using the transition classes' time_warp arguments is much more easy when access is given to the already-defined functions.
The keys to the dict are already documented.

### DIFF
--- a/renpy/common/000atl.rpy
+++ b/renpy/common/000atl.rpy
@@ -24,6 +24,8 @@
 
 python early in _warper:
 
+    from renpy.atl import pause, instant
+
     # pause is defined internally, but would look like:
     #
     # @renpy.atl_warper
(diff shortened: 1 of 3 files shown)


Gouvernathor:
I'm trying to give a better access by turning the renpy.atl.warpers dict into a renpy.warpers (pseudo-)module.
I added `warpers = renpy.python.StoreModule(renpy.atl.warpers)` to the end of exports.py and for the life of me I can't understand why it won't work.
The launcher fails to open as if there was an exception, but
```py
try:
    warpers = renpy.python.StoreModule(renpy.atl.warpers)
except Exception as e:
    warpers = e
```
keeps the launcher from opening too.
I did a sanity check for a syntax error by putting the line in renpy's console and it works fine, so... what the hell.

Gouvernathor:
Switched to the already existing _warper module

PR #2918, opened 2021-07-19
Add DynamicLayeredImageProxy

Gouvernathor:
Like #2532, but good.
The resulting class could (theoretically) work in replacement of DynamicImage.
Maybe it could be smoothened around the edges, and it should probably be subject to more thorough testing.

For the moment I spotted some positioning differences between it and dynamicimage, when the underlying displayable changes size, but it seems the main objective of live-refreshing a layeredimageproxy is reached.

The `allow_layim` callback part is quite nasty per se, but I wanted to avoid code duplication (which was unavoidable for the find_target method) and I lack knowledge about that kind of functions researching and converting displayables.

### DIFF
--- a/renpy/common/00layeredimage.rpy
+++ b/renpy/common/00layeredimage.rpy
@@ -1,4 +1,4 @@
-﻿init offset = -100
+init offset = -100
 
 python early in layeredimage:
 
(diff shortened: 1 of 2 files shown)


renpytom:
I'm still not convinced that going back to this sort of dynamism is a  good thing, given that the point of LayeredImage was to get away from computing images like this. We're kind of stuck with DynamicImage, but I don't want to encourage the use of this programming style going forwards. 

This might make sense in certain games, but I don't think this is something I want in Ren'Py.

Gouvernathor:
The fact is that the current LayeredImageProxy enables this kind of dynamism with string substitution, and that given it's not checked on a per-interaction basis, it can leave an image in a state of limbo. If the state of the variables orders it to change what it's showing but no refresh happened, it creates situations where reloading, or advancing and rolling back, change what is displayed.
I'm not advocating using this instead of DynamicImage, but instead of the currently imperfect LayeredImageProxy.

PR #2921, opened 2021-07-20
Avoid using a dict as a default argument value

mal:
It's dangerous as it becomes trivial to accidentally pollute what is expected to be a clean dictionary. The default `group_properties` as it stands never leaves this function, and is never written to, and so the current code is safe in that regard, however leaving it in the codebase becomes a landmine just waiting to be stepped on in future.

```py
>>> def foo(bar={}):
...     return bar
... 
>>> foo()
{}
>>> foo()['qux'] = 'leaky'
>>> foo()
{'qux': 'leaky'}
>>>
```

### DIFF
--- a/renpy/common/00layeredimage.rpy
+++ b/renpy/common/00layeredimage.rpy
@@ -261,7 +261,9 @@ python early in layeredimage:
             self.image = None
             self.properties = OrderedDict()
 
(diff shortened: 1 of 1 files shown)


renpytom:
I don't think this is a real problem, but since it's recently changed, I'll take the fix.

PR #2922, opened 2021-07-21
Typos

Gouvernathor:
In addition to this, the top of the file describes the default for config.gl2 to be False, but it's not the case anymore. I understand leaving advice on the meshed transform properties to remind people that meshing needs to be enabled, but shouldn't *this* be changed ?


renpytom:
Merged, and I'll remove the disclaimer.

PR #2927, opened 2021-07-21
Search archives in all searchpaths, not just gamedir

korden32:
It makes possible to put rpa archives in Android overwrite folder

### DIFF
--- a/renpy/main.py
+++ b/renpy/main.py
@@ -367,14 +367,15 @@ def main():
                 archive_extensions.append(ext)
 
     # The basename is the final component of the path to the gamedir.
-    for i in sorted(os.listdir(renpy.config.gamedir)):
-        base, ext = os.path.splitext(i)
+    for dir in renpy.config.searchpath:
+        for i in sorted(os.listdir(dir)):
+            base, ext = os.path.splitext(i)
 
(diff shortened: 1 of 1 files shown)


Gouvernathor:
gamedir should probably be added to that list because since the user can modify it at will and it's specified to be *additional* search paths. And I believe it should be searched first, to ensure backwards-compatibility.
So I'd advocate something like `for dir in [renpy.config.gamedir] + renpy.config.searchpath`

korden32:
> gamedir should probably be added to that list 

Isn't it included in line 329? `renpy.config.searchpath = [ renpy.config.gamedir ]`

Gouvernathor:
Indeed, it is, but it can be modified by the user, which can remove it from the list.
That being said, and unless I'm mistaken, the line you quoted and the one you modified folloxing each other, whatever the user have put in the list won't be taken in consideration, rendering your edit non-working. Have you tested it ?

renpytom:
7.4.8 will be very android focused, and out soon after 7.4.7, so consideration of this PR will wait until after 7.4.7 is out.

renpytom:
Merged, and I'll clean it up a bit. Thank you.

PR #2936, opened 2021-07-25
Update loadsave.py

Meithal:
While stuck in an infinite loop during saving, trying to break from it with a keyboard interrupt causes a crash because that exception args is an empty tuple, so we get an uncaught exception here, this checks that looking args[0] or args[1:] won't lead to a crash.

### DIFF
--- a/renpy/loadsave.py
+++ b/renpy/loadsave.py
@@ -414,7 +414,9 @@ def save(slotname, extra_info='', mutate_flag=False):
         if bad is None:
             reraise(t, e, tb)
 
-        e.args = (e.args[0] + ' (perhaps {})'.format(bad),) + e.args[1:]
+        if len(e.args) > 1:
+            e.args = (e.args[0] + ' (perhaps {})'.format(bad),) + e.args[1:]
+
         reraise(t, e, tb)
 
     if mutate_flag and renpy.python.mutate_flag:


renpytom:
How did you get into an infinite loop during saving?

Meithal:
That's what I'm trying to understand right now, right now I have to hit Ctrl+C twenty times before the game really exits, and prints an other traceback because e.args[0] raised an IndexError, with this patch it is easier. I have set use_cpickle to False and pasted a pickle.py file into the game library to see what is going on, I will try to submit an other PR if the freeze can be avoided

Meithal:
Actually saving just takes a lot of time, the infinite loop seems to be here

*** <type 'exceptions.KeyboardInterrupt'>  ()
******  ()
  File "..\..\renpy-7.4.6-sdk\renpy.py", line 198, in <module>
    main()
  File "..\..\renpy-7.4.6-sdk\renpy.py", line 194, in main
    renpy.bootstrap.bootstrap(renpy_base)
  File "D:\urp\renpy-7.4.6-sdk/renpy/bootstrap.py", line 370, in bootstrap
    renpy.audio.audio.quit()
  File "D:\urp\renpy-7.4.6-sdk/renpy/audio/audio.py", line 987, in quit
    renpysound.quit()
KeyboardInterrupt
Error in atexit._run_exitfuncs:
  File "/home/tom/ab/renpy-build/tmp/install.android-x86_64/lib/python2.7/atexit.py", line 24, in _run_exitfuncs
KeyboardInterrupt
Error in sys.exitfunc:
  File "/home/tom/ab/renpy-build/tmp/install.android-x86_64/lib/python2.7/atexit.py", line 24, in _run_exitfuncs
KeyboardInterrupt

Meithal:
Removing weakrefs from the game store seems to have fixed the issue on my side, I have no idea how to reproduce the issue but I suspect it has something to do with weakrefs making cyclical references with a very big game store (854458 log = <RollbackLog>).

renpy/loadsave.py
```diff
@@ -414,7 +414,9 @@ def save(slotname, extra_info='', mutate_flag=False):
         if bad is None:
             reraise(t, e, tb)
 
-        e.args = (e.args[0] + ' (perhaps {})'.format(bad),) + e.args[1:]
+        if len(e.args) > 1:
```
mal, line 417: Think the condition here should be `if len(e.args) > 0:` or even just `if len(e.args):` . As it is we lose the error augmentation in the event that the tuple only has a single item.
The `[1:]` slice is safe on both empty and single item lists, so it's only the direct `[0]` access that needs the guard.
```py
>>> [1][1:]
[]
>>> [][1:]
[]
```

PR #2972, opened 2021-08-11
Function to get the files inside the folder.

qTich:
The goal is to be able to receive the contents of the folder on the Android OS.

### DIFF
--- a/renpy/loader.py
+++ b/renpy/loader.py
@@ -293,6 +293,8 @@ loadable_cache = { }
 # A map from filename to if the file is downloadable.
 remote_files = { }
 
+# A map with found files during scanning. 
+folder_files = { }  # type: dict[str, list]
 
 def cleardirfiles():
     """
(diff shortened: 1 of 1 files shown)


renpytom:
Isn't this redundant to renpy.list_files() ?

renpytom:
I'm not sure what you mean by "seems sad".  renpy.list_files is fast, as it's using game_files and common_files, not hitting the disk.

KiloOscarSix:
Correct me if I'm wrong, but I think he means that `list_files()` returns all files and dirs while his function you can include a path argument.

qTich:
I think that I agree with you about redundancy, since I received a value of `0.05-0.1 seconds` (on my phone) for `7717 files` when filtered based on `2-conditions`.

Rather, the question is in convenience, because for a personal computer you can use `os.listdir(path)`, and it will spend less time. Based on this, if I want to get a maximum speed optimization, then it should do so:
```py
files = [file for file in renpy.list_files() if file.startswith(path)] if renpy.android else os.listdir(path)
```
And it looks not very beautiful. In the decision that I proposed you can avoid such a case.

I think that it is possible to close the request, as this decision is not necessary.

renpytom:
Files don't exist on android, so I'm closing this.

Issue #2978, opened 2021-08-15
Lint isn't config.adjust_attributes aware


mal:
The new `config.adjust_attributes` feature isn't applied by lint when checking image validity leading to cases where lint reports issues that work correctly in game.

PR #2979, opened 2021-08-15
Make lint config.adjust_attributes aware

mal:
Applies the function when checking a recognised image, and stashes the original name so that the error report will match the line in the script being referenced (rather than reporting the adjusted version).

I believe this is the correct ordering, as it means it's the "real" tag that makes it's way into `image_prefixes`, but I'm not 100% sure.

Seeks to address https://github.com/renpy/renpy/issues/2978.

### DIFF
--- a/renpy/lint.py
+++ b/renpy/lint.py
@@ -271,6 +271,11 @@ def image_exists(name, expression, tag, precise=True):
     Checks a scene or show statement for image existence.
     """
 
+    orig = name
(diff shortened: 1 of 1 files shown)


Gouvernathor:
Maybe you should put this in a try/except structure. I don't really know the python context when this is executed, but the function call may have side-effects, or reference variables defined through the game script.

mal:
It's explicitly called out in the documentation of this function that it should not rely on game state due to being used during prediction, and that it shouldn't have side effects holds true for the same reason.

PR #2982, opened 2021-08-16
Add default projects directory for Ren'Py and move starter projects into projects dir

TheShermanTanker:
This is a minor commit adding a change I personally found useful while working with the engine, having a default projects directory, while also removing the hardcoded project scanning in renpy's root dir

Despite the large size of the PR, most of it comes from the moving of the starter projects into the new dedicated projects directory, the actual file changes themselves are very small


renpytom:
I'm not sure what the justification for this change is.

Issue #2139, opened 2020-03-19
with statements after call screen behaving


Gouvernathor:
Given the following screen :
```
screen whatever:
    textbutton _("Continue") action [Return()]
```
The following code makes the screen disappear immediately after the click on the button, then waits 3 seconds, then shows the say dialog :
```
call screen whatever with None
with Dissolve(3)
e "hey"
```
Which is a contradiction to the last paragraph of [this section](https://www.renpy.org/doc/html/screens.html#call-screen) of the doc.

Furthermore, the following unexpectedly works exactly as the doc's specifications says :
```
call screen whatever with None
with fade
e "hey"
```

Andykl:
It's controlled by `config.implicit_with_none`. If it's True - Ren'py did a hidden with None after each interaction. You may disable it in general by set this to False or just in case with `call screen whatever(_with_none=False) with None`

PR #2993, opened 2021-08-23
Document screens' _with_none kwarg and the call screen's with behavior

Gouvernathor:
adresses #2139 
The fade transition has been replaced with pixellate, because it's clearer to see what's happening in the example.


renpytom:
Merged, and I'll be editing it a bit.

Issue #2895, opened 2021-07-10
[Feature Request] size transform property should accept positions instead of ints


Gouvernathor:
Currently, `Transform(size=(100.0, 100.0))(Solid("#f00"))` amounts to a red 100x100px square.
There is no actual way of telling a displayable to take half or any fraction of the available room as height or width.
With a Solid `crop_relative` could be used but not for a proper image.
`xzoom=.5` would tell it to take half of its native size, not half of the maximum available size.

I think it could be a nice (retro-incompatible) addition to enable floats in `size`, `xsize` and `ysize` to mean a fraction of the screen instead of a number of pixels, thus paving the way for px and coordinates in a near future.
Except for `offset`, `size` is the only transform property to only allow numbers of pixels instead of any position value.
Furthermore, the corresponding `xsize`, `ysize` and `xysize` style properties support it (and have recently been fixed about it).

renpytom:
It probably makes sense to rename the TransformProperty to xysize, and make the change then,.

PR #3003, opened 2021-08-27
Add xycenter and xysize, modify xsize and ysize

Gouvernathor:


Gouvernathor:
Adresses #2895

renpy/config.py
```diff
@@ -1177,6 +1177,9 @@ def say_attribute_transition_callback(*args):
 # Should Ren'Py raise exceptions when finding an image?
 raise_image_exceptions = True
 
+# Should the size transform property only accept numbers of pixels ?
+old_size_tpref = False
```
renpytom, line 1181: This should probably be positive (defaulting to true), and tpref is something that's only used in the documentation - it's short for transform property reference. How about:

```
relative_transform_size = True
```

renpytom (commented):
Looks mostly good, though some names should be changed. 

This needs documentation and changelog entries, too.

Gouvernathor:
I took the liberty to move `size` down a bit in the list of transform properties, to store it next to `maxsize`.
I'll mark this as ready once I have tested it.

renpytom:
I merged this in, so that it gets built by the nightly build system, and so I can test it here. It looked ready to me.

PR #3007, opened 2021-08-30
Enable all updates in the daily check

Gouvernathor:
Adds a preference which disables the check whether updates are Releases or not, so that nightly updates (and all others) can trigger the red button.

Also moves the preference up a bit so that theme-related options are back on their own.
The nightly option is not displayed when daily checks are disabled, but it can just be rendered inactive to avoid making the height of the buttons list move back and forth.

Concern should be given to the fact that with this preference button, adding another will make the list overflow, which is already the case when set to piglatin language.
Maybe an advanced version of this could be a single button which would cycle between "check for none", "check for releases" and "check for any" ? But I'm not sure how to implement that.

### DIFF
--- a/launcher/game/preferences.rpy
+++ b/launcher/game/preferences.rpy
@@ -62,6 +62,7 @@ default persistent.legacy = False
 default persistent.force_new_tutorial = False
 default persistent.sponsor_message = True
 default persistent.daily_update_check = False
+default persistent.daily_check_nightly = False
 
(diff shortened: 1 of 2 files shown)


renpytom:
I think it might make sense to allow this for prereleases, but it's a bad idea for nightlies. There's no guarantee a nightly is going to even run. So we really don't want to automate updates to it.

Issue #3011, opened 2021-09-02
layeredimage not displayed as intended when the same attribute is repeated


magicnutsgames:
Hi. I recently updated my Ren'py from 7.3.5 to 7.4.8 and noticed that when showing layered images, repeating the same attribute twice messes up the images, something that didn't cause any issues in the previous version.
For example:

`show lana mouthsmile mouthsmile eyesclosed`

The repetition of the smiling mouth in this case will cause the layered image to display the 'default' attribute for all of the groups that have a default attribute set for them (in addition to showing the repeated attribute). Going with this example, the character Lana will have her smiling mouth shown correctly, but the closed eyes will not be shown and instead, the sprite's neutral eyes will be displayed. 

[image files](https://whatalegendgame.com/wp-content/uploads/2021/09/pov.zip)
The layeredimage:

```
layeredimage pov:
    always:
        "pov/povnull.webp"

    group extraback:
        attribute xbshadowhwall2
        attribute xbshadowhwall1
        attribute xbnone:
            "null"

    group pants:
        attribute pneu default:
            "pov_base"
        attribute pdown0
        attribute pdown1
        attribute pdown2_1
        attribute pdown2_2
        attribute pdown3
        attribute ptopless

    group erect:
        attribute erect0:
            "null"
        attribute erect1
        attribute erect2
        attribute erect1topless
        attribute erect2topless

    group eyes auto:
        attribute eneu default
        attribute eclosed
        attribute ewrite
        attribute ecross
        attribute ecrossdown
        attribute ecrossup
        attribute eflat
        attribute eflatb
        attribute eflatc
        attribute ehappy
        attribute enarrow
        attribute enarrowb
        attribute enarrowc
        attribute eneub
        attribute eneubdown
        attribute eneubup
        attribute eneuc
        attribute eneucdown1
        attribute eneucdown2
        attribute eneucup
        attribute eneudown
        attribute eneuup
        attribute eshut
        attribute ewide
        attribute ewideb
        attribute ewidebdown
        attribute ewidebup
        attribute ewidec
        attribute ewidecdown
        attribute ewidecup
        attribute ewidedown
        attribute ewideup
        attribute ewink
        attribute ewinkc
        attribute ewinkdown
        attribute ewinkbdown
        attribute ethink
        attribute ethinknarrow
        attribute ethinkflat
        attribute ethinkwide
        attribute ethinkdown
        attribute ethinkwidedown
        attribute esides
        attribute esideswide
[...]

PR #3015, opened 2021-09-02
Prevent duplicate attributes breaking LIs

mal:
Attempt at fixing #3011 solely within the confines of LayeredImage.

- Keeps any duplicates present in `attributes`. This is necessary to pass the match checking code in `image.py`.
- Extends them with any de-duplicated optional attributes.
- Checks all the attributes it's about to return are valid.


renpytom:
Merged, thanks for tracking this one down.

PR #3038, opened 2021-09-15
Add the color equivalent of the text image

Gouvernathor:
Adds ParameterizedColor, in the same fashion as ParameterizedText, to simplify access to the Solid displayable.
For renpy developpers and bugsweepers, this helps a lot when testing stuff like transitions, attributes...
For game developpers, this can help reduce the amount of superfluous image files simply filled with a single color.

It can also take arguments like `color (1., .0, .0)` instead of `color "#f00"`, but I thought it best not to document this, since it could prompt creators to inject variables or other expressions as arguments.
I added ParameterizedColor equivalents everywhere I could find ParameterizedText.
The lint part of the _duplicate function was extrapolated from the equivalent in ParameterizedText, but it's not tested.
Finally, I placed it next to the Solid displayable in the doc, but there may be a better place for this.


renpytom:
I don't believe that this has enough of an advantage over

```
show expression "#f003"
```

or 

```
show expression Color(...)
```

to be worth adding to the engine. I also really don't like ParameterizedText. The only reason it exists is that it existed before ``show expression`` was a thing. So I don't think this is something that makes sense to add to Ren'Py.

PR #3049, opened 2021-09-21
Use marshal module for the persistent fields responding seen data

### LINKED ISSUE #3042: Slow persistent pickle with large persistent._seen_ever
Our game has 310_000 nodes in `renpy.game.script.namemap` and if I pass the whole game, the size of `persistent._seen_...` (of all sorts) becomes so large that it takes over 5 seconds on my machine to perform `renpy.save_persistent`, whereas with these sets empty it's only 0.1-0.2 seconds.

My suggestion is to put all the RenPy defined persistent data into a separate class (RenpyPersistent or another name) and instead of pickle this data, use marshal. This is possible because only the built-in types are used there. Using marshal even on fully populated sets takes on average 0.2 seconds on my machine.

The format for saving data has 2 options

1. Leave a single file. Approximate new `persistent` file format
```
RENPY PERSISTENT (header)
struct.pack("I", len(Pickled Persistent object))
Pickled Persistent object
Maybe some separator
Marshaled RenpyPersistent data
```

2. Add a new file (renpy_persistent or some other name)

I personally prefer the second option, because it's easier to implement, and as a sideeffect it allows developers to distribute a "100% persistent file" which will not break user settings.

@renpytom if you generally approve the idea, I may come with implementation shortly.

Andykl:
[test_persistent.zip](https://github.com/renpy/renpy/files/7206736/test_persistent.zip) Test project with 250k dialogue lines

This works slower for games with less than 3000 lines, but that amount is saved in less than 0.1 second so regression is not important.
Starting from 5000 seen lines it works the same and for 25000 marshal is ~2x faster on my machine. Further increases in size widen the gap.

To be honest, I don't like the implementation because it is obscure and tricky, but it works in general.

This fix #3042

### DIFF
--- a/renpy/persistent.py
+++ b/renpy/persistent.py
@@ -27,6 +27,7 @@ import copy
 import time
 import zlib
 import weakref
+import marshal
 
 import renpy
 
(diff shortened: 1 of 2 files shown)


Andykl:
It is draft for now. I would like to make it less obscure and also check if converting dicts to sets for marshaling will increase the speed.

renpytom:
I may make a pass at this myself, as I feel that savelocation shouldn't have to know about the format of the save files. (This might mean having separate save_seen, load_seen, and unlink_seen methods.)

uyjulian:
I think it might be a better to save the seen data as bytearray inside the pickled persistent data to avoid separate files.  


Additionally, I think it may be better to convert the marshal data to pickle data e.g. when saving on thread or when closing the game for portability reasons.  

This should allow quicker saves when calling the save function while providing more portability when more time for saving is provided (e.g. saving on thread or closing the game).

Issue #3054, opened 2021-09-24
Request: Pause on MusicRoom


Neyunse:
I think it would be a good idea to add a pause option for the music box; so the player if he is listening to the music gallery can pause it and replay it from where he stopped listening to it when he clicked a pause button.

renpytom:
This has been added to 7.4.9 already.

Neyunse:
But I am getting this error, or was it added in a night version? @renpytom ```
I'm sorry, but an uncaught exception occurred.

While running game code:
  File "game/script.rpy", line 91, in script call
    call expression state from _call_expression
  File "game/story/script.rpy", line 178, in script
    call screen music_m(md=True)
  File "renpy/common/000statements.rpy", line 569, in execute_call_screen
    store._return = renpy.call_screen(name, *args, **kwargs)
  File "game/scripts/PhoneUI.rpy", line 348, in execute
    screen music_m(md=False):
  File "game/scripts/PhoneUI.rpy", line 348, in execute
    screen music_m(md=False):
  File "game/scripts/PhoneUI.rpy", line 350, in execute
    if md:
  File "game/scripts/PhoneUI.rpy", line 355, in execute
    fixed at ajust_center_b:
  File "game/scripts/PhoneUI.rpy", line 376, in execute
    hbox:
  File "game/scripts/PhoneUI.rpy", line 378, in execute
    if not renpy.music.get_pause("music"):
  File "game/scripts/PhoneUI.rpy", line 379, in execute
    imagebutton:
  File "game/scripts/PhoneUI.rpy", line 379, in keywords
    imagebutton:
  File "game/scripts/PhoneUI.rpy", line 384, in <module>
    action [jb.Pause()]
AttributeError: 'MusicRoom' object has no attribute 'Pause'

-- Full Traceback ------------------------------------------------------------

Full traceback:
  File "game/script.rpy", line 91, in script call
    call expression state from _call_expression
  File "game/story/script.rpy", line 178, in script
    call screen music_m(md=True)
  File "/home/neyunse/renpy-7.4.6-sdk/renpy/ast.py", line 2015, in execute
    self.call("execute")
  File "/home/neyunse/renpy-7.4.6-sdk/renpy/ast.py", line 2003, in call
    return renpy.statements.call(method, parsed, *args, **kwargs)
  File "/home/neyunse/renpy-7.4.6-sdk/renpy/statements.py", line 278, in call
    return method(parsed, *args, **kwargs)
[...]

PR #3055, opened 2021-09-24
Added TogglePause on MusicRoom

Neyunse:
If music is currently playing it will pause, otherwise it will resume from where it stopped.
In case a music has already finished, the queue will play the music again from the beginning the next time it is called.
If the music is playing the button will receive the selected status and when paused it will be removed.

Note:
`__MusicRoomTogglePause` is based on `__MusicRoomStop`, because in some tests in my project, if `__MusicRoomTogglePause` was the same as `__MusicRoomTogglePlay` it caused some data about the music being played added using `add(action=[SetVariable("jb_title", title)])` not to be displayed.

Inspired by the report I did before #3054

### DIFF
--- a/renpy/common/00musicroom.rpy
+++ b/renpy/common/00musicroom.rpy
@@ -1,4 +1,4 @@
-﻿# Copyright 2004-2021 Tom Rothamel <pytom@bishoujo.us>
+# Copyright 2004-2021 Tom Rothamel <pytom@bishoujo.us>
 #
 # Permission is hereby granted, free of charge, to any person
 # obtaining a copy of this software and associated documentation files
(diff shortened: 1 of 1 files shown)


renpytom:
Ah, shoot - before I saw this, I already implemented MusicRoom.TogglePause. 

Can you try out  ad8ff33bd and make sure it fixes the SetVariable issue?

Neyunse:
> Ah, shoot - before I saw this, I already implemented MusicRoom.TogglePause.
> 
> Can you try out [ad8ff33](https://github.com/renpy/renpy/commit/ad8ff33bd4ac1692fccbb70a0be5d264486c59fb) and make sure it fixes the SetVariable issue?

No, the data is not displayed. Appears only "()"

screenshot-0004

This is how it should look

screenshot-0005

Neyunse:
@renpytom I discovered a solution, and it is to add ```
def periodic(self, st):
            
            self.mr.periodic(st)

            return .1
```

https://github.com/Neyunse/renpy/blob/master/renpy/common/00musicroom.rpy#L142

renpytom:
Merged, thanks.

PR #3075, opened 2021-10-04
Fix scope deletion in screen.py

multimokia:
the `try`/`finally` block in screen.py still had exception possibilities, and when raised due to an unsafe dict key removal in the `finally`  block, the original exception was completely lost, leaving what was shown as something somewhat unrelated and very misleading.

This PR adds the `except` block and prints its exception info to the log file, making it easier to see what went wrong and where, additionally I have changed the `del` statement to be safer, using `dict.pop` with a default value of `None` as to not raise a further exception.

Examples from traceback:
image
image

Where `_scope` isn't used at all in our game's code.

image
Rest of the trace leads back here.

### DIFF
--- a/renpy/display/screen.py
+++ b/renpy/display/screen.py
@@ -638,16 +638,8 @@ class ScreenDisplayable(renpy.display.layout.Container):
 
             renpy.ui.close()
 
-        # Avoid suppressing errors during evaluation, we should log what happened and report it
-        except:
-            import traceback
-
-            print("While evaluating screen ", self.screen_name)
-            traceback.print_exc()
-            print()
-
         finally:
-            # Safe removal as to not reraise another exception
+            # Safe removal as to not reraise another exception and lose the last one
             self.scope.pop("_scope", None)
 
             renpy.ui.screen = old_ui_screen
--- a/renpy/display/screen.py
+++ b/renpy/display/screen.py
@@ -1238,7 +1238,7 @@ def predict_screen(_screen_name, *_args, **kwargs):
             print()
 
     finally:
-        del scope["_scope"]
+        scope.pop("_scope", None)
 
     renpy.ui.reset()
 
@@ -1293,7 +1293,7 @@ def use_screen(_screen_name, *_args, **kwargs):
     try:
         screen.function(**scope)
     finally:
-        del scope["_scope"]
+        scope.pop("_scope", None)
 
     _current_screen.old_transfers = old_transfers
 
--- a/renpy/display/screen.py
+++ b/renpy/display/screen.py
@@ -638,8 +638,17 @@ class ScreenDisplayable(renpy.display.layout.Container):
 
             renpy.ui.close()
 
+        # Avoid suppressing errors during evaluation, we should log what happened and report it
+        except:
+            import traceback
+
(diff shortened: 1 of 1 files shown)


renpytom:
I'm fine with the change to .pop. I don't like the except block - I don't think screen exceptions should be entirely hidden like that.

multimokia:
I did that mainly because without it, it's suppressed, ideally we should be logging it somewhere to note what's going on so it can be fixed, plus I took from example where we already do the same:

image

Though I think I'll also change the `del scope["_scope"]` from here as well

multimokia:
Also if I recall, `stdout` is also mirrored to `log.txt`, right?

renpytom:
I guess I don't understand what you're trying to do here, but supressing all screen errors (during real execution) is not the way to fix it. 

I like the use of .pop, but not the traceback.print_exc part.

multimokia:
My point is that it _already_ suppresses it because there's no `except` block.

The issue here is actually never shown to the user or even shows up in logs. The reason the tracebacks above occurred was because of `del self.scope["_scope"]` raising a `KeyError` in the `finally` block. However that exception didn't include the context of the original issue in the `try` block, so it lead to an ambiguous exception where the internal issue is not reported, but also mixes the current execution context, hence the actual game scripts being reported in the traceback, even if they have no mention of `_scope` at all in them.

Moving that to be a `pop` would therefore entirely suppress that error, since the `finally` wouldn't raise it.

As a result of that, I added the exception catch, and print to stdout as a way of logging the real issue instead of completely suppressing and hiding it, and then trying to salvage through `finally`.

multimokia:
I can move it to a `raise` if you'd like? That way we formally report it

multimokia:
My bad, try/finally was not the part which suppresses.

It stems from higher up -- `self.screen.function(**self.scope)`

Basically we only raise the final exception since we never `raise new_exception from old_exception`
So what seems to be happening upon closer inspection is:

1. We run a screen function
2. An error is raised
3. We execute finally
4. Another error is raised

Since we only raise individual errors, the original exception is lost.

Given that, maybe it's best we keep the `except` but also `raise`  after the stdout? That way we pass it back to the upper level caller

renpytom:
Isn't the point of using pop that another error won't be raised?

multimokia:
Fair point, will remove except

Andykl:
@multimokia I do not understand what the problem you think to solve by that.
If screen scope have no reference to itself after screen execute, it's probably something bad with your code or with RenPy. Shading it is not a solution.

renpytom:
I think this is a change worth merging, since the del (vs pop) is probably hiding the real problem. It would be nice to know what the real problem was, though.

Andykl:
Aha, now I see. To get such traceback some can do
```
init python:
    class Buggy(Action):
        def __init__(self):
            scope = renpy.current_screen().scope
            del scope["_scope"]
            raise Exception("I'm here")

screen test():
    textbutton "Test" action Buggy()

label start:
    call screen test
    return
```
And yes, original exception is lost. Applying this fix allows the original exception to be seen.

multimokia:
With that issue fixed, I can report it next time I see it so it can be squashed too.

PR #3077, opened 2021-10-05
Report translate nodes with duplicate id clause

Andykl:
Such code will be executed correctly on default language, but chained translations will have corrupted return point.
```
label start:
    "Line 1" id start_00000000
    "Line 1" id start_00000000
    return
```

### DIFF
--- a/renpy/translation/__init__.py
+++ b/renpy/translation/__init__.py
@@ -156,6 +156,13 @@ class ScriptTranslator(object):
             elif type_n is Translate:
 
                 if n.language is None:
+                    if n.identifier in self.default_translates:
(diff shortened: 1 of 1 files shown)


Gouvernathor:
Does this PR cover this case ?
```rpy
label start;
    "Line A" # naturally generated id : start_000
    ...
    "Line B" id start_000
```
If not, it's still a very good PR, but this case should get covered as well in the future.

Also, just my humble opinion, I would rephrase the exception text as `Lines at (location1) and (location2) have/share the same id : (id)`.
Since both lines may not be the same at all, phrasing it as being one single line may induce misinterpretation on the user's side. (even though it's very likely for the two lines to be similar).

Issue #3069, opened 2021-10-03
Repr of Displayable subclasses is ambiguous


Andykl:
Repr of `im.Data("a" * 10, "test.png")` is `<'Data' u'aaaaaaaaaa' u'test.png'>` (this is most confusing as it really prints data field which is really long.
Repr of `im.Composite((300, 600), (0, 0), "girl_body.png")` is `<'Composite' (300, 600) (0, 0) u'girl_body.png'>`
Repr of `ImageReference(Image("test.png"))` is `<<ImageReference <'Image' u'test.png'> at 00000000>`
and so on.

Current implementation is
```py
class Displayable:

    def __unicode__(self):
        return self.__class__.__name__

    def __str__(self):
        return self.__class__.__name__

    def __repr__(self):
        return "<{} at {:x}>".format(str(self), id(self))
```
This is ambiguous with py3 compability, as unicode become str, and subclasses that redefine unicode to change repr should be changed anyway. But also, couple of times I was really confused to met just "Image" in string substitution when I was expecting path, which happened because substitution used im.Image.__unicode__. Get rid of `__unicde__` and `__str__` prevent this.

I propose to review repr of all Displayable subclasses and make them into the common form 
```py
f"<{type(self).__module__}.{type(self)} at 0x{hex(id(self))} {self.repr_info()}>
```
or something like that.

So, default implementation in class Displayable become something like
```py
class Displayable:
    # No __str__ or __unicode__. Subclasses that really need that should define it themselves.

    def repr_info(self):
        return None

    def __repr__(self):
        klass = type(self).__module__ + "." + type(self).__name__

        at = hex(id(self))
        at = "0x" + at.zfill(16 if renpy.exports.bits == 64 else 8)

        repr_info = self.repr_info()
        if repr_info is not None:
            repr_info = " " + repr_info

        return "<{} at {}{}>".format(klass, at, repr_info)
```

renpytom:
I think this is a reasonable thing to do.

Gouvernathor:
About the implementation, I think it would be simpler to use `object.__repr__` to obtain the type name (which will automatically be prepended with the module name in PY3 and will not in PY2) and the address.
Then if `self._repr_info()` is not None, it can be included like
```py
return object.__repr__(self)[:-1] + " " + repr_info + ">"
```
This would avoid the issue of re-implementing something Python already provides, so it would solve two inconsistencies : the missing `object` word before the `at`, and the module name, which is only supposed to be included in PY3.

[...]

PR #3087, opened 2021-10-12
Fix repr for Displayable subclasses

Gouvernathor:
As per #3069.

The template I used in Displayable subclasses is slightly different from @Andykl's.
Andy's : `<MyModule.MySubclass object at 0x00000000 optional_repr_info>`
Mine : `<MyModule.MySubclass object (optional_repr_info) at 0x00000000>`
The parentheses were chosen to separate the aditional info from the class name and other text such as the "at" and "object" words.
I didn't choose brackets or square-brackets to avoid substitution issues (although they're unlikely to happen).
The placement, between the "object" and the "at", should prevent people from interpreting it as constructor arguments, and also placing it after the address seemed weird.
If you prefer Andy's style, it's available in my code, just commented out.

As it were, MultiBox's repr is a special case which overrides the class name, and DragGroup's is a special case of MultiBox which reverts the override. ~~DragGroup now uses `object`'s repr, instead of Displayable's, which is arguable.~~
I think a better version would be to make it relate to a property such as `__classname` (without trailing underscores) in MultiBox so that the behavior will only apply for MultiBox and not its subclasses. It would special-case only the MultiBox class, instead of having a special-case of a special-case. ~~I'm not sure PY2 will allow it, but I'll try to make it work.~~ **Edit :** Implemented, actually it worked in another way.

ImageBase's subclasses had a modified repr. They used to display every argument passed to the constructor, now they don't. I think this change was intended in Andy's issue, and I believe that's a good thing. However, arguments could be made to modify the `_repr_info` of these classes to include some of the now-omitted data.
But they're all `im.` stuff, so I don't think it really matters anyway.

I also removed Character and Displayable's \_\_unicode__ methods, as they only returned a unicode version of the \_\_str__ method, which is already done by python in the background. I also did it to the parser's ParseError, for which no feature was removed (it's still castable to unicode un PY2 and in PY3) but now it works in PY3, thanks to the new \_\_str__ method.
I can revert this, since it doesn't exactly relate to the base repr issue, but I think \_\_unicode__ has nothing to do in a python code which is interpreted not only by versions superior to 2.7, but also by python 3.

### DIFF
--- a/renpy/character.py
+++ b/renpy/character.py
@@ -1060,18 +1060,6 @@ class ADVCharacter(object):
         else:
             images.predict_show(None, image_with_attrs, show=False)
 
-    def __unicode__(self):
-
-        who = self.name
-
-        if self.dynamic:
-            if callable(who):
-                who = who()
-            else:
-                who = renpy.python.py_eval(who)
-
-        return renpy.substitutions.substitute(who)[0]
-
     def __str__(self):
 
         who = self.name
--- a/renpy/common/_developer/inspector.rpym
+++ b/renpy/common/_developer/inspector.rpym
@@ -24,6 +24,22 @@
 style _hyperlink_button is _default
 style _hyperlink_button_text is _hyperlink
 
+init python:
+    def _inspector_repr(d):
+        if type(d) is renpy.display.screen.ScreenDisplayable:
+            clname =  "Screen"
+        elif type(d) is renpy.display.layout.MultiBox:
+            clname =  d._classname()
+        else:
+            clname =  type(d).__name__
+
+        info = d._repr_info
+        if info is None:
+            return clname
+        if callable(info):
+            info = info()
+        return clname + " " + info
+
 screen _inspector:
     zorder 1010
     modal True
@@ -75,7 +91,7 @@ screen _inspector:
 
                 for depth, width, height, d in tree:
 
-                    $ t = "  " * depth + u"\u2022 " + unicode(d)
+                    $ t = "  " * depth + u"\u2022 " + _inspector_repr(d)
                     $ s = __format_style(d.style.parent)
 
 
@@ -118,7 +134,7 @@ screen _style_inspector:
             xfill True
             spacing gui._scale(10)
 
-        $ displayable_name = unicode(d)
+        $ displayable_name = _inspector_repr(d)
         label _("Inspecting Styles of [displayable_name!q]")
 
         $ styles = d.style.inspect()
--- a/renpy/common/_developer/inspector.rpym
+++ b/renpy/common/_developer/inspector.rpym
@@ -201,9 +201,7 @@ init python:
         else:
             clname = type(d).__name__
 
-        info = d._repr_info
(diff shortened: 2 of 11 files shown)


Andykl:
1. I think the part with ParseError should be removed. It needs separate investigation where it is used and I think it's better to do in separate PR.
2. I think for objects with `_repr_info` it is better to also remove `object` part, as it is obvius that this is an object because of parentheses. Otherwise repr of ImageReference become too long `<renpy.display.image.ImageReference object (<renpy.display.im.Image object (u'test.png') at 0x000000000667ea90>) at 0x00000000063b1c10>`. However I personally prefer omit `object` completly, because classes have repr `<class 'type.__qualname__'>` and have not `at` part.
3. `im.Data("0" * 1000, "test.png")` returns `<renpy.display.im.D(u'test.png') ata object at 0x00000000066817d0>`. You need to use rpartition to split `at` part.
4. `Text("some \"text\" and 'text'\n")` returns `<renpy.text.text.Text object (u'some \\"text\\" and \'text\'\\n') at 0x00000000064736d0>`. Here the problem is that we replace quotes and `\n` (the thing that repr do) and then put this modified string to repr again. I think the solution is just remove `replace` thing completly and just return repr of created short string.
5. `renpy.get_screen("quick_menu")` returns `<renpy.display.screen.ScreenDisplayable object (quick_menu) at 0x00000000066df810>`. Here the problem is that screen name is not repred as a string, but as a name, which is wrong. It's `_repr_info` should be wrapped into `repr`.
6. I would like to add `_repr_info` to Transform class, the same way as ImageReference. It is okay to add `None` if Trasnform have not child. This also applyable to ATL Transform, which have its repr which may be updated to `_repr_info`

Gouvernathor:
1. Agreed, removed
2. In that case the parentheses can be mistaken for arguments to give to the constructor to generate a similar object, which is not the case for ScreenDisplayable for example (and many others). Plus it adds a difference with python's base repr, which I find misleading and unnecessary.
3. Agreed, I used rpartition and also checked for " at " instead of "at", to be double-sure.
4. I would agree, but I'd rather have @renpytom's opinion about it.
5. Agreed, done.
6. I'll explore that possibility, and what relevant info could be chosen to be integrated in _repr_info.

Andykl:
And also. I think you may omit `if callable(reprinfo)` as it should be a function if not None anyway.

Gouvernathor:
One underscore means the function is only used locally or internally. As far as I know, two underscores only makes a method inheritance-proof in Python, so I don't know why the other functions of the inspector file have two underscores.

I added the callable test if a class were to have a constant _repr_info... the opposite of what MultiBox does, in a way. But I guess it's unlikely to be needed.

Also (unrelated), with my 568779f, if _repr_info is something like `(onething) (oneotherthing)`, the repr will end up having two separate sets of parentheses instead of one, because the _repr_info will not be enclosed in supplementary parentheses. I don't think it's a bad thing, but if you do then that commit can be reverted or tweaked more (although it would become quite complicated).

PR #3094, opened 2021-10-17
Add Live2D dummy

### LINKED ISSUE #3093: Make it possible to init Live2D without gl2 or with broken libs
Live2D is the only thing which requires config.gl2 and make an error if it is not, when other gl2 things like matrixcolor, matrixtransform or camera silently ignore it and do nothing without gl2. The idea is to return a dummy thing, which could return `Text("Live2D is broken")` in its render, in case `live2d.init` results in error for some reason (which could be broken lib).
Right now to workaround it I have to do:
```py
init 500:
    if renpy.has_live2d():
        image a = Live2D(...)
        image b = Live2D(...)
    else:
        # Actually there is a CDD which ignore passed _duplicate args
        image a = Text("Live2D is broken")
        image b = Text("Live2D is broken")
```

Gouvernathor:
If the live2d init hasn't gone through successfully, the Live2D constructor returns a placeholder.
Fixes #3093.

It's not a Placeholder instance. Placeholder is created soon enough for Live2D to access it when images are created, but even though I'm setting its name to ("Live2D Dummy", filename), the name to be displayed is always overriden when the Placeholder object is duplicated.
Subclassing Placeholder does't work, because it's not created soon enough.
So, I'm using the Fixed with text used inside the Placeholder class.
If anyone know how to make the placeholder remember its name,..

Other than that, please check this doesn't break normal Live2D behavior. I don't have the Cubism nor examples of live2D structures to play with.

### DIFF
--- a/renpy/gl2/live2d.py
+++ b/renpy/gl2/live2d.py
@@ -38,6 +38,8 @@ import collections
 
 did_onetime_init = False
 
+successful_init = False
+
 
 def onetime_init():
     global did_onetime_init
@@ -60,6 +62,8 @@ def onetime_init():
 
     if not renpy.gl2.live2dmodel.load(dll):
         raise Exception("Could not load Live2D. {} was not found.".format(dll))
+
+    successful_init = True
 
 
 did_init = False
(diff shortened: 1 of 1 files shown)


renpytom:
Using ``__new__`` to change the class of an object on construction is way too magical. This isn't something I could merge.

PR #3139, opened 2021-11-23
Update German translation in common.rpy

kigyo:
I noticed many of the "German" translations were left in English. I tried to fix as many as I could, but a few I was still unsure about. ("Clipboard voicing", in particular.)
As a general note, these older translation files seem to be missing strings from the renderer (shift+G) screen.

### DIFF
--- a/launcher/game/tl/german/common.rpy
+++ b/launcher/game/tl/german/common.rpy
@@ -3,71 +3,71 @@ translate german strings:
 
     # 00action_file.rpy:26
     old "{#weekday}Monday"
-    new "{#weekday}Monday"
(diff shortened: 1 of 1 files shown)


renpytom:
I'll merge this in as is. 

"Clipboard voicing" is a mode where each line is copied to the clipboard, so a user's screen-reader can read it. 

To update all translations, go to preferences in the launcher, open the launcher project, and then generate a new set of translations - https://www.renpy.org/doc/html/translating_renpy.html talks about how to do it. That will add new translations to the bottom of each file. 

Thank you!

PR #3160, opened 2021-12-07
Add text to Placeholder

Gouvernathor:
Allows for Placeholders with constant names, such as the error message ("no live2D") mentioned in #3093.

This extends (and simplify) the Placeholder displayable to use it outside of an image name call.

The implementation is a middle-ground between an outside function calling the Placeholder class, and a supplementary name_override argument to the Placeholder constructor. Either version could arguably be preferred to this one.

### DIFF
--- a/renpy/common/00placeholder.rpy
+++ b/renpy/common/00placeholder.rpy
@@ -32,7 +32,7 @@ init -1500 python:
         def after_setstate(self):
             self.child = None
 
-        def __init__(self, base=None, full=False, flip=None, **properties):
+        def __init__(self, base=None, full=False, flip=None, name=None, **properties):
             """
             `base`
                 The type of image to display. This should be one of:
@@ -63,6 +63,11 @@ init -1500 python:
 
             `flip`
                 If true, the sprite is flipped horizontally.
+
+            `name`
+                The name the placeholder will be given. If provided, no other
+                text than this will be displayed on the placeholder. If not,
+                the text will reflect the show instruction used to display it.
             """
 
 
@@ -74,6 +79,8 @@ init -1500 python:
 
             # A list of name components.
             self.name = [ ]
+            if name:
+                self.name.append(name)
 
             # The child of this placeholder, if known.
             self.child = None
@@ -193,7 +200,10 @@ init -1500 python:
             args = args or self._args
 
             rv = Placeholder(self.base, self.full, self.flip)
-            rv.name = list(args.name) + list(args.args)
+            if self.name:
+                rv.name = self.name
+            else:
+                rv.name = list(args.name) + list(args.args)
             rv._duplicatable = False
 
             return rv
--- a/renpy/common/00placeholder.rpy
+++ b/renpy/common/00placeholder.rpy
@@ -32,7 +32,7 @@ init -1500 python:
         def after_setstate(self):
             self.child = None
 
-        def __init__(self, base=None, full=False, flip=None, name=None, **properties):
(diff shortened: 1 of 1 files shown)


renpytom:
I'd rather this be a parameter to Placeholder, than its own displayable. 

```
Placeholder(name="eileen")
```

is what I'd like.

mal:
I think it's possible to do what you're looking for with a much lighter touch than adding extra class and instance variables, and branches around the `text` construction with something like:

```diff
--- a/renpy/common/00placeholder.rpy
+++ b/renpy/common/00placeholder.rpy
@@ -32,7 +32,7 @@ init -1500 python:
         def after_setstate(self):
             self.child = None
 
-        def __init__(self, base=None, full=False, flip=None, **properties):
+        def __init__(self, base=None, full=False, flip=None, name=None, **properties):
             """
             `base`
                 The type of image to display. This should be one of:
@@ -73,7 +73,7 @@ init -1500 python:
             self.full = full
 
             # A list of name components.
-            self.name = [ ]
+            self.name = [ name ] if name is not None else [ ]
 
             # The child of this placeholder, if known.
             self.child = None
@@ -193,7 +193,7 @@ init -1500 python:
             args = args or self._args
 
             rv = Placeholder(self.base, self.full, self.flip)
-            rv.name = list(args.name) + list(args.args)
+            rv.name = self.name if self.name else list(args.name) + list(args.args)
             rv._duplicatable = False
 
             return rv
```

Gouvernathor:
Thanks ! I added your version, along with the doc addition and and adapting of the guess_base mechanism.
(The adaptation is because formerly the first element of self.name could not contain any space, whereas now it's possible.)

mal:
I'm not sure that `guess_base` really makes sense the context of explicitly setting a name. If a specific placeholder image is desired then that can be set explicitly too: `Placeholder('boy', name='Bill')` or `Placeholder('bg', name='Beautiful Landscape')`, and that also prevents splitting explicitly given names with newlines when using the `boy` and `girl` images.

renpy/common/00placeholder.rpy
```diff
@@ -193,7 +200,10 @@ init -1500 python:
             args = args or self._args
 
             rv = Placeholder(self.base, self.full, self.flip)
-            rv.name = list(args.name) + list(args.args)
+            if self.name:
+                rv.name = self.name
+            else:
+                rv.name = list(args.name) + list(args.args)
```
mal, line 206: Just realised this can be condensed down to:
```suggestion
            rv.name = self.name or list(args.name) + list(args.args)
```

Gouvernathor:
I think it's as relevent as it was before. Whether you use the name argument or not, the Placeholder tries to guess its base from the data it has, unless it's explicitely given one with the `base` argument/attribute.

Gouvernathor:
Actually before this is merged, I'd suggest we consider renaming the parameter to "text", or "content", to make fully understandable that it can be more than just a handful of valid python identifiers, but rather any potentially line-breaking string.

renpytom:
I think 'text' would be fine.

mal:
FWIW, I still think `text` should responsible for text only, and if the creator wants a specific base image they have the `base` argument at their finger tips for exactly that. Trying to guess the base from the text arg still seems out of place given that it could be handed an arbitrary string from pretty anything the creator is interested in - not only a pre-canned static string - but I've more than said my piece, and probably outstayed my welcome on this issue, so I'll leave it at that. 😧

PR #3162, opened 2021-12-08
[Internal]Move DictEquality and FieldEquality from rpy to py

Gouvernathor:
That way it can be used earlier in other py files, such as for Transform and TransformState.

~~WIP~~ Done.
~~The config.debug_equality has been removed, moving to readd it.
The license is missing from the new file.~~


Gouvernathor:
Now, kinda open question : could there be pickling errors if the class descriptor object (DictEquality) still exists in the namespace it was previously defined in, but does not *originate* from the same module ?
Doesn't seem to cause any errors in my game, but that's a question worth asking, since I'm not sure I can do exhaustive tests on this one.

renpytom:
Can you give a better justificatioon than "can be used earlier"? Do you have plan to use this change for something?

Gouvernathor:
This was intended for use in the Transform equality PR, which is now moot.
Dismissed for lack of usability in pure python, due to the RevertableObject being used.

Issue #3114, opened 2021-11-01
[Request] Counting words by characters in the Lint statistics.


Neyunse:
I was wondering if this might be interesting; the idea would be a separate stat count that shows per character total dialog blocks, words and characters.

I don't know how you could get the name; since I didn't look at the Launcher code thoroughly, but I think you could get it from the `Character()`. 

This could be useful for the developer to have an idea of how many words each character has, an example of use would be for voice actors. Each voice actor is charged for a certain amount of words, this could help to give an idea of how much would be the cost for each character.

Example:
```
Statistics:

The game contains 843 dialogue blocks, containing 9,245 words and 51,218
characters, for an average of 11.0 words and 61 characters per block.

The game contains 3 menus, 101 images, and 61 screens.

Characters statistics:

Eileen has: 400 dialogue blocks, containing 500 words

Aki has: 700 dialogue blocks, containing 1,000 words
```

Gouvernathor:
A simple version could be to check what's written before the dialogue line, which would miss `extend` or treat it as one single character. Just putting a name between quotes, or defining a Character object in-place, or having a more complex expression returning a str or Character object, would be difficult if not impossible to scan properly.

A more complex version would require evaluating each of those expressions, which would obviously be complex, if not impossible in cases such as when a variable implicated in those expressions is not defined with a `define` nor `default` statement.

An arguable compromise could be to only consider stats about `define`d characters, plus the narrator, and ignoring all other dialogue lines.

Neyunse:
@Gouvernathor It would be nice to add it using the simple form, so that it is not so complicated, as I said above, this feature would make it easier for the devs to have an approximate of how many words per character they have and use it as a reference for the voice actors.

PR #3170, opened 2021-12-10
Add character stats in lint

Gouvernathor:
Adresses #3114

### DIFF
--- a/renpy/lint.py
+++ b/renpy/lint.py
@@ -916,14 +916,6 @@ characters per block. """.format(
 
         lines.append(s)
 
-        if language is None:
-            sl = ["Character stats for default language :"]
-
-            for cha, coun in charastats.items():
-                sl.append("    {0} : {1}".format(cha, humanize(coun)))
-
-            lines.extend(sl)
-
     print("")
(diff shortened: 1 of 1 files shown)


Gouvernathor:
At the moment, there is some line breaks between the characters in the end file, which seem unnecessary.
For simplicity reasons, the stats are only computed for the None language.

Gouvernathor:
The characters are also ranked by order of frequentness.

renpy/lint.py
```diff
@@ -924,9 +928,17 @@ def report_language(language):
     lines.append("The game contains {0} menus, {1} images, and {2} screens.".format(
         humanize(menu_count), humanize(image_count), humanize(screen_count)))
 
+    sl = ["Character stats for default language :"]
+    for cha, coun in sorted(charastats.items(), key=lambda a:-a[1]):
+        sl.append("    {0} : {1}".format(cha, humanize(coun)))
+    lines.append(sl)
```
Gouvernathor, line 934: This is not a mistake, I do mean append and not extend.

PR #3172, opened 2021-12-11
Raise shadowing errors related to substores and the default statement

Gouvernathor:
When reading `default a.b.c = value`, renpy will check for :
- a.b.c being a store (added in 2d2649666912c06531940e0f8e9797804b9094f0)
- a.b.c being already defaulted (case already covered)
- ~~a.b being already defaulted (added in 966016306d364949bd84d1db0953bee781191618)~~
- ~~a being already defaulted (added in 966016306d364949bd84d1db0953bee781191618)~~
(Actually these last two are not necessary since `default a.b` will have already triggered case 1)

If something else should be checked, please let me know (I'll look into define-default concurrence and lint later).


mal:
Do you have sample code that triggers the exception inside the while loop? As far as I can tell it's not possible to end up in that state.

renpy/ast.py
```diff
@@ -2521,19 +2521,23 @@ def set_default(self, start):
...
-        if self.varname not in defaults_set:
+        if self.varname in defaults_set:
+            if start and renpy.config.developer:
+                raise Exception("{}.{} is being given a default a second time.".format(self.store, self.varname))
+            return
 
-            if start or (self.varname not in d.ever_been_changed):
-                d[self.varname] = renpy.python.py_eval_bytecode(self.code.bytecode)
+        # do the variable shadowing if not in this case, for compatibility reasons
+        if start and (renpy.config.developer is True):
```
mal, line 2530:
```suggestion
        if start and renpy.config.developer:
```

renpy/ast.py
```diff
@@ -2521,19 +2521,23 @@ def set_default(self, start):
...
+            fullname = '.'.join((self.store, self.varname))
+            if fullname in renpy.python.store_dicts.keys():
```
mal, line 2532:
```suggestion
            if fullname in renpy.python.store_dicts:
```

renpy/ast.py
```diff
@@ -2521,19 +2521,23 @@ def set_default(self, start):
...
+                raise Exception("The {} default variable is shadowing a store with the same name.".format(fullname))
```
mal, line 2533:
```suggestion
                raise Exception("{} is being given a default but is already a store.".format(fullname))
```
Some creators won't understand what shadowing is in this context, so copying the style of the existing exception and explaining that it's already a store should hopefully be more immediately understandable to them.

renpy/ast.py
Gouvernathor: This was a request from @renpytom

PR #3176, opened 2021-12-11
Some pylint and py3 preparation

Gouvernathor:
See commit comments for more specific details.

The json.load matter may be more problematic than the others, since the `encoding` (kw)arg is not explicitely documented anymore.


Gouvernathor:
Other than low-importance code simplification and occasional bug sweeping, two major finds :
The json.load matter, which would deserve to be checked (by anyone knowing anything about json), and the imp module matter, which was checked to effectively return the same value as the imp stuff that was used.

PR #3177, opened 2021-12-12
Adress submoduling persistent and special namespaces

Gouvernathor:


Gouvernathor:
My tries to implement the submoduling feature to persistent are more than shaky, due to the fact that submodules are created before the persistent object exists.
So I would advocate merging this as-is, maybe with a compat or a config.developer try/except, before moving forward with any persistent manipulation.

PR #3189, opened 2021-12-20
Remove bare excepts from the repository

Gouvernathor:
This first version removes them all to replace them with `except Exception:`, as per [the PEP8](https://www.python.org/dev/peps/pep-0008/#id49)'s recommendation (see points n°6 and 8 from that section).
The goal is not to intercept BaseException instances, such as SystemExit and KeyboardInterrupt
Some refining and review would be good. I did some and will continue.

In two instances it was not as needed as otherwise :
- in the cds page of the doc, I chose to replace it in a code snippet as well, to avoid showcasing bad practices.
- in the commented-out section of the placeholder code (which should probably be removed ? since this is git), there was another instance


Gouvernathor:
I found one typo and no instance where KeyboardInterrupt or SystemExit would be legitimately caught.
Given that the only REPL (renpy's console) doesn't support Ctrl+C, I think this is normal.

PR #3191, opened 2021-12-22
Add config.default_attributes callback dictionary

mal:
This is to support images having a "weak" set of default attributes
applied to them when they are first shown, allowing for more complex use
cases with a minimum of boilerplate in scripts.

While similar in nature to the existing `config.adjust_attributes`, this
feature only applies to images when they are first shown, and is "weak"
in the sense that it may not countermand attributes from the script.

### DIFF
--- a/renpy/config.py
+++ b/renpy/config.py
@@ -1149,7 +1149,7 @@ adjust_attributes = { }
 
 # A dictionary from a tag to a function that produces default attributes
 # for that tag.
-default_attributes = {}
+default_attribute_callbacks = { }
 
 # The compatibility mode for who/what substitutions.
 # 0: ver < 7.4
--- a/renpy/config.py
+++ b/renpy/config.py
@@ -1147,6 +1147,10 @@ gl_lod_bias = -.5
 # of that tag.
 adjust_attributes = { }
 
+# A dictionary from a tag to a function that produces default attributes
+# for that tag.
+default_attributes = {}
+
 # The compatibility mode for who/what substitutions.
 # 0: ver < 7.4
 # 1: 7.4 <= ver <= 7.4.4
--- a/renpy/display/image.py
+++ b/renpy/display/image.py
@@ -921,8 +921,21 @@ class ShownImageInfo(renpy.object.Object):
 
         nametag = name[0]
 
-        # The set of attributes a matching image may have.
-        optional = list(wanted) + list(self.attributes.get((layer, tag), [ ]))
+        # Start building the set of attributes a matching image may have.
+        optional = list(wanted)
+
+        # Find any attributes applied previously.
+        defaults = self.attributes.get((layer, tag), None)
+
+        # If no record, it's the first show, so try to fetch defaults.
+        if defaults is None:
+            f = renpy.config.default_attributes.get(name[0], None)
+            if f is not None:
+                defaults = f(name)
+
+        # Add any defaults to the set of attributes a matching image may have.
+        if defaults:
+            optional += list(defaults)
 
         # The list of attributes a matching image must have.
         required = [ ]
--- a/renpy/display/image.py
+++ b/renpy/display/image.py
@@ -929,7 +929,8 @@ class ShownImageInfo(renpy.object.Object):
 
         # If no record, it's the first show, so try to fetch defaults.
         if defaults is None:
-            f = renpy.config.default_attributes.get(name[0], None)
+            f = renpy.config.default_attribute_callbacks.get(name[0], None) \
+                or renpy.config.default_attribute_callbacks.get(None, None)
             if f is not None:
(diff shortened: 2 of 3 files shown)


Gouvernathor:
I feel like this should be covered by checking the None-ness of `renpy.get_attributes(tag)` (or even just `renpy.showing(tag)`) within the adjust_attributes function. It doesn't feel necessary to add a dedicated function for this.
Does this count as "relying on the image's state" ? Maybe.

What's more, I think this is a bad thing to include in renpy, because displaying the image with no attributes (the default behavior) becomes shadowed and unreachable. In other words, there's no opt-out.

A compromise (if the recipe I gave earlier doesn't work/interferes with prediction because of relying on the image's state), a new keyword-argument could be passed to the adjust_attributes functions to specify whether this is the first time the image is shown. Along with a compat config variable to ensure backwards-compatibility, of course.

mal:
Funnily enough, both of your suggestions (using `renpy.showing`, and the keyword argument to `adjust_attributes`) mirror those that I approached Tom with initially when considering this feature! 😁

Ultimately the conclusion was that changing the signature of `adjust_attributes` while technically possible wasn't a particularly appealing proposition, or as it turns out the most ideal solution. It also transpired that calling `showing` from within the callback wasn't feasible without this either, as it is `name[0]` that gets passed to the callback and so use of `as xyz` in the script would render it unusable, and was further compounded because in order for the check to be reliable the layer would also need to be passed - both of which would only serve as a temptation to rely on further state.

Using a new callback to facilitate fetching attributes as optional when no others already existed, and which won't countermand those expressed by the script was presented as a good alternative and where the absence of the tag in self.attributes provides the first shown condition more or less "for free".

The feature is an explicit opt-in, if a creator does not configure any callbacks then there is no change in behaviour. Being able to change the `show myimage` behaviour to automatically apply additional attributes is the feature, so I'm not quite sure why you seem to view it as a breaking change.

Your point about hidden attributes seems correct, however I think it reasonable that this callback not be activated in that scenario as the creator is already doing their own attribute management via other means, and interceding on their behalf in such a case seems like it would be unwarranted.

Gouvernathor:
I don't mean that it's a breaking change, I mean that if you enable this default-attributing for a tag, you can't decide to show this tag without the default attributes going with it.
However, you could answer that the adjust_attribute way of doing it (were it working) would suffer from the same issue, which is right. And in fact, both could be inhibited by some exterior variable being toggled on or off, since both are functions and not some predetermined frozenset of attributes. So, legit.
However I don't really follow the part about `name[0]` and `as xyz`. Both this and adjust_attributes are based upon the tag being shown, unless I misunderstood, so how does this default_attributes way solve the issue ? And... how is it an issue actually ?

I still feel like this is a too-particular callback to add, and that the attributes-managing callbacks are already a bit messy (the adjust_attributes and the attributes_function from layeredimage), but Tom's royal sanction is all you need.

mal:
Tag in this context pulls two different duties. In the majority of cases (where `as` is not being used) including at define time, `tag` is referring to the first part of the image name. This makes sense as the callbacks are linked to image definitions. However when using the `as` feature the two diverge internally with the value given to `as` serving as the internal identifier allowing creators to show the same asset twice under different names, and with different attributes. The `showing` function is checking against this internal list of tags, but the callbacks are using the name because they're linked to the image definition, not it's usage.

<details><summary>Probably best shown with an example:</summary>

```renpy
image abc = Solid('f77', xysize=(.3, .3))


init python:
    adjusted = None

    def adjust(name):
        global adjusted
        adjusted = name
        return name

    config.adjust_attributes['abc'] = adjust


label start:
    show abc as xyz
    $ abc_showing = renpy.showing('abc', layer='master')
    $ xyz_showing = renpy.showing('xyz', layer='master')
    '''
    showing abc: [abc_showing]\n
    showing xyz: [xyz_showing]\n
    adjusted: [adjusted]
    '''
```
</details>

As you can see the `adjust_attributes` callback gets the name, of which the first part is often, but crucially not always, the same as the internal tag, and also lacks the layer information required to make an accurate check.

Gouvernathor:
So, you're saying that if I have an "eileen happy" and an "eileen sad" defined images, if I show one then the other then either of them `as badeileen`, your function would trigger in all three cases, even though the adjust_attributes one would trigger only on the two former ones ? I didn't understand it that way.

If so, then this is another reason not to complicate the attributes callbacks any more than they already are. Having the two functions trigger under different conditions is a negative mark to me.

mal:
No, that's not at all what I'm saying. The crux of the matter is that there is insufficient data to be able to provide this functionality in `adjust_attributes`, there's no appetite for changing the signature of that callback, and the use case is better served by this new callback which not only has better visibility on the state of existing attributes, but can apply them weakly.

PR #3194, opened 2021-12-23
Hide current screen

via-my:
What about adding an easier way to hide current screen? This is really efficient for projects that use a lot of them.

### DIFF
--- a/renpy/common/00action_control.rpy
+++ b/renpy/common/00action_control.rpy
@@ -198,6 +198,7 @@ init -1500 python:
 
         `_layer`
             This is passed as the layer argument to :func:`renpy.hide_screen`.
+            Ignored if `screen` is None.
         """
 
         _layer = None
@@ -214,10 +215,10 @@ init -1500 python:
                 if cs is None:
                     return
 
-                self.screen = cs.screen_name
-                self._layer = cs.layer
+                renpy.hide_screen(cs.screen_name, layer=cs.layer)
 
-            renpy.hide_screen(self.screen, layer=self._layer)
+            else:
+                renpy.hide_screen(self.screen, layer=self._layer)
 
             if self.transition is not None:
                 renpy.transition(self.transition)
--- a/renpy/common/00action_control.rpy
+++ b/renpy/common/00action_control.rpy
@@ -191,6 +191,7 @@ init -1500 python:
         :doc: control_action
 
         This causes the screen named `screen` to be hidden, if it is shown.
+        Otherwise if `screen` is None, the current screen is used instead.
 
         `transition`
             If not None, a transition that occurs when hiding the screen.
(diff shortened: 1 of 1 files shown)


renpytom:
I think I like the idea of it, not 100% sold on the name, as self is used a lot in Python. Perhaps HideCurrent() ?

Gouvernathor:
What does this do when screen A `use`s screen B, thelatter including a button with this action on it ? Does it rightfully hide screen A ?

Also, about the name, how about making this the behavior of the Hide action when no arguments is passed to it ? Which it doesn't currently support.

renpytom:
It would hide screen A, which is the right behavior. Hide with no arguments could work as well. @via-my, what do you think about that option?

via-my:
Sounds much better :)
Meh, it looks a little bit messy... Sorry.

renpy/common/00action_control.rpy
```diff
@@ -201,16 +202,24 @@ init -1500 python:
...
             self._layer = _layer
 
         def __call__(self):
+            if self.screen is None:
+                cs = renpy.current_screen()
+
+                if cs is None:
+                    return
+
+                self.screen = cs.screen_name
```
mal, line 217: In order to work with screens that declare a preferred `layer`, and also those called with `_layer=`, I think we also need to set the layer from the current screen before calling `hide_screen`.
```suggestion
                self.screen = cs.screen_name
                self._layer = cs.layer
```

via-my: Well spotted!

Andykl:
I do not like the idea that `__call__` mutates the object itself.
I can have such code
```
define hide_current_screen = [Hide(), With(Dissolve(0.5), maybe some other actions]

screen some1():
    timer 1.0 action hide_current_screen

screen some2():
    timer 1.0 action hide_current_screen
```

And your code after first use with screen some1 will not work with screen some2.

The solution is easy, just compute the screen and layer as local variables instead of setting self.screen in `__init__`.

via-my:
I guess this should work now. Thank you!

renpy/common/00action_control.rpy
```diff
@@ -191,26 +191,37 @@ init -1500 python:
...
+
+            else:
+                renpy.hide_screen(self.screen, layer=self._layer)
 
             if self.transition is not None:
                 renpy.transition(self.transition)
 
             renpy.restart_interaction()
-
+
```
Andykl, line 227: Spurious spaces

Andykl (approved):
Nit thing. Except for that LGTM.

renpy/common/00action_control.rpy
```diff
@@ -191,26 +191,37 @@ init -1500 python:
...
-        def __init__(self, screen, transition=None, _layer=None):
+        def __init__(self, screen=None, transition=None, _layer=None):
             self.screen = screen
             self.transition = transition
             self._layer = _layer
 
         def __call__(self):
-            renpy.hide_screen(self.screen, layer=self._layer)
+            if self.screen is None:
+                cs = renpy.current_screen()
```
Gouvernathor, line 213: This function does not look documented. If I have two (independant) screens being displayed, A and B, how is it supposed to tell the two apart ?

via-my: Seems that renpy.current_screen() only works in the context of screen, so it shouldn't be misleading.

Gouvernathor: Ok, I tested for myself and I don't know how, but it works.

Gouvernathor (approved):

renpytom:
Merged, thanks. I made minor fixes in 90d5ee811.

PR #3196, opened 2021-12-23
Delete temporary renaming file if it exists

Andykl:
After we added the use of save rename function, we have started to get errors referring to WindowsError: [Error 183] Cannot create a file when that file already exists. I don't know why it happens only for some of them, but this should fix the problem.

### DIFF
--- a/renpy/savelocation.py
+++ b/renpy/savelocation.py
@@ -287,6 +287,9 @@ class FileLocation(object):
             if not os.path.exists(old):
                 return
 
+            if os.path.exists(old + ".tmp"):
+                os.unlink(old + ".tmp")
+
             os.rename(old, old + ".tmp")
             old = old + ".tmp"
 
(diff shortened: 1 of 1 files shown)


renpytom:
I'd suggest doing it in an exception handler (probably for OSError), so that it's atomic when it doesn't fail.

renpy/savelocation.py
```diff
@@ -287,13 +287,9 @@ def rename(self, old, new):
...
                 return
 
-            os.rename(old, old + ".tmp")
-            old = old + ".tmp"
-
-            if os.path.exists(new):
-                os.unlink(new)
-
-            os.rename(old, new)
+            old_tmp = old + tmp
```
Gouvernathor, line 290: You probably mean `".tmp"` instead of `tmp`

Andykl: It is the same as below with persistent

PR #3201, opened 2021-12-27
Move overfull check in the Grid's add method

### LINKED ISSUE #2795: Traceback could contain line number for Grid Overfull in the script file that has it
Sometimes, I get an exception for Grid overfull. I will admit this is my fault and something I need to fix in my code. However (for example), I am using Grid in a screen that will show multiple character sheets. I did this so I can have a row of information nicely aligned. :)  Unfortunately, I got a grid overfull after entering something, not in a grid (as far as I am aware), but the exception tells me that testScreen() with displayCharacterSheets() had this exception (with line numbers) but will not tell me the line number in the screen for the Grid that overfull. Having this information would help determine which one is throwing the exception so it can be corrected. :)

Gouvernathor:
fix #2795

Having the line number for the *grid* is "simple" (don't quote me on this), having the line number for the *element* making it overfull is not.
I guess that by writing the SL parser differently, it could report the line of the extraneous element by itself, but I wouldn't recommand going that way just for more accurate error reports.

### DIFF
--- a/renpy/display/layout.py
+++ b/renpy/display/layout.py
@@ -535,8 +535,8 @@ class Grid(Container):
 
         return rv
 
-    def per_interact(self):
-        super(Grid, self).per_interact()
+    def add(self, d):
+        super(Grid, self).add(d)
 
         delta = (self.cols * self.rows) - len(self.children)
         if delta == 0:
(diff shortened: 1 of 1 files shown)


renpy/display/layout.py
```diff
@@ -545,16 +545,20 @@ def per_interact(self):
         elif delta < 0:
             raise Exception("Grid overfull.")
```
Andykl, line 546:
```py
if len(self.children) > (self.cols * self.rows):
    raise Exception("Grid overfull.")
```

Gouvernathor:
I have another version more similar to #3203 available if and when Grid allows None as rows or cols.

PR #3202, opened 2021-12-27
Allow cols or rows to be `None` in `Grid` 

KiloOscarSix:
Changes:
 - Allow dynamic `cols` or `rows` definition based on number of children
 - Changed `allow_underfull` to `True` by default to inhibit behavior similar to vpgrid

Errors:
 - Parsing the script expects the exact number of positional arguments

TODO:
 - If approved update docs
 - Test if transpose works correctly with `allow_underfull` grid?

### DIFF
--- a/sphinx/source/cds.rst
+++ b/sphinx/source/cds.rst
@@ -368,7 +368,7 @@ A custom statement's parse function takes an instance of a Lexer object.
                 ll = l.subblock_lexer()
 
                 while ll.advance():
-                    with ll.catch_errors():
+                    with ll.catch_error():
                         strings.append(ll.require(ll.string))
                         ll.expect_noblock("string inside mystatement")
                         ll.expect_eol()


renpytom:
This seems to reformat unrelated code, doing things like changing single quotes to double quotes. Please fix that and resubmit the changes, as it's not reallly possible to see what's being changed. Please do not run black on Ren'Py.

KiloOscarSix:
Certainly, will update

Gouvernathor:
Grid.allow_underfull cannot be changed due to compatibility with before that class attribute was added. If you want to change the default, you need to change the value of the config variable (because it will trigger when the instance has not received a specific value) and compat it. However I doubt Tom would agree with that.

As far as I understand pickling, the @property stuff won't work - because the saved instances' values will be summarily replaced with property objects, and the _rows and _cols woll be None (or even worse, non-existent).
You could try checking the None-ness of the attribute at the last moment (so in the render method I'd guess) and add them as None class-attributes (for default and compat).

Andykl:
> As far as I understand pickling, the `@property` stuff won't work - because the saved instances' values will be summarily replaced with property objects, and the _rows and _cols woll be None (or even worse, non-existent).
> You could try checking the None-ness of the attribute at the last moment (so in the render and per_interact methods) and add them as None class-attributes (for default and compat).

This is what `__version__` and `after_upgrade` stard for.

renpytom:
I could see perhaps going with something like this change, and perhaps only triggering an error if the grid is incomplete.

The idea would be that if rows and cols are defined, we check for a specific number of elements. If rows or cols are defined, we'd have to check that the number of children are a multiple of the rows and columns. 

Backwards compatibility is reasonably important.

renpy/config.py
```diff
@@ -460,7 +460,7 @@
 debug_text_overflow = False
 
 # Should underfull grids raise an exception?
-allow_underfull_grids = False
+allow_underfull_grids = True
```
Gouvernathor, line 463: You will need to compat this in renpy/common/00compat.rpy

renpytom:
I don't think we want to change allow_underfull_grids. What we want to do is to change our definition of what an underfull grid is.

KiloOscarSix:
> Backwards compatibility is reasonably important.

Absolutely, the change from positional args to keyword args is the biggest concern in regards to compatibility

renpytom:
Yeah, we can't accept this until there's a better solution for that change - probably something like changing sl2 to have optional positional arguments.

Gouvernathor:
The `store.Grid` will also have to be updated, as it is actually not the same object as `renpy.display.layout.Grid`.
It's located in renpy/defaultstore.py:294, and its `nargs=2` is a problem (and removing it is not enough, for the same reason as for the screen language issue).

renpytom:
I'd suggest we start off by allowing either rows or cols (but not both) to be None. I'm not sure optional positional arguments are possible - I need to think about if it's possible to make a parser that can deal with them.

Gouvernathor:
I'm not sure it's safe to have the per_interact and render methods mutate the object, or at least not that way.
It could be better to have your function return a computed number of rows and cols, like a property would, without changing the state of the object's attribute (without writing to self.rows or self.cols).

KiloOscarSix:
> I'm not sure it's safe to have the per_interact and render methods mutate the object, or at least not that way. It could be better to have your function return a computed number of rows and cols, like a property would, without changing the state of the object's attribute (without writing to self.rows or self.cols).

aye, planning on doing so, however suddenly got rather late here so bundled it in a function, will continue to work on the idea tomorrow

PR #3214, opened 2022-01-04
Add px and coordinate, and central computing of positions

Gouvernathor:
As introduced in #2827 ; replaces #2832.

This implementation uses namedtuple because it's light-weight and offers readonly security for no visible drawbacks.
The code was also made clearer.

The `absolute.compute` function seeks to regroup all coordinate calculations in one central place, to prevent code duplication as well as to provide a testing ground for the coordinate class.
Returning an absolute number of pixels is the very purpose of this function, that's why I stored it in the `absolute` class and made it return an `absolute` instance, to remove uncertainty as to the measuring unit of the value.
However, since creating and storing `absolute` instances is more expensive than manipulating native types, I added the short-circuit `absolute.compute_native` function, which returns the same value but performs no conversion whatsoever. That one is to be favored in contexts where the type of the value may not lead to confusion, or where best performance is sought.

# Caveat
Addition and substraction aren't supported between coordinates, so [atl interpolation](https://github.com/renpy/renpy/blob/faf684495468871e987db2d6f3ac3e4eae7919be/renpy/atl.py#L136) will interpolate coordinates in a tuple-like fashion. That means interpolating the absolute value and the relative value separately, and uniting them back in a pure tuple. So either we special-case the coordinate type and don't interpolate it tuple-like, or we replace the `tuple` in [this line](https://github.com/renpy/renpy/blob/faf684495468871e987db2d6f3ac3e4eae7919be/renpy/atl.py#L143) with `type(b)`, and **test that it works**.
Another solution must be found for interpolation between a coordinate and a non-tuple instance, including in [the interpolate_spline function](https://github.com/renpy/renpy/blob/faf684495468871e987db2d6f3ac3e4eae7919be/renpy/atl.py#L163). This is for interpolating between different types of positions, which was officially never supported, but 1) I think it should, and 2) it never raised exceptions, and only ever gave weird behaviors.


Gouvernathor:
Testing coordinate, px and absolute.compute can be done with the following code (copy/paste it in a function and call it from the console, if it returns izgud) :
```py
## testing for internal errors
coordinate(offset=1.)
coordinate(offset=1., relative=.5)
10*px
10*px+.7

## value checking
if not (absolute.compute(10, 500) == 10.):
    raise Exception("absolute.compute failing for integer value")
if not (absolute.compute(.25, 500) == 125.):
    raise Exception("absolute.compute failing for float value")
if not (absolute.compute(absolute(6.25), 500) == 6.25):
    raise Exception("absolute.compute failing for absolute value")

# disabled because the offset argument is not optional
# if not (absolute.compute(coordinate(relative=.25), 500) == 125.):
#     raise Exception("absolute.compute failing for relative coordinate")
if not (absolute.compute(coordinate(offset=6.25), 500) == 6.25):
    raise Exception("absolute.compute failing for offset coordinate")
if not (absolute.compute(coordinate(offset=4.25, relative=.5), 500) == 254.25):
    raise Exception("absolute.compute failing for relative+offset coordinate")
if not (absolute.compute(coordinate(offset=4.25)+coordinate(offset=-50), 500) == -45.75):
    raise Exception("absolute.compute failing for sum of offset coordinates")

if not (absolute.compute(10*px, 500) == 10.):
    raise Exception("absolute.compute failing for literal offset coordinate")
if not (absolute.compute(10*px+.5, 500) == 260.):
    raise Exception("absolute.compute failing for literal relative+offset coordinate")
if not (absolute.compute(10*px+absolute(20)*px, 500) == 30.):
    raise Exception("absolute.compute failing for sum of literal offset coordinates")
if not (absolute.compute(10*px+absolute(20)*px+.5, 500) == 280.):
    raise Exception("absolute.compute failing for sum of literal offset coordinates with a relative value")
# disabled because summing coordinates with relative values is disabled
# if not (absolute.compute(10*px+.5+absolute(20)*px, 500) == 280.):
#     raise Exception("absolute.compute failing for sum of literal coordinates")

## testing for errors which should be there
try:
    1.*px
except Exception: pass
else:
    raise Exception("Pure-float multiplication should be disabled")

try:
    (10*px+.5)*10
except Exception: pass
else:
    raise Exception("Multiplication on instances with relative values should be disabled")

try:
    10*px+10
except Exception: pass
else:
    raise Exception("Adding integers should be disabled")

try:
    10*px+absolute(10)
except Exception: pass
else:
    raise Exception("Adding absolutes should be disabled")
```

Gouvernathor:
I searched the repo for uses of absolute and float type-checks, looking to replace all occurrences of various position conversions with either compute or compute_native.
Among them, I spotted a number of doubtful cases (which are flagged in 67b1c3d and 59278c497257e3f4b42ae494a2d616e16721cc0b) which would probably deserve a more specific examination.
Some examples :
- [layout.py line 1216](https://github.com/renpy/renpy/blob/f8869d098187be65c68bfcca5106c11fb7fb0824/renpy/display/layout.py#L1216) does something weird : after a strict type-check, instead of multiplying the value with the available room, it *replaces* it with it...?
- [render.pyx line 228](https://github.com/renpy/renpy/blob/f8869d098187be65c68bfcca5106c11fb7fb0824/renpy/display/render.pyx#L227), [live2D.py line 929](https://github.com/renpy/renpy/blob/f8869d098187be65c68bfcca5106c11fb7fb0824/renpy/gl2/live2d.py#L929), [testfocus.py line 79](https://github.com/renpy/renpy/blob/f8869d098187be65c68bfcca5106c11fb7fb0824/renpy/test/testfocus.py#L79) and multiple others don't take the `absolute` case into account, they use `isinstance` instead of a strict type-check. In other cases (such as [here](https://github.com/renpy/renpy/blob/f8869d098187be65c68bfcca5106c11fb7fb0824/renpy/ui.py#L980)) it's obvious the context has nothing to do with the matter at hand, but in the flagged cases it's not so clear.

Gouvernathor:
Considering that any operation involving an `absolute` returns a pure float, and nobody complained, I preferred disabling non-necessary operations with `coordinate`, namely addition between coordinates and left-multiplication.
The only allowed operations are as follows (see the test file for use cases) :
- multiplication between an int or absolute and a coordinate, in that specific order
- addition or substraction between a coordinate and a non-absolute float, in any order
- equality between coordinates

Comparisons between instances and length were left as tuple implements them because I feel it's not necessary to disable, but if necessary it can be done with `__gt__ = __ge__ = __lt__ = __le__ = __len__ = None` in the class body.

renpy/display/render.pyx
```diff
@@ -223,6 +223,7 @@ cpdef render(d, object widtho, object heighto, double st, double at):
     xmaximum = style.xmaximum
     ymaximum = style.ymaximum
 
+    # TODO: check this regarding the new absolute.compute
```
Gouvernathor, line 226: xmaximum and ymaximum officially only support integers, but this piece of code which seems to be used by all displayables proves otherwise. If there is no incompatibility with another part of the code, I'd advocate upgrading the style-properties doc (which by the way says offset support all position types, and I think it's not actually the case), and implement the new system here.

renpy/gl2/live2d.py
```diff
@@ -928,6 +928,7 @@ def render(self, width, height, st, at):
 
         zoom = self.zoom
 
+        # TODO : check this regarding the new absolute.compute
```
Gouvernathor, line 931: The doc officially only supports floats. I don't know why integers seem technically supported. Upgrading to all/any position type is left as an exercise to the reader (who hopefully knows anything about live2D).

renpy/test/testfocus.py
```diff
@@ -76,6 +76,7 @@ def match(f):
 
 def relative_position(x, posx, width):
     if posx is not None:
+        # TODO : check this regarding the new absolute.compute
```
Gouvernathor, line 79: This has little to do with the rest, as it's not about transform or style properties. I'd advocate converting it to the new system, but it's really a separate issue.

renpy/common/00splines.rpy
```diff
@@ -62,6 +62,7 @@ init -1500 python:
             self.initialized = None
 
         def init_values(self, sizes):
+            # TODO : check this regarding the new absolute.compute
```
Gouvernathor, line 65: This seems to be used in the part where ATL warps between three or more key positions. I think it should definitely support all types of position, and that it's current lack of support for absolutes is a bug in and of itself.
While we're at it, it could be good to support interpolating between positions of different types, as it is not currently supported between only two values, and this code proves it can be done.

renpytom:
Delaying this, just to lower the risk of 8.0 a bit. (Per discussion on discord.)

renpytom:
Closing this, in favor of the upcoming position type.

Issue #3197, opened 2021-12-23
config.name with a colon breaks on iOS


rdurfee:
If the game's config.name contains a colon character (":"), the game will not start on iOS. The Xcode project builds fine, but when run it produces an error:

`
ImportError: "No module named site"
`

renpytom:
Can you post any more of the error? I don't see an obvious way it could be related, but iOS is weird. You can probably get around this by setting build.directory_name to something different, without the :.

rdurfee:
The game starts, showing the launch screen, then it quits. It's not actually an in-game error, but an output message, displayed in the All or Package Output window.

Setting the directory_name like this has no effect.
```
define config.name = _("Test: The Sequel")
define build.directory_name = "TestTheSequel"
```

rdurfee:
I don't know if it's related, but when it creates the xcodeproject file, it names it with a / character instead of a colon....

`Test/ The Sequel.xcodeproj`

renpytom:
Fixed by renpy/renpy-build@df89ba03abea4466c9e286c5df6e2db870301224 .

PR #3220, opened 2022-01-07
renpy/__init__: make site import optional

uyjulian:


renpytom:
This introduces a bug if site won't import - official won't be set. But site's a standard module, so why wouldn't it be importable?

Gouvernathor:
In #3197 the `site` module fails to import. But if that is the issue this PR is trying to solve, I believe it's hiding a symptom instead of solving the actual problem.

PR #3245, opened 2022-01-21
Readonly RevertableObject with `__slots__`

Gouvernathor:
Makes `object().arbitrary = value` as illegal as for the actual python object, by perpetuating its empty `__slots__`.

This would not have yielded any consequence in the code if object() had been used as the native object in the code, unfortunately it was not.
I changed some code using an instance of object as a namespace, something which is illegal in regular python, and made it use namespaces from `init python in ...`. This, in my opinion, yields better code.
In one instance it wasn't possible so I used OpenRevertableObject, an empty subclass of RevertableObject which reopens its `__slots__` and is essentially the same as what RevertableObject initially was.
I also changed some code using (revertable)object instances as sentinels, and while it probably wouldn't yield any bug, it was unnecessary to have them use a revertable object instead of a lighter-weight native object.

You may wonder, why do any of this ? It makes pure instances of RevertableObject more lightweight, but it also makes them useless, so what's the point ?
The point is dual.
First, it enjoins junior creators discovering python not to take bad habits regarding object(), by making python's object and renpy's object more alike.
Second, it allows for `__slots__`-defining classes to be declared in renpy and be actually worth it (they currently inherit an empty `__dict__`, defeating the purpose of having lightweight instances). This also applies to subclassing `__slots__`-using types, such as native types or namedtuple.

Compat :
There are two ways to do it. The one I chose (but it could be reversed) is to make pickle use OpenRevertableObject for pickled RevertableObjects in python3, which is simple, but not doable in py2. So I simply disabled the effects of this whole change in py2, by aliasing RevertableObject to OpenRevertableObject.
Another approach (which I haven't tested) would be to rename the old class BaseRevertableObject, have OpenRevertableObject be named RevertableObject, and have `object` alias the Base one instead of the Open one. That way, pickle finds the equivalent type under the original name and doesn't have to be tricked. It would create some level of benign inconsistency between classes explicitely inheriting RevertableObject and classes defined in rpy, inheriting only BaseRevertableObject. It may actually be the best option, but I'm leaving myself some time to sleep on it.

### DIFF
--- a/renpy/common/000namespaces.rpy
+++ b/renpy/common/000namespaces.rpy
@@ -53,7 +53,7 @@
 
         def set_default(self, name, value):
 
-            undefined = object()
+            undefined = python_object()
 
             if not isinstance(persistent._preference_default, dict):
                 persistent._preference_default = { }
--- a/renpy/common/00action_data.rpy
+++ b/renpy/common/00action_data.rpy
@@ -24,7 +24,7 @@ init -1600 python:
    ##########################################################################
     # Functions that set variables or fields.
 
-    __FieldNotFound = object()
+    __FieldNotFound = python_object()
 
     def __get_field(obj, name, kind):
 
--- a/renpy/common/00definitions.rpy
+++ b/renpy/common/00definitions.rpy
@@ -101,9 +101,12 @@ init -1400:
 
 # Transitions ##################################################################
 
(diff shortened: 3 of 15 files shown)


renpytom:
I don't think I want to break something that has been working for years for this reason.

PR #3246, opened 2022-01-22
Subclass functools.partial for better performance

Gouvernathor:
Improves currying performance.
Curry is not widely used in renpy, but it's used for very commonly used things (every transition class, for one, some others, and every time a creator defines a custom transition by "instancing" a transition class).

This uses functools.partial as a base, because it's builtin and yields far better performance. It is subclassed to provide the same features as the current Curry class does, including the clever hash caching.

With this version, I'm computing a x7-x9 timer performance upgrade, and a -25% memory use (arguable, memory computation in python are messy).
Time computation is calculated as follows :
```py
def partial_timer_test():
    from renpy.curry import Curry, Partial
    def donothing(*args, **kwargs):
        pass
    stamp0 = time.time()
    for _ in range(100000):
        Curry(donothing)()
    stamp1 = time.time()
    for _ in range(100000):
        Partial(donothing)()
    stamp2 = time.time()
    return (stamp1-stamp0) / (stamp2-stamp1)
```
**_Merging and compatibility considerations:_**
I only kept the Curry class in for pickling reasons, but if both Partial and Curry are used coincidentally, we'll end up having to keep both no matter what. Some solution should be found to convert Curry instances to Partial instances at unpickling time (which is not possible in py2 pickle, as far as I know). So we may better wait until py2 is definitively not a part of renpy.
Another solution could be to keep both classes side-by-side while renpy7 still exists, then to remove the Curry class when it's discontinued.
Other than that, this works.


renpy/curry.py
```diff
@@ -73,6 +70,42 @@ def __hash__(self):
...
 
 
+class Partial(functools.partial):
+    """
+    Stores a callable and some arguments. When called, calls the
+    callable with the stored arguments and the additional arguments
+    supplied to the call.
+    """
+
+    __slots__ = ("hash",)
```
renpytom, line 80: We probably don't want slots here, as that means that something that  worked before: 

```
t1 = renpy.curry(test):
t1.predict = predict_t1 

Would stop working. 
```

Gouvernathor: I would still work because functools.partial includes "\_\_dict__" in the \_\_slots__

geomat0101:
> This uses functools.partial as a base, because it's builtin and yields far better performance. It is subclassed to provide the same features as the current Curry class does, including the clever hash caching.

Your comment about the hash caching piqued my curiosity and ultimately led me to this, which i thought was a really good read.  Not particularly relevant in this case because we don't have any concerns about values matching each other in the hash chain, and commutability is actually desirable in this case, but still interesting nonetheless so I figured I'd share: https://stackoverflow.com/questions/5889238/why-is-xor-the-default-way-to-combine-hashes

Gouvernathor:
The [python doc](https://docs.python.org/3/reference/datamodel.html#object.__hash__) advises the hashing of tuples containing the different values, and it certainly seems to work since `hash((1, 2, 3)) != hash((1, 3, 2))`.
Returning `hash((self.func, self.args, frozenset(self.kwargs.items())))` could be a good improvement of the hash mechanism (I used frozenset because items() returns an unhashable object in py3, tuple is order-dependant, and set is unhashable).

But I'm not sure whether it would break the existing class instances or just under-optimize them. Weighing for the second option in that case, but I'm not sure.
Anyway, to minimize risks, it could be preferable to do this in a separate PR.

PR #3251, opened 2022-01-24
Fix datetime pickle hack

Gouvernathor:
This fixes an error occuring when reloading a py3 game defining datetime objects.
Why it works, with the partial objects being class attributes and not instance attributes defined in `__init__`, is anyone's guess.


renpytom:
I'm not 100% sure myself, but if it works, that's good.

PR #3262, opened 2022-01-29
Clean up disused kwargs on attribute functions

mal:
The `wanted` and `remove` kwargs on `Character.resolve_say_attributes` and
`images.apply_attributes` were previously used to support the speaking
attribute feature, however that has since moved over to using the
temporary attribute functionality and these parameters are never passed
as anything beside empty lists, so we can clean them up.

### DIFF
--- a/renpy/display/image.py
+++ b/renpy/display/image.py
@@ -900,7 +900,7 @@ class ShownImageInfo(renpy.object.Object):
 
         self.shown.discard((layer, tag))
 
-    def apply_attributes(self, layer, tag, name, wanted=[], remove=[]):
+    def apply_attributes(self, layer, tag, name):
         """
         Given a layer, tag, and an image name (with attributes),
         returns the canonical name of an image, if one exists. Raises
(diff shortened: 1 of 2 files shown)


renpy/display/image.py
```diff
@@ -937,9 +934,8 @@ def apply_attributes(self, layer, tag, name, wanted=[], remove=[]):
             if f is not None:
                 defaults = f(name)
 
-        # Add any defaults to the set of attributes a matching image may have.
-        if defaults is not None:
-            optional.extend(defaults)
+        # The list of attributes a matching image may have.
+        optional = list(defaults) if defaults else [ ]
```
Gouvernathor, line 938: Maybe keeping this a None-check instead of a -check ? Like `[] if defaults is None else list(defaults)`

Gouvernathor (approved):

renpy/display/image.py
mal: This was done intentionally to avoid passing empty iterables to the list constructor. It's quite common for `()` to show up here when the callback is not being used (any image without active attributes will get an empty tuple), and the creator could also choose to return `set()`, `[]`, or `()` from the callback, which seems reasonable to me if they're already working with those data types in their function.

Conversely, I can't think of any valid scenario in which we'd want to send a falsey value to the list constructor over using the more optimised empty list literal. The worst possible outcome I can think of at the moment would be that non-iterable falsey values would be swallowed rather than triggering a stack trace, which in this scenario feels acceptable to avoid calling the list constructor unnecessarily for the vast majority of show statements.

PR #3265, opened 2022-02-01
Use At instead of transforms' `__call__`

### LINKED ISSUE #3264: Layeredimage with transform with default parameters does not pass the parameters, instead Image is received
This is related to
https://github.com/renpy/renpy/issues/2990

I am using latest Release, version 7.4.11.2266, but the issue was found on 7.4.6.

This code (simplified):
```
python: 
    def x (trans, st, at, a, b=0, c=0, d=0):
      trans.rotate = c+d
      return 0

transform y (a, b, c=0, d=0):
  function renpy.curry(x)(a=a, b=b, c=c, d=d)
  repeat


layeredimage li:
  always:
    "a.png"
  always:
    "b.png"
    at y (1, 2, 3)

show li at y(1,2,3)
```

gives:
```
TypeError: unsupported operand type(s) for +: 'float' and 'Image'
```

Changing it to 
```
layeredimage li:
  always:
    "a.png"
  always:
    "b.png"
    at y (1, 2, 3, 0)
```
produces working code.
Can you confirm the issue?

Gouvernathor:
fixes #3264

### DIFF
--- a/renpy/common/00layeredimage.rpy
+++ b/renpy/common/00layeredimage.rpy
@@ -143,8 +143,6 @@ python early in layeredimage:
             Wraps a displayable in the at list and transform arguments.
             """
 
-            d = renpy.displayable(d)
-
             d = At(d, *self.at)
 
             if self.group_args or self.transform_args:
--- a/renpy/common/00layeredimage.rpy
+++ b/renpy/common/00layeredimage.rpy
@@ -145,8 +145,7 @@ python early in layeredimage:
 
             d = renpy.displayable(d)
 
-            for i in self.at:
-                d = At(d, i)
+            d = At(d, *self.at)
 
             if self.group_args or self.transform_args:
                 d = Transform(d)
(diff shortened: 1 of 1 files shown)


mal:
The `At` function already calls `displayable` and handles application of multiple transforms, so if we're going to use it here rather than call it inside a loop the call should probably replace both the existing call to `displayable` and the transform application loop.

renpy/common/00layeredimage.rpy
```diff
@@ -145,8 +145,7 @@ python early in layeredimage:
 
             d = renpy.displayable(d)
```
mal, line 146: This can also be emitted as it's the first thing that `At` will do.

Gouvernathor: Only if at least one of self.at, self.group_args and self.transform_args is non-empty. In other cases this changes the return type of the method. I'm not sure of the consequences of that, so I'd abstain.

mal: I don't follow. The call to `displayable` inside `At` isn't contingent on anything. You could call `At(d)` and it would still call `displayable`. :/

PR #3268, opened 2022-02-01
Add SlottedNoRollback, slotted version of NoRollback

Siege-Wizard:


renpy/python.py
```diff
@@ -353,7 +353,19 @@ def reset_store_changes(name):
...
+class SlottedNoRollback(object):
+    """
+    :doc: slottednorollback class
+
+    Instances of classes inheriting from this class do not participate in
+    rollback. Objects reachable through an instance of a SlottedNoRollback
+    class only participate in rollback if they are reachable through other
+    paths.
+    """
+    __slots__ = tuple()
```
Gouvernathor, line 365: The use of the `()` litteral is safer and faster than a call to the tuple constructor.

Gouvernathor (approved):

renpy/python.py
Siege-Wizard: Done

renpy/python.py
```diff
@@ -353,7 +353,19 @@ def reset_store_changes(name):
 # the rollback list before rollback or serialization.
 
 
-class NoRollback(object):
+class SlottedNoRollback(object):
+    """
+    :doc: norollback class
```
Gouvernathor, line 358: There should not be two identical entries in the seme place in the doc for different classes. Either the new one is explained in a way highlighting its specificities, or it doesn't get documented.

Siege-Wizard: Would adding the following sentence be appropiate?
> SlottedNoRollback is designed to be used by classes that want to benefit from the reduced memory footprint that a class with `__slots__` offers.

mal:
I'm wondering if, given that `NoRollback` is (in effect) an abstract class, there is any reason we wouldn't just simply add `__slots__ = ()` to `NoRollback` instead of create a whole new class and expand the MRO? For what it's worth, several of the abstract base classes in the `stdlib` collections module do this, I believe for exactly the same reason, facilitating the creation slotted subclasses.

Any classes inheriting from it without defining `__slots__` would be allocated a `__dict__` and `__weakref__` automatically, and any that did would not. It also wouldn't conflict with other slotted classes in multi-inheritance scenarios as it wouldn't be allocating any slots.

We'd probably want to confirm there's no detriment, but it seems like a potentially much cleaner way to address this with less overhead, as well as simplifying usage for creators who wouldn't need to worry about which class to inherit from, and reduce accidents if a creator were to add `__slots__` to their `NoRollback` subclass at a later date and forget to update the superclass in the definition.

renpytom:
It's not an abstract class. It's fine to do:

```
python:
    nr = NoRollback()
    nr.field =42
```

For a pseudo-namespace.

mal:
Ahh, that's unfortunate. 🙁

Gouvernathor:
Just adding `__slots__` is not backwards-compatible, because of what Tom said. However, I'm thinking about only documenting SlottedNoRollback (maybe a shorter name would be better ?) and de-documenting NoRollback. That way current/older games using NoRollback won't change behavior, and new users will progressively use the better version.

Siege-Wizard:
@mal I had your same line of thought initially, so I asked @renpytom beforehand which approach he considered was better.

@Gouvernathor if we think of the new class as a future replacement of `NoRollback`, but keeping the backwards compatibility, what name would you suggest for the new class? Brainstorm: `NonRollback`, `NotRollback` as similar generic names that don't conflict with the old name? `Unrollbackable`? Sounds a bit weird.

Gouvernathor:
`IgnoreRollback`. Not too long, understandable, not mistakable with the other one.

renpytom:
SlottedNoRollback sounds good. We don't want to confuse the special weird one  with the one 90% of people will want to use.

PR #3228, opened 2022-01-08
Allow Transform-defined transforms to take strings

Gouvernathor:
This allows
```rpy
define centr = Transform(align=(0.5, 1.0))
image test = centr("#fff")
```
following the example of
```rpy
transform centr:
    align (0.5, 1.0)
image test = centr("#fff")
```

Issue #2102, opened 2020-01-22
Ren'Py transforms issues.


Andykl:
1. It's impossible to wrap string in predefined `Transform`:
```
define centr = Transform(align=(0.5, 1.0))
image test = centr("#fff")
```
It's at odds with `ATLTransform`:
```
transform centr:
    align (0.5, 1.0)
image test = centr("#fff")
```

2. When we have ATL based image and wrap it in another ATL transform (or via `At`) like:
```
image base:
    "image 1"
    pause 1.0
    "image 2"
    pause 1.0
    repeat

transform centr:
    align (0.5, 1.0)

image test = centr("base")
```
It starts animation on the first show. But after hide and show again the animation doesn't starts from the begin, but starts from the point it was hidden.
The same thing with first point:
```
define centr = Transform(align=(0.5, 1.0))
image test = centr(ImageReference("base"))  # ImageReference to avoid first point
```
It's at odds with another kinds of transform usage, where animation starts from the beggining:
```
image test = Transform("base", align=(0.5, 1.0))

image test:
    centr
    "base"

show base at centr
```

3. Manual wrapping an ATLTransform in ATLTransform (or via `At`) like:
```
transform test_zoom:
    zoom 1.0
    linear 1.0 yzoom 0.25
    linear 1.0 yzoom 1.0
    pause 0.1
    repeat

transform test_alpha:
    alpha 1.0
    linear 1.0 alpha 0.25
    linear 1.0 alpha 1.0
    pause 0.1
    repeat

image test = test_zoom(test_alpha("base"))
```
It starts with propertly animation on the first show. But after hide and show again the animation of the inner transform (which is not the final child, for child it behave as second point) freeze on the point it was hidden and and idle some time. After some time of observation I guess that inner transform waits untill game runtime reached.
This behaviour totally confusing and at odds with the only way that works:
```
image test:
    "base"
    parallel:
        test_zoom
    parallel:
        test_alpha
```
4. If a `Transform` without child passed as expression statement, it will be interpreted as displayable and will be rendered as Null. Assuming that Transform is a simplified version of ATLTransform, it is expected that the parent transform will only take state from this.
5. And vice-versa, if ATLTransform with the child passed to another transform, it tries to execute it ignoring the child. This is related to #1957.
6. I'm not sure that's the right option. If an expression statement returns None, it will be exception "Not a displayable: None". It would be useful to interpret it as a pass. Right now to simulate this I should write:
```
transform _pass:
    pass
```
[...]

PR #3269, opened 2022-02-02
Improve duplicatable check to consolidate coercion

mal:
Trying to coerce the `child` arg to `Transform.__call__` into a displayable
adds an avoidable call and could lead to redundant duplication (in the
case where the coercion yields a duplicatable displayable).

Instead, use `getattr` in the duplicatable check to prevent raising and
allow `__init__` (which already does arg coercion) to handle it as normal.

Related to #2102, and #3228. /cc @Gouvernathor

### DIFF
--- a/renpy/display/transform.py
+++ b/renpy/display/transform.py
@@ -774,7 +774,7 @@ class Transform(Container):
         if child is None:
             child = self.child
 
-        if (child is not None) and getattr(child, '_duplicatable', False):
+        if getattr(child, '_duplicatable', False):
             child = child._duplicate(_args)
 
         rv = Transform(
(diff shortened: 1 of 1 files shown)


renpy/display/transform.py
```diff
@@ -770,12 +770,11 @@ def event(self, ev, x, y, st):
         return None
 
     def __call__(self, child=None, take_state=True, _args=None):
-        child = renpy.easy.displayable_or_none(child)
 
         if child is None:
             child = self.child
 
-        if (child is not None) and (child._duplicatable):
+        if (child is not None) and getattr(child, '_duplicatable', False):
```
Gouvernathor, line 777: You can even get rid of the `child is not None` part now. It's great ! Hadn't realized how simple it is now.

Gouvernathor (approved):

renpy/display/transform.py
mal: Good call, overwhelming majority of calls are `not None` so optimising for that case makes sense. 👍

PR #3282, opened 2022-02-07
Add SlottedRevertableObject, the base type that have no `__dict__`

Andykl:


renpy/revertable.py
```diff
@@ -411,10 +411,54 @@ def _rollback(self, compressed):
...
+        rv = super(RevertableObjectMeta, cls).__new__(cls, name, bases, dct)
+
+        # Collect all slots from type MRO
+        slots = set()
+        for t in rv.__mro__:
+            # Stop on object as it has no slots but it is special
+            if t is object:
+                break
+
+            if "__slots__" in t.__dict__:
```
Gouvernathor, line 425: maybe this could be `slots.update(getattr(t, "__slots__", {"__dict__"}))`

renpy/revertable.py
```diff
@@ -411,10 +411,54 @@ def _rollback(self, compressed):
...
+                slots.update(t.__slots__)
+            else:
+                slots.add("__dict__")
```
mal, line 428: Proposed fix to avoid capturing slots defined in subclasses whose ancestors have already dropped `__slots__`.
```suggestion
        for t in reversed(rv.__mro__):
            # Skip object as it has no slots but it is special
            if t is object:
                continue

            if '__slots__' not in t.__dict__:
                slots.add('__dict__')
                break

            slots.update(t.__slots__)
```

renpy/revertable.py
```diff
@@ -411,10 +411,54 @@ def _rollback(self, compressed):
...
+
+        # Collect all slots from type MRO
+        slots = set()
+        for t in rv.__mro__:
+            # Stop on object as it has no slots but it is special
+            if t is object:
+                break
+
+            if "__slots__" in t.__dict__:
+                slots.update(t.__slots__)
```
Gouvernathor, line 426: This misbehaves if a subclass has a string \_\_slots__.
Something like `slots.update((t.__slots__,) if isinstance(t.__slots__, str) else t.__slots__)`.

Andykl:
Code with which I tested this PR
```py
init python:
    class Descriptor:
        def __set_name__(self, instance, name):
            self.name = name

        def __get__(self, instance, owner):
            if owner is None:
                return self

            return instance._descr[self.name]

        def __set__(self, instance, value):
            instance._descr[self.name] = value

        def __delete__(self, instance):
            del instance._descr[self.name]

    class NoDict(renpy.revertable.SlottedRevertableObject):
        __slots__ = "b"
        a = Descriptor()

        def __init__(self):
            self._descr = {}
            self.a = 1
            self.b = 1
            self.c = 1

    class NoDictChild(NoDict):
        __slots__ = "d"

        def __init__(self):
            super().__init__()
            self.d = 1

    class NoDictChildWithDict(NoDictChild):
        def __init__(self):
            super().__init__()
            self.e = 1

    class Dict(object):
        __slots__ = "b"
        a = Descriptor()

        def __init__(self):
            self._descr = {}
            self.a = 1
            self.b = 1
            self.c = 1

    class DictChild(Dict):
        __slots__ = "d"

        def __init__(self):
            super().__init__()
            self.d = 1

    class DictChildWithDict(DictChild):
        def __init__(self):
            super().__init__()
            self.e = 1

default no_dict = NoDictChildWithDict()
default with_dict = DictChildWithDict()
label start:
    $renpy.unwatchall()
    $renpy.watch("no_dict.a")
    $renpy.watch("no_dict.b")
    $renpy.watch("no_dict.c")
    $renpy.watch("no_dict.d")
    $renpy.watch("no_dict.e")
    $renpy.watch("with_dict.a")
    $renpy.watch("with_dict.b")
    $renpy.watch("with_dict.c")
    $renpy.watch("with_dict.d")
    $renpy.watch("with_dict.e")
    pause
    $no_dict.a = 3
    $no_dict.b = 3
    $no_dict.c = 3
    $no_dict.d = 3
    $no_dict.e = 3
    $with_dict.a = 3
    $with_dict.b = 3
    $with_dict.c = 3
    $with_dict.d = 3
    $with_dict.e = 3
    pause
    $del no_dict.a
    $del no_dict.b
    $del no_dict.c
    $del no_dict.d
    $del no_dict.e
    $del with_dict.a
    $del with_dict.b
    $del with_dict.c
    $del with_dict.d
    $del with_dict.e
    pause
    return
```

Andykl:
More sophisticated code
```py

init python:
    class Descriptor:
        def __set_name__(self, instance, name):
            self.name = name

        def __get__(self, instance, owner):
            if owner is None:
                return self

            return instance._descr[self.name]

        def __set__(self, instance, value):
            instance._descr[self.name] = value

        def __delete__(self, instance):
            del instance._descr[self.name]

    class NoDict(renpy.revertable.SlottedRevertableObject):
        __slots__ = "__b", "_descr", "c"
        a = Descriptor()

        def __init__(self):
            self._descr = {}
            self.a = 1
            self.b = 1
            self.c = 1

        @property
        def b(self):
            return self.__b

        @b.setter
        def b(self, value):
            self.__b = value

        @b.deleter
        def b(self):
            del self.__b

    class NoDictChild(NoDict):
        __slots__ = "__d"

        def __init__(self):
            super().__init__()
            self.d = 1

        @property
        def d(self):
            return self.__d

        @d.setter
        def d(self, value):
            self.__d = value

        @d.deleter
        def d(self):
            del self.__d

    class NoDictChildWithDict(NoDictChild):
        def __init__(self):
            super().__init__()
            self.e = 1

    class Dict(object):
        __slots__ = "__b"
        a = Descriptor()

        def __init__(self):
            self._descr = {}
            self.a = 1
            self.__b = 1
            self.c = 1

        @property
        def b(self):
            return self.__b

        @b.setter
        def b(self, value):
            self.__b = value

        @b.deleter
        def b(self):
            del self.__b

    class DictChild(Dict):
        __slots__ = "__d"

        def __init__(self):
            super().__init__()
            self.d = 1

        @property
        def d(self):
            return self.__d

        @d.setter
        def d(self, value):
            self.__d = value

        @d.deleter
        def d(self):
            del self.__d

    class DictChildWithDict(DictChild):
        def __init__(self):
            super().__init__()
            self.e = 1

default no_dict = NoDictChildWithDict()
default with_dict = DictChildWithDict()
label start:
    $renpy.unwatchall()
    $renpy.watch("no_dict.a")
    $renpy.watch("no_dict.b")
    $renpy.watch("no_dict.c")
    $renpy.watch("no_dict.d")
    $renpy.watch("no_dict.e")
    $renpy.watch("with_dict.a")
    $renpy.watch("with_dict.b")
    $renpy.watch("with_dict.c")
    $renpy.watch("with_dict.d")
    $renpy.watch("with_dict.e")
    pause
    $no_dict.a = 3
    $no_dict.b = 3
    $no_dict.c = 3
    $no_dict.d = 3
    $no_dict.e = 3
    $with_dict.a = 3
    $with_dict.b = 3
    $with_dict.c = 3
    $with_dict.d = 3
    $with_dict.e = 3
    pause
    $del no_dict.a
    $del no_dict.b
    $del no_dict.c
    $del no_dict.d
    $del no_dict.e
    $del with_dict.a
    $del with_dict.b
    $del with_dict.c
    $del with_dict.d
    $del with_dict.e
    pause
    $no_dict.a = 5
    $no_dict.b = 5
    $no_dict.c = 5
    $no_dict.d = 5
    $no_dict.e = 5
    $with_dict.a = 5
    $with_dict.b = 5
    $with_dict.c = 5
    $with_dict.d = 5
    $with_dict.e = 5
    pause
    return
```

Booplicate:
This looks very ineffective. Devs would have to make their classes explicitly inherit from the new class. You could avoid that and also don't make codebase this complex if you just made the original class use slots. That way developer can just define `__slots__` and get the expected result (or not define and get a class with `__dict__`). That's also how it is in Python.

renpytom:
We've discussed that already, and the problem with it is that it would break existing games, which expect the default object to have a __dict__.

Booplicate:
Can't it be made compat? Like catching if the object is expected to have `__dict__` during unpickling?

Gouvernathor:
Not with python 2 (which is still the base for renpy 7), and it would wtill breax things like
```py
define mynamespace = object()
define munamespace.variable = value
```
in the code of some games. That code was supported, so it needs to continue being supported.

Booplicate:
I see now, that is inconvenient... I don't know how many people used code like that, to me it looks like a bad practice. I also think some breaking changes should be expected when you update old games from py2 to py3. It's also very easy to fix - subclass `object`.

But if you want to keep that behaviour, then, what if we kept `object` as is so people who used it as a "namespace" could still do it. But make classes inherit from the slotted version by default?

renpytom:
Right now, the transition from Python 2 to Python 3 has been pretty seamless. I don't know about very large games, but I have been able to run many small and medium sized games with no changes, or very trivial ones. 

I don't think that introducing support for ``__slots__`` by default is what I want to do, to the point where I'm leaning against merging this as a whole. What I'm leaning towards is documenting ``python_object``, the name for Python's object in Ren'Py, and making adding ``__slots__`` to an object an error. 

The reason for this is that there's basically an inherent conflict in SlottedRevertableObject. On one hand you're saying that these objects take up so much memory that we're worried about how much memory they take up, so we need to try to reduce that. And at the same time, being revertable implies the objects will participate in save, load, and rollback. 

That doesn't make sense to me.

PR #3285, opened 2022-02-09
update 7z sfx with dpi aware

qinlili23333:
the old 7z sfx has no dpi aware that it looks blur on high dpi screens
i compiled a new 7z sfx with dpi aware using latest 7z source (2107)
image
it looks clear now


renpytom:
Merged, thank you.

renpytom:
I'm going to be moving this file into renpy/renpy-build, but that's purely a project organization thing.

PR #3295, opened 2022-02-19
Allow Python statements to be run from JS (emscripten)

Tey:
This PR allows Python statements to be executed in Ren'Py process from Javascript. Currently, the interface between JS and Ren'Py is pretty limited and adding new functionalities to it requires to recompile the emscripten application, as there is no feature in emscripten/SDL to allow generic communications between JS and the application. Being able to execute Python statements in Ren'Py process directly allows more powerful controls from JS without having to recompile anything.

The changes make Ren'Py look for a specific MEMFS file during idle frames, and execute its content as a Python script. Using a file in MEMFS makes sure the lookup will be fast and no FS sync is required. The file creation and execution errors are handled by JS (see the related PR in renpyweb).
I don't know Ren'Py architecture well enough to see if there is a better place in the code to check for that file. The main problem with doing it during idle frames is that no such execution can happen when the browser window is not visible.

It should be safe to execute any external Python statements when running from the browser (emscripten) as the Python process cannot do more harm than what can be done from JS already. It is unsafe for other platforms though.

I'll post a PR in renpyweb with the JS interface to run Python statements, and a concrete use case (see renpy/renpyweb#23).

### DIFF
--- a/renpy/display/core.py
+++ b/renpy/display/core.py
@@ -4303,25 +4303,18 @@ class Interface(object):
         """
 
         if not renpy.emscripten:
-            # Allowing execution of untrusted Python statements from outside
-            # the browser would be a security risk
             return
 
-        # Use a command file in MEMFS so that no sync is required
-        cmd_file = '/_js_cmd.py'
-
-        # Check if a command needs to be executed
-        if not os.path.exists(cmd_file):
+        # Retrieve the command to be executed from a global JS variable
+        # (an empty string is returned if the variable is not defined)
+        cmd = emscripten.run_script_string("window._renpy_cmd")
+        if len(cmd) == 0:
             return
 
-        # Read command file content
-        with open(cmd_file, "rt") as f:
-            cmd = f.read()
+        # Delete command variable to prevent executing the command again
+        emscripten.run_script("delete window._renpy_cmd")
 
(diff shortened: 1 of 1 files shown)


renpytom:
I like the idea of this, though I'm not sure about the implementation. Wouldn't it be simpler to just have a field, something like window.renpy_exec, that can be accessed without going through the filesystem?

Tey:
Do you mean something like renpy checking this field periodically to see if there is something to execute? If so, I have the feeling this would be slower than reading the FS because it would rely on multiple calls to `emscripten.run_script*()`, but I can compare the performances if this is what you mean.

Otherwise, if you mean having a JS function that would execute something in Renpy's Python process without having Renpy to pool for anything, this is not possible AFAIK.

renpytom:
I'm thinking of something like:

```python
command = emscripten.run_script_string("window.renpy_command")
if command:
    emscripten.run_script("window.renpy_command = '';")
    exec(command)
```

Tey:
OK, I'll try that and compare the performances.

Tey:
I was wrong: pooling is at least 5 times faster when using a global variable instead of a file. I've updated the branches of both PRs.

renpytom:
I've merged them both, thank you.

PR #3296, opened 2022-02-20
Allow function list and None for function statement in ATL ant parameter of Transform

kyouryuukunn:
I allowed function list for function statement in ATL ant parameter of Transform.
This allows below code.  the minimum number is used when some functions return a number .
```
show test:
    function [func1, func2]
```

I also allowed None for function statement in ATL.
Now, below code can disable function.
```
show test:
    function function_test
    time 2.
    function None
```

### DIFF
--- a/renpy/atl.py
+++ b/renpy/atl.py
@@ -1702,10 +1702,11 @@ class Function(Statement):
 
     def execute(self, trans, st, state, events):
         min_fr = None
-        for f in self.function:
-            fr = f(trans, st, trans.at)
-            if fr is not None and (min_fr is None or fr < min_fr):
-                min_fr = fr
+        if self.function is not None:
+            for f in self.function:
+                fr = f(trans, st, trans.at)
+                if fr is not None and (min_fr is None or fr < min_fr):
+                    min_fr = fr
 
         if min_fr is not None:
             return "continue", None, min_fr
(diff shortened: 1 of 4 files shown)


Gouvernathor:
Can this load a save made with the current `Transform(function=whatever)`, without raising exceptions ?
Considering the code, I'm 99% sure it errors out. Luckily, \_\_version__ and after_upgrade can solve it (I'll do it if this is accepted).

mal:
I think this is the sort of thing better left for creators that need it to implement for themselves. In the provided implementation all the functions are re-run based on the one that wants to re-run the most, which may not always be desirable.

Here's an example creator-side implementation to support this use-case. It would require some Python knowledge to create initially, but once in place would present a simple to use interface:

```py
class multifn(object):
    def __init__(*args):
        self.fns = args

    def __call__(trans, st, at):
        min_fr = None
        for f in self.fns:
            fr = f(trans, st, at)
            if fr is not None and (min_fr is None or fr < min_fr):
                min_fr = fr
        return min_fr
```
```rpy
show test:
    function multifn(func1, func2)
```

renpytom:
Mal's right. Composing functions is easy to do in Python, and I don't think there's a clear way of doing it like there is for Actions.

kyouryuukunn:
OK, @renpytom, @Mal, I close.

PR #3297, opened 2022-02-21
Fix a crash in dev menu

Booplicate:
Since `file` doesn't exist in py3, I replaced it with `open` instead. During my testing the menu worked again.


renpytom:
Thanks!

PR #3298, opened 2022-02-21
add ShowAlternateMenu action

kyouryuukunn:
I added ShowAlternateMenu action and _alternate_menu screen

example:
alternate ShowAlternateMenu( [ ("button1", MainMenu()), ("button2"), NullAction()) ],  style_prefix="alternatemenu" )

    class ShowAlternateMenu(Action, DictEquality):
        """
         :doc: menu_action
         Show alternate menu screen at the bottom right corner of the cursor.
         That screen can include textbuttons.
         `button_list`
             The list of tuple which has two elements. first is a string used
             as text for textbutton and second is action.
         `menu_width`
             If the x-coordinate of the cursor plus this number is greater than
             :var:`config.scren_witdth`, the menu will be displayed to the left
             of the cursor. default is 300
         `menu_height`
             If the y-coordinate of the cursor plus this number is greater than
             :var:`config.scren_witdth`, the menu will be displayed to the upper
             of the cursor. default is 300
         `style_prefix`
             If not None, this is used for menu screen as style_prefix.

image

### DIFF
--- a/renpy/common/00action_menu.rpy
+++ b/renpy/common/00action_menu.rpy
@@ -355,45 +355,47 @@ init -1500 python:
 
          """
         def __init__(self, button_list, menu_width=300, menu_height=200, style_prefix=None):
-            self.button_list = button_list
+            self.button_list = []
+            for text, action in button_list:
+                if not isinstance(action, list):
+                    action = [action]
+                action += [Hide("_alternate_menu")]
+                self.button_list.append((text, action))
             self.style_prefix = style_prefix
             self.menu_width = menu_width
             self.menu_height = menu_height
 
         def predict(self):
-            renpy.predict_screen("_alternate_menu", self.button_list, 
+            renpy.predict_screen("_alternate_menu", self.button_list, pos = (x, y), anchor=(xanchor, yanchor),
                 menu_width=self.menu_width, menu_height=self.menu_height, style_prefix=self.style_prefix, _transient=True)
 
         def __call__(self):
 
-            renpy.show_screen("_alternate_menu", self.button_list, 
+            x, y = renpy.get_mouse_pos()
+            if x + self.menu_width > config.screen_width:
+                xanchor = 1.0
+            else:
+                xanchor = 0.0
+            if y + self.menu_height > config.screen_height:
+                yanchor = 1.0
+            else:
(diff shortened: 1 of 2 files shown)


Andykl:
I like the idea in general.
3 things I think should be fixed
1. I think it is better to make all computations in action instead of screen
2. If I understand correctly `$(x,y) = renpy.get_mouse_pos()` will make this window follow the cursor, I think it should be calculated once in point 1.
3. To make this window customizable it should be
```py
gui.rpy template file
screen alternate_menu(button_list):
    for text, action in button_list:
        textbutton "[text!ti]" action action

00action_menu.rpy
init -1500:
    screen _alternate_menu(...):
        ...
        use alternate_menu(button_list)
```

renpy/common/00action_menu.rpy
```diff
@@ -329,3 +329,71 @@ init -1500 python:
...
+         `menu_height`
+             If the y-coordinate of the cursor plus this number is greater than
+             :var:`config.scren_witdth`, the menu will be displayed to the upper
+             of the cursor. default is 300
+         `style_prefix`
+             If not None, this is used for menu screen as style_prefix.
+
+         """
+        def __init__(self, button_list, menu_width=300, menu_height=200, style=None):
+            self.button_list = button_list
```
Gouvernathor, line 358: The list could be converted here to a list of TextButtons. It would save time, because it would be computed once per ShowAlternateMenu instance instead of once per right-click. It would also open the door for the programmers to pass Buttons in the list directly.
You could then add each one in the screen with `add` instead of `textbutton`.
(I wrote this before Andy but forgot to send it)

kyouryuukunn:
I fixed these issues.
I add gui/game/screen.rpy alternate_menu screen instead of gui.rpy

renpytom:
I don't think this is going to go into Ren'Py as-is - I agree that we should probably have this functionality, but this implementation doesn't seem right to me. I think this should probably be some sort displayable, rather than something implemented in terms of screens. 

(Basically, this feels fine now - but with the ability to change Ren'Py, I don't think this is something I want to support forever in the engine.)

kyouryuukunn:
OK, @renpytom 
I'll close

PR #3304, opened 2022-02-26
Document config.atl_one_frame

Gouvernathor:


renpytom:
I don't think we want to document this, outside of incompatible changes.

PR #3308, opened 2022-03-01
Relative designate for ATL properties

kyouryuukunn:
I allowed relative value for ATL properties which are numerical type and not None by adding property `rel_` prefix.

for example, The below 2 code is equivalent.

```
    show bg:
        xpos 0.5 
        linear 1.15 rel_xpos 0.1  knot -0.87 
```

```
    show bg:
        xpos 0.5 
        linear 1.15 xpos 0.6  knot -0.37 
```

### DIFF
--- a/renpy/atl.py
+++ b/renpy/atl.py
@@ -1270,10 +1270,9 @@ class Interpolation(Statement):
                             raise Exception("tuple isn't available for relative designation")
                         value += old_v
                         abs_values.append(value)
-                    abs_values = tuple(abs_values)
                 else:
                     abs_values = values
-                splines.append((name, [ getattr(trans.state, name) ] + values))
+                splines.append((name, [ getattr(trans.state, name) ] + abs_values))
 
             state = (linear, revolution, splines)
 
--- a/renpy/atl.py
+++ b/renpy/atl.py
@@ -950,7 +950,7 @@ class RawMultipurpose(RawStatement):
         properties = [ ]
 
         for name, expr in self.properties:
-            if name not in PROPERTIES:
+            if name not in PROPERTIES and (not name.startswith("rel_") or name[len("rel_"):] not in PROPERTIES):
                 raise Exception("ATL Property %s is unknown at runtime." % name)
 
             value = ctx.eval(expr)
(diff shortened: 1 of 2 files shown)


via-my:
Wow! That's really useful.

renpytom:
I like the idea of this, but I'm not a huge fan of the syntax. I think a version of the syntax that might work better would be: 

```
xpos current + 10
ypos current - 20
```

Something like that, maybe?

JSearUK:
Hmm. Is there a reason why the following might not be useable?
```py
xpos += 10
ypos -= 20
```

PastryIRL:
Funnily enough, I just opened an issue #3311 that this PR addresses.
I've checked out the fork and it appears to be working perfectly. I'll adopt it until it's merged into master :)
As far as the syntax goes, I think the standard decrement/increment notation of `-=`/`+=` would be the best approach as it's the most commonly used.

Booplicate:
I like the `+=`/`-=` approach. Short, informative, doesn't require to learn new keywords/syntax as it already exists in Python and other languages.

Andykl:
I would like it, if it is also possible to use `xpos = 100`. Otherwise it may be a problem to remember `xpos += 100` to shift, but `xpos 100` to set.

Andykl:
Tom's variant is backward incompatible due to
```
transform some(current):
    xpos current + 10
```

renpytom:
My only issue with +=/-= is when you have more than one of them on a line, which Ren'Py encourages. 

```
linear 2.0 xpos += 5 ypos += 10
```

Looks weird to me. I'd be fine with allowing an optional = there, and perhaps it makes sense to allow the interpolations to take a block. This would look good as:

linear 2.0:
    alpha = .5
    xpos += 5
    ypos += 10

I think. @Andykl, what do you think?

Andykl:
Yeah, it sounds great and acutally I was thinking to add this possibility because long lines are terrible.
@kyouryuukunn Do you will to make it or should I?

renpytom:
(I am tired enough I didn't realize who was working on this.)

kyouryuukunn:
Hello.

I think Pytom's idea is good.
Thank you, @Andykl. please make it.

The reason I chosed my syntax is it's difficult to change parser and it's easy to just add properties prefix

renpytom:
I'd suggest for the implementation, have a new type, relative(), such that  += 100 is represented as relative(100) after the parser step. That could then be used by the Interpolation node to determine the new value based on the old one.

Andykl:
Closing in favor of #3358

PR #3314, opened 2022-03-02
add drag dragging callback

kyouryuukunn:
I added drag displayable and drag statement in screen lang dragging callback.
I also added document start_x, start_y

    `dragging`
        A callback (or list of callbacks) that is called when the Drag is being
        dragged. It is called with one argument, a a list of Drags that are
        being dragged. If the callback returns a value other than None, that
        value is returned as the result of the interaction.

    `start_x`, `start_y`
         The dragg start position of the Drag relative to its parent, in pixels.


renpytom:
Merged, thank you. I'll make a few backwards compatibility fixes to this, but the basics seem good.

PR #3320, opened 2022-03-08
Add the stack console command

Gouvernathor:


renpytom:
What does the output of this look like? Can it be formatted better?

Gouvernathor:
image
An indent could be added to the lines to separate them from the first one, but I think the parentheses and the colon make it clear enough.
I don't think you can be in the console with an empty stack, since accessing the console is a call in itself (afaik), but just in case I added an "empty stack" line.

renpytom:
I don't think this is particularly useful as is. What would anyone do with this information. 

What would probably work is something like:

* anonymous <game/front_page.rpy:1234>
* _call__foo_bar <game/other_place.rpy:6789>

But without any formatting of the results, this isn't particularly useful. 

It might also make sense to remove the top entry, as it's the location the console was called from, which isn't overly interesting.

Gouvernathor:
image
From a game with a `call a from b` line.
The anonymous call to the console is kept, because you actually **can** go to the console without calling it : from the main menu, from the game menu, when the interface is being hidden (with the H key)... Whenever it says "ren'py script disabled" instead of "ren'py script enabled", it seems.
I'd rather keep this as-is to avoid misleading users.

renpytom:
I can see that. I think this is better output. Anything else before I merge it?

Gouvernathor:
In my code I assumed entry[2] to be the line number since entry[0] is a filename, but there isn't anything of relevance in the 15th line of my script.rpy file. So, I'm not so sure about it now...

PR #3327, opened 2022-03-11
add doc about movie with alpha channel

kyouryuukunn:
I add the below sentence doc.

It depends on the ffmpeg native decoders whether movies with alpha channel work
correctly. We recommend using mask or side_mask for that purpose.

### DIFF
--- a/sphinx/source/movie.rst
+++ b/sphinx/source/movie.rst
@@ -37,7 +37,12 @@ Movies can be displayed fullscreen or in a displayable. Fullscreen movies
 are more efficient.
 
 It depends on the ffmpeg native decoders whether movies with alpha channel work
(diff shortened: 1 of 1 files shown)


kyouryuukunn:
I added ffmpeg example.


It depends on the ffmpeg native decoders whether movies with alpha channel work
correctly. We recommend using mask or side_mask for that purpose. Here is an
ffmpeg example of encoding a webm file for side-by-side mask mode from a mov
file with an alpha channel. ::

        ffmpeg -i original.mov -filter:v alphaextract mask.mov
        ffmpeg -i original.mov -i mask.mov -filter_complex "hstack" -codec:v vp8 -crf 10 output.webm

PR #3330, opened 2022-03-11
add doc cosole commands

kyouryuukunn:
I added doc console command helps


renpytom:
Does it make sense to auto-generate this? I'd be a bit worried about this going out of date over time.

kyouryuukunn:
I fixed. now doc is auto-generated.

result is here
image

PR #3336, opened 2022-03-16
Change the outlines_scaling's default

Gouvernathor:


renpy/common/00style.rpy
```diff
@@ -150,7 +150,7 @@ init -1800:
         drop_shadow None
         drop_shadow_color (0, 0, 0, 255)
         outlines [ ]
-        outline_scaling "step"
+        outline_scaling "integer"
```
mal, line 153: Is this meant to be `linear`?

Issue #1992, opened 2019-08-31
feature request: extract dialogue for another language than the default one


zeograd:
When performing voice dubbing for another language than the native/default one, it would be interesting to have the spreadsheet export with id, actor and translated text
I guess that adding a language selector on the Extract dialog screen and applying translation during the export could answer this use case.

So far, we can rely on the tl/*/script.rpy at best which isn't adapted for spreading work across multiple voice actors nor tracking progression.

Thanks in advance.

kyouryuukunn:
I sent pull request.
#3339

PR #3339, opened 2022-03-19
extract dialogue in preferences.language in that game

kyouryuukunn:
This make lancher extract dialogue in preferences.language in that game
#1992

### DIFF
--- a/launcher/game/translations.rpy
+++ b/launcher/game/translations.rpy
@@ -27,6 +27,9 @@ init python:
     if persistent.translate_language is None:
         persistent.translate_language = "english"
 
+    if persistent.extract_language is None:
+        persistent.extract_language = "None"
+
     if persistent.generate_empty_strings is None:
         persistent.generate_empty_strings = True
 
@@ -391,7 +394,7 @@ screen extract_dialogue:
                     text _("Language:")
 
                     input style "l_default":
-                        value FieldInputValue(persistent, "translate_language")
+                        value FieldInputValue(persistent, "extract_language")
                         size 24
                         color INPUT_COLOR
                         allow interface.TRANSLATE_LETTERS
@@ -408,7 +411,7 @@ label start_extract_dialogue:
 
     python:
 
-        language = persistent.translate_language
+        language = persistent.extract_language
 
         args = [ "dialogue", language]
 
--- a/launcher/game/translations.rpy
+++ b/launcher/game/translations.rpy
@@ -27,8 +27,6 @@ init python:
     if persistent.translate_language is None:
         persistent.translate_language = "english"
 
-    persistent.extract_language = persistent.extract_language or "None"
-
     if persistent.generate_empty_strings is None:
         persistent.generate_empty_strings = True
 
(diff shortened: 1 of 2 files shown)


kyouryuukunn:
maked language selectable
image

renpytom:
I like  the general idea of this, but I think it should more explicitly tell the user they can type in None to export dialogue for the default language, and it should probably use it's own persistent variable, not the translate language. I think despite where it's located in Ren'Py's implementation, this is mostly used to export text for voice acting, and we don't want to break that case.

kyouryuukunn:
I added persistent.extract_language. By default , this is "None" so user look "None" for the first time in extract dialogue screen and they should understand they can export the default language in None.

launcher/game/translations.rpy
```diff
@@ -381,6 +384,21 @@ screen extract_dialogue:
...
                     textbutton _("Extract all translatable strings, not just dialogue.") action ToggleField(persistent, "dialogue_strings") style "l_checkbox"
 
+                add SPACER
+                add SEPARATOR2
+
+                frame:
+                    style "l_indent"
+                    has vbox
+
+                    text _("Language:")
```
Gouvernathor, line 394:
```suggestion
                    text _("Language (or None):")
```
This should be simple enough to show creators that they can use None here. If they have put another language and forgotten about the default value, for example.

launcher/game/translations.rpy
```diff
@@ -27,6 +27,9 @@ init python:
     if persistent.translate_language is None:
         persistent.translate_language = "english"
 
+    if persistent.extract_language is None:
```
Gouvernathor, line 30: This should probably be replaced with `persistent.extract_language = persistent.extract_language or "None"`.
That way, if you simply remove all characters in the input and end up with "" (the empty string), it also gets replaces with "None".
I also think that replacement should not only happen at init time, which is on launcher boot and reload, but every time it may come to use, which means doing it inside the extract_dialogue label.

Gouvernathor (changes requested):

renpy/translation/dialogue.py
```diff
@@ -336,7 +352,11 @@ def dialogue_command():
             continue
 
         filename = os.path.normpath(filename)
-        DialogueFile(filename, output, tdf=tdf, strings=args.strings, notags=args.notags, escape=args.escape)
+        language = args.language
+        if language == "None":
```
Gouvernathor, line 356:
```suggestion
        if language in ("None", ""):
```
just in case

kyouryuukunn:
I fixed.

launcher/game/translations.rpy
```diff
@@ -27,6 +27,8 @@ init python:
     if persistent.translate_language is None:
         persistent.translate_language = "english"
 
+    persistent.extract_language = persistent.extract_language or "None"
```
Gouvernathor, line 30: I think this would better be placed in the "extract_dialogue" label, or even in the "extract_dialogue" screen, so that it's updated every time the variable is read, used or set.

kyouryuukunn:
I placed it into "extract_dialogue" label.

Issue #2012, opened 2019-10-02
Sidebar glitch on documentation site


lunalucid:
Scrolling to the bottom of a page causes the sidebar to glitch/shrink

PR #3351, opened 2022-03-24
Allow ATL atl_expression to be a None

Andykl:
6th point in #2102. I'm not sure it is a good idea, because it can lead to implicitly ignoring None returns from functions.

### DIFF
--- a/renpy/atl.py
+++ b/renpy/atl.py
@@ -964,6 +964,9 @@ class RawMultipurpose(RawStatement):
             else:
                 transition = None
 
+            if child is None:
+                return Block(self.loc, [ ])
+
             if isinstance(child, (int, float)):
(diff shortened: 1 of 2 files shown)


mal:
This feels likely to create more issues than it solves. If a function is being called here and not returning a displayable that feels like it'd be an unintentional mistake more often than not, obfuscating the error behind a silent no-op does not seem helpful.

There is already a mechanism for updating the child displayable, which would presumably be the reason for wanting something like this - use the `function` keyword to execute the logic necessary and then `t.set_child` inside that function to update (or not) as appropriate.

renpytom:
I think the workaround in #2012 - defining an explicit _pass transform - and using that is probably the right solution for this, at least until the eventual ATL redesign. So I'll close this for now.

PR #3355, opened 2022-03-24
Forbid to use `define persistent.var`

### LINKED ISSUE #3354: Warn about define with persistent in lint
Related to #2166.
As said [here](https://www.renpy.org/doc/html/changelog6.html#define-improvements) in the changelog6 page, `define persistent.x = val` behaves exactly as `default persistent.x = val` would, which is especially different from the way the following code works :
```rpy
init python:
    persistent.x = value
```
The behavior of the define statement is usually equivalent (with minor differences mostly relative to substores) to the init python version.
Also, as the documentation for the define statement states :
**Variables that are defined using the define statement are treated as constant, are not saved or loaded, and should not be changed.**
Considering that the persistent store is meant to be changed by its very purpose, it is nonsensical to use define with the persistent store.
For these two reasons, it is most likely a mistake for the define statement to be used in cunjunction with the persistent store.

Of course, a compatibility config variable and/or a config.developer exception should be added in order not to break older games. However, such compat variable should not be documentedin the changelog, as creators should be encouraged to use `default` instead of `define` when it comes to persistent. Unless a difference is found in how default and define work with regard to the persistent store.

Andykl:
Fixes #3354.
Misses changes to documentation and mention in incompatible changes.

### DIFF
--- a/renpy/common/000namespaces.rpy
+++ b/renpy/common/000namespaces.rpy
@@ -20,8 +20,10 @@
         pure = False
 
         def set(self, name, value):
-            if getattr(persistent, name) is None:
-                setattr(persistent, name, value)
+            if config.allow_define_persistent:
+                self.set_default(persistent, name, value)
+            else:
+                raise Exception("The define statement can not be used with the persistent namespace.")
 
         def set_default(self, name, value):
             if getattr(persistent, name) is None:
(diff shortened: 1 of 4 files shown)


Gouvernathor:
I don't know if it would be compatible with the use of a config variable, but maybe it would be better to implement this in renpy/common/000namespaces.rpy ?

renpytom:
Isn't there a chicken-and-egg problem here, in that the define statements can run before the persistent statements can? 

This should be, at most, a lint warning. This change doesn't  really benefit creators - it just requires them to change their games, and for what benefit to them?

renpytom:
Closing this, as I don't think it's the right approach, and the change to 01example.rpy is in.

PR #3358, opened 2022-03-26
Add ability to use relative ATL property values

Andykl:
Working, but a bit messy implementation of the idea from #3308


PastryIRL:
Any plans on getting this merged?

PR #3365, opened 2022-03-27
add matrix_info property in transform to give current matrix information

kyouryuukunn:
It is difficult to understand current matrixtransform and matrixcolor status directly and there is no manner to get them.
I added matrix_info properties in transform 

```
.. transform-property:: matrixcolor_info

    :type: list
    :default: None

    If not None, This gives list that includes a tuple of MatrixColor object
    and that parameters in the given order. This is just for informational
    purposes only and it makes no sense to change this property.


.. transform-property:: matrixtransform_info

    :type: list
    :default: None

    If not None, This gives list that includes a tuple of TransformMatrix
    object and that parameters in the given order. This is just for
    informational purposes only and it makes no sense to change this property.

```

example
```
    show sylvie green normal:
        perspective True
        matrixtransform ScaleMatrix(1, 1, 1)*RotateMatrix(0, 1, 0)*OffsetMatrix(0, 0, 0)
        linear 10 matrixtransform ScaleMatrix(1, 1, 1)*RotateMatrix(0, 1, 0)*OffsetMatrix(300, 0, 0)
        repeat

```
image


renpytom:
Why not just access state.matrixtransform.origin?

renpytom:
Fundamentally, I don't think that I'd want this to be a stable API. I'm fine with your tools accessing this information, since it's likely that that your tools will be updated when Ren'Py is. 

I don't want to make this a documented API that can be used by games, since doing so would mean that we'd have to commit to a particular implementation of ScaleMatrix, etc - and I don't think we want to do that, since it would limit how Ren'Py can change.

PR #3374, opened 2022-03-28
Changes to the crop_relative transform property

Gouvernathor:
It now takes two values, either "child" or "room", which select what crop should be relative to.
The former False value and the related behavior seem to me like a compat-only useful option, as if a creator wants elements of crop to be non-relative the proper way should be to use ints or to cast the values with ``absolute``. So I removed it from the doc (but the code still supports it). The `crop` documentation has been shortened accordingly.
I chose to document the type as only string, with True being now equivalent to "child".
This should probably get an entry in the changelog.

An argument could be made for the code in accelerator.pyx to set `state.crop_relative` to the default value at the same time as setting `crop_relative`. It would make subsequent passes faster, but I'm not sure it's worth it. If such behavior (where the state is updated) is implemented, accelerator.pyx arguably shouldn't be the place for it.

As usual, this code hasn't been tested because its main features are in a pyx file.


mal:
I'm not sure `child` and `room` are great for this, since `child` would (I think) actually mean relative to it's own size, and `room` seems too ambiguous and difficult to intuit. I don't have any amazing alternative suggestions at the moment, but still something it seems worth considering. _Maaaaybe_ `self` and `parent` but I'm not really sold on those either. :/

PR #3387, opened 2022-04-05
Sphinx-directory-traversal

geomat0101:
This fixes sphinx in-game project I/O issues related to relative paths when renpy.py is launched from its base directory instead of from within the sphinx subdirectory.  Older file I/O patterns (open/close) have been updated to their newer forms (with open() as)


geomat0101:
sorry this got tangled up with PR #3386

PR #3390, opened 2022-04-05
ignore MTLc3m if present

geomat0101:


renpytom:
What's the purpose of this? MTL3cm shouldn't be used anymore.

geomat0101:
https://github.com/renpy/renpy/blob/master/testcases/game/script.rpy#:~:text=JAPANESE%20%3D%20%22../../launcher/game/fonts/MTLc3m.ttf%22

sorry if this is irrelevant / extra / not particularly useful, I've just been going through the basics in an effort to get better acquainted with the project and submitting a couple things as i come across them

renpytom:
Ah, gotcha. I think the problem is really that some older projects use MTLc3m, when the main stuff has changed fonts. So it probably makes sense to change that to uses of SourceHanSans.

geomat0101:
right.  will fix and resubmit, thanks

PR #3397, opened 2022-04-08
Document #2462

Gouvernathor:
document #2462

### DIFF
--- a/sphinx/source/atl.rst
+++ b/sphinx/source/atl.rst
@@ -993,9 +993,12 @@ both horizontal and vertical positions.
    :type: None or string
    :default: None
 
-   If not None, causes the displayable to be sized according to the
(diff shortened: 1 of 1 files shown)


mal:
This behaviour is incidental, not explicit, as it relies on the behaviour of other properties. How things act in the absence of the `fit` setting isn't a property of fit, but rather of whatever other properties are defined. The "if not None" caveat is already used several times in docs to allow for this sort of thing.

mal:
Perhaps something like?
```
   If not None, the displayable will be sized according to the
   table below. If neither ``xsize`` nor ``ysize`` are set, they
   will take their values from containing area.

   In this context "dimensions" refers to one or more of ``xsize``
   and ``ysize`` when they are not None.
```

PR #3419, opened 2022-04-16
Update 00console.rpy

ji123345:
config.console=True


renpytom:
I don't think this is a correct default.

Issue #3342, opened 2022-03-20
Open files in a lazy manner.


renpytom:
Most files that Ren'Py opens aren't actually used by Ren'Py. In many cases, Ren'Py opens a file, only to pass it to SDL2, which reads the filename and re-opens that file as an SDL_RWops. 

What makes sense is to create a LazyFile object, that doesn't actually open the file until the first read. In the usual case that this is just being used to create an RWops, that open doesn't need to happen at all, and so won't hit the disk.

PR #3423, opened 2022-04-18
loader: use RWops object from pygame_sdl2

uyjulian:
Depends on https://github.com/renpy/pygame_sdl2/pull/135  
A possible solution to https://github.com/renpy/renpy/issues/3342

### DIFF
--- a/renpy/loader.py
+++ b/renpy/loader.py
@@ -34,6 +34,8 @@ import re
 import io
 import unicodedata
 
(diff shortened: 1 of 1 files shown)


renpytom:
At least in theory, I like the idea of this. I'll want to look into it more, though, and I think it's an intrusive change, so this should probably wait for early in the 7.6/8.1 cycle.

renpytom:
Merged, thank you.

Issue #2645, opened 2021-02-04
[Feature request] Stop all sounds & music if window is minimized


Vadim-Karpenko:
Hello everyone,
Is there a way to stop playing all the sounds from the game if the window is minimized and then continue from the same spot when game is maximized again?

Did try to find how this can be achieved, but no luck yet. Any help would be appreciated

renpytom:
This should be controlled by a preference that defaults to False.

renpytom:
@jsakaa asked to take this on as part of a university project.

Andykl:
Fixed by #3431.

PR #3431, opened 2022-04-19
Pause audio when window is minimized

jsakaa:
Attempted implementation of #2645. Governed by a preference; should pause all audio tracks when the game window is minimized and resume them when it is maximized by default.

### DIFF
--- a/renpy/common/00preferences.rpy
+++ b/renpy/common/00preferences.rpy
@@ -218,6 +218,9 @@ init -1500 python:
          * Preference("high contrast text", "toggle") - Toggles high contrast text.
 
 
+         * Preference("audio when minimized", "enable") - Enable sounds playing when the window is not in focus.
+         * Preference("audio when minimized", "disable") - Disable sounds playing when the window is not in focus.
+         * Preference("audio when minimized", "toggle") - Toggle sounds playing when the window is not in focus.
 
          Values that can be used with bars are:
 
@@ -499,6 +502,14 @@ init -1500 python:
                 elif value == "toggle":
                     return [ ToggleField(_preferences, "high_contrast"), _DisplayReset() ]
 
+            elif name == _("audio when minimized"):
+
+                if value == "enable":
+                    return SetField(_preferences, "audio_when_minimized", True)
+                elif value == "disable":
+                    return SetField(_preferences, "audio_when_minimized", False)
+                elif value == "toggle":
+                    return SetField(_preferences, "audio_when_minimized")
 
             mixer_names = {
                 "music" : "music",
--- a/renpy/display/core.py
+++ b/renpy/display/core.py
@@ -4177,7 +4177,17 @@ class Interface(object):
 
                     if ev.state & 2:
                         self.keyboard_focused = ev.gain
-
+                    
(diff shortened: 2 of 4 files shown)


Booplicate:
Let's say developer (or user via a button action) paused music. Would changing focus unpause music? Looking at the code I think it would. But I'd say it's not expected - we shouldn't unpause music which was intentionally paused before.

Gouvernathor:
If parts of the code need to be re-done for the reasons stated above (and that's a big if), and if there are no technical advantages to pausing, I would slightly prefer the sound be muted rather than paused.
One reason would be that fullscreen games work more that way, in my experience. Another would be that ATL and screen timers and time-taking elements of renpy won't be paused in the meantime.

Booplicate:
Actually, I agree, I'd also prefer the sound to be muted than paused. If I recall correctly there's also one special channel (`audio`?) which can't be paused either?

jsakaa:
> Let's say developer (or user via a button action) paused music. Would changing focus unpause music?

I had the same concern, but it shouldn't as it is currently. 
It's a little ambiguously written but the boolean that governs whether the audio should unpause upon changing focus only becomes True if there was a matching minimization of the window beforehand. I should probably give that value a better name, since 'audio_paused' is somewhat misleading in hindsight.
So that is to say that just changing mouse or keyboard focus for instance shouldn't trigger audio to unpause.

> But I'd say it's not expected - we shouldn't unpause music which was intentionally paused before.

I don't believe it would, since this sort of mirrors the functionality of the 'check_suspend()' function earlier in the file for Android. I brought this worry up to Tom over Discord and that was the conclusion he came to as well.

I do also agree that muting would be preferable to pausing, though, since that does model the behavior of most games in my experience as well. I'll try to see to making that change.

Gouvernathor:
Just to dig in on that risk of mistakenly unpausing things, and this may be an unnecessary concern - what if I have a paused channel, and another upaused channel, and I minimize and reopen the game ? It will leave the fist channel as it was, and unpause the second channel ?

jsakaa:
> It will leave the fist channel as it was, and unpause the second channel ?

I believe that would be the behavior, looking at this snippet of relevant code from 'periodic()' in the 'channel' class in 'audio.py'.

As per my understanding,
after 'pause_all()' is called, I think that for the channel that was already paused by the developer/user, 'self.context.pause' would be True, as would be 'global_pause', making 'want_pause' equal to True as well. Then, 'self.paused' would equal 'want_pause' meaning the condition is not met and that this already paused channel would remain paused.
After, when the user maximizes the window and 'unpause_all()' is called, 'self.context.pause' would still be true even if 'global_pause' is now false, meaning that 'want_pause' would be equal to True. 'self.paused' would still be equal to True as it was not changed earlier, meaning that once again the condition is not met and the channel is not unpaused.

Though please let me know if I've had a misunderstanding! 

```
want_pause = self.context.pause or global_pause

if self.paused != want_pause:

            if want_pause:
                self.pause()
            else:
                self.unpause()

            self.paused = want_pause
```

renpytom:
This looks good to me - I'm merging it now. Thank you for the contribution.

PR #3433, opened 2022-04-20
Document how the fixed properties work

Gouvernathor:
That is, when offer_screen isn't a thing.

### DIFF
--- a/sphinx/source/style_properties.rst
+++ b/sphinx/source/style_properties.rst
@@ -1061,11 +1061,11 @@ These are used with the fixed layout.
     bottom side of the bottommost child of the fixed.
 
 Note that children with relative positions or variable sizes can
-induce surprising behavior with boxes, when these properties are
-applied. This is because the children are placed and sized as if
+induce surprising behavior with Fixeds, when these properties are
+applied. This is because the children are first placed and sized as if
 these properties didn't exist, and only then, from the way the children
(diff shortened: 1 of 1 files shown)


sphinx/source/style_properties.rst
```diff
@@ -1060,6 +1060,13 @@ These are used with the fixed layout.
     If True, the size of the fixed layout is shrunk vertically to match the
     bottom side of the bottommost child of the fixed.
 
+Note that children with relative positions or variable sizes can
+induce surprising behavior with Fixeds, when these properties are
+applied. This is because the children are first placed and sized as if
```
renpytom, line 1065: I don't like the wording of this. Everything is surprising if you don't understand it. It makes sense to document this, but we don't need to say it's surprising.

mal:
I've read this several times, and still am not clear on what this is seeking to clarify. It doesn't tell the reader anything actionable or otherwise of use, and seems to simply amount to "this may not work how you think", which is entirely subjective and circles back to Tom's point.

Gouvernathor:
Describing how things work is the purpose of documentation, and in that goal I find it safer to presume that someone doesn't know something rather than the opposite, as it results to more and clearer information available.
This clarifies the reason why offer_screen was added (without saying it), and what happens without it.
Somewhere there's a contradiction though : you can't say that this information is unnecessary for certain users, and that it's an implementation detail at the same time.

PR #3438, opened 2022-04-21
Rewrite docs for Creator-Defined Statements

jsfehler:
This PR makes the following changes to the CDS docs:

- Move the example section to the top of the page. This established a hierarchy for the information from broad to increasing in specific detail.
- Adds return values to the Lexer class docs.
- Change Lexer method docs from a passive voice ("Returns the apple") to an active voice ("Return the apple")


renpytom:
Merged, looks great, thank you.

Issue #3439, opened 2022-04-21
Some atl transforms are broken


Gouvernathor:
The following works in the last Release, but not in the nightlies (since as soon as 7.5.0.22040201n) :
```rpy
transform chromatic_focus(child, chzoom=1.01, center=(.5, .5)):
    child

label start:
    pause
    show eileen at chromatic_focus(chzoom=1.02)
    pause
```
The error is this :
```
Full traceback:
  File "game/script.rpy", line 37, in script call
    call shader_test
  File "game/nightly.rpy", line 270, in script
    pause
  File "renpy/ast.py", line 2203, in execute
    self.call("execute")
  File "renpy/ast.py", line 2191, in call
    return renpy.statements.call(method, parsed, *args, **kwargs)
  File "renpy/statements.py", line 281, in call
    return method(parsed, *args, **kwargs)
  File "renpy/common/000statements.rpy", line 455, in execute_pause
    renpy.pause()
  File "renpy/exports.py", line 1565, in pause
    rv = renpy.ui.interact(mouse='pause', type='pause', roll_forward=roll_forward, pause=delay)
  File "renpy/ui.py", line 299, in interact
    rv = renpy.game.interface.interact(roll_forward=roll_forward, **kwargs)
  File "renpy/display/core.py", line 3344, in interact
    repeat, rv = self.interact_core(preloads=preloads, trans_pause=trans_pause, pause=pause, pause_start=pause_start, **kwargs)
  File "renpy/display/core.py", line 3764, in interact_core
    root_widget.visit_all(lambda i : i.per_interact())
  File "renpy/display/core.py", line 575, in visit_all
    d.visit_all(callback, seen)
  File "renpy/display/core.py", line 575, in visit_all
    d.visit_all(callback, seen)
  File "renpy/display/core.py", line 575, in visit_all
    d.visit_all(callback, seen)
  File "renpy/display/core.py", line 565, in visit_all
    for d in self.visit():
  File "renpy/atl.py", line 606, in visit
    block = self.compile()
  File "renpy/atl.py", line 513, in compile
    block = self.atl.compile(self.context)
  File "renpy/atl.py", line 713, in compile
    statements = [ i.compile(ctx) for i in self.statements ]
  File "renpy/atl.py", line 961, in compile
    child = ctx.eval(expr)
  File "renpy/atl.py", line 260, in eval
    return eval(expr, renpy.store.__dict__, self.context) # @UndefinedVariable
  File "<string>", line 2, in <module>
[...]

PR #3440, opened 2022-04-21
Restore positional parameter info

mal:
Fix for #3439 via small targeted reversion of #3370.


Gouvernathor:
I'll test that tonight (unless you already did), but while we're at editing 3370, wouldn't it be better to put empty tuples in EMPTY_PARAMETERS instead of empty lists ? That way we're more impervious to unforeseen mutations in the future.
I'm guessing changing the type wouldn't be a problem since an empty dict did the same as an empty list, at the place you edited.

PR #3450, opened 2022-04-25
Clarify and add internal reference for automatic image definition

Gouvernathor:


renpytom:
This isn't right. Auto-defined images are the same thing as other images at this level.

Gouvernathor:
That's true, but the rules of auto-defined images aren't the same as for the manually defined ones. I think I was the one who added "or to a filename" and it isn't correct.
Lowercasing, and the replacement of underscores (and dashs ?) with spaces are one of the differences, and that's what I wanted to reference. Also the manually-defined images take precedence, which is why I put them on a line before.

For a different phrasing, we could add "... This includes (ref)automatically-defined images from the image directory.".

renpytom:
From this perspective, images are the same no matter how much it's defined. I do think it's fine to mention auto-defined images, but they should be called out as something overly special.

PR #3464, opened 2022-04-30
Subclass functools.partial for better performance

mal:
Builds on @Gouvernathor's work in #3246 to address some of the compatibility concerns and provide a clean migration path when deserialising old save data, without introducing further dependency on the old Curry class.

To help give confidence in the unpickling approach, a small demonstration of pickles being created with the old code, and unpickled with the new code (in py2, and py3) can be found here: https://gist.github.com/mal/65d2a5aa3dc22480b712b2088c452282. 

Relevant notes from commit messages:

> The underlying functools.partial doesn't have an \_\_init__ method, so the
> super call bounces up to object which doesn't take arguments. This is
> fine in normal use but presented issues in some pickling scenarios.
> 
> Since we can be slightly more efficient without it, and it avoids the
> edge cases with pickle, nuke it.

> [Migration] is done by having [Curry]'s \_\_new__ method return a Partial instance
> created with a placeholder callable (hash), and rebound setstate method
> capable of converting the state of the old Curry instance into the
> format that functools.partial expects.
> 
> Any attempt to create a new instance of the Curry class directly will
> result in an instance of Partial, avoiding further propagation of the
> deprecated class.
> 
> As all of the extra work for compatibility only occurs when attempting
> to deserialise Curry instances, we don't impact performance of the new
> Partial class, or compromise it's API into the future.

### DIFF
--- a/renpy/curry.py
+++ b/renpy/curry.py
@@ -47,6 +47,7 @@ class Curry(object):
         return rv
 
     def __setstate__(self, state):
+        state.pop('hash', None)
         state = (state.pop('callable'), state.pop('args'), state.pop('kwargs'), state)
         return super(Partial, self).__setstate__(state)
 
--- a/renpy/curry.py
+++ b/renpy/curry.py
@@ -27,47 +27,25 @@ import functools
 
 
 class Curry(object):
-    # essentialy the same as Partial, kept for compatibility
+    """
+    :undocumented:
 
-    hash = None
+    Compatibilty class that exists solely to deserialise instances from
+    old saves into instances of the replacement Partial type below.
+    """
 
-    def __init__(self, callable, *args, **kwargs):  # @ReservedAssignment
-        self.callable = callable
-        self.args = args
-        self.kwargs = kwargs
-        self.__doc__ = getattr(self.callable, "__doc__", None)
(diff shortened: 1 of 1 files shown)


renpy/curry.py
```diff
@@ -23,54 +23,68 @@
...
+            _hash = hash(self.func) ^ hash(self.args)
 
-        if self.hash is None:
-            self.hash = hash(self.callable) ^ hash(self.args)
+            for i in self.keywords.items():
+                _hash ^= hash(i)
 
-            for i in self.kwargs.items():
-                self.hash ^= hash(i)
+            setattr(self, 'hash', _hash)
```
Gouvernathor, line 85:
```suggestion
            self.hash = _hash
```

mal: Wrote it this way to have it mirror the `getattr` used the start of the function which I find helps in recognising that the same thing being accessed/written. Also ends up having (nominally) better performance.

mal:
Closing in favour of https://github.com/renpy/renpy/pull/3246, with https://github.com/renpy/renpy/pull/3464/commits/dbc0fe93c1fd9b282a44c5205b5a54c6c26cd11a to be re-submitted as a standalone PR.

PR #3466, opened 2022-05-01
Make OpenDirectory public and compat it in launcher

### LINKED ISSUE #3458: Make OpenDirectory() available.
An usecase could be a button that opens the screenshot directory. The player could be clueless or bothered to search it.

I'm using the action from the launcher and it works fine.

Andykl:
This fixes #3458

### DIFF
--- a/launcher/game/android.rpy
+++ b/launcher/game/android.rpy
@@ -283,7 +283,8 @@ init python:
                     pass
 
             if opendir:
-                store.OpenDirectory(dir_to_open)()
+                dir_to_open = os.path.join(p.path, dir_to_open)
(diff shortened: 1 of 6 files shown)


mal:
lgtm, two minor queries:
- All `renpy.run` does for a single action is to call it, so why introduce the extra function hops, is there some extra magic to it?
- Are the in-function imports just to avoid leaking the modules into `store`? If so, I believe it may be possible to omit/hoist the `os` ones, as `os` is [already in `store`](https://github.com/renpy/renpy/blob/69e6ffd09d40208a59c1aee72ad4029988488537/renpy/defaultstore.py#L466).

PR #3483, opened 2022-05-06
Add Master volume

kyouryuukunn:
I added master volume to Preference.

     * Preference("master volume", 0.5) - Set the adjustment applied to all channels.
     * Preference("master volume")
     * Preference("master mute", "enable") - Mute master.
     * Preference("master mute", "disable") - Un-mute master.  
     * Preference("master mute", "toggle") - Toggle master.

### DIFF
--- a/renpy/audio/audio.py
+++ b/renpy/audio/audio.py
@@ -418,12 +418,13 @@ class Channel(object):
         # Update the channel volume.
 
         mixer_volume = renpy.game.preferences.volumes.get(self.mixer, 1.0)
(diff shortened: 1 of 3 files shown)


renpytom:
I merged this in as 694d6e82a762f44cbbba02708e10580937dfe799, renaming it from master to main. The weird commit is that it was easier to edit the diff than to merge, then edit.

(I may want to limit the uses of the word 'master' in Ren'Py. That's a longer-term thing, but for now I want to avoid adding more uses of it.)

PR #3484, opened 2022-05-06
add config.show_volume_in_mute

kyouryuukunn:
Currently, volume bars show that value as 0 when it is mute and changing the bar disables mute.
I added config.show_volume_in_mute.
If False, the default, the behavior isn't changed. Otherwise, volumes are shown and adjustable while keeping mute when it is mute.


renpytom:
Can you change it to config.preserve_volume_when_muted ?

kyouryuukunn:
I changed.

renpytom:
Thanks. I'll merge this once I finish what I'm working on.

PR #3491, opened 2022-05-12
fix (tutorial): Correct typo in dialogue

jacobgkau:
Just noticed this while going through the tutorial. I'm not sure if the second "to" is actually needed, but I am fairly sure the first one is supposed to be "do."


renpytom:
Thanks for the fix.

PR #3494, opened 2022-05-12
fix (tutorial): Correct typos

jacobgkau:
I found a few more typos after submitting https://github.com/renpy/renpy/pull/3491. I waited until I was done with the rest of the tutorial so they could all be handled in the same commit. They're all in the dialogue except one that's in an example's comment (just an extra space between two words).

### DIFF
--- a/tutorial/game/tutorial_screen_displayables.rpy
+++ b/tutorial/game/tutorial_screen_displayables.rpy
@@ -173,7 +173,7 @@ label add_displayable:
                 xalign 0.5 ypos 50
                 add "images/logo base.png"
 
-    e "An image can also be referred to by it's filename, relative to the game directory."
+    e "An image can also be referred to by its filename, relative to the game directory."
 
     example large:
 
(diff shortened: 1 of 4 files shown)


tutorial/game/tutorial_screen_displayables.rpy
```diff
@@ -470,7 +470,7 @@ label button_displayables:
                         text _("Heal") style "button_text" yalign 0.5
                         bar value AnimatedValue(health, 100, 1.0) yalign 0.5 xsize 200
 
-    e "A button takes another displayable as children. Since that child can be a layout, it can takes as many children as you want."
+    e "A button takes another displayable as children. Since that child can be a layout, it can take as many children as you want."
```
Gouvernathor, line 473: "another" (singular) also doesn't go well with "children" (plural)

jacobgkau: Agreed, I was only changing the most blatant errors since I didn't intend to actually modify any of the content, but I've added that fix with 63ba851. I turned it singular in the first sentence since the next sentence refers to "that child" (singular).

renpytom:
Committed as cc293a3e7 , as it  was easier to edit the diff to add translation ids that add them after a normal merge. Thank you!

PR #3509, opened 2022-05-15
SChinese update for new Ren'Py

zedraxlo:


renpytom:
Is this not supposed to go in?

zedraxlo:
> Is this not supposed to go in?

I accidentally changed the line break from LF to CRLF, which caused us to not be able to see the difference.
I think I should re-upload it once using LF to see the difference.

renpytom:
Ah, makes sense. Thanks. Let me know when it's ready and I'll merge it.

zedraxlo:
> Ah, makes sense. Thanks. Let me know when it's ready and I'll merge it.

New one ready. Changes can now be displayed.

Issue #3508, opened 2022-05-15
crop_relative should default to False for integers


loafylemon:
Version: 7.5.0.22051404n

In the following example, I'm making a square with four colours (RGBY), then attempting to crop it with corner properties such that just the yellow square is visible, however once corner properties are given, the entire displayable disappears.

```renpy
image sqr = Solid("#f00", xysize=(256, 256))
image sqg = Solid("#0f0", xysize=(512, 256))
image sqb = Solid("#00f", xysize=(256, 512))
image sqy = Solid("#ff0", xysize=(512, 512))

image square = Fixed("sqy", "sqb", "sqg", "sqr", xysize=(512, 512))

screen square():
    add "square":
        corner1 (256, 256)
        corner2 (512, 512)

    textbutton "Exit" action Return()

label start:
    call screen square
```

Booplicate:
Apparently this:
```renpy
screen square():
    add "square":
        crop_relative False
        corner1 (256, 256)
        corner2 (512, 512)

    textbutton "Exit" action Return()
```
works.
And this:
```renpy
screen square():
    add "square":
        corner1 (0.5, 0.5)
        corner2 (512, 512)

    textbutton "Exit" action Return()
```
also works (although the documentation says `corner` must be `None` or `int`).

Since you didn't specify `crop_relative`, it takes the value from `config.crop_relative_default` which defaults to `True`, so it tries to do relative cropping.

loafylemon:
I must have missed the change that defaulted `config.crop_relative_default` to `True`. Thank you!

Booplicate:
Note, this might be a bug, unsure.

renpytom:
This looks like an issue for me, as crop_relative_default really shouldn't come into play when everything appears to be an integer.

PR #3510, opened 2022-05-15
Set crop_relative to False for integers.

loafylemon:
Attempts fixing #3508

### DIFF
--- a/renpy/display/accelerator.pyx
+++ b/renpy/display/accelerator.pyx
@@ -265,6 +265,7 @@ def transform_render(self, widtho, heighto, st, at):
     yo = 0
 
     crop = state.crop
+    crop_relative = state.crop_relative
 
     # Cropping.
     if (state.corner1 is not None) and (crop is None) and (state.corner2 is not None):
@@ -282,9 +283,10 @@ def transform_render(self, widtho, heighto, st, at):
 
         crop = (x1, y1, x2-x1, y2-y1)
 
-    if crop is not None:
+        if crop_relative is None:
+            crop_relative = any(isinstance(i, float) for i in crop)
 
-        crop_relative = state.crop_relative
+    if crop is not None:
 
         if crop_relative is None:
             crop_relative = renpy.config.crop_relative_default


Booplicate:
I think there's some other unknown issue, and by fixing it this way you might just hide it.

Technically speaking, the `relative` function defined below should handle `int` and `absolute`. And it even works for the `crop` property, but not for the `corner`.

I think we should focus on why `crop (10, 10, 100, 100)` works, but `corner1 (10, 10)` + `corner2 (100, 100)` doesn't. Those should give the same result (unless I misunderstood).

Unsure if any of the following has any effect on this bug, but:

- Kind of weird the code does `crop = (x1, y1, x2-x1, y2-y1)` (reduces width/height) for corners, but not for crop?
- Also there's this code in `transform.py`:
```py
add_property("corner1", (float, float), None)
add_property("corner2", (float, float), None)
```
Maybe it should be changed to `(position, position)`?

renpytom:
I'm working on this myself.

Gouvernathor:
The fix in this PR is incorrect for the reasons stated above. I'll try to answer your points in order @Booplicate .

- The corners are documented as taking ints, not positions, so making them respond to relative coordinate would be a feature request. A legit one, in my view, but it's not a required fix for `corner1 (.5, .5)` to take a quarter of the child.
- corner2 is documented as being the coordinates of the bottom-right corner of the kept area, while crop[2:4] is the *size*, in pixels, of the cropped area. Not the same. So your example is inexact : `crop (10, 20, 100, 150)` should equate to `corner1 (10, 20)`, `corner2 (120, 170)`. Unless *I* misunderstood.
- The first code you quote is accurate because of what I just said, but the second one, with the floats, may be the answer to our problem. Will test it asap (not now). Feel free to lodge it as a PR (replacing them with `position`, might as well kill two birds with one stone).

renpytom:
Obsoleted by 1de4027d7b5b481238c8f59c32dba36dcac04e20 .

PR #3518, opened 2022-05-17
Translate language names in their specific languages in the launcher

### LINKED ISSUE #3517: Translate language names in launcher.
I think the list of languages should have each language in its own language instead of a list of languages in English, similar to the list of languages available on wikipedia.

Gouvernathor:
fixes #3517

### DIFF
--- a/launcher/game/new_project.rpy
+++ b/launcher/game/new_project.rpy
@@ -74,7 +74,7 @@ label new_project:
                 cancel=Jump("front_page"),
                 )
         else:
-            new_project_language = renpy.translate_string(_preferences.language or "English")
+            new_project_language = __("{#language name and font}")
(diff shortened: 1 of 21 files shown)


renpytom:
Also, we probably wouldn't have the translation of "Japanese" -> "日本語" already, etc. This is a bigger project than it looks.

Gouvernathor:
Two issues remain.
First, I need to put the
```py
translate None strings:
    old "{#language name and font}"
    new "English"
```
block somewhere, and there's no tl/None directory. At the moment it's in the preferences.rpy file.
Second, when some language's translation is missing, and when the None translation is here file for example), the empty string isn't displayed, it's replaced with "English", the None translation. I'm not sure why that happens, but it can be misleading.

PR #3534, opened 2022-05-18
Tweak substitution

Gouvernathor:
- tweak 1ba50f11da46e6f3519f0dbb8a9419cd4df18588 to still disallow "[var!!]" the way "[var!]" is rejected, and to allow the `not conversion` optimization when there's multiple bangs
- tweak 461bddf7f3dd8396daf2b20a7825dde2720583b3 to apply capitalize, not upper, on the first character (see https://docs.python.org/3/library/stdtypes.html#str.capitalize as for why it's different)

### DIFF
--- a/renpy/substitutions.py
+++ b/renpy/substitutions.py
@@ -175,10 +175,12 @@ class Formatter(string.Formatter):
         if conversion is None:
             return value
 
+        conversion = conversion.replace("!", "")
+
         if not conversion:
             raise ValueError("Conversion specifier can't be empty.")
 
-        if set(conversion) - set("rstqulci!"):
+        if set(conversion) - set("rstqulci"):
             raise ValueError("Unknown symbols in conversion specifier, this must use only the \"rstqulci\".")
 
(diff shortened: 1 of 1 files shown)


mal:
The changes here seem reasonable and should probably be merged as is, but we should note the following corner-cases:
- `[var!!]` would now raise where it didn't previously.
- Applying `c` to a string with an already uppercased leading digraph would lowercase the second half of it.
  (I'm guessing that digraphs were also the motivation for this?)

If either of those would rise to the level of requiring compat, then I'd flip to being more in favour of avoiding this change altogether - neither of these tweaks would justify the added complexity.

mal:
After further thought, I think that the `!!` part probably shouldn't be merged. It's aimed at optimising and supporting misuse which doesn't seem justified here as it hold no benefit for the typical/expected/supported case.

renpytom:
I simplified the second part of this. The first part felt a little too risky to me, so I didn't merge it - it seems like it might break things with no benefit.

PR #3538, opened 2022-05-19
Revert some launcher languages' font size to normal

Gouvernathor:
One outlier : Indonesian uses the DejaVuSans font override, and a .9 text size decrease. Using DejaVuSans without the size decrease results in text overflows.
However, I'm not convinced Indonesian actually needs a font override. Malay doesn't have a font override, and it's a very similar language.


Gouvernathor:
Closing this, to be checked and added by actual malay and japanese speakers.
@kyouryuukunn if you see this, could you try reproducing 1930abc and see if there's any problem somewhere in the launcher ?

PR #3547, opened 2022-05-21
Remove creating screens with python and most ui functions from the documentation

Gouvernathor:


sphinx/source/screen_python.rst
```diff
@@ -2,127 +2,15 @@
...
-* ui.hbox
-* ui.side
-* ui.vbox
-* ui.imagemap
-* ui.draggroup
-
-There are a few UI functions that do not correspond to screen language
-statements, as they correspond to concepts that are not present in the
-screen language.
+There are a few UI functions, which are not equivalents to statements
```
renpytom, line 12: This feels like a non-sequitur, as it explains what the UI functions are not, rather than what they are.

PR #3551, opened 2022-05-21
Implement show/call/hide screen expression

Booplicate:
I've seen quite a few people asking how to call/show a screen using an expression, they often try to do it using the syntax close to `call expression` for labels. I also had to use `renpy.call_screen` when I only had a variable with the screen name.

While there's a way to do it directly using Python, I think having a way to do it from RenPy which is intuitive (since the syntax is very similar to `call expression`) will be beneficial.

At first I only wanted `call screen expression` and `show screen expression`, but I think having `hide screen expression` won't do harm.

### DIFF
--- a/renpy/common/000statements.rpy
+++ b/renpy/common/000statements.rpy
@@ -468,13 +468,48 @@ python early hide:
     # Should we predict screens?
     config.predict_screen_statements = True
 
+    def _get_screen_name(p):
+        """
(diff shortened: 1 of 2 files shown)


renpytom:
This looks basically good, but the documentation needs to be updated to match the change.

Booplicate:
@renpytom, @Gouvernathor I added some basic documentation, feel free to edit.

PR #3556, opened 2022-05-21
Allow to omit `audio/` when specifying the path to the sound.

### LINKED ISSUE #2077: Iconsistent use of images and audio directory
There is an inconsistency in the use of the images and audio directory. For the images directory the 'images' part of the directory doesn't need to be specified in the path, however for the 'audio' directory it's mandatory to be specified in the path otherwise it can't find the file.

images path: image bg office1="story/office/spy_office1.png"
audio path: play sound "audio/soundfx/telephone_ring.ogg"

Andykl:
This fix #2077.

### DIFF
--- a/renpy/config.py
+++ b/renpy/config.py
@@ -645,7 +645,7 @@ after_replay_callback = None
 wrap_shown_transforms = True
 
 # A list of prefixes Ren'Py will search for assets.
-search_prefixes = [ "", "images/" ]
+search_prefixes = [ "", "images/", "audio/" ]
 
 # Should Ren'Py clear the database of code lines?
 clear_lines = True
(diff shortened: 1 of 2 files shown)


Gouvernathor:
Perhaps this should be separated in two : have audio be searched for in audio/, and images in images/.
It would require a bit of internal rework, and most likely two additional config variables (be they documented or not). And "images" to be removed from config.search_prefixes.

Also, when "images" was added to search_prefixes, that change was compated in 00compat, so regardless of what we do to config.search_prefixes, it should probably get a compat.

renpytom:
I've taken a slightly different approach to this in  b6d51b7b0bffb5ab1b08deeadfbea9cad273ceda  .

PR #3566, opened 2022-05-23
Add Update Actions

Booplicate:
I've seen a few cases where people have tried to change a variable using the `SetVariable` action (or similar), but since the value is being evaluated during creation of the action, the variable will only change once.
Example:
```renpy
screen MyScreen():
    default my_variable = 0

    text "My variable: [my_variable]"
    textbutton "Increment variable":
        # This won't work
        action SetScreenVariable("my_variable", my_variable + 1)
```
Usually it's not a big deal, you just create a custom function and change your value there. It's not too much trouble for a global variable, but to do it with a screen variable, the developer has to access the screen scope, which is a quite advanced thing to do.

With my changes one will be able to do:
```renpy
screen MyScreen():
    default my_variable = 0

    text "My variable: [my_variable]"
    textbutton "Increment variable":
        # Works!
        action UpdateScreenVariable("my_variable", lambda i: i+1)
```
which I think is very elegant and easy to do.

The following code allows to test all new actions:
```renpy
screen TestNewUpdateActions():
    default var_a = 0

    vbox:
        anchor (0.5, 0.5)
        pos (0.5, 0.5)

        text "screen var a: [var_a]"
        textbutton "Update a":
            action UpdateScreenVariable("var_a", lambda i: i+1)

        use _TestUsedScreen

        text "global var c: [var_c]"
        textbutton "Update c":
            action UpdateVariable("var_c", lambda i: i+1)

        text "global dict d: {}".format(store.var_d).replace("{", "{{")
        textbutton "Update d":
            action UpdateDict(store.var_d, "my_key", lambda i: i+1)

        text "global object field e: [var_e]"
        textbutton "Update e":
            action UpdateField(store.var_e, "field", lambda i: i+1)

        textbutton "Exit":
            action Return()

screen _TestUsedScreen():
    default var_b = 0

    text "used screen var b: [var_b]"
    textbutton "Update b":
        action UpdateLocalVariable("var_b", lambda i: i+1)

init python:
    class TestObj():
        def __init__(self, field=0):
            self.field = field
        def __repr__(self):
            return "{}(field={})".format(
                type(self).__name__,
                self.field
            )

default var_c = 0
default var_d = {"my_key": 0}
default var_e = TestObj()

label main_menu:
    return

label start:
    $ renpy.retain_after_load()
    call screen TestNewUpdateActions()
    return
```

### DIFF
--- a/renpy/common/00action_data.rpy
+++ b/renpy/common/00action_data.rpy
@@ -21,7 +21,7 @@
 
 init -1600 python:
 
-   ##########################################################################
+    ##########################################################################
     # Functions that set variables or fields.
 
     __FieldNotFound = object()
@@ -52,13 +52,13 @@ init -1600 python:
     @renpy.pure
     class SetField(Action, FieldEquality):
         """
-         :doc: data_action
-         :args: (object, field, value)
+        :doc: data_action
+        :args: (object, field, value)
 
-         Causes the a field on an object to be set to a given value.
-         `object` is the object, `field` is a string giving the name of the
-         field to set, and `value` is the value to set it to.
-         """
+        Causes the a field on an object to be set to a given value.
+        `object` is the object, `field` is a string giving the name of the
+        field to set, and `value` is the value to set it to.
+        """
 
         identity_fields = [ "object" ]
         equality_fields = [ "field", "value" ]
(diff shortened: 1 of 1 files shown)


mal:
Your first example (despite the comment) already appears to work, so I'm a little confused as to what's being addressed with this change. Moreover, passing lambdas into actions is generally a bad idea as they cannot be serialised and so are likely to lead to problems when saving.

Booplicate:
> Your first example (despite the comment) already appears to work

Hmm last time I tried the action didn't get updated with the next value to set.

renpytom:
Unless something is going wrong, the first example should work. Lambdas can't be loaded or saved, so we don't want to encourage them.

Booplicate:
Yep, tried and it seems to work now, unsure what happened before then.

PR #3573, opened 2022-05-24
Rmove bare excepts from the repository - 2

Gouvernathor:
I'm not too sure about the C file though, maybe it should be excluded.

### DIFF
--- a/module/tinyfiledialogs/tinyfiledialogs.c
+++ b/module/tinyfiledialogs/tinyfiledialogs.c
@@ -1136,7 +1136,7 @@ int tinyfd_notifyPopupW(
         size_t lDialogStringLen;
 
         if (aTitle && !wcscmp(aTitle, L"tinyfd_query")) { strcpy(tinyfd_response, "windows_wchar"); return 1; }
-        
+
         if (quoteDetectedW(aTitle)) return tinyfd_notifyPopupW(L"INVALID TITLE WITH QUOTES", aMessage, aIconType);
 		if (quoteDetectedW(aMessage)) return tinyfd_notifyPopupW(aTitle, L"INVALID MESSAGE WITH QUOTES", aIconType);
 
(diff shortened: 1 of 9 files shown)


module/tinyfiledialogs/tinyfiledialogs.c
```diff
@@ -4053,7 +4053,7 @@ static int tkinter2Present(void)
 	static int lTkinter2Present = -1 ;
 	char lPythonCommand[256];
 	char lPythonParams[128] =
-		"-S -c \"try:\n\timport Tkinter;\nexcept:\n\tprint 0;\"";
+		"-S -c \"try:\n\timport Tkinter;\nexcept Exception:\n\tprint 0;\"";
```
Gouvernathor, line 4056: this old-style `print 0` is very strange too

Numerlor:
Do these need to be broad excepts? While better they can still be a PITA.
Most of them look like they could be catching OSError/NotImplementedError, maybe with some handling on switching it if Python 2 raises something different in certain cases

Gouvernathor:
I tried that once in #3189, just once, and it was a mistake that broke the android port, so I'm not doing it again.

Numerlor:
> I tried that once in #3189, just once, and it was a mistake that broke the android port (fixed in [0ac623b](https://github.com/renpy/renpy/commit/0ac623bb7eaad727fa219ef8d53d8c45ba5bf9fd)), so I'm not doing it again.

While I'm not sure on the OS functions (though I think the OSError and not implemented should be most of what's needed), blocks like these https://github.com/renpy/renpy/blob/e36ac33d407a5e1732a87c0873bfdd9486a69547/renpy/common/00steam.rpy#L772-L776 definitely shouldn't need broad excepts , or if they do that needs to be addressed in the code that raises them

Numerlor:
> I don't see much of a benefit to precise error handling - it require more mental effort to try and figure out what errors can happen, and it many to all of these cases, it's not clear what benefit that mental effort brings, since we're mostly trying to guard against the "something didn't work" case.

I'd think that a bit more time spent on figuring out what errors to catch would be worth it to prevent issues in the future. A broad except will catch errors that can aid during development and hide them, and specifying the errors will clearly state to the reader which errors it is supposed to be handling and aren't caught just because it's broad even if the handling may be unwanted or could go a different path

renpytom:
I understand that argument and also disagree with it.

PR #3575, opened 2022-05-24
Implement `ToggleFocus`

Booplicate:
This is a shortcut to
```renpy
If(
    GetFocusRect("my_focus"),
    ClearFocus("my_focus"),
    CaptureFocus("my_focus")
)
```
So one can do
```renpy
ToggleFocus("my_focus")
```

### DIFF
--- a/renpy/common/00action_other.rpy
+++ b/renpy/common/00action_other.rpy
@@ -739,7 +739,7 @@ init -1500 python:
         """
         :doc: focus_action
 
-        If the focus exist, clears it, otherwise captures it.
+        If the focus exists, clears it, otherwise captures it.
         """
 
         def __init__(self, name="default"):
--- a/renpy/common/00action_other.rpy
+++ b/renpy/common/00action_other.rpy
@@ -734,6 +734,28 @@ init -1500 python:
             renpy.clear_capture_focus(self.name)
             renpy.restart_interaction()
 
+    @renpy.pure
+    class ToggleFocus(Action, DictEquality):
+        """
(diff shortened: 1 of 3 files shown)


renpytom:
Do you see this as having a use? I'm not immediately seeing it, but I'd be happy to be wrong.

renpy/common/00action_other.rpy
```diff
@@ -734,6 +734,28 @@ init -1500 python:
...
+        def __call__(self):
+            name = self.name
+
+            if renpy.get_focus_rect(name) is not None:
+                renpy.clear_capture_focus(name)
+
+            else:
+                renpy.capture_focus(name)
+
+            renpy.restart_interaction()
```
mal, line 757: Based purely on the existing actions, does this only need to happen when clearing, rather than always?

mal: Ah, I was reading the `GetFocusRect` action. Nevermind. /o\

renpytom:
exist -> exists, please. 

Also, if you can add this to the changelog, I'd appreciate it.

Gouvernathor:
If the given use-case is correct, then maybe the said example from the docs should be using ToggleFocus ?

Booplicate:
I wasn't sure, for a simple example on how use `nearrect` it does its job. But I can update it if needed!

Booplicate:
Done.

renpytom:
99f8cdd74 Adds this, without the documentation changes, which I don't think reflect how I want things used.

PR #3576, opened 2022-05-24
Add warning about names interference between shaders

Gouvernathor:


sphinx/source/model.rst
```diff
@@ -215,6 +221,11 @@ The custom shader can then be applied using a transform::
 
     show eileen happy at gradient
 
+As stated before, the ``gradient_done`` variable from the exemple.gradient shader
```
renpytom, line 224: exemple -> example

PR #3579, opened 2022-05-24
NearRect fixes

Booplicate:
This fixes 2 problems, documents another, and there's potentially one more, but I'm not sure about it.

The first issue is `self.offsets = [ None ]` causes a crash, that's because it's being unpacked as a tuple in the event method. Since it's `None`, it can't be unpacked. I set it to `[(0, 0)]`, but maybe it should be set to an empty list?

The second issue is `if self.focus`, `focus` is a method of the parent class (`Displayable`), that `if` check will always be evaluated to `True`. I think it was intended to be `if self.focus_rect`.

The third issue is using `focus` in the `NearRect` constructor. The `Displayable` class it inherits has a parameter named `focus`, but it seems to be doing something else. In the parent class `focus` sets `focus_name`, in `NearRect` it sets `focus_rect`. Right now it's impossible to pass `focus` to the parent class. This looks wrong to me and I'd rename the parameter while this version has not been released, but it could be done intentionally, so I'm just bringing your attention to this.

Found another issue, `NearRect` won't ever be redrawn. If it returned an empty render, it will always be rendered "empty." I tried to check if it the rect has changed since last time in `per_interact` and redraw if it has, but that makes using transforms impossible (since we just redraw into empty render instantly). So that does not work.
Update: I added that to the docs and also added an exception if `config.developer` is `True` and we have no focus rect to render to.

### DIFF
--- a/renpy/display/layout.py
+++ b/renpy/display/layout.py
@@ -2292,6 +2292,9 @@ class NearRect(Container):
             rect = self.parent_rect
 
         if rect is None:
+            if renpy.config.developer:
+                raise Exception("NearRect is being displayed without focus rect.")
+
             self.offsets = [ (0, 0) ] # type: ignore
 
             return rv
--- a/renpy/display/layout.py
+++ b/renpy/display/layout.py
@@ -2286,7 +2286,7 @@ class NearRect(Container):
 
         rv = renpy.display.render.Render(width, height)
 
-        if self.focus:
+        if self.focus_rect:
             rect = renpy.display.focus.get_focus_rect(self.focus_rect)
         else:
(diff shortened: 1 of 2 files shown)


Booplicate:
Okay, fixed another thing. Still not sure what to do with the `focus` parameter.

renpytom:
Let me have a go at this, since NearRect is designed to show/hide using focus. I'll pull in some of the technology underlying ShowIf to make sure it does things correctly.

Booplicate:
Okay, I'll leave this open as a reference for you until you fix it (don't forget about the other 3 issues).

PR #3589, opened 2022-05-26
Link Language and other scattered actions to the main list of actions

Gouvernathor:


Gouvernathor:
Other documented Actions could/should be part of this : Replay, EndReplay, SetPreference, TogglePreference, StylePreference, and the [voice actions](https://www.renpy.org/dev-doc/html/voice.html?highlight=setvoicemute#voice-actions).

Purchase and Restore (which are related to in-app purchases), Sync (related to achievements), Update (related to games' self-updates) and ui.ChoiceJump and ui.ChoiceReturn (related to blocking rollback) should probably not be included in the general list.

Issue #3577, opened 2022-05-24
Bring the show/call/hide screen up to date with their statement equivalents


Gouvernathor:
The [renpy.show_screen function](https://www.renpy.org/doc/html/screen_python.html?highlight=show_screen#renpy.show_screen) takes parameters to set the zorder and the layer the screen will be shown at, and some other properties (_tag, _widget_properties and _transient).
The renpy.call_screen function is documented not to pass the corresponding parameters to the screen (because they start with an underscore), but it's unsaid whether the parameters have the same effect as in show_screen.
renpy.hide_screen only takes a parameter for the layer - and it's consistent since the zorder doesn't matter when hiding a screen.

The show screen, call screen and hide screen don't include any of these controls. `show screen skreen(_layer="master")` works, but is undocumented (I guess it also works with zorder, but it's just as undocumented). There's no other way of then hiding skreen with hide screen. The only ways I found were `$ renpy.hide_screen("skreen", layer="master")`, and `hide skreen` but it probably shouldn't be documented as working that way.

There should be a `layer` clause for the hide screen statement (or rather an `onlayer` clause, to match the simple hide statement).
There should be the same and a `zorder` clause for the show screen and call screen statements.
There should be an `as` clause for the show screen and call screen statements, matching the `_tag` parameter of the renpy.show_screen function, ~~and arguably _widget_properties and _transient equivalents~~.

mal:
I'd be tempted to view those `_`-prefixed args working as an implementation detail, with the preferred method being to set those values inside the screen itself with the appropriate keyword. I'm not sure that complicating the basic show/call/hide screen statements with them to support hypothetical(?) cases (has this ever come up in the wild?) would be all that advantageous. Meanwhile anyone requiring the additional power can still use things as they are with acceptable caveats.

renpytom:
I can see a case for as/onlayer/zorder (the first for all three, the last two for just call/show screen.) 

_transient and _widget_properties seem very python-specific for me, and so I don't think those need clauses.

Gouvernathor:
The Show, ShowTransient, ToggleScreen and Hide actions (which apply to screens) should also be part of this update.

The _layer kwarg is already taken into account by all four, but only documented for Hide,

Show and ShowTransient would pass any kwarg related to zorder or to `as`/_tag to the renpy.show_screen function, but the code doesn't explicitly support it, nor is it documented as such.
ToggleScreen and Hide don't support it, but ToggleScreen probably should.

[...]

PR #3595, opened 2022-05-28
New screen clauses

Booplicate:
Attempt to implement #3577
Allows to do:
```renpy
show screen screen_name as screen_tag
show screen some_overlay onlayer overlay zorder 15
```
and similar.

### DIFF
--- a/renpy/common/000statements.rpy
+++ b/renpy/common/000statements.rpy
@@ -678,7 +678,7 @@ python early hide:
             renpy.with_statement(eval(transition_expr))
 
 
-    def lint_screen(p):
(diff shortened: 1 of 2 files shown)


Gouvernathor:
This should get documented in the related section of screens.rst, similar to the phrasing of other clauses from that file, or from the simple show/hide statements, and ideally referencing them.
I can do it, if you want.

Booplicate:
> This should get documented in the related section of screens.rst, similar to the phrasing of other clauses from that file, or from the simple show/hide statements, and ideally referencing them. I can do it, if you want.

I'd appreciate if you do it!

PR #3605, opened 2022-05-31
Updated the channel selection screen.

Neyunse:
The channels are separated into "Release, Prerelease and Nightly", this is based on the preferences screen.


renpytom:
Thank you for this, but I don't think this is a change I want to add to Ren'Py. 

The reason is that right now, apart from translations, creating a new channel is something I do entirely server-side. For example, when I created the Ren'Py 8 channel, I didn't have to update Ren'Py 7.4 to let people upgrade to it. 

So, my basic instinct is that there isn't really a need for organization like this, especially since the Ren'Py 7 releases will end after a year or two.

Neyunse:
Oh, I get it. This occurred to me more to separate between "launch, pre-release and nightly" (outside of what are 7/8 versions).
I'll close this...

PR #3609, opened 2022-06-03
add poi, orientation and x,y,z rotate

kyouryuukunn:
I added point of interest, orientation and x,y,zrotate transform properties like AfterEffect.

I think matrixtransform is flexible but complex.
It is difficult to rotate the camera or an image keeping that coordinates.
It is impossible to change an element separately. For example, changing x rotate during looping changing z rotate.

I added below properties like AfterEffect.

.. transform-property:: poi

    :type: (float, float, float)
    :default: None

    This gives the position of the point of interest. the camera or an image
    are rotated then face that point evenif that pos or zpos are changed.

.. transform-property:: orientation

    :type: (float, float, float)
    :default: None

    This rotates an image or the camera without change these position. Only the
    camera rotates in z, y, x order for convenience.

    The difference between orientation and xrotate, yrotate or zrotate are the
    way of interpolation.  orientation rotates through the shortest root.

.. transform-property:: xrotate

    :type: float
    :default: None

    This rotates an image or the camera without change these position. Only the
    camera rotates in z, y, x order for convenience.

The difference between orientation and x,y,zrotate are same as ones of AfterEffect.
https://www.youtube.com/watch?v=8Nn19xx37Mw&t=251s

I uploaded sample movie. This is on VirtualBox and perfomance isn't goold.
https://youtu.be/XBPUa33O0hU

This still has two issues.

A. Rotating system in Ren'Py are x, y, z order, but rotating stage in x, y, z order is equals to rotate the camera in z, y, x order. Z rotation are not around the center of screen when rotating camera in x, y, z order . It is inconvenient. There are two resolutions.
1. only the camera rotates in z, y, x order.
2. Change rotating system from x, y, z order to z, y, x order. It is an imcompatible change.
I select 1 in this pull request.

 B. For using 2D images in 3D space, they are desirable to be possible to point the camera at the image and point an image at the camera. I wanted to allow poi property to take "camera" and image tag, but these coordinates aren't known until calling transform_render. I tried saving it in Tranform and TransformState instances in transform_render, but these seem to be overwritten. And I don't know the way to get the own image tag and layer inside transform_render. How should I resolve this?


Booplicate:
Just my two cents.

Judging by the video, I don't see any new functionality added, pretty sure you can do all that already with `matrixtransform`:

https://user-images.githubusercontent.com/53382877/171866746-d7a75794-2977-40c7-86df-86bedcc0e893.mp4

> I think matrixtransform is flexible but complex.

They are harder than just using basic `xpos` or `anchor` or `rotate`, but not *too* complex. I believe adding more ways of doing the same thing would make RenPy more complex (e.g. look at C#). Developers would have to question themselves, should they use this or that? What's the difference? What's better? Can you use both at once? Or with other transform properties?
Then it's also more code for you to maintain later on.

> It is difficult to rotate the camera or an image keeping that coordinates.
It is impossible to change an element separately. For example, changing x rotate during looping changing z rotate.

I don't think I've had any of those problems with rotation. The first sounds like a problem with the wrong anchor. The second, unsure, but I'm able to rotate by x, y, and z both separately and simultaneously.

kyouryuukunn:
Sorry, I uploaded new movie.
https://youtu.be/4_hfJsTwMWY

Yes, x, y, zrotate is like using OffsetMatrix and RotateMatrix in matrixtransform. The advantage is that the user does not have to consider matrix configuration and the current zpos, especially when zpos is not fixed.

Setting x, y, zrotate of the camera to (a, b, c) is equal to set matrixtransform of it to OffsetMatrix(0, 0, config.perspective[1]+zpos)*RotateMatrix(a, b, c)*OffsetMatrix(0, 0, -config.perspective[1]-zpos)

    camera:
        zpos camera_zpos
        xrotate 20 yrotate 30 zrotate 45
    camera:
        zpos camera_zpos
        xrotate 20 yrotate 30 zrotate 45
        matrixtransform OffsetMatrix(0, 0, config.perspective[1]+camera_zpos)*RotateMatrix(-20, -30, -45)*OffsetMatrix(0, 0, -config.perspective[1]-camera_zpos)

Setting x, y, zrotate of the image to (a, b, c) is equal to set matrixtransform of it to OffsetMatrix(0, 0, zpos)*RotateMatrix(a, b, c)*OffsetMatrix(0, 0, -zpos).

    show sylvie:
        zpos zpos
        xrotate 60 yrotate 40 zrotate 20
    show sylvie:
        zpos zpos
        matrixtransform OffsetMatrix(0, 0, zpos)*RotateMatrix(60, 40, 20)*OffsetMatrix(0, 0, -zpos)

>>It is impossible to change an element separately. For example, changing x rotate during looping changing z rotate.

This means codes like below.

    show sylvie:
        parallel:
            yrotate 0
            linear 3 yrotate 360
            repeat
        parallel:
            linear 1 xrotate 10
            ....


Orientation cannot be substituted for "matrixtransform" because it rotates by the shortest path, not by interpolation of arguments.
Poi allows to specify the orientation of the camera and image in coordinates rather than angles. This is intended to be used when the image is pointed at the camera or when the camera is moving while facing a specific target.


Below is new movie script.rpy.

    label start:
        camera:
            perspective (0, 843, 100000)

        show sylvie green normal at left
        with fade
        "image poi sample"
        "This points the image to that point."
        show sylvie:
            poi (640, 370, 300)
        "set poi of sylvie to (640, 370, 100)"
        show sylvie:
            linear 3 poi (0, 370, 100)
        "set poi of sylvie to (0, 370, 100)"
        show sylvie:
            linear 3 poi (0, 0, 100)
        "set poi of sylvie to (0, 0, 100)"
        show sylvie:
            linear 3 poi (1280, 0, 100)
        "set poi of sylvie to (1280, 0, 100)"
        show sylvie:
            linear 3 poi (1280, 760, 100)
        "set poi of sylvie to (1280, 760, 100)"
        show sylvie:
            linear 3 poi (0, 760, 100)
        "set poi of sylvie to (0, 760, 100)"
        show sylvie:
            linear 3 poi (0, 760, -100)
        "set poi of sylvie to (0, 760, -100)"
        show sylvie:
            linear 3 poi (1280, 760, -100)
        "set poi of sylvie to (1280, 760, -100)"
        show sylvie:
            linear 3 poi (1280, 0, -100)
        "set poi of sylvie to (1280, 0, -100)"
        show sylvie:
            linear 3 poi (0, 0, -100)
        "set poi of sylvie to (0, 0, -100)"
        show sylvie:
            linear 3 poi (0, 0, 100)
        "set poi of sylvie to (0, 0, 100)"
        show sylvie:
            linear 3 poi (640, 370, 100)
        "set poi of sylvie to (640, 370, 100)"
        show sylvie:
            poi (640, 360, 843)
            xpos -0.1
            linear 3 xpos 0.9
            linear 3 xpos -0.1
            repeat
        "Setting poi of the image to the point of camera means the image faces camera always."
        show sylvie:
            poi None
            pos (.6, 1.0)
            zpos -300

        "image rotate test"
        "This rotates the image keeping that position."
        show sylvie:
            linear 1 zrotate 20
        "set zrotate of sylvie to 20"
        show sylvie:
            linear 1 yrotate 40
        "set yrotate of sylvie to 40"
        show sylvie:
            linear 1 xrotate 60
        "set xrotate of sylvie to 60"
        show sylvie:
            xrotate 0 yrotate 0 zrotate 0
            matrixtransform OffsetMatrix(0, 0, -300)*RotateMatrix(60, 40, 20)*OffsetMatrix(0, 0, 300)
        "Setting x, y, zrotate to (a, b, c) of the image is equal to set matrixtransform OffsetMatrix(0, 0, zpos)*RotateMatrix(a, b, c)*OffsetMatrix(0, 0, -zpos)"

        show sylvie:
            matrixtransform None
        "The difference between rotation and orientation is the way to interpolate. orientation rotates through the shortest root."
        show sylvie:
            orientation (0, 0, 0)
            xrotate 0 yrotate 0 zrotate 0
            linear 3 xrotate 300 yrotate 200 zrotate 150
        "set x, y, zrotation to (300, 200, 150)"
        show sylvie:
            orientation (0, 0, 0)
            xrotate 0 yrotate 0 zrotate 0
            linear 3 orientation (300, 200, 150)
        "set orientation to (300, 200, 150)"



        hide sylvie
        scene bg lecturehall
        show sylvie green normal at left as green:
            zpos 100
        show sylvie blue normal at default as blue:
            xpos 0.7
            zpos 300
        with fade
        "camera poi sample"
        "This points the camera to that point."
        camera:
            poi (640, 360, 0)
        "set poi of camera to (640, 360, 0)"
        camera:
            linear 1 poi (0, 360, 0)
        "set poi of camera to (0, 360, 0)"
        camera:
            linear 1 poi (0, 0, 0)
        "set poi of camera to (0, 0, 0)"
        camera:
            linear 1 poi (1280, 0, 0)
        "set poi of camera to (1280, 0, 0)"
        camera:
            linear 1 poi (640, 360, 0)
        "set poi of camera to (640, 360, 0)"
        camera:
            poi (640, 360, 0)
            linear 3 pos (-1.0, -1.0)
            linear 3 pos (1.0, -1.0)
            linear 3 pos (1.0, 1.0)
            linear 3 pos (-1.0, 1.0)
            linear 3 zpos -843*2
            linear 3 pos (1.0, -1.0)
            linear 3 zpos 0
            linear 3 pos (.0, .0)
        "Setting poi of the camera also means the camera faces that point even if it moves."

        camera:
            poi None
            pos (0., 0.)
        "camera rotate test"
        "This rotates the camera keeping that position."
        camera:
            linear 3 xrotate 20
        "set xrotate of camera to 20"
        camera:
            linear 3 yrotate 30
        "set yrotate of camera to 30"
        camera:
            linear 3 zrotate 45
        "set zrotate of camera to 45"
        camera:
            xrotate 0 yrotate 0 zrotate 0
            matrixtransform OffsetMatrix(0, 0, 843)*RotateMatrix(-20, -30, -45)*OffsetMatrix(0, 0, -843)
        "Setting x, y, zrotate of the camera to (a, b, c) is equal to set matrixtransform OffsetMatrix(0, 0, config.perspective\[1]+zpos)*RotateMatrix(a, b, c)*OffsetMatrix(0, 0, -config.perspective\[1]-zpos)"

        return

renpytom:
One of the purposes of Ren'Py is to make life easier on creators. Insofar as we can improve Ren'Py to do that, and that's a change that we understand and can live with long term, we should make that change. Specifically, I don't think that matrixtransform is a complete solution - it's more of a tool to let us experiment with what can be done on the 3d state. 

I like the idea of this, but I don't think the primitives that are supplied are, necessarily, the right ones. Specifically, I think that there might be operations that we want to do that aren't supported by the poi/orientation/x/y/zrotate. 

One example of this assumes that Ren'Py can display a 3d cube, something I'd like to add in a future release. (My eventual goal for this would be to load in gltf models, so people can have 3d scenery in their game - people are making 3d backgrounds now, and I'd like to be able to move the camera through them.) 

I want to pick a corner of the cube, find the vector from the center of that cube to the corner, and rotate 360 degrees around that vector, then repeat. I don't think there's a good way to express this in this PR/proposal. 

I'm tired now, so I'm not sure what the answer is, but my feeling is it has to be extensible (by us, and by the creator), so that when Ren'Py can show more things, more transforms are possible. 

I'm off to bed now, will consider this more in the morning.

kyouryuukunn:
Thank you.

I chose these properties because AfterEffect which has been used for many years has them so I figured they were a no-brainer. I figured that I could use these properties for simple operations and matrixtransform for more advanced ones.

PR #3614, opened 2022-06-04
Update updater documentation

Moshibit:
As of Ren'Py 7.4.9 it is possible  to download updates over https.


renpytom:
I think the current form of the documentation is correct. Https works, but it doesn't patch the game, just re-downloads it.

Moshibit:
ok

Issue #3622, opened 2022-06-07
Improve ATL `Context.eval`


Andykl:
It uses `renpy.python.escape_unicode` and `eval`. Given that this is almost always `PyExpr` then `renpy.python.py_eval` can speed things up by caching the result.

PR #3630, opened 2022-06-08
atl: `Context.eval` now uses `renpy.python.py_eval` instead of Python `eval`

Andykl:
This makes any expression used in ATL cached in bytecode.rpyb


Andykl:
This fixes #3622

PR #3642, opened 2022-06-10
fix doc perspective defaut to None

kyouryuukunn:
In document, default perspective is False, but actually None.

### DIFF
--- a/sphinx/source/3dstage.rst
+++ b/sphinx/source/3dstage.rst
@@ -316,7 +316,7 @@ The following transform properties are used by the 3D Stage.
 .. transform-property:: perspective
 
     :type: True or False or Float or (Float, Float, Float)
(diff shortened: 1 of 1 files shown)


Gouvernathor:
It looks like False and None are treated the same way internally:
https://github.com/renpy/renpy/blob/131a9b810aacf87c256289cc99706bb8c62a59e4/renpy/display/accelerator.pyx#L176-L177
So I'm not sure whether it matters if it's documented or not. Did you encounter a bug where having that information in the doc would have helped you ?

kyouryuukunn:
I checked it in 8.0.0.22060901 and the results are different.

camera:
    perspective False xpos 0.5

result is
image

camera:
    perspective None xpos 0.5

result is
image

PR #3644, opened 2022-06-11
doc: Document the init block

### LINKED ISSUE #3639: init statement and init phase are not documented
Init statement used to be documented in Quickstart section, but now it is absent from documentation (only `init python` is documented), only broken links on forums and entry in historic wiki remains.

Likewise there are multiple references to "init time" and "init phase", but no explanation provided (before it used to be shortly explained in init statement documentation, explanation of order of init blocks execution currently present in `init python` documentation)

Gouvernathor:
fix #3639

Some ways this could be improved :
- The interaction between init offset and init blocks could be left uncharted, and it could be said just to use one of them per file (only init offsets, or only init blocks). But init python blocks are kinda unavoidable, and some people have taken the habit of using init offset (since it was the only one documented), so... perhaps better this way ?
- The "priority" value works upside-down : blocks with a higher "priority" are run after blocks with a lesser one, that's absurd. Given that the word "priority" is only used in the doc, it's not a property name or keyword or something, perhaps it should be renamed to something like ["niceness"](https://manpages.ubuntu.com/manpages/trusty/man1/nice.1.html) or "politeness" ?
- Find a link to the original init-block section of the doc, to make old links go to the new location (not much hope since it's not the same file as before, but well)


Moshibit:
I found in the changelog that some statements like image and define can be used with init, I don't know if this is the case in modern Ren'Py as this is only documented in the version 6 changelog. https://www.renpy. org/doc/html/changelog6.html#syntax-changes.

Gouvernathor:
Init can contain define, default, image, style, label, call, jump, and probably screen, layeredimage and transform too. What it *can't* contain (like say, show, with or play statements) are the exception rather than the rule, so I'd rather not try to list any of these two lists exhaustively.

Moshibit:
Perhaps maintaining code readability is the reason init was removed from the documentation. It is true that this use does not seem adequate, it makes more sense to use init offset to put something like` init 3 define name = value` to each statement

Gouvernathor:
To be honest `init offset` seems to be the biggest obstacle to code readability : it has effect on code placed under it, but that code is not visibly bundled in a block, like it is with the init block. In the inline version of the init block, `init 5 define ` for example, the effect is visibly and explicitly applied to one single line.
So, if we're to resurrect the inline init, I would advocate de-documenting init offset.

Booplicate:
`init offset` is indeed a bad one, very hard to follow what it affects.
Although, unsure if `init 5 define` is needed since you can do `define 5 my_var = "value"`, which does the same (but maybe the former is better because it's more explicit).

mal:
I'm wondering if this discussion is potentially riding the same train of thought that lead to it becoming undocumented, I guess Tom will know for sure.

The `init offset` statement is useful as a file header when the entire contents of a file should have a non-default, but uniform, priority (indeed that's how Ren'Py itself [uses it](https://github.com/renpy/renpy/blob/782df7d83f5e79704e4d8eb1d55f80228afaf19e/gui/game/screens.rpy)); and any odd statements in other files should probably explicitly apply priority directly, as in the case noted above regarding `define 5 my_var`.

After that, besides `init python`, `init` doesn't seem to add much beyond, as has also been noted, the potential to make things harder to read.

Just my opinion based on reading this thread, and my own (infrequent) use of `init`, but it might be worth checking with Tom before expending more effort writing docs for something that _might_ be a consciously undocumented legacy syntax. 🤷

Gouvernathor:
In that case, maybe adding to the doc that the init offset statements should only be used like this : once at the top of the file.
But then, you can't have different things defined at different init priorities in the same file 🤔.

I don't like the `define 5 var` statement because it's undocumented, and because `init 5 define var` is a more clear way to do the same thing, due to its resemblence with the other init things.
But if we document that (for define and default for example), the resulting code could seem weird : why can we have a block with python but not with define ?
```py
init 5 python:
    class A: pass

    class B(A): pass

init 5 define me = A()
init 5 define her = B()
```

I don't know if I'm clear enough : I understand how one could choose to undocument the init blocks, but I think that it's not the best way to go, documentation-wise, and that it doesn't result in the clearest code.
The weird way init offset and the init blocks interact (with the offsets summing up instead of replacing one another) are definitely a good argument not to have both being documented.

JSearUK:
Is there any way (in the far, far future, perhaps) to have Ren'Py attempt to handle the ordering of things itself? For example, by  making sure that anything that *appears* to have not been defined yet is searched for and, if found, automatically assigned a priority that enables it to work? Or - classes could have `early` attached to them if later code tries to put it in the persistent store, in much the same way that `from` clauses can be added to calls.
The point, after all, is to make all this as easy for the novelist as possible, and the less they have to worry about the wider an audience you'll enable. I certainly wouldn't want to go back to manually releasing memory, for example - the garbage collector does that for me. I don't have to worry too much about explicit typing, either; Python and Ren'Py abstract most of that. Why not this, too?

Booplicate:
> I don't know if I'm clear enough : I understand how one could choose to undocument the init blocks

I don't use `init i define x = z` or `define i x = z` often (actually, never used the former), but I do use init blocks, so I'd choose them over the rest if you had too keep one.

> For example, by making sure that anything that appears to have not been defined yet is searched for and, if found, automatically assigned a priority that enables it to work? Or - classes could have early attached to them if later code tries to put it in the persistent store

This would be very bad imo, it's already easy enough and most of the time you don't have to manage init order anyway. If you're working on a more complex project, I think it's fair you have to adjust init order yourself. What you suggest looks like a huge crutch which would attempt to fix messed up init order. If you know how to use classes, I think it's ok to assume you can handle their definition.

Gouvernathor:
Trying to have Ren'py fix code by itself makes its behavior far more complex and hard to understand and to predict. Thus, bugs are much harder to understand (because all easy bugs are fixed automatically). Ambiguous code risks being "fixed" into one way rather than another for no understandable reason, adding to the complexity of the system.

JSearUK:
In that case - since I *have* seen people query why their variable is not defined when "It's right there!" - perhaps the relevant error message can be adjusted to state which init level it was at at the time, and that said definition needs to have occurred in a lower one?

I agree that self-correction is bad for any number of reasons (though it would be awesome if it worked); I mentioned it only as a notion with regard to perhaps entirely eliminating the need for documentation regarding `init` and `init offset`. I use blocks rather than offset myself.

> why can we have a block with python but not with define ?

Why, indeed? I like the idea of batch defaulting and defining!

Moshibit:
I like to use init offset as the header of my files, for me it is clear, all statements have a priority of zero, their priority is only added to the one indicated by init ofsset. One of the good features of Python and Ren'Py is that you can have different solutions for the same problem, the important thing is that the creator or the development team agree on how they write their code.

According to the documentation, the previous code should look like this, because define has its equivalent in init python,
`
init 5 python:
    class A: pass

    class B(A): pass

    me = A()
    her = B()
`
for me it is more understandable like this. each one looks for the way in which he can better read his own code.

That's why I don't think it's necessary for Ren'Py to try to tidy up the creator's code, or that we have to write a style guide and best practices.

Gouvernathor:
Closing this, because a good use of init offset is better than a regular use of init blocks (Moshibit is right) and essentially makes them not required. Also, a chunk of this (including the part about default vs. define trigger times) was added in the lifecycle page.

PR #3656, opened 2022-06-13
log: Revert to using None to denote unset dirs

mal:
This once again avoids attempting to open a log file before the relevant directories have been set.

Alternative to #3655 which reverts a small part of 69ca7cc44c19ca493f6776a1934e1348f5f5de02 which introduced the issue.


renpytom:
I did my own version of this in 86576cf07 .

PR #3658, opened 2022-06-14
copypaste in renpy.input() just like the input in screens

Neyunse:
```
label start:
     code = renpy.input("Paste the secret code", copypaste=True)
     "The secret code is [code]"
```

### DIFF
--- a/renpy/exports.py
+++ b/renpy/exports.py
@@ -778,7 +778,7 @@ def scene(layer='master'):
     renpy.display.interface.ongoing_transition.pop(layer, None)
 
 
-def input(prompt, default='', allow=None, exclude='{}', length=None, with_none=None, pixel_width=None, screen="input", mask=None, **kwargs): # @ReservedAssignment
+def input(prompt, default='', allow=None, exclude='{}', length=None, with_none=None, pixel_width=None, screen="input", mask=None,copypaste=False, **kwargs): # @ReservedAssignment
     """
     :doc: input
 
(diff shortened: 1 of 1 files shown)


mal:
Seems to be missing the update to the `input` method signature?

Neyunse:
> Seems to be missing the update to the `input` method signature?

I saw this function calls the class where the argument is, I don't know where an updater should go, I assume it will be updated in class. But I'm not sure, in a test I did it worked.

Neyunse:
> > Seems to be missing the update to the `input` method signature?
> 
> I saw this function calls the class where the argument is, I don't know where an updater should go, I assume it will be updated in class. But I'm not sure, in a test I did it worked.

> It was a bit tricky to keep track of where the `input` was...

What I'm talking about is that I think it is updated here.

https://github.com/renpy/renpy/blob/2a66adb496f6b47786847325b930664a622a1448/renpy/display/behavior.py#L1359

And as I understand, this part sends the input properties to the screen.

https://github.com/renpy/renpy/blob/2a66adb496f6b47786847325b930664a622a1448/renpy/exports.py#L853

https://github.com/renpy/renpy/blob/2a66adb496f6b47786847325b930664a622a1448/renpy/display/screen.py#L1139

After that, I think it then calls the input class to pass it the arguments, but I'm not sure where.

renpytom:
Merged, thank you. I'm also going to default this to true.

PR #3663, opened 2022-06-15
atl: Add an insensitive event for buttons

mal:
Unsure if this fits with the longer-term vision for atl/styles/events/states, so if this is not "the way" it can just be nuked. Filed on the off chance that it makes sense and is more or less the direction we want to head in. May be one to kick to 7.6/8.1 if we want to think on it more.

---

Fires only when changing the `focusable` property to avoid event spam. Also refactors `Button.set_transform_event` to handle dispatching events to the child (if there is one) and prepending the `role` property automatically (reducing copy-paste).

I wasn't able to recreate a scenario that would lead to the button transitioning from `focusable = False` to `focusable = True`, but I've included the `idle` event there on the assumption that it is possible to trigger that code path.

### DIFF
--- a/renpy/display/behavior.py
+++ b/renpy/display/behavior.py
@@ -900,6 +900,12 @@ class Button(renpy.display.layout.Window):
 
         return rv
 
+    def set_transform_event(self, event):
+        super(Button, self).set_transform_event(self.role + event)
+
+        if self.child is not None:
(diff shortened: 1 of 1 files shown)


mal:
Functionality landed as part of addressing another issue in https://github.com/renpy/renpy/commit/a0f4ed2ca613176324a1e017a225ee17c8dbb256.

Issue #3671, opened 2022-06-17
Raise subclasses of Exception more often in internal code


Gouvernathor:
This is to be more informative about the error being raised, in addition to the message being given to the exception, i.e a TypeError or a ValueError are usually better and more relevent. Searching for `raise Exception` in the repo returns 456 occurences, I think we can do better.

Elevating the exception from one class to one subclass is safe for any existing try-catch, but the other way around (reverting one for example) can break a try/except.
This is less dangerous than modifying the classes taken in an `except`.

PR #3672, opened 2022-06-17
Use exception subclasses in `raise`s

Gouvernathor:
As per #3671 
This is just a first batch of what it could look like.
The commits sort the edits in different use-cases, basically one per subclass.

### DIFF
--- a/distribute.py
+++ b/distribute.py
@@ -153,7 +153,7 @@ def main():
         destination += "-" + args.variant
 
     if os.path.exists(os.path.join(destination, "checksums.txt")):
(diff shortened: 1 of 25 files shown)


renpytom:
This isn't something that I want for Ren'Py. I don't think it adds enough value to be worth the effort.

multimokia:
Out of curiosity, and with all due respect, why are we making it actively harder to debug an issue? If something is going wrong, knowing exactly what it is goes a _long_ way in debugging as it can reduce hours to a few minutes. Imo these changes would help significantly.

Issue #3292, opened 2022-02-15
[BUG] Transform doesn't update


Booplicate:
When using a transform with default parameters changing the transform's parameters outside of its function and using `update` doesn't update the transform. However, that works if the transform doesn't have default parameters.

```renpy
label test_transforms:
    python:
        transform_with_args = Transform("a", rotate=0.0, pos=(0.25, 0.5))

        transform_pure = Transform("b")
        transform_pure.rotate = 0.0
        transform_pure.pos = (0.75, 0.5)

    "Showing both transforms:"

    show expression transform_with_args zorder 15
    show expression transform_pure zorder 15

    "Trying to change the one with default args:"

    python:
        transform_with_args.rotate# 0.0
        transform_with_args.rotate = 45.0
        transform_with_args.update()# Nothing happens, why?
        transform_with_args.rotate# 0.0, why?
        # Obviously we also can't change pos since we passed in xpos and ypos:
        transform_with_args.pos = (0.25, 0.0)
        transform_with_args.update()# Nothing

    "The image on the left did not rotate nor moved. A bug?"
    "Trying to change the one without args:"

    python:
        transform_pure.rotate# 0.0
        transform_pure.rotate = 45.0
        transform_pure.update()# Worked!
        transform_pure.rotate# 45.0 as expected
        # And this, of course, works
        transform_pure.pos = (0.75, 0.0)
        transform_pure.update()# Moved

    "The image on the right rotated and moved. Works as expected."
    return
```

renpytom:
The default parameters are given to the transform every frame if present.

Booplicate:
I'm not sure if this is correct.
If I changed the code above a bit:
```renpy
    python:
        def _transform_func(transform, st, at):
            if st > 5.0:
                return None

            roto_speed = -10
            transform.rotate = at % 360 * roto_speed
            return 0.0

        transform_with_args = Transform("a", function=_transform_func, rotate=45.0, pos=(0.25, 0.5))
```
And this transform works as expected, the function updates it, even when it has default parameters. I understand *why* it happens, perhaps you're applying the default parameters and then the function? But that doesn't work with `update` because it just redraws the transforms and resets the parameters. This is very counter-intuitive, though, and probably should be documented (unless I missed it somewhere?). But I'd argue that `update` should *update* the transform. For example, my use case was:
```py
# I have multiple images defined like this
[...]

PR #3674, opened 2022-06-17
Clarify Transform's fields' allowed use

Gouvernathor:
The doc initially says that the fields of Transform corresponding to transform properties "may not be modified outside of the constructor", so after object creation.

This is contradicted by two things. First, the `function` argument, whose purpose is to mutate these attributes.

Second, the `update` method, which "should be called when a transform property field is updated outside of the callback method". Although the "callback method" wording is dubious (and I included a correction), it clearly implies the fields are modified outside of the constructor.

So I added a clarification listing these three cases as times when tprop fields can be set.

### DIFF
--- a/sphinx/source/trans_trans_python.rst
+++ b/sphinx/source/trans_trans_python.rst
@@ -53,7 +53,9 @@ The Python equivalent of an ATL transform is a Transform object.
 
     Additional keyword arguments are values that transform properties are set
     to. These transform properties are set each time the transform is drawn,
-    and so may not be changed outside the constructor.
+    and so may not be changed after the Transform object is created, unless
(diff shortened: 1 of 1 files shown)


Booplicate:
I reported that as a bug in #3292, `update` doesn't update default parameters. Apparently it's a feature.

> Basically, if you include a property as part of the Transform, it's fixed for the life of the transform. That's just how transforms are defined to work.

Unsure if even `function` should be able to update default parameters either per Tom's response. A bug?

Gouvernathor:
Oh, ok I understand. So, Transform property fields can be modified either in the `function` or before calling .update, EXCEPT for transform properties set by the constructor.

PR #3675, opened 2022-06-18
mofify alpha, zoom type in default and reset transform

kyouryuukunn:
The values of alpha and zoom properties are int In default & reset transform.
But these type should be float in the document.
https://www.renpy.org/doc/html/atl.html#transform-property-alpha


renpytom:
Let's hold this off for 8.1 and 7.6.

renpytom:
Merged, thank you.

PR #3676, opened 2022-06-18
Add lint warning for unreachable statements

Gouvernathor:
Two possible changes :
- the TranslateBlock node takes a block, and it could be added alongside Init and Translate as reaching containers, but I don't know what it's used for (yet), so I don't know
- who-less say statements are sometimes used as docstrings in cookbook files, so it could be considered not to warn about who-less Say nodes. That can be done very easily with a property on the Say node class. The question is, should we adapt lint not to flag bad but common practices, or should we adapt the parser to consider them as docstrings instead of say statements ?

### DIFF
--- a/renpy/ast.py
+++ b/renpy/ast.py
@@ -1755,7 +1755,8 @@ class Return(Node):
 
     __slots__ = [ 'expression']
 
-    should_be_reachable = True
+    # should_be_reachable = None
+    # there are implicit return statements in various places
 
     def __new__(cls, *args, **kwargs):
         self = Node.__new__(cls)
--- a/renpy/ast.py
+++ b/renpy/ast.py
@@ -2097,6 +2097,14 @@ class If(Node):
 
     should_be_reachable = True
 
+    @property
+    def block(self):
+        """
+        Aligns If with other containers for easier recursion.
+        Uses tuple instead of list to prevent the illusion of possible side-effects.
+        """
+        return tuple(n for (_c, nodlist) in self.entries for n in nodlist)
+
     def __init__(self, loc, entries):
         """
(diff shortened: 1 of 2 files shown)


mal:
I haven't looked into the reachability aspect itself yet, but I'd recommend encapsulating all the logic for this within `lint.py`. 

Rather than add `should_be_reachable` to the functional ast code, I'd suggest keeping a tuple of "should be reachable" types inside `lint.py` and making the check `isinstance(node, reachable_type)`. This would give one central point of reference to update/augment in future.

Similarly with the `If.block` property, I think it makes more sense to have that logic within a `lint.py` helper that could be extended in future, should it be required, without impact to the AST. Organisationally it also positions us better should the AST ever be refactored, as we can have confidence that things in AST are required functionality, while code in `lint.py` just needs to support any changes.

Gouvernathor:
I think I agree about your If proposition, but less about the first. Information about objects, related to their types, should be in their classes. If we have a simple list of nodes considered should-be-reachable in a boolean way, you lose the ability to fine-tune it within a given type using a property, what I said about Say nodes.
Maybe making an ABC then, just for custom instancechecks then ? But if we do that, we can't have user statements decide their own value for if they should be reachable or not (that's not currently implemented but it would become more complicated to add).
And in the other hand I understand your point about making ast.py feature-only.

In any case, before partially reimplementing it (or in a different way), I'd like to have Tom's confirmation that the basic thing is good, in its purpose and what it does.

renpytom:
As mentioned before, I don't believe this is a good approach compared to a proper reachability analysis. (Possibly one that leverages the statement prediction infrastructure.)

PR #3698, opened 2022-06-26
Fix CropMove transitions

Gouvernathor:
The reason behind this is interesting : for the crop and the position of the cropped layer (here, the one being appeared), the value for the beginning and the value for the end, both between 0 and 1, are interpolated relative to when we are within the transition, then that is multiplied by the with and heigth to render to. This obviously returns a float, which will very likely not have an integer value. In the case of this transition (as for many others), the float of the position and the float of the crop size add up to the total size (width or height).
These floats (those of the crop and those of the pos) are then clamped *down* to integers, using `int`. And that's where the bug happens. Say we have a width of 1920, and the x pos is 10.6, the x crop is then 1909.4, since they add up to 1920. But when both are clamped down with `int`, it gives 10 and 1909, which add up to 1919, **one (constant) pixel short**.

I suspect this is a dormant bug likely to happen any time a float representing a position is clamped down to an integer. That's not the case every time a float position is used, but it is done sometimes.
Interestingly, removing the `int` cast didn't solve the problem, I suspect the .subsurface and/or .blit methods do another cast under the hood.
Anyway, `round` works, because it rounds to the nearest integer, so in the above example 10.6 would be rounded to 11, solving the missing pixel problem.

### DIFF
--- a/renpy/display/transition.py
+++ b/renpy/display/transition.py
@@ -697,6 +697,10 @@ class AlphaDissolve(Transition):
         return rv
 
 
+def interpolate_tuple(t0, t1, time, scales):
+    return tuple([ round(s * (a * (1.0 - time) + b * time))
+                    for a, b, s in zip(t0, t1, scales) ])
+
 class CropMove(Transition):
     """
     :doc: transition function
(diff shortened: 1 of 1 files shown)


Booplicate:
`round` returns a float in Py2, so you might still need to cast to `int`.
Also what about the other transitions? Like `PushMove` below uses a similar function, but it defines its own version for some reason. Maybe `interpolate_tuple` could be moved to the outer scope and then be used by all the classes that need it? So the fix would apply to everything.

Gouvernathor:
`round` won't budge because we import a compat version at the top of all files, it works like the PY3 version and returns an int when called with one argument. And even if it did return a float, the function we pass the values to accept them.

PR #3699, opened 2022-06-26
add watch per-interact and watch real-time

kyouryuukunn:
I added two console commands.
watch per-interact: update traced expressions per interaction (default)
watch real-time: update traced expressions as soon as possible

### DIFF
--- a/renpy/common/00console.rpy
+++ b/renpy/common/00console.rpy
@@ -120,6 +120,7 @@ init -1500 python:
 
 default persistent._console_short = True
 default persistent._console_traced_short = True
+default persistent._console_traced_per_interact = True
 default persistent._console_unicode_escaping = False
 
 init -1500 python in _console:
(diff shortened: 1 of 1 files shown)


Booplicate:
The same can already be achieved with
```py
if config.developer:
    config.per_frame_screens.append("_trace_screen")
```

Gouvernathor:
Yes, but @kyouryuukunn's implementation has the advantage of enabling the switch (between real-time and per-interact) to be done without reloading the game, whereas config variables are not supposed to be changed after init.
Although if the per-interact mode is deemed not particularly useful by itself, that per-frame setting can be always enabled.

Andykl:
I actually always use in my projects that `per_frame_screens` thing. I've wanted to add that to RenPy to use by default for a long time, but kept forgetting. As this is only for developers, I don't think performance is a concern here.

Booplicate:
> has the advantage of enabling the switch

I guess if the point is this.

Myself:
- Sometimes don't care how often it updates
- Most of the time want it to be updated asap to see the change

I do understand some people might use heavy functions there, so they probably want those to be called only once per interaction, but if anything I'd just make that behaviour default.

renpytom:
This was obsoleted by #3734, as I think it isn't necessary to make this configurable.

PR #3703, opened 2022-06-26
Add missing field to Module node

Numerlor:
the field is required for it to compile in python 3.8+


renpytom:
Merged, and a good catch in a rarely-exercised code path. Thank you.

PR #3704, opened 2022-06-26
Enable fonts to be looked for in a specific store

Gouvernathor:
Works the same as for speakers in say statements, with the character store.


KiloOscarSix:
Will this work with the `font` property on `text` objects and others?

Gouvernathor:
Yep, it works

renpytom:
I don't believe that this is worth including when we already have config.font_replacement_map.

PR #3715, opened 2022-06-28
Change conditions for Movie._play being valid

Gouvernathor:
In other conditions (the at clause for layeredimages for example), the `play` value would have been turned into a singleton list if not already a list. However, considering that there's already two _play attributes (_play and _original_play) to manage, I think it would have been a bit too complicated. And, the decision to support and document Movie taking lists has not actually been taken, so...

Also, the question was raised of when to consider the argument as being valid or not, if it's a list containing valid and non-valid values. I chose to expedite it as if there's even one invalid, the whole bunch is considered tainted. Another solution such as using `filter` to have a list of only the loadable ones would require adapting the after_setstate function, and I didn't take that risk.

So only the renpy.loadable calls have been protected, other than that, Movie processes the argument just like it did before.

### DIFF
--- a/renpy/display/video.py
+++ b/renpy/display/video.py
@@ -328,7 +328,7 @@ class Movie(renpy.display.core.Displayable):
 
     def after_setstate(self):
         play = self._original_play or self._play
-        if (play is not None) and (not isinstance(play, list) and renpy.loader.loadable(play)
+        if (play is not None) and (isinstance(play, basestring) and renpy.loader.loadable(play)
                                    or all(renpy.loadable(p) for p in play)):
             self._original_play = self._play = play
         else:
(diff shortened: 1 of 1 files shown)


mal:
While I understand the reasons for choosing guards over coercion, I do think that perhaps it would be a nicer way to go, ideally mirroring the implementation of the underlying `music.play` and being done in `__init__` prior to setting either of the instance variables:
https://github.com/renpy/renpy/blob/f198e4b94e77d1781463bb380d980501437be41e/renpy/audio/music.py#L94-L95

If that's not possible, however, then I'd at least suggest trading `not isinstance(play, list)` for `isinstance(play, basestring)` so as to allow through other types of sequence, as in `music.play`.

Your inclination towards treating the argument as invalid if anything in the sequence isn't loadable seems right; as a creator it would be much harder to find a problem if it was quietly filtered.

PR #3717, opened 2022-06-29
Fix GetStat & SetStat.

zedraxlo:


renpytom:
I don't know what I was thinking here, thanks for the fix.

PR #3730, opened 2022-07-02
Add a dict for font aliases

Gouvernathor:
I named it config.font_names, because font_map didn't seem very clear (and similar to font_replacement_map). font_alias would be great but it would create confusion with anti-aliasing.

I added a blocker in FontGroup, because there is a bug triggering if "a" is the alias of a fontgroup and "a" is also included in a fontgroup.
Without this block, the behavior is weird to document because having "a" in an aliased fontgroup if "a" is the alias of a simple font worked fine, BUT having "a" be the alias of "b" and "b" the alias of an actual font, that didn't work. So, it's clearer to disable both direct and indirect alias chaining.

I'll wait for a pre-approval before documenting this.


renpy/config.py
```diff
@@ -1264,6 +1264,9 @@ def say_attribute_transition_callback(*args):
 # the style inspector.)
 alternate_unelide_path = None
 
+# An alias -> font map.
+font_names = {}
```
renpytom, line 1268: Perhaps a good name would be font_name_map. Font_names doesn't say anything about what it does - I'd expect this to be a list or set.

renpy/text/text.py
```diff
@@ -352,6 +352,14 @@ def bounds(self, glyphs, bounds, layout):
         fo = font.get_font(self.font, self.size, self.bold, self.italic, 0, self.antialias, self.vertical, self.hinting, layout.oversample)
         return fo.bounds(glyphs, bounds)
 
+    @property
```
renpytom, line 355: I don't think a property is the right way to do this. Let's have a set_font method, which would be used in TextSegment.take_style, and then again in the implementation of the font tag.

renpytom (changes requested):

renpytom:
Looks good now.

sphinx/source/text.rst
```diff
@@ -134,13 +134,13 @@ It should be noted that:
...
 - Supplementarly exclamation marks will be ignored, and will not circumvent
-  the previuous rule : ``!l!c`` will do the same as ``!c!l`` or ``!cl``.
+  the previous rule : ``!l!c`` will do the same as ``!c!l`` or ``!cl``.
 
 The transformations are done in the following order:
 
 #. ``r``/``s`` (repr or str)
 #. ``t`` (translate)
-#. ``i`` (recursive interpolation)
+#. ``i`` (inner interpolation)
```
mal, line 143: This _is_ recursive. It doesn't stop at the first level.

Gouvernathor: I had done that to have a word in "i", but "interpolation" is enough

PR #3734, opened 2022-07-03
Add _trace_screen to per_frame_screens

Gouvernathor:
Alternative to #3699, per discussion there.

### DIFF
--- a/renpy/common/00console.rpy
+++ b/renpy/common/00console.rpy
@@ -1099,3 +1099,6 @@ label _console:
 
 label _console_return:
     return
+
+init -1010 python:
+    config.per_frame_screens.append("_trace_screen")
--- a/renpy/config.py
+++ b/renpy/config.py
@@ -864,7 +864,7 @@ fast_redraw_frames = 4
 gl_clear_color = "#000"
 
 # Screens that are updated once per frame rather than once per interaction.
-per_frame_screens = [ ]
+per_frame_screens = ["_trace_screen",]
 
 # How long we store performance data for.
 performance_window = 5.0
--- a/renpy/config.py
+++ b/renpy/config.py
@@ -864,7 +864,7 @@ fast_redraw_frames = 4
 gl_clear_color = "#000"
 
 # Screens that are updated once per frame rather than once per interaction.
-per_frame_screens = ["_trace_screen",]
+per_frame_screens = [ ]
 
 # How long we store performance data for.
 performance_window = 5.0


mal:
I think there's a significant risk here of being more confusing to most creators. In the overwhelming majority of cases this would cause a watched value will be updated far more frequently than it would be by any screen using it. While the current default isn't perfect, it does lead creators to discover when and why values get updated, this would show them a value updating like they expect, yet having no impact where they would expect. That seems much more confusing. Conversely, creators of sufficient expertise to want this behaviour in certain situations are capable of adding the trace screen to this config var themselves.

Separate to the concerns above, if this change was to be made then it should probably be done in the same manner that the `_performance` screen is appended to this list inside `00performance`, i.e. in `00console` near the `_trace_screen` definition.

Booplicate:
> more confusing to most creators. In the overwhelming majority of cases this would cause a watched value will be updated far more frequently than it would be by any screen using it. While the current default isn't perfect, it does lead creators to discover when and why values get updated, this would show them a value updating like they expect, yet having no impact where they would expect. That seems much more confusing.

To me it was more confusing why the value wasn't updating when it should, is my code wrong? Is it a renpy bug?

> yet having no impact where they would expect

The code is still being executed, so it means there is impact? Like if I watch some object state and it's being changed and I can see it on my screen, there is definitely impact. What is worse is when there's impact, but the screen doesn't show me, making me think something is wrong.

Gouvernathor:
I disagree about the last part, about experienced creators. `_trace_screen` is not documented, and we should not require or push people to read the inner code in order to master renpy in a meaningful way. So, I think either this PR or the original one should be merged.

As for the first argument, I also disagree, because not every value you may want to watch has something to do with screens. Some may be used in custom displayables for examples, or persistent, or time flow itself, or during a transition - that particular one is impossible to check without a per-frame trace.
So, the defaut behavior should not get flawed as it is now in its representation of how python works inside, even if it's flawed just like (other) screens. For people not knowing renpy that much, it is even *more* confusing, because we're lying about when things actually get updated (if it's a value being watched and not a function call).

Gouvernathor:
@Booplicate "The code is still being executed, so it means there is impact?"
Mal means that a screen will not instantly update based upon the state of a variable, typically a screen's if clause will not instantly execute (or at least be displayed), it will rather wait for the next interaction.
But I think that's only one more reason to make watch per-frame, because if your trigger something and it doesn't instantly apply, you may believe something is wrong with your trigger, when it's actually because of how screens work. Making the watch/trace show values in real time allows you to that.

Booplicate:
> Mal means that a screen will not instantly update based upon the state of a variable, typically a screen's if clause will not instantly execute (or at least be displayed), it will rather wait for the next interaction.

Oh I see, sorry. That makes more sense, but I don't think it should be a factor here.

I still remember times when renpy didn't have per frame screens, it was so confusing and inconvenient to track object states. I had to hide/show interface to trigger an interaction which would update the screens. Once renpy got per frame screens, I figured it'd be incredible for the watch screen, but I was also surprised it wasn't in per frame screens by default. Since then all my projects use it, I don't remember a single time when I needed it to be updated per interaction.

renpytom:
I think the argument for reflecting the actual state of Ren'Py is convincing, so I'm merging this. Since it's very minor, I'll put it in 8.0.1.

PR #3750, opened 2022-07-12
Optimizations for building Web packages

Tey:
When packaging a Web build, all the files are put in a game.zip file first before being selectively removed based on whether they should be downloaded progressively or not. This is a waste of time and disk space (the ZIP file content is duplicated), especially for novels with lots of images.

This PR adds a generic file filtering feature to the Distributor class so that builder scripts can filter out the files to package based on their specific needs. The Web builder script uses that to prevent the files to be downloaded progressively from being put in the ZIP archive at all, and copy them to the destination folder on the fly. The original progressive download filtering logic is preserved, but it happens sooner.

With these changes, empty folders are put in the ZIP archive though, but it should be harmless. Detecting empty folders is tricky without relying on a second pass.

### DIFF
--- a/launcher/game/distribute.rpy
+++ b/launcher/game/distribute.rpy
@@ -436,7 +436,7 @@ change_renpy_executable()
         This manages the process of building distributions.
         """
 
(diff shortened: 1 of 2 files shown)


renpytom:
Merged, thanks as always.

PR #3766, opened 2022-07-15
add combine option to extract dialogue

kyouryuukunn:
I added combine option to extract dialogue.
This combine special characters (%, {, [) which need to be double in dialogue into one. 
image


Gouvernathor:
I'm not sure this should be optional. Maybe there should be a middle-ground happening all the time, extracting the dialogue lines exactly like they appear on screen, which means with {{ turned into {, but (unescaped) text tags removed.
This looks like what you programmed, so I think it should always be enabled, not with a checkbox and a command-line argument.
But, maybe don't remove them before getting a second opinion about it 😅

kyouryuukunn:
[[ now doesn't combine. [[ and [ aren't distinct if combined.
"[test]" and "[[test]" is the same result

renpytom:
I'm fine making this normal behavior, not an option.

kyouryuukunn:
I removed the combine option. %% and {{ is always combined.

PR #3789, opened 2022-07-19
File statistics & Word counts for character statistics

kigyo:
Elaborates on #3114


renpytom:
This is too much detail, I think.

PR #3790, opened 2022-07-20
Fix modal behavior

midgethetree:
Renpy games were being allowed to advance in the following two scenarios even when a modal screen was being shown: when using auto-forward, and when using the renpy.pause function. With these changes a modal screen will properly block progression in those scenarios.

### DIFF
--- a/renpy/display/behavior.py
+++ b/renpy/display/behavior.py
@@ -519,24 +519,25 @@ class PauseBehavior(renpy.display.layout.Null):
 
     def event(self, ev, x, y, st):
 
-        if ev.type == renpy.display.core.TIMEEVENT and ev.modal:
-            renpy.game.interface.timeout(max(self.delay - st, 0))
-            return
+        if ev.type == renpy.display.core.TIMEEVENT:
+            if ev.modal:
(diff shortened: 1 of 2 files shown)


renpytom:
I've backed out  c1a75910405bbe7b1dd4cdbec1f6d8cbca926eff . I don't think this did anything, since it shouldn't be possible for an event other than a modal TIMEEVENT to reach a pause that's behind a modal screen. (If it is, that's the issue we should fixed - do you have a replication of this failing?)

midgethetree:
The problem I was originally fixing in c1a7591 was TIMEEVENTs reaching the PauseBehavior with their modal property set to false even when a modal screen was being shown, hence the switching around in interact_core of when the PauseBehavior was added to the root widget so the TIMEEVENTs would reach the modal screen and get properly flagged as modal. After that pauses were still timing out under a modal confirm screen if I moved my mouse, which is why I also edited PauseBehavior's event method to ignore nontime events. I was able to replicate the issue while testing all this by just adding "pause 3" followed by a line of dialogue to the beginning of a new project and clicking X to making the confirm screen show up in the middle of the pause - the pause would end and the dialogue would show up despite the presence of a modal screen without my edits.

PR #3804, opened 2022-07-24
Fixed mouse area hovering

midgethetree:
Mouseareas are supposed to not hover if the keyboard is being used for focusing, but this was broken by an if/else statement that changed the is_hovered value again without regard to if it had already been set to false because of keyboard focusing. Changing it to elif/else ensures the value of is_hovered is only changed if it hasn't already been set to false because of keyboard focusing.


renpytom:
Can you please retarget this to the maint-8.0 branch? Looks good otherwise.

renpytom:
Never mind - figured out I can do it (and that doing that doesn't work well). I'll just merge to master and cherry-pick.

PR #3821, opened 2022-07-29
Check conflicting properties in SL and ATL

Gouvernathor:
Before this is viable, the errors shouldn't be raised with .error but rather stored, and a check after init (?) should raise them or ignore them based upon the config variable.

The style part was tested, although not extensively.


renpy/sl2/slparser.py
```diff
@@ -117,6 +117,35 @@ def __init__(self, prefix, name):
             parser.add(self)
 
 
+incompatible_props = {"alignaround" : {"xaround", "yaround", "xanchoraround", "yanchoraround"},
```
Gouvernathor, line 120: The actual association table is here https://github.com/renpy/renpy/blob/198fc935327c185d317d6713fc82429f6613db17/module/generate_styles.py#L284
Or almost : it would actually be `{p : {el[0] for el in l} for p, l in synthetic_properties.items()}`.
I'm not sure how best to import that into renpy - the possibilities are numerous, and a lot of them are bad.

Gouvernathor:
Tested, seems to work.

PR #3827, opened 2022-07-30
Fix and simplify the "List filenames" dev action

Gouvernathor:
In py3 it resulted in a bunch of `b'somepath'` being written to the file and in the console, now it's good, back to `somepath`.
I also changed things internally to simplify it.


KiloOscarSix:
I would be interested to see this expanded to allow a sub directory argument that allows more precise file list instead of everything.

PR #3838, opened 2022-08-01
use character namespace for narrator

### LINKED ISSUE #3837: Character namespace doesn't work on Narrator object.
As documented you can style the narrator with `narrator = Character(**properties)`

However, unlike other character objects using the character namespace:
`character.narrator = Character(**properties)`
won't style the narrator character as the `say()` function is using `renpy.store.narrator` 

Not sure if this is an oversight or intended

KiloOscarSix:
Resolves #3837

### DIFF
--- a/renpy/common/00library.rpy
+++ b/renpy/common/00library.rpy
@@ -161,7 +161,10 @@ init -1700 python:
             who = None
 
         if who is None:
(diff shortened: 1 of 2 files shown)


renpytom:
Closing this per my comment on #3837.

PR #3839, opened 2022-08-02
Modified ConditionSwitch to include a 'continuous' argument.

chocoblu:
Modified *ConditionSwitch* to include a new argument: `continuous`

If `continuous` is true, the *ConditionSwitch* displayable continuously rechecks its conditions every `config.conditionswitch_refresh_rate` seconds, and responds to changes in variables, etc. in between interactions, instead of just at the start.

It's cleaner and easier than *DynamicDisplayable* for some situations- could be useful for e.g. a character whose eyes actively follow a sprite, or current mouse position; or with python code that runs in between say statements, or in the background as a thread. 

Could also be useful for creators who don't quite grasp, or need something as flexible as, *DynamicDisplayable*.

------
*(Either way, DynamicDisplayables were getting too cumbersome. I couldn't implement this in the script, so I had to alter the engine. Now hopefully, I wont have to re-appy this feature whenever RenPy gets updated)*

### DIFF
--- a/renpy/config.py
+++ b/renpy/config.py
@@ -915,6 +915,9 @@ optimize_texture_bounds = True
 # Should we predict everything in a ConditionSwitch?
 conditionswitch_predict_all = False
 
+# How long should a continuous ConditionSwitch wait before updating itself? (seconds)
+conditionswitch_refresh_rate = 0.1
+
 # Transform events to deliver each time one happens.
 repeat_transform_events = [ "show", "replace", "update" ]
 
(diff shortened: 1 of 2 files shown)


renpytom:
I like the idea of this, though not some of the naming. I think my preferred version would be: 

```
image cs = ConditionSwitch(..., period=.1)
```

I like the name period, rather than continuous. That also means we wouldn't need the config variable, and could have different refresh rates. Would you want to make these changes, or should I?

chocoblu:
> Would you want to make these changes, or should I?

I think that it would be best for you to handle this...

I don't actually have a working RenPy dev setup- I just cloned the repo and copied over the alterations I made to my RenPy installation lmao

PR #3855, opened 2022-08-10
Use c-boosted mul function instead of python-defined one

Gouvernathor:
I'm not sure when these are called, I just saw the code and tested an improvement.
Timeit shows a 2x faster calculation.


renpytom:
Seems fine-ish. It's unlikely to be a major improvement, but doesn't hurt.

PR #3866, opened 2022-08-13
Lint out-of-range init priorities

Gouvernathor:
I didn't phrase it like "the init *statement's* priority" because it also applies to statements like `define 1000 bluh = "oh"`, which aren't understood by your common creator as init statements.

### DIFF
--- a/renpy/lint.py
+++ b/renpy/lint.py
@@ -708,6 +708,11 @@ def check_styles():
         check_style("Style " + name, s)
 
 
+def check_init(node):
+    if not (-999 <= node.priority <= 999):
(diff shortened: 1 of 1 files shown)


KiloOscarSix:
Does having high init priority have any negative effects?
i.e if a creator is using `init 1000` `init 2000` etc

Gouvernathor:
Yep, it can make the block execute after built-in blocks that are meant to execute after any and all creator init code

KiloOscarSix:
Does having high init priority have any negative effects?
i.e if a creator is using `init 1000` `init 2000` etc

KiloOscarSix:
Does having high init priority have any negative effects?
i.e if a creator is using `init 1000` `init 2000` etc

Issue #3755, opened 2022-07-13
use a NamedTuple for version_tuple


Numerlor:
Getting data from the version tuple is somewhat unclear with the indexed access, a named tuple can give meaningful names to the version fields, while also keeping compat with the plain tuple used previously

GiovanH:
Should be resolved with https://github.com/renpy/renpy/pull/3883,

PR #3883, opened 2022-08-20
Use a NamedTuple for version_tuple (for #3755)

GiovanH:
Very simple draft that enables the convenience methods for indexing properties while remaining backwards compatible.

### DIFF
--- a/renpy/__init__.py
+++ b/renpy/__init__.py
@@ -79,12 +79,12 @@ except ImportError:
 
 official = official and getattr(site, "renpy_build_official", False)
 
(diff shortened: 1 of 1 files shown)


GiovanH:
image

image

GiovanH:
Only question I can think of is if `vc_version` is the best name for the fourth field. Maybe just `vc`?

renpytom:
Probably 'commit' would be a good name for the last field. 

I'd also call it VersionTuple, and not SemVer. Ren'Py doesn't use Semantic Versioning.

renpytom:
Merged. Thank you.

PR #3899, opened 2022-08-23
Launcher options to add id clauses to all dialogue lines

Gouvernathor:
Based upon the add_from implementation.

### DIFF
--- a/launcher/game/translations.rpy
+++ b/launcher/game/translations.rpy
@@ -208,6 +208,11 @@ screen translate:
                         textbutton _("Update Default Interface Translations"):
                             action [ Jump("update_renpy_strings") ]
 
+                        add HALF_SPACER
+
+                        textbutton _("Generate id clauses for dialogue lines"):
+                            action Jump("add_id_clauses")
+
 
 
                 frame:
@@ -229,6 +234,11 @@ screen translate:
 
     textbutton _("Return") action Jump("front_page") style "l_left_button"
 
+label add_id_clauses:
+    python:
+        interface.processing(_("Adding id clauses to say statements that do not have them."))
+        project.current.launch([ "add_id" ], wait=True)
+    jump translate
 
 label translate:
     call screen translate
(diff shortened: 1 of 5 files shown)


mal:
Script parsing is already one of the slowest parts of starting larger games, so making it slower by building a huge array of all dialogue blocks without an explicit ID doesn't seem like a great idea, even before considering the memory consumed to store it. :/

Neyunse:
I think this is important, because if I change all the dialogs and move other script files with existing translations, it won't be used.  This would help avoid having to go line by line fixing the IDs that were moved to another file, of course the IDs must be included before the file is changed.

This will probably fix several issues with translations that renpy missed or probably misses when a translation change is generated, maybe there are more examples to add to this. but well, this is just my point of view.

renpytom:
The script ids don't consider filename, only the label that precedes them. (With label hide available if you introduce a label you don't want considered by the translation system.) That being said, moving game script between files after release is bad, as it can break the save/load system - the statements will be given new names. (Which are different from the translation ids.)

Neyunse:
> The script ids don't consider filename, only the label that precedes them. (With label hide available if you introduce a label you don't want considered by the translation system.) That being said, moving game script between files after release is bad, as it can break the save/load system - the statements will be given new names. (Which are different from the translation ids.)
> 
> I tend to think the benefit is small, and there's a large risk of someone hitting this by mistake and making their script very ugly.

But with this adding the ids, wouldn't the statements be preserved when moving the dialogs to another file?  

And perhaps, so that it is not checked by accident, the user should be asked if they really want to make this change, with a brief explanation of what will happen and that they do so at their own risk. But as you say, maybe the benefit is small.

Gouvernathor:
> wouldn't the statements be preserved when moving the dialogs to another file?

There are two ids systems, for two separate uses. The translation ids system do not break when you rename a file, so adding translation ids to dialogue before doing that won't change anything.
The second id system is about saving and loading (basically), and it's unrelated to the translation ids system. Renaming files will cause save/load issues, regardless of whether you added translation ids or not.

Gouvernathor:
I coded this under false premises, with an inaccurate idea of how translation ids were generated (it's much more reliable than I thought). Closing this for now, unless some day we find a good enough reason to reopen.

PR #3901, opened 2022-08-24
Improve lint for say attributes

midgethetree:
Applies config.adjust_attributes to say attributes when linting, so lint won't report errors for say statements which work in-game.

### DIFF
--- a/renpy/lint.py
+++ b/renpy/lint.py
@@ -510,13 +510,18 @@ def check_say(node):
 
         name = (char.image_tag,) + attributes
 
+        orig = name
+        f = renpy.config.adjust_attributes.get(name[0], None) or renpy.config.adjust_attributes.get(None, None)
+        if f is not None:
+            name = f(name)
+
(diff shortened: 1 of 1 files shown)


mal:
Rather than duplicate the code from `image_exists` in `check_say`, I wonder if instead the calls to `image_exists_imprecise` in `check_say` ought to become `image_exists(..., precise=False)` calls. Not 100% on that, more floating the idea just in case.

Gouvernathor:
@mal I think the duplicate code is fine, because it's pretty straightforward and it's already a duplicate from [renpy.image.ShownImageInfo.apply_attributes](https://github.com/renpy/renpy/blob/1daceab8eea7842535ebc95f16dfa12e1a194b78/renpy/display/image.py#L937). So I think we should merge this, and possibly improve it afterwards if we want.

renpytom:
I agree. I think this makes sense to go in, and I'll merge it into the maint-8.0 branch.

PR #3904, opened 2022-08-25
Allow config.version to be any object

Andykl:
The idea is to allow developers to add full version information to the log file and (more importantly, since you can't write to that file) the traceback file, but the human-readable version may be different from that.
For example, in the git version I add the hex of the current commit, and in the release versions I add the additional hotfix number used when updating, which allows me to see if the bug report is out of date.

This is intended to be used like
```py
python early:
    class GameVersion(object):  # or str subclass
        def __str__(self):
            return "0.999"

        def __repr__(self):
            return "0.999.999"

    config.version = GameVersion()
```

Alternative solution could be to add another config.variable.
TODO: document it

I also want to fix the issue with `[config.name] [config.version]` not being printed in log.txt because it happens before any script read. I think we could add additional file like `script_version.txt` with that info during build, so if people send `log.txt` instead of `traceback.txt` I know the version they use wilthout usage of undocumented `renpy.write_log(config.version)`. Alternatively we could write it in the middle of the log before execute init code.

### DIFF
--- a/gui/game/options.rpy
+++ b/gui/game/options.rpy
@@ -23,9 +23,13 @@ define config.name = _("Ren'Py 7 Default GUI")
 define gui.show_name = True
 
 
-## The version of the game.
+## The version of the game used in log file and traceback file. This could be any object.
 
 define config.version = "1.0"
+
+## The version of the game displayed to user in about screen.
+
+define gui.version = "1.0"
 
 
 ## Text that is placed on the game's about screen. Place the text between
--- a/gui/game/screens.rpy
+++ b/gui/game/screens.rpy
@@ -370,7 +370,7 @@ screen main_menu():
             text "[config.name!t]":
                 style "main_menu_title"
 
-            text "[config.version]":
+            text "[gui.version]":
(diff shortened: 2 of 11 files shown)


renpytom:
Hm... perhaps have a build_info.json (perhaps in cache/), that is written to if config.developer is True, and initializes config.version (And any other fields that might come in handy in the future.)

mal:
Interpolating `config.version` using `%r` will add quotes around the most common use case which doesn't seem ideal.

More generally my question would be why not simply use the more specific string as `config.version` and forgo the less specific one? Is there a reason you wouldn't want to expose the finer version points in the UI?

If UI use is the issue, then perhaps a `gui.version` that by default takes the value of `config.version` but can be set explicitly to something else for use in UI situations?

Andykl:
I updated that to Mal's idea.

renpytom:
I'd suggest that gui.version should default to config.version if not explicitly set. This way, we don't need to introduce it in the default options.rpy.

Issue #2977, opened 2021-08-14
[Feature Request] avif images support


suphamster:
Avif about 25% more effecient at the same size than webp (and about same amount slower).

renpytom:
Do any web browsers support avif? I tend to use that as my proxy for a format becoming popular enough that tools exist to make it worth using?

uyjulian:
https://caniuse.com/avif  

`sdl2_image` and `ffmpeg` do not support AVIF container format yet.

PhiSYS:
The downside is that the encoding is hell of slow... but encoding is a one time process and it worth it. I can confirm in my tests that AVIF is far by superior from WebP, it retains much more fine details while it has practically no artifacts/ringing and it's.
It also supports Animations and Alpha althought I didn't tested them yet.

I definitely upvote AVIF

renpytom:
I'll close this for now, as I probably won't do anything until it's SDL2_image,  at which point we'll get it by default. I don't think I need this to stay open until then.

suphamster:
> I'll close this for now, as I probably won't do anything until it's SDL2_image, at which point we'll get it by default. I don't think I need this to stay open until then.
Hi, it looks like devs of this library finally added support by my request https://github.com/libsdl-org/SDL_image/commit/00b71a12a5b0ea376a890fbcdd22475ae9694a39
https://github.com/libsdl-org/SDL_image/issues/188

thelordofcorruption:
AVIF support has been added to SDL_image, so I think this issue should be reopened. https://github.com/libsdl-org/SDL_image/commit/00b71a12a5b0ea376a890fbcdd22475ae9694a39

renpytom:
Not sure which advanced format is best, as I need to investigate which dependencies these have.

renpytom:
This should be done now.

PR #3906, opened 2022-08-26
Add JPEG-XL, AVIF support

### LINKED ISSUE #3865: [Feature Request] Implement support for JPEG XL image format
JPEG XL is a new image format made by the "JPEG" committee intended as a replacement for jpeg. It support both lossless and lossy compression while having a smaller size. This would be awesome for people like me who like to use lossless images in their VN for the best quality possible and with JPEG XL we will have lower sizes too.

About JPEG XL:
- free, open source and royalty-free
- bitstream was frozen a year ago, December 2020
- standard formally approved 17.December.2021 [twitter](https://twitter.com/jonsneyers/status/1471942556468584454)
- beats every other standardized format in compression and features (HDR, progressive decoding, precision, animation, channels, speed), only AVIF wins in compression for very low quality, low fidelity images
- old JPEGs can be encoded to 20% smaller JPEG XLs lossless and 100% restored to old JPEGs again
- extension is .jxl

About it's compression:
- Lossy: about 60% smaller than JPEG for same visual quality
- Lossless JPEG transcoding: about 20% smaller
- Lossless: about 35% smaller than PNG (50% smaller for HDR)

And as for it's support all major browsers support it but it's not yet enabled by default. Many image viewers and picture editors already do support viewing JPEG XLs.

JPEG XL has also been added in to SDL_image: https://github.com/libsdl-org/SDL_image/commit/742cd132b18c3612e00f5f7d47bc24dbae2e76f9

Some links:
- Most of info about JPEG XL taken from: https://github.com/libsdl-org/SDL_image/issues/204
- Wikipedia: https://en.wikipedia.org/wiki/JPEG_XL

Orochimarufan:
Plumb loading of .jxl and .avif files.

The second commit adds some heuristics to unofficial build to hopefully help with issues due to missing (and non-default) support for these formats in users' SDL2_image. It can be removed if unwanted.

Depends on renpy/pygame_sdl2#142

Fixes #3865 #2977 

Tested by converting the_question/images/* to JXL and running the game.

I'm not overly familiar with the codebase, so I may have missed somewhere that needs new image formats mentioned, please point me in the correct direction if so. In pgrender.py, there exists a list of reentrant format loaders; I didn't add the new ones as I don't know if they are and couldn't easily find any information on the topic.

This also assumes that official builds will include support in their bundled SDL2_image build. I haven't looked into any build scripts that may be used for official builds yet so this still needs to be addressed before an official release.

### DIFF
--- a/renpy/common/00images.rpy
+++ b/renpy/common/00images.rpy
@@ -39,7 +39,7 @@ init -1900 python:
             basename = os.path.basename(fn)
             base, ext = os.path.splitext(basename)
 
-            if not ext.lower() in [ ".jpg", ".jpeg", ".png", ".webp" ]:
+            if not ext.lower() in [ ".jpg", ".jpeg", ".png", ".webp", ".jxl", ".avif" ]:
                 continue
 
             base = base.lower()
--- a/renpy/common/00images.rpy
+++ b/renpy/common/00images.rpy
@@ -58,3 +58,57 @@ init 1900 python:
 
     if config.late_images_scan:
         _scan_images_directory()
+
+init 3000 python:
+    def _heuristic_image_format_check():
+        formats = {
+            # PNG
+            ".png": pygame_sdl2.image.INIT_PNG,
+            # JPEG
+            ".jpg": pygame_sdl2.image.INIT_JPG,
+            ".jpeg": pygame_sdl2.image.INIT_JPG,
+            # TIFF
+            ".tif": pygame_sdl2.image.INIT_TIF,
+            ".tiff": pygame_sdl2.image.INIT_TIF,
+            # WebP
+            ".webp": pygame_sdl2.image.INIT_WEBP,
+            # JPEG-XL
+            ".jxl": pygame_sdl2.image.INIT_JXL,
+            # AVIF
+            ".avif": pygame_sdl2.image.INIT_AVIF,
+            ## There is no real way of checking the below,
+            ## but they are built into SDL2_image by default
+            # QOI
+            ".qoi": 0,
+            # BPM
+            ".bmp": 0, ".ico": 0, ".cur": 0,
+            # GIF
+            ".gif": 0,
+            # TGA
+            ".tga": 0,
+            # XCF
+            ".xcf": 0,
+        }
+
+        from collections import defaultdict
+        counter = defaultdict(lambda: 0)
+
+        for _, d in renpy.display.image.images.items():
+            if isinstance(d, renpy.display.im.Image):
+                _, ext = os.path.splitext(d.filename)
+                if ext:
+                    counter[ext.lower()] += 1
+
+        for ext, count in counter.items():
+            if count > 0:
+                itag = formats.get(ext, None)
+                if itag is None:
+                    renpy.log("Possibly unknown image format: %s (found %d images)" % (ext, count))
+                elif not pygame_sdl2.image.has_init(itag):
+                    renpy.log("Possibly missing image support in Ren'Py/SDL2_image build: %s (found %d images)"
+                        % (ext, count))
+
+        return counter
+    
+    if not renpy.official:
+        _heuristic_image_format_check()
--- a/renpy/config.py
+++ b/renpy/config.py
@@ -1300,7 +1300,7 @@ def init():
 
     global autoreload_functions
     autoreload_functions = [
-        (r'\.(png|jpg|jpeg|webp|gif|tif|tiff|bmp)$', renpy.exports.flush_cache_file),
+        (r'\.(png|jpg|jpeg|webp|gif|tif|tiff|bmp|jxl|avif)$', renpy.exports.flush_cache_file),
         (r'\.(mp2|mp3|ogg|opus|wav)$', renpy.audio.audio.autoreload),
         ]
 


renpytom:
Do you know how big AVIF and all it's libraries is? I'm worried that adding tons of minority image formats will bloat Ren'Py for little benefit. (And Jpeg XL seems better.)

Orochimarufan:
Indeed there is considerable overhead in including these libraries. I've compiled a few variations of SDL2_image:

```
752K	libSDL2_image-2.0.so.0.700.0-webp
3.4M	libSDL2_image-2.0.so.0.700.0-webp+avif
2.7M	libSDL2_image-2.0.so.0.700.0-webp+jxl
5.3M	libSDL2_image-2.0.so.0.700.0-webp+jxl+avif
5.8M	libSDL2_image-2.0.so.0.700.0-webp+jxl+avif+tiff
```

While I'm personally also somewhat partial to JPEG-XL over AVIF, I decided to do both while I was at it and they are in SDL2_image. A decision can still be made to only support one or even none of both.

I should also note that libavif seems to be rather annoying to vendor and I haven't yet figured out how to keep it from pulling in system libjpeg and libyuv. Upstream SDL2_image doesn't yet support vendoring libavif at all. See [my branch](https://github.com/Orochimarufan/SDL_image/tree/feat-vendor-avif) for what was used for the builds above.

If you could point me in the right direction, I could try building a couple of full Ren'Py distributions with either or both formats included and see what the size difference comes down to.

-- Slightly off-topic from here --
Incidentally, while testing the above builds of SDL2_image with some sample images, it seems that defaults for lossy JPEG-XL are rather more conservative than the others (resulting in sometimes significantly larger files), while it seems to consistently beat them in lossless mode.

renpytom:
Hm... the ren'py build process consists of the various tasks in renpy-build/tasks. Ideally, thinks would be built with configure, and cross-compiled. I'll probably have a go at this after 8.0.3 comes out. 

How much bigger are the jxl files, and are they bigger than jpegs at equivalent quality? My feeling is the target for this are advanced users, and they're likely to pick their wn image quality settings. 

Looking at the numbers and various other things, I think JXL only (and not adding support for tiff or avif) is probably the way to go. I'd probably want to add support for SVG at the same time - I think SDL_image has a renderer.  Does SDL_image have any support for HDR decoding?

Orochimarufan:
I might look into the build at some point, but the requirements seem to be rather hefty as far as disk space and time go. Maybe i'll spin up a temporary cloud machine or something.

I think my test image may have been something of a worst-case scenario for JXL (using jxl 0.6.1 and aom 3.4). Of course, the big problem here is "equivalent quality". FWIW, Butteraugli favors jxl quite significantly (which isn't surprising, seeing as how it was tuned to that metric), while more naive metrics tend to favor avif. I didn't find an avifenc option to compare -q60 to. In the end, all of this circumstantial though.

|   Size   | Ratio  |  MSE   |  PSNR  |  SSIM | BtrAugli |        File        |
|---------:|-------:|-------:|-------:|------:|---------:|:-------------------|
|  14.2MiB |        |        |        |       |          | test.png           |
| 388.1KiB |  2.67% |  3.642 | 42.517 | 0.976 | 5.560896 | test-default.avif  |
| 613.1KiB |  4.22% | 20.894 | 34.931 | 0.953 | 8.502783 | test-default.jpg   |
| 663.7KiB |  4.57% |  7.045 | 39.652 | 0.973 | 2.048718 | test-default.jxl   |
| 333.5KiB |  2.30% | 18.500 | 35.459 | 0.959 | 9.212794 | test-default.webp  |
|   8.2MiB | 57.84% |  0.000 |    inf | 1.000 | 0.000000 | test-lossless.avif |
|   7.3MiB | 51.27% |  0.000 |    inf | 1.000 | 0.000000 | test-lossless.jxl  |
|   8.1MiB | 56.76% |  0.000 |    inf | 1.000 | 0.000000 | test-lossless.webp |
| 474.8KiB |  3.27% | 27.020 | 33.814 | 0.945 | 9.826682 | test-q60.jpg       |
| 274.9KiB |  1.89% | 19.271 | 35.282 | 0.958 | 4.927706 | test-q60.jxl       |
| 281.6KiB |  1.94% | 21.329 | 34.841 | 0.955 | 9.804398 | test-q60.webp      |
| 700.9KiB | 4.83% | 6.829 | 39.787 | 0.973 | 2.385626 | test-d1e9E3.jxl  |
| 408.5KiB | 2.81% | 11.750 | 37.431 | 0.966 | 2.953185 | test-d2.jxl |
| 311.7KiB | 2.15% | 16.447 | 35.970 | 0.961 | 3.821957 | test-d3.jxl |

Between WebP and JXL, there may indeed not be any need for also having AVIF. SVG is indeed in SDL_image, though I'm not sure how complete it's implementation is. From what I can tell, it should even work right now, provided there is an explicit image statement or .svg is added to 00images.rpy.

As to HDR, [I don't think there is (currently) any HDR in SDL at all](https://wiki.libsdl.org/SDL_PixelFormatEnum). JXL itself supports it just fine. I don't know how much of Ren'Py's drawing pipeline relies on SDL, but HDR seems like a major undertaking to me.

Perhaps it would be more beneficial to deal with libjxl directly instead of going through SDL_image anyway. It may also allow us to use JXL's frame feature to pack multiple variations of a base image into the same file with basic intra-frame compression down the line. This could bring significant savings for Ren'Py specifically, i think, as a lot of VN assets have variations with only small differences like character sprites only differing in facial expression, or backgrounds in different lighting conditions.

thelordofcorruption:
Another cool feature that JPEG XL has and AVIF lacks is also progressive decoding which I think would be an awesome feature in the future on renpy web https://www.youtube.com/watch?v=UphN1_7nP8U

suphamster:
> Plumb loading of .jxl and .avif files.
> 
> The second commit adds some heuristics to unofficial build to hopefully help with issues due to missing (and non-default) support for these formats in users' SDL2_image. It can be removed if unwanted.
> 
> Depends on [renpy/pygame_sdl2#142](https://github.com/renpy/pygame_sdl2/pull/142)
> 
> Fixes #3865 #2977
> 
> Tested by converting the_question/images/* to JXL and running the game.
> 
> I'm not overly familiar with the codebase, so I may have missed somewhere that needs new image formats mentioned, please point me in the correct direction if so. In pgrender.py, there exists a list of reentrant format loaders; I didn't add the new ones as I don't know if they are and couldn't easily find any information on the topic.
> 
> This also assumes that official builds will include support in their bundled SDL2_image build. I haven't looked into any build scripts that may be used for official builds yet so this still needs to be addressed before an official release.

Hi, thx for this very much! Can u compile plz if possible renpy sdk for Windows with jpeg-xl and avif support?

Quackdoc:
I do not know how most renpy games works, but being linked this issue I thought I would give my thoughts on the benefits and cons of AVIF and JXL to help make a decision if one or both are worth it. both AVIF and JXL have various benefits, but typically for still images I would use JXL, it wins on most cases over avif and only typically looses at the slowest settings in terms of compression.

JXL can losslessly compress jpeg images, and will be smaller then most jpegs out in the wild at any given quality.

AVIF on the otherhand is pretty much just a video file (an MP4 to be specific) coded in the AV1 video codec. if it's a still, it's a single image, if it's an animated AVIF, it's a standard video with a flag set in the mp4 container with no audio and a different extension. this means that while AVIF is still better then webp and jpeg in typical scenarios. it will usually lose to JXL in cases where still images are involved. (there are always exceptions) 

The key strength of AVIF is it's animation support. (I have no idea if this applies here. hence my earlier disclaimer) being based on the AV1 codec, it makes this a very efficient replacement for gifs animated webps, and animated pngs. 

In the case of AVIF vs Gif vs Webp, Gif is obviously no contender, when encoding from source, you can get similar qualities of webp animations at still much lower filesizes. webp animations are made using the video codec VP8, av1 is significantly smaller then VP8 at similar qualities (sometimes 2-8x smaller) this means something like a 20Mb webp animation could be achieved with something like an 8mb avif. (for further reading simply comparing vp8 to av1 video efficacy will get you the idea)

AVIF animations come with a significant downside though, at resolutions greater then 1080p significantly lower end devices may not be able to smoothly play them back, for instance I have a lenovo winbook N22 equiped with a celeron N3050 (the cheapest laptop from 2015, was also the cheapest chromebook availible at the time) will typically stuggle to playback 1080p30 avif animations unless they are made with a fastdecode preset. 

this is a very minor subset of hardware that cannot play back these files. as even a raspberry pi4 should be able to playback 1080p30 smoothly. 

Hopefully this can help decide whether or not JXL and/or AVIF are worth while.

VXsz:
To add to what [Quackdoc](https://github.com/Quackdoc) said, I personally really like JXL, not only because of it's great features, but rather how good it compresses and easily decodes, sometimes even faster than good old jpeg. 


I've done *a lot* of testing, and JXL easily wins over avif in terms of image compression, here's way:

- Decoding & encoding is much better/faster than AVIF, and runs easier with JXL, Making it a much better candidate for real-world usage.
- For intense images where there's a lot of information and very fine details everywhere, JXL wins by a huge margin, that's because AVIF just tends to smooth over and keep edges rather than just keeping every detail it can, so this could be problematic to some specific scenes or renders. Could even change the art direction.
- As much as slow the encoding is, AVIF only wins on about %20 of the cases, the good times where AVIF wins is obviously flat images or 2D art, but for the usual 3D render or even a photo, most of the time either the size gets too big, or too small & blurry, that sweetspot between those spots is where it wins over JXL, but it's a very narrow window so it's not constant or something really useful to do.

A few pros of AVIF
- It really compress flat images well, 2D stills and such are easily compressed but I haven't tried comparing those (wasnt my target images), but I'm sure it wins over JXL in these cases.
- For video, AVIF is the new codec in the block, and it should be the greatest so far (tho the encoders need to get a bit more mature for quality encodes), and with how wide it's getting supported even on TVs and most phones, it would be great to have for video content, specially 2D animated ones where AVIF's strengths appear.

Eagerly waiting for a JXL build :D

Orochimarufan:
I've been thinking more about this recently and was wondering how video decoding is handled in Ren'Py.

Sooner or later it may be beneficial to have AV1 support for videos, at which point all the AVIF dependencies are basically required anyway. (This was the argument for quick adoption of AVIF in browsers too; This also applies to HEVC/HEIF, though I'd personally skip those in favor of AV1)

The same is already true, to an extent, for WebM/VP8 and WebP. I'm guessing that we currently duplicate relevant dependencies between SDL_image and FFmpeg. Perhaps dropping SDL_image entirely and folding image decoding into the same path as video could save some duplicate dependencies and may even ease implementation of more advanced features like HDR (at least as far as decoding goes) and/or layered images (OTOH, I'm not at all sure about progressive loading).

Admittedly, this would be a far larger project than flipping a few switches in SDL_image. FWIW, FFmpeg also has support for all of {JXL, AV1, AVIF, SVG}, though I can't say how heavy the respective implementations are compared to SDL_image (especially for SVG).

VXsz:
> was wondering how video decoding is handled in Ren'Py

As far as I can tell, it's mostly handled by [ffmpeg](https://github.com/renpy/renpy-build/blob/master/tasks/ffmpeg.py#L126), licensing and related stuff need to be checked before adding a new codec to avoid legal issues.

suphamster:
> Plumb loading of .jxl and .avif files.
> 
> The second commit adds some heuristics to unofficial build to hopefully help with issues due to missing (and non-default) support for these formats in users' SDL2_image. It can be removed if unwanted.
> 
> Depends on [renpy/pygame_sdl2#142](https://github.com/renpy/pygame_sdl2/pull/142)
> 
> Fixes #3865 #2977
> 
> Tested by converting the_question/images/* to JXL and running the game.
> 
> I'm not overly familiar with the codebase, so I may have missed somewhere that needs new image formats mentioned, please point me in the correct direction if so. In pgrender.py, there exists a list of reentrant format loaders; I didn't add the new ones as I don't know if they are and couldn't easily find any information on the topic.
> 
> This also assumes that official builds will include support in their bundled SDL2_image build. I haven't looked into any build scripts that may be used for official builds yet so this still needs to be addressed before an official release.

Hi, can u tell me how to compile renpy/py_sdl2 properly to add jxl/avif support? I've replaced old ver with latest SDL_image version, added in build config jxl/avif (--enable-jxl, --enable-avif), downloaded external libs to folder with same name but when SDL_image compiling (along with renpy using renpy-build script) it says that can't find jxl/avif libs but there are exist/downloaded in https://github.com/libsdl-org/SDL_image/tree/main/external with download.sh script and I've make sure that external folder wouldn't be cleared by build script in tasks...

renpytom:
I've manually merged a decent fraction of this in, to support svgs and avif files, so I'm closing this pull request.  Thank you, @Orochimarufan, for the contribution.

PR #3911, opened 2022-08-27
Add build_info.json during distribution to help log.txt get the game name and version

Andykl:
This actually does not work because `repny.loader.load` don't have searchpath yet and in time it does we already have log.txt written.


renpytom:
Implemented in 270b4bbb07eb5b0801804bffd167cf4535adeebd and a few commits before it, so done.

PR #3913, opened 2022-08-28
Allow quotes inside of triple strings

Phxntxm:
Unless I am missing some case this covers, I do not see why this regex should disallow quoting inside of triple strings at all. This allows something like the following:

```
label start:
    """
    Test "with" quote in it
    """
```

Which currently does not work.

### DIFF
--- a/renpy/lexer.py
+++ b/renpy/lexer.py
@@ -905,13 +905,13 @@ class Lexer(object):
         newline.
         """
 
-        s = self.match(r'r?"""([^\\"]|\\.)*"""')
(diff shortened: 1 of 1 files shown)


renpytom:
A good catch, but this is the wrong solution. The regex you proposed would match all of:

```
""" foo """ bar """
```

Including all three delimiters. I'll change to:

```python
r?"""([^\\"]|\\.|"(?!""))*"""
```

Which allows a " when not followed by two more quotes, and I think it's correct.

PR #3922, opened 2022-08-30
Allow multipliers in {size=<value>}

methanoliver:
Sometimes, your tag doesn't know what size it will be, but `+`/`-` is not enough, and turned out that in some circumstances I actually need a multiplier, rather than an addition, never mind that it's an int in the end.

Could this be added?

```
"{size=*2}This is double size{/size} and this is {size=*0.5}half size{/size}"
```

### DIFF
--- a/renpy/text/text.py
+++ b/renpy/text/text.py
@@ -1184,6 +1184,8 @@ class Layout(object):
 
                     if value[0] in "+-":
                         push().size += int(value)
+                    elif value[0] == "*":
+                        push().size = int(float(value[1:]) * push().size)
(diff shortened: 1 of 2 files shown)


JSearUK:
I've wanted this a couple of times, too. I write on the assumption that the user may adjust text size themselves at any time via the Accessibility screen, so I never assume that hardcoded sizes will stay that way. The lack of relative size adjustment on this tag always perplexed me slightly.

renpytom:
Makes sense, so I've merged this. Thank you for the contribution.

PR #3938, opened 2022-09-02
Add renpy.clear_attributes

Gouvernathor:
This is a high-level implementation, using only documented functions. It has the advantage not to rely on existing implementation, making it virtually obsolescence-proof, but it may not be the most efficient implementation.

Anyway, this can be merged now and improved afterwards.


renpytom:
This needs a changelog entry, but otherwise looks good.

PR #3943, opened 2022-09-04
Add min, max arguments for actions

via-my:
A simple idea that will make our life easier.

From:
`SetVariable('value', min(5, max(0, value + 1)))`
To:
`SetVariable('value', value + 1, max=5, min=0)`

### DIFF
--- a/renpy/common/00action_data.rpy
+++ b/renpy/common/00action_data.rpy
@@ -61,17 +61,27 @@ init -1600 python:
         """
 
         identity_fields = [ "object" ]
-        equality_fields = [ "field", "value" ]
+        equality_fields = [ "field", "value", "min", "max" ]
 
         kind = "field"
 
-        def __init__(self, object, field, value, kind="field"):
+        def __init__(self, object, field, value, kind="field", **kwargs):
             self.object = object
             self.field = field
             self.value = value
             self.kind = kind
+            self.min = kwargs.get("min")
+            self.max = kwargs.get("max")
 
         def __call__(self):
+            if isinstance(self.value, (int, float)):
+
+                if self.min is not None:
+                    self.value = max(self.value, self.min)
+
+                if self.max is not None:
+                    self.value = min(self.value, self.max)
+
             __set_field(self.object, self.field, self.value, self.kind)
             renpy.restart_interaction()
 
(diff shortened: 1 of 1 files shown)


Andykl:
I don't think we should add such thing to the language. I can imagine some other "usefull once per 69 cases" keywords.
Moreover, because I also needed invrement/decrement actions I made my own `IncrementVariable(variable_name, amount=1, max=None)` and `DecrementVariable(variable_name, amount=1, min=None)`, and their additional benefit is that they are constant, whereas `SetVariable(variable_name, value, min=None, max=None)` is not (in case `value` is not a constant).

via-my:
Good idea. Let's propose `IncrementVariable` and `DecrementVariable` then.

via-my:
But then we should write `IncrementVariable`, `IncrementFieldVariable`, `IncrementScreenVariable`, `IncrementLocalVariable`, `IncrementDict` and `Decrement`...

Numerlor:
What exactly is wrong with `SetVariable('value', min(5, max(0, value + 1)))`? If the nested min/max is the issue, you can just define your own clamp function

via-my:
Basically, this is more about simplification, time saving and code readability for a common novel developer.
It will be a little bit faster to read and edit code, especially when there will be no need to worry about the closing braces or any other non eye-catching symbols. Sometimes, it may be a time consuming task for big projects that require 30sec> to load, since if there will be a small syntax mistake, the whole project will be closed.
```
action SetScreenVariable('page', clamp(page + 1, 0, int(len(GALLERY_DICT) / per_page)))
```
```
action SetScreenVariable('page', page + 1, min=0, max=int(len(GALLERY_DICT) / per_page))
```

Booplicate:
It doesn't improve readability to me. It's more explicit, but not in a good way.
Also since it's just one of use cases of `SetVariable`, it feels weird having a set of parameters just for that.

renpytom:
I think for this sort of thing, it's best to keep the built-in actions simple, and then let the creators define their own actions (or functions that return actions). 

For example:

```renpy
init python:
    def SetClampedVariable(variable, value, minimum=0, maximum=100):
         value = max(value, minimum)
         value = min(value, maximum)
         return SetVariable(variable, value)
```

Ren'Py is designed to make this really simple (some of the built-in actions are just functions returning actions), and so I think this sort of thing should be encouraged.

via-my:
If the renpy coding philosophy is like so, then I quite understand it. Thanks for your time, guys.

PR #3944, opened 2022-09-04
Allow general iterables in some places where only lists were allowed

Gouvernathor:
`config.tag_transform` and `renpy.show_layer_at` take any iterable of transforms rather than only lists.
Same with the layeredimage's `at` property and argument, the layeredimage layers' `at` property, and the transform parameter to LayeredImageProxy.

The `if_x` properties for layeredimages' layers now take iterables of strings rather than lists of strings. Only this one was granted an str check, because for transforms it makes no sense to use strings anyway.

### DIFF
--- a/renpy/common/00layeredimage.rpy
+++ b/renpy/common/00layeredimage.rpy
@@ -4,6 +4,7 @@ python early in layeredimage:
 
     from store import Transform, ConditionSwitch, Fixed, Null, config, Text, eval, At
     from collections import OrderedDict
+    from collections.abc import Iterable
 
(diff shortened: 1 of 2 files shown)


renpytom:
This doesn't work on Python 2.7, which we'll be supporting for the next year or so. 

Although I don't see that as a problem here, Iterable also matches strings, and I really don't like that. 

What might make sense is to have two functions, renpy.easy.to_list and renpy.easy.to_tuple that take care of the "if something is a list/tuple, keep it, otherwise wrap it" pattern that's pretty common in Ren'Py.

PR #3948, opened 2022-09-05
Use easy.to_list and easy.to_tuple

Gouvernathor:
replaces #3944 
I chose to test for str and not basestring, so that the behavior is exactly the same on py2 and py3, but that's arguable (now it only applies to unicode, to be clear).
The default value of `copy` is also arguable.
Another or_none parameter could be added, to avoid returning [None] in some cases.

### DIFF
--- a/renpy/display/behavior.py
+++ b/renpy/display/behavior.py
@@ -222,8 +222,7 @@ def map_event(ev, keysym):
 
         return False
 
-    if isinstance(keysym, list):
-        keysym = tuple(keysym)
+    keysym = renpy.easy.to_tuple(keysym)
 
     check_code = event_cache.get(keysym, None)
     if check_code is None:
@@ -240,8 +239,7 @@ def map_keyup(ev, keysym):
         if (keysym in ev.eventnames) and ev.up:
             return True
 
-    if isinstance(keysym, list):
-        keysym = tuple(keysym)
+    keysym = renpy.easy.to_tuple(keysym)
 
     check_code = keyup_cache.get(keysym, None)
     if check_code is None:
--- a/renpy/display/behavior.py
+++ b/renpy/display/behavior.py
@@ -222,7 +222,8 @@ def map_event(ev, keysym):
 
         return False
 
-    keysym = renpy.easy.to_tuple(keysym)
+    if isinstance(keysym, list):
+        keysym = tuple(keysym)
 
     check_code = event_cache.get(keysym, None)
     if check_code is None:
@@ -239,7 +240,8 @@ def map_keyup(ev, keysym):
         if (keysym in ev.eventnames) and ev.up:
             return True
 
-    keysym = renpy.easy.to_tuple(keysym)
+    if isinstance(keysym, list):
+        keysym = tuple(keysym)
 
     check_code = keyup_cache.get(keysym, None)
     if check_code is None:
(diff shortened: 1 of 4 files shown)


renpy/easy.py
```diff
@@ -258,3 +263,22 @@ def split_properties(properties, *prefixes):
...
+def to_list(value, copy=False):
+    """
+    If the value is an iterable, turns it into a list, otherwise wraps it into one.
+    If a list is provided and `copy` is True, a new list will be returned.
+    """
+    if isinstance(value, Iterable) and not isinstance(value, str):
+        if (not copy) and isinstance(value, list):
+            return value
+        return list(value)
+    return [value]
```
mal, line 276:
```suggestion
    if isinstance(value, list):
        return list(value) if copy else value
        
    if not isinstance(value, str) and isinstance(value, Iterable):
        return list(value)

    return [value]
```

renpy/easy.py
```diff
@@ -258,3 +263,22 @@ def split_properties(properties, *prefixes):
...
+
+def to_tuple(value):
+    """
+    Same as to_list, but with tuples.
+    """
+    if isinstance(value, Iterable) and not isinstance(value, str):
+        return tuple(value)
+    return (value,)
```
mal, line 284:
```suggestion
    if isinstance(value, tuple):
        return value

    if not isinstance(value, str) and isinstance(value, Iterable):
        return tuple(value)

    return (value,)
```

renpy/display/behavior.py
```diff
@@ -222,8 +222,7 @@ def map_event(ev, keysym):
 
         return False
 
-    if isinstance(keysym, list):
-        keysym = tuple(keysym)
+    keysym = renpy.easy.to_tuple(keysym)
```
mal, line 225: There doesn't appear to be any expectation here that the `keysym` argument wouldn't already be an `Iterable` of some kind, and I don't think we should be accidentally adding more expensive code here to allow it. These are internal functions that already work, in contrast to the proposed changes to `layeredimage` and `at_list` code which is dealing directly with creator input and providing a direct benefit.

renpy/display/behavior.py
```diff
@@ -240,8 +239,7 @@ def map_keyup(ev, keysym):
         if (keysym in ev.eventnames) and ev.up:
             return True
 
-    if isinstance(keysym, list):
-        keysym = tuple(keysym)
+    keysym = renpy.easy.to_tuple(keysym)
```
mal, line 242: Same opinion here as with the proposed `map_event` change.

mal (commented):
I think it would be good in the proposed helpers to optimize for the expected cases of the type being asked for, then `str`, and then `Iterable`; short-circuiting as necessary to minimise the number of checks. Somewhat more subjectively, I'd consider that approach to be much easier to reason about when reading the code. With that in mind, for your consideration:

renpy/display/behavior.py
Gouvernathor: I was under the impression that the keysym directly came from the user, even though the function itself is not public ? I don't know, but if non-list iterables of keysyms are currently accepted as well as lists, then we can surely kick it off.

renpy/easy.py
Gouvernathor: Yes, real instance checks are faster than abc checks, let's do them first.

renpy/display/behavior.py
mal: I didn't get the rationale quite right, but did, I believe, get the right conclusion. The correct explanation is that it comes down to how `compile_event` works, and that converting non-iterables into iterables here needlessly creates more work in that function, and results in more complex dynamic code being generated there.

PR #3959, opened 2022-09-09
Fix `renpy.load_module` not respecting init order

Booplicate:
Before AST nodes would run in the order they were in the script. Now we're sorting them before running them.


renpytom:
Makes sense. Thank you.

PR #3961, opened 2022-09-09
Add `renpy.include_module`

### LINKED ISSUE #3960: Smart loading of renpy scripts
So, `renpy.load_module` exists, but it loads ASTs the moment it's being called. So if I execute it at init -10, all ASTs inside will be executed at -10, even if I had stuff inside for init 500.
I need something that would allow me to "plug in" a renpy script (read module) and then just let renpy load it as usual, along other scripts. Is it possible to achieve? Or can be considered as a possible feature?

Booplicate:
## Main change
This PR closes #3960 by adding a new function `renpy.include_module`. It allows the developer to control which module will be *included* in the script during init (where `renpy.load_module` loads the module right away, "ignoring" the init order).

## Additional changes
This also moves the sort I added in #3959 directly in `Script.load_module` and also adds sort to `Script.load_string`, now they are consistent with `Script.load_script`. Sorry, I missed it the first time.
Since timsort is a stable sorting algorithm (as of Python 2.3), I removed the old code with decorated sort to make it run a bit faster.

## Notes
I had to add `heapq_merge` which looks ugly and scary, but it's just a port from the Python 3.10 sources that works in Python 2.7. It can be removed down the line when you decide to no longer support py2.

If any edits are required, I can do them.  Otherwise if this PR looks good to you, I can add the changelog.

### DIFF
--- a/renpy/main.py
+++ b/renpy/main.py
@@ -550,7 +550,9 @@ def main():
 
         renpy.game.exception_info = 'While executing init code:'
 
-        for _prio, node in game.script.initcode:
+        for id_, (_prio, node) in enumerate(game.script.initcode):
+
+            renpy.game.initcode_ast_id = id_
 
             if isinstance(node, renpy.ast.Node):
                 node_start = time.time()
--- a/renpy/exports.py
+++ b/renpy/exports.py
@@ -2777,7 +2777,6 @@ def load_module(name, **kwargs):
     renpy.config.locked = False
 
     initcode = renpy.game.script.load_module(name)
-    initcode.sort(key=lambda i: i[0])
 
     context = renpy.execution.Context(False)
     context.init_phase = True
@@ -2844,6 +2843,26 @@ def load_string(s, filename="<string>"):
 
     finally:
         renpy.game.exception_info = old_exception_info
+
+
+def include_module(name):
+    """
+    :doc: other
+
+    Similar to :func:`renpy.load_module`, but instead of loading the module right away,
+    inserts it into the init queue somewhere after the current AST node.
+
+    The module may not contain init blocks lower than the block that includes the module.
+    For example, if your module contains an init 10 block, the latest you can load it is
+    init 10.
+
+    Module loading may only occur from inside an init block.
+    """
+
+    if not renpy.game.context().init_phase:
+        raise Exception("Module loading is only allowed in init code.")
+
+    renpy.game.script.include_module(name)
 
 
 def pop_call():
--- a/renpy/game.py
+++ b/renpy/game.py
@@ -98,6 +98,9 @@ persistent = None # type: Any
 # The current preferences.
 preferences = None # type: Any
 
+# Current id of the AST node in script initcode
+initcode_ast_id = 0
+
 
 class ExceptionInfo(object):
     """
--- a/renpy/script.py
+++ b/renpy/script.py
@@ -487,6 +487,8 @@ class Script(object):
 
         stmts = self.finish_load(stmts, initcode, False)
 
+        initcode.sort(key=lambda i: i[0])
+
         return stmts, initcode
 
     def finish_load(self, stmts, initcode, check_names=True, filename=None):
--- a/renpy/script.py
+++ b/renpy/script.py
@@ -410,12 +410,14 @@ class Script(object):
         initcode into the script current initcode
         """
         module_initcode = self.load_module(name)
+        if not module_initcode:
+            return
 
         # We may not insert elements at or prior the current id!
         current_id = renpy.game.initcode_ast_id
 
         if module_initcode[0][0] < self.initcode[current_id][0]:
-            raise Exception("Module %s contains nodes with priority lower than the node that loads it" %name)
+            raise Exception("Module %s contains nodes with priority lower than the node that loads it" % name)
 
         merge_id = current_id + 1
         current_tail = self.initcode[merge_id:]
--- a/renpy/script.py
+++ b/renpy/script.py
@@ -417,7 +417,7 @@ class Script(object):
 
         # We may not insert elements at or prior the current id!
         current_id = renpy.game.initcode_ast_id
-        print("{} < {}".format(module_initcode[0][0], self.initcode[current_id][0]))
+
         if module_initcode[0][0] < self.initcode[current_id][0]:
             raise Exception("Module %s contains nodes with priority lower than the node that loads it" %name)
 
--- a/renpy/script.py
+++ b/renpy/script.py
@@ -374,13 +374,9 @@ class Script(object):
 
             self.load_appropriate_file(".rpyc", [ "_ren.py", ".rpy" ], dir, fn, initcode)
 
-        # Make the sort stable.
-        initcode = [ (prio, index, code) for index, (prio, code) in
-                     enumerate(initcode) ]
+        initcode.sort(key=lambda i: i[0])
 
-        initcode.sort(key=lambda i: (i[0], i[1]))
-
-        self.initcode = [ (prio, code) for prio, index, code in initcode ]
+        self.initcode = initcode
 
         self.translator.chain_translates()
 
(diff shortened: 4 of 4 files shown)


renpy/main.py
```diff
@@ -550,7 +550,9 @@ def main():
 
         renpy.game.exception_info = 'While executing init code:'
 
-        for _prio, node in game.script.initcode:
+        for id_, (_prio, node) in enumerate(game.script.initcode):
```
Gouvernathor, line 553: Maybe I'm stupid, but isn't this code adding things to a list we're already iterating ?
This seems to work in the console, but I'm not sure it's documented and/or reliable. In particular, sets and dicts react badly to that, and references I could find seem to indicate it as a very bad thing to do.
If it is the case, we could solve the problem by iterating with while, and either incrementing an index variable, or doing .pop(0) if we don't care about destroying the list.

Booplicate: Indeed, we're changing the list we're iterating through, but in a smart way! You know the common pattern where you iterate a list in the reverse order to be able to `pop` elements from it? Similar approach is used here, we're adding elements to the list, but we're adding them *ahead* of the current index, which is totally safe.

Booplicate:
Any possibility of adding this to RenPy? My project *really* needs this and I'd like to avoid adding hacks to RenPy.
If backporting `heapq_merge` doesn't feel right, we could avoid using heap on py2 (or even py3 all together), but it'd cause some performance impact due to extra sorting.
Here's a chart to explain the idea visually:
chart
Assuming we're loading a module with 3 nodes (-600, -100, and 600 priorities) at init -600. On the top is how `renpy.load_module` works, on the bottom is how `renpy.include_module` would work.

mal:
This really doesn't feel like the right approach to what is perhaps best described as "conditional `.rpy` loading".

The problem you describe seems roughly equivalent to the concept of feature flagging, something that can be achieved with existing syntax (now whether that's an intentional feature, I'm not sure), never-the-less you can do this (see example below) which seems like it would cover the use-case you described.

<details><summary>Example (click to expand)</summary>

**main.rpy**
```rpy
init python early:                    # set feature/dlc flags in an early block
    flags = {'feature1', 'feature3'}  # turn on features 1 and 3, but not 2
    active_dlc = 2                    # activate dlc 2

label main_menu:
    return                            # skip main menu in this toy project

label start:                          # mini script to show values that have been set
    scene foo
    "feature 1 expected to load.\nbar = [bar]"

    scene qux
    "feature 2 not expected to load. grey missing undefined image qux."

    scene
    "feature 3 expected to load.\nmessage = [message]"

    scene splash
    "DLC? [title]"
    return
```

**feature1.rpy**
```rpy
init -600 if 'feature1' in flags:     # only execute if feature1 flag is set
    image foo = Placeholder(text='feature1-foo.png')
    define bar = 'f1'
```

**feature2.rpy**
```rpy
init 100 if 'feature2' in flags:     # only execute if feature2 flag is set
    image qux = Placeholder('feature2-qux.png')
```

**feature3.rpy**
```rpy
init 600 if 'feature3' in flags:     # only execute if feature3 flag is set
    $ message = 'feature 3 is active!'
```

**dlc1.rpy**
```rpy
init -600 if active_dlc == 1:         # only execute if active dlc is 1
    image splash = Placeholder(text='dlc1/splash.png')
    define title = _('DLC 1.0')

init -100 if active_dlc == 1:         # only execute if active dlc is 1
    python:
        print('DLC 1 is active!')
```

**dlc2.rpy**
```rpy
init -600 if active_dlc == 2:         # only execute if active dlc is 2
    image splash = Placeholder(text='dlc2/splash.png')
    define title = _('DLC 2.0')

init -100 if active_dlc == 2:         # only execute if active dlc is 2
    python:
        print('DLC 2 is active!')
```
</details>

Booplicate:
I actually wasn't aware of the ability to use  `if` with `init`. That's cool.

Although, that's not something I could utilise. It assumes that the rpy knows/controls whether or not it should be loaded, when what I'm proposing allows to load modules from the outside. If we put aside that that's how usually "modules" in different systems work, there's a few things to think about:
 - It'd allow to make a "plug and use" kind of things. I don't need to define and then add all those conditions - I can just load the module when needed.
 - Another reason why I don't want the modules to control the loading is because if I control it, I can make my system more backward compatible - I can adjust some things or not load the module at all. This is probably the most important part for me, as my project is constantly delivers new content to the users.
 - As you could see you have to duplicate your conditions. Now if we're thinking of big modules with 50+ init blocks, you would have to write `if active_dlc == 1` in each of them, along with having to use `init` in the first place (usually you wouldn't need it for screens or transforms). It's kind of cumbersome.
 - There's also labels which I'm not sure could be put inside `init`, at least it doesn't make sense to me.

You say
>  "conditional .rpy loading".

But I'd call it *dynamic loading*. If we could dynamically (aka *when needed*) load any module w/o modifying the code (by adding those conditions), it would mean I could reuse the code in different projects. For example, this could open the door to more proper renpy plugins where right now it's just drop in rpys without any signatures of being external "libraries" basically, or any way to control their loading.

> without introducing the init-mangling and sort overhead in this PR

The overhead should be minimal with heap merge, is there anything I can do to make the merging look more applicable? Maybe change `for` into `while`?

Gouvernathor:
The init-if feature isn't documented, so it would need to get documented before solving the issue. And raises the question : do we want it to be, or is the incluce_module way better ?
I'm keeping an open mind, but I'm tending against the init blocks and/or their combination with other statements, because it's a really bad thing to mix with `init offset` statements (which are now solidly documented).

> that's how usually "modules" in different systems work

It's not the case in python though. Python import works more or less the same as the current load_module.

mal:
Putting aside implementation for a moment, `include_module` also seems like a really _really_ confusing foot-gun for anyone coming to it fresh. Being able to include modules that, depending on when (in the init order) you include them, might execute in their entirety, partially, or not at all seems like a really bad and confusing thing.

I don't think that the proposed `include_module` or the undocumented `init if` functionality are particularly good fits for either Ren'Py or this problem respectively. It kinda feels like we've fallen foul of the XY problem, where we're now only looking at the proposed solution, rather than the underlying feature request. 😦

Booplicate:
> confusing foot-gun for anyone coming to it fresh.

To me the fact that `renpy.load_module` basically ignores init order was confusing. Where `renpy.include_module` actually respects it.

> Being able to include modules that, depending on when (in the init order) you include them, might execute in their entirety, partially, or not at all seems like a really bad and confusing thing.

None of that is true. It's consistent in how it works, you just call the function and it adds the nodes from the module to `initcode`. All of them will always be executed. Unless the code is incorrect (e.g. attempt to load a -500 node at init 0), in that case the developer would get a crash.

> It's not the case in python though. Python import works more or less the same as the current load_module.

Both `renpy.load_module` and `renpy.include_module` are close to how Python modules work. But the latter respects RenPy's init system which makes sense within RenPy. My point was that the proposed alternative (where the module controls whether or not it should be loaded) isn't the case. Not in Python.

mal:
> None of that is true. It's consistent in how it works, you just call the function and it adds the nodes from the module to initcode. All of them will always be executed. Unless the code is incorrect (e.g. attempt to load a -500 node at init 0), in that case the developer would get a crash.

My mistake, I missed the raise when reading through.

Gouvernathor:
Maybe we should be discussing the concept in the issue and the implementation here ? Idk,
I sort-of agree with the motivation and use-case given in the issue (the XY problem you were referring to @mal ) : if you have a multi-priorities init phase, as supported by renpy, the nominal version of adding a code-altering DLC is to add this version of rpym loading. The other would be to add a large number of ifs, which is very unhandy, and partly undocumented (for transforms and images and the like).
So, this could be deemed too much to support for too little benefit, but if not, I think it's the best solution.

renpytom:
I think I like this. 

I'm going to merge it, but then I'm going to remove heapq_merge, as I don't think it's any better than just sorting a second time.

PR #3964, opened 2022-09-09
Allow repair updates from the launcher GUI

Gouvernathor:
Supersedes #3954, if we want to open that option to any creator.


renpytom:
Probably not needed, unless there is some large rash of this that people aren't reporting.

PR #3977, opened 2022-09-16
Give Fade some of Dissolve's args

Gouvernathor:
Document the alpha argument, add a time_warp argument. (could be duplicated though, for the in and out dissolves, but I'm keeping it simple for the moment).
Also I moved the displayable_or_none call for circumstances when it's really useful.

### DIFF
--- a/renpy/display/transition.py
+++ b/renpy/display/transition.py
@@ -204,15 +204,19 @@ def Fade(out_time,
          color=None,
          widget=None,
          alpha=False,
+         time_warp=None,
          ):
     """
     :doc: transition function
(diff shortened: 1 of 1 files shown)


mal:
This seems like a good potential case for creators to use `MultipleTransition`, which would better support (for example) using different warps for the two dissolves (i.e. `easein`, then `easeout` - to give symmetry). In contrast, adding (and documenting) these additional params for the generic `Fade` limits options for any future implementation changes the engine may one day want to make.

renpytom:
I don't think this is necessary, and the time_warp parameter is a bit confusing, so I'm going to leave Fade as-is.

PR #3983, opened 2022-09-18
Documenting the pitfall of animated shaders.

### LINKED ISSUE #3982: Shader transform does not appear to get animation timebase updates when there is no ATL-animated object on screen
Sorry for the long title, but I think that's the most correct description of what is happening. Or at least what I *think* is happening.

This is a script.rpy meant to be dropped into a fresh new project which illustrates the problem, and is the most concise expression of it I could work out:

```

init python:

    renpy.register_shader("testanim",
        variables="""
            uniform vec2 u_model_size;
            uniform float u_lod_bias;
            uniform sampler2D tex0;
            uniform float u_time;
            varying vec2 v_uv;

            uniform vec2 u_speed;
            uniform float u_scale;
            uniform sampler2D tex1;
            uniform vec2 res1;

            attribute vec2 a_tex_coord;
        """,
        vertex_300="""
            v_uv = a_tex_coord;
        """,
        fragment_300="""
            vec2 warp = texture2D(tex1, fract(v_uv + u_time * u_speed), u_lod_bias).rg;
            vec2 st = v_uv + warp * u_scale;
            gl_FragColor = texture2D(tex0, st, u_lod_bias);
        """
    )

init:

    image bg test = Tile("blindstile.png")
    
    transform anishader(child):
        shader "testanim"
        Model().child(child, fit=True).texture("squarestile.png")
        u_speed (0.05,0.05)
        u_scale 0.5

        # This is the workaround.
        # Uncomment this to see the shader animation stop stuttering.

        # zoom 1.0
        # block:
        #     linear 60 zoom 1.000001
        #     linear 60 zoom 1.0
        #     repeat

    transform imageanim:
        align (0.5,0.5)
        alpha 0
        ease 0.2 alpha 1.0
        block:
            ease 1.0 alpha 0.0
            pause 1
            ease 1.0 alpha 1.0
            pause 1
            repeat

    image testimage = At(Text("FOO, I say.", size=80, color="#000"), imageanim)

label start:

    scene bg test at anishader

    "Watch the stuttering shader on the background. It's not supposed to stutter."

    "Now we're going to put up an animated image."

    show testimage at truecenter

    "Notice how the shader animation stutters when there 
    is {i}no{/i} object constantly animated by ATL on screen,
    but runs smoothly when there is -- and starts stuttering again
    when the animation pauses."

    return
```

Initially I thought it was the reverse -- that pauses in ATL (I had a CTC blinking in the corner) cause the shader to stop, which turned out not to be the case. The obvious workaround is to have the shader ATL constantly animate a parameter that does not actually affect the display, which will keep the timebase spinning. However, this strikes me as a very crude solution.

I suspect this is intended behavior, but if so, it would be lovely if there was a legitimate way to override it, and it definitely needs to be documented as a caveat.

**P.S.** Eventually I found [this issue](https://github.com/renpy/renpy/issues/3041), which describes the same situation. As well as the fact that `pause 0` `repeat` is the legit way to handle it and the behavior is intended.

If everyone is too busy to do so, I could send a PR?...

methanoliver:
Fixes #3982 by documenting the cause and recommended solution.

### DIFF
--- a/sphinx/source/model.rst
+++ b/sphinx/source/model.rst
@@ -487,6 +487,28 @@ a built-in shader to create the Dissolve transform::
 Using the Model displayable as the child of a displayable is incompatible
 with :tpref:`mesh`, as the two both create models inside Ren'Py.
 
+Animated Shaders
+----------------
+
+When using shaders that depend on `u_time` to animate, one must be aware, that
+even though every shader on screen will run on every frame displayed, Ren'Py
+does not run on constant FPS, and will fall back to the minimum frame rate of 5
+FPS if no displayables require to be redrawn.
+
+When using an animating shader in an ATL transform, this can cause that shader
+to "stutter" and only animate properly while some other object on screen
+animates as well, in case the transform you're using it in does not cause
+redraws otherwise. In these cases, you can introduce an empty ATL loop to force
+redraws to happen::
+
+    transform fancy_shader:
+        shader 'my_fancy_shader'
+        pause 0
+        repeat
+
+`pause 0` will cycle the frames as fast as possible. You can also set different
+values for `pause` to specify a minimum frame rate, like `pause 1/30`.
+
 Default Shader Parts
 --------------------
 
--- a/sphinx/source/model.rst
+++ b/sphinx/source/model.rst
@@ -490,10 +490,10 @@ with :tpref:`mesh`, as the two both create models inside Ren'Py.
 Animated Shaders
 ----------------
 
-When using shaders that depend on `u_time` to animate, one must be aware, that
-even though every shader on screen will run on every frame displayed, Ren'Py
-does not run on constant FPS, and will fall back to the minimum frame rate of 5
-FPS if no displayables require to be redrawn.
+When using shaders that depend on ``u_time`` to animate, one must be aware,
+that even though every shader on screen will run on every frame displayed,
+Ren'Py does not run on constant FPS, and will fall back to the minimum frame
+rate of 5 FPS if no displayables require to be redrawn.
 
 When using an animating shader in an ATL transform, this can cause that shader
 to "stutter" and only animate properly while some other object on screen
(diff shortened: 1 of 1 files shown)


sphinx/source/model.rst
```diff
@@ -487,6 +487,28 @@ a built-in shader to create the Dissolve transform::
...
+
+When using shaders that depend on `u_time` to animate, one must be aware, that
+even though every shader on screen will run on every frame displayed, Ren'Py
+does not run on constant FPS, and will fall back to the minimum frame rate of 5
+FPS if no displayables require to be redrawn.
+
+When using an animating shader in an ATL transform, this can cause that shader
+to "stutter" and only animate properly while some other object on screen
+animates as well, in case the transform you're using it in does not cause
+redraws otherwise. In these cases, you can introduce an empty ATL loop to force
```
Gouvernathor, line 501: "In this case, an empty ATL loop can be introduced..." I would prefer the passive form

sphinx/source/model.rst
```diff
@@ -487,6 +487,28 @@ a built-in shader to create the Dissolve transform::
...
+redraws to happen::
+
+    transform fancy_shader:
+        shader 'my_fancy_shader'
+        pause 0
+        repeat
+
+`pause 0` will cycle the frames as fast as possible. You can also set different
+values for `pause` to specify a minimum frame rate, like `pause 1/30`.
```
Gouvernathor, line 510: 1/30 will return 0 in py2. I'm not sure we care, though (it wouldn't break the transform, just make it redraw faster).

Gouvernathor (commented):
I think this is fine

sphinx/source/model.rst
methanoliver: I suppose. Fixed.

sphinx/source/model.rst
methanoliver: No reason not to make it universal by writing `1.0/30`...

Gouvernathor: It's less classy, but works better 🤷 However you want.

sphinx/source/model.rst
```diff
@@ -487,6 +487,28 @@ a built-in shader to create the Dissolve transform::
 Using the Model displayable as the child of a displayable is incompatible
 with :tpref:`mesh`, as the two both create models inside Ren'Py.
 
+Animated Shaders
+----------------
+
+When using shaders that depend on `u_time` to animate, one must be aware, that
```
Gouvernathor, line 493: Oh, I forgot : in shinx you need double-backquotes for `fixed-width text`, not simple backquotes like in markdown. This needs to be \``u_time``, and so on.

methanoliver: Right, single quotes are italics... Fixed.

PR #3986, opened 2022-09-19
Persistent devtool

Gouvernathor:
Very similar to the existing variable viewer, I'm sure there is a way to simplify the code and make it less autosimilar.

The persistent._keys() method could get documented.
If this is merged, I'll add a `__contains__` method to persistent in a separate PR, which will allow `"key" in persistent` to check what `hasattr(persistent, "key")` fails to.


renpy/common/_developer/developer.rpym
```diff
@@ -208,6 +210,66 @@ label _debugger_screen:
...
+    call _enter_game_menu
+
+    python hide:
+        import reprlib
+        aRepr = reprlib.Repr()
+        aRepr.maxstring = 120
+
+        entries = []
+
+        for vn in sorted(persistent._keys()):
```
mal, line 259:
```suggestion
        for vn in sorted(persistent.__dict__):
```

This would alleviate the need to alter `renpy/persistent.py`, and allow the display of variables with a leading double underscore as line 263 seeks to do.

Gouvernathor: The screen and label are mainly copypaste from the variable viewer that I edited the least possible to make replacing them easier. But I don't intend any underscored things to be visible. I'll probably comment these lines.
Unless you have a reason to make double-underscored elements visible, but since the persistent inner process doesn't treat them normally (see former line 69) I don't think we should support it.

mal: While probably unintentional, it's possible use a leading double underscore to set an (as per line 69) persistent value that won't be cleared unless done so explicitly, there are reasonable uses for that, especially for games that provide the ability for users to reset their persistent data ("reset statistics" type of thing) where it makes sense to clear practically but perhaps not quite all data.

It's certainly not a "take it under our wing and say it's supported" thing, so much as a "it currently works and there's no real need to break it" thing, and with that being the case showing such values in the inspector wouldn't seem to do any harm while being potentially useful for those making use of said quirk. That it also aligns with the existing logic of the current variable viewer is a happy bonus.

Gouvernathor: I disagree about showcasing undocumented things in a devtool. If you want to document the __ thing (with the name-mangling mechanism on top of it as always), it should be done separately.

mal: IMO hiding values that could exist for purely idealogical reasons seems to run counter to it being a tool to help creators.

Booplicate: As a developer, I'd like to be able to see everything via the dev tool so I can debug every possible case. I think it might be beneficent even for renpy itself.

renpy/persistent.py
```diff
@@ -118,6 +113,12 @@ def _update(self):
                 "_seen_translates" : 0,
             }
 
+    def _keys(self):
+        """
+        Returns the set of keys which the creator should be aware of.
+        """
+        return [k for k in self.__dict__ if not k.startswith("_")]
```
mal, line 120: This function doesn't seem to add much besides materialisation of a list that doesn't ever need materialising.

Gouvernathor: I suggest you read the opening message of the PR before making suggestions. The idea is to make it accessible for creators, either directly or through `__contains__`. And in the meantime, to factorize code.

renpy/common/_developer/developer.rpym
Gouvernathor: Some persistent values are used internally and should not be visible for creators, for the same kind of reasons the `config`, `preferences` and so on stores are not visible in the normal variable viewer.
If you want to start considering __-started persistent variables as a documented features for creators, then create a PR proposing it, documenting it and adressing how it works with the name-mangling mechanism (which is all but simple to understand).
Until then, we should not advertize or facilitate the use of undocumented features.

renpy/persistent.py
mal: I did read it, but even if I did see the value in giving persistent a `__contains__` method (I don't, a key point of `persistent` is that unset attributes come back as `None`), adding a method that materialises a list isn't necessary to support it, or any thing else in this PR.

renpy/common/_developer/developer.rpym
mal: I think you're conflating two things here. Just because a something isn't documented isn't a reason to hide it, it's a reason to stay neutral/unopinionated about it. Actively hiding things because support for them isn't documented isn't helpful to anyone, and having it be displayed isn't any form of documentation or binding agreement to support it, especially in a devtool. It's just allowing people to see what values they put there, even if nothing said they could (because it's not documented).

renpy/persistent.py
Gouvernathor: A list was already created before my edits, so this is not a regression.
The reason I won't change it is because not creating a list (returning a parenthesized iterator for example) would yield a risk (in later separate uses of the function) that the contents change between the function call and the iteration, and/or related to the single-use nature of an iterator, which are not obvious bugs.

mal: I'm not suggesting that it should return an iterator, simply that the function doesn't need to exist at all.

renpy/common/_developer/developer.rpym
Gouvernathor: I understand your argument, but I also disagree with it. I sort-of agree the __-starting keys should be adressed one way or another, but it won't be in this PR.

renpytom:
I'd want to show fields starting with __, as that's usually a user-owned pace. And I'd want to hide fields starting with _, since those are often big and uninteresting.

mal:
I don't think it's possible to propose a sub-pull request, so rather than creating a separate issue I'll just link the commit containing some proposed changes here: https://github.com/mal/renpy/commit/fe6277dcfaa180b326a021f9db03637181f46ff1.

If accepted, it can be cherry-picked or I can push it to this branch on the `renpy` repo.

- Changes an internal label name to better reflect it's purpose - `_debugger_screen` -> `_variable_viewer`.
- Aligns the new label with the change above - `_persistent_explorer` -> `_persistent_viewer`.
- Adds a `title` argument to the `_variable_viewer` screen.
- Removes the `_persistent_data_viewer` screen in favour of reusing the existing `_variable_viewer` screen.
- Updates the (now) `_persistent_viewer` label to align with Tom's comment on `__` variables, and more closely mirror the code flow in (now) `_variable_viewer`.
- Reverts changes to `persistent.py` to keep separation between business and display logic.

Gouvernathor:
Closed in favor of #4195

PR #3989, opened 2022-09-21
Document some quality-of-life functions in devtools

Gouvernathor:
Some functions from the Other page could end up there too, I'm thinking of split_properties for example.


Gouvernathor:
I'm not convinced this is any more useful than functools for creators, since the Partial object doesn't participate in rollback. So, better not add a feature burden on future versions of renpy.

The PR and the branch can be resurrected later if someone wants to defend it.

PR #3994, opened 2022-09-24
Pylint

Gouvernathor:


renpy/atl.py
```diff
@@ -1183,11 +1183,6 @@ def execute(self, trans, st, state, events):
 
         warper = warpers.get(self.warper, self.warper)
 
-        if trans.atl.animation:
```
Gouvernathor, line 1193: It's very weird that this statement ends up not being used, while it was before we fixed the ATL fast-forward

renpy/translation/__init__.py
```diff
@@ -945,7 +945,7 @@ def detect_user_locale():
         if isinstance(locale_name, bytes):
             locale_name = locale_name.decode("utf-8")
 
-        local_name = locale_name.replace("-", "_")
```
Gouvernathor, line 950: This should be checked. The `local_name` variable wasn't used, which is weird, but this assignment on the `locale_name` variable changes the behavior of the function.

renpy/translation/dialogue.py
```diff
@@ -311,7 +311,7 @@ def get_strings(self):
             if self.notags:
                 s = notags_filter(s)
 
-            what = combine_filter(s)
```
Gouvernathor, line 314: The combine_filter function is pure, so this seems very much pointless.

mal:
This reads like a half automated, half manual changeset, and it's not always clear which is which (the commit messages don't offer much insight). There are also several inconsistencies between different sections of changes, and subtle behaviour changes to existing code, which may ultimately be fine, but they are completely buried in the diff. :/

Gouvernathor:
This is all manual (hence where some of the inconsistencies probably come from), based on pylint results. For example I sometimes left an unnecessary return, or left functions with inconsistent return values, because it made more human sense when reading the code or because the risk of breaking something was too big.

There is one behavior change I flagged in comments which is probably a typo correction but should be checked. I remember another where I removed an `== True`, and in that case I studied the function to decide whether it should be removed or replaced with `is True`. And there's some `a = a - x` I replaced with `a -= x`, and one or two cases of argument-passing I knowingly simplified.
If you see others, leave a review and we can study it.

Theoretically the different types of changes were ordered in different commits so that if some were considered too dangerous the more "no-brainers" could be cherrypicked.

renpytom:
Can I have the changes that actually fix things? I'm not likely to take a large diff like this just to shut a tool I don't use up - my first instinct is that this is too risky for too little benefit.

renpytom:
I'm reopening, as there are probably things worth merging here, but certainly not the whole thing.

Gouvernathor:
82e8bc347cf5a11050b80a694045d2d57db24a9c, 8ec35e949d60a744a87e7817c60cad1ec371c91b and 8e188cc5b6d7d074a2954523c4a69caf2a17a0b0 just improve code readability, and while not absolutely necessary, it's very easy to check they're safe
fae053e0205e218c98ce0084bef2572643eb5baa also clarifies the code, and while a bit less easy to check (but I checked it already), it goes a long way improving readability
73c85754a9531d4d248e613e39ca0ec8fd9f7f76 removes an ambiguity with another variable (the parameter) improving readability
038e86bedbaf029e2d26d2bcd9fbc71416580732 is removing a bare except and making it except Exception instead, so basic stuff
9b884d3e01a1585ad95831d78084c09a5aab75fa removes unnecessary operations (like calls to .keys()) and optimizes loops, while it probably deserves checking (and I did check it before commiting it), it has actual benefits
2904149287655e5315c17967b058912f71e5df78 is some reimports of modules already imported in the global scope
58a29501542f2433c7f9d7d53d4b65a7868a5828 some protection over opening files to read/write them
bde9d18cbcb2f017d0e29f3a81c4a16fd2db83d4 this one is actually wasteful of memory, read the commit comment
7edb69b64d5e528616478dd3753128881926a595 improves code readability, revealed one definite error (the exception thing at the end), one probable error (the `local_name` thing), one possible error (the atl thing) and a waste of time as a pure function being called for no reason
e7d14e381e5705a25cbe00eb1ebc72605d9ed2b8 corrects a dangerous practice of having mutable objects in function signatures, in cases which I checked were safe (there are others in the code which I didn't change), or using the `=None` method which is safe
250b4328be03791cdba50df74950e837d3851752 this one is wasteful in processing time, parsing a constructor call is worse than parsing a literal
1d088175d04f7753b78fe2181fd4213b35bbbdb8 this one is wasteful too by uselessly converting a generator into a list before passing it

PR #3996, opened 2022-09-25
More documentation for named stores

Gouvernathor:


Booplicate:
May I suggest documenting being able to do `python early in some_storemodule:`. It can be handy, but it's not straightforward.

Gouvernathor:
I don't know how that's possible, because currently the substores don't exist until the end of the parsing phase. In any case, this should probably be discussed in a new feature request issue.

PR #4012, opened 2022-10-06
allow: `textsubsegment_callbacks` for `FontGroup`

qTich:
This allows you to manipulate `text.TextSegment` relative to the fonts used in `font.FontGroup`.
Let's say one font in paragraphs may be larger than another, but simply changing the styles won't help here.
Example:
```python
def _ts_callback(segment, ss):
    if segment.font in ("example.ttf", ):
        segment.size *= 0.9
    return segment
```

Using a language for the key is justified by the fact that fonts are overwhelmingly used differently for each language.

### DIFF
--- a/renpy/config.py
+++ b/renpy/config.py
@@ -1285,6 +1285,10 @@ viewport_drag_radius = 10
 # function is run.
 scene_callbacks = [ ]
 
+# This is a dict of list with functions that accept the key to which is language,
+# (object of `text.TextSegment``, phrase by `font.FontGroup.segment``),
+# Only called when using `font.FontGroup` instead of a font.
+textsubsegment_callbacks = { }
 
 del os
 del collections
--- a/renpy/text/text.py
+++ b/renpy/text/text.py
@@ -326,16 +326,25 @@ class TextSegment(object):
             return
 
         segs = { }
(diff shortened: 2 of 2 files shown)


Gouvernathor:
I think documenting config.ftfont_scale or adding a more general size-tweaker mapping would be a better solution than general callbacks.

qTich:
> I think documenting config.ftfont_scale or adding a more general size-tweaker mapping would be a better solution than general callbacks.

The size is just an example, fonts may be used the same for different languages, but the initial font size itself may be different. As far as I understand `ftfont_scale` does not do this, it uses zooming relative to the base text size (maybe I'm wrong).

The main purpose of this pool reqest is to have a direct ability to manipulate `TextSegment` if needed, and also to be able - based on the string - to check if it makes sense to do so.

Gouvernathor:
I understand that the callbacks are more powerful than a simple zoom/size change (which ftfont_scale or your example do), but I don't think there's a need for such a tool. Especially I don't see why it should only apply on FontGroups.

renpytom:
This is not something that can be added as a documented API - it's too dependent on the current implementation of Ren'Py, which might change. (Font handling has changed multiple times in the past.) What's the actual requirement here?

PR #4013, opened 2022-10-08
Fix RENPY_DEPS_INSTALL docs for armhf, ppc64, and ppc64le

JeremyRand:
The correct triplets for various arches are:

- `x86_64-linux-gnu`
- `arm-linux-gnueabihf`
- `powerpc64-linux-gnu`
- `powerpc64le-linux-gnu`

The `uname`-based instructions produced the following:

- `x86_64-linux-gnu` (correct)
- `armv7l-linux-gnu` (wrong)
- `ppc64-linux-gnu` (wrong)
- `ppc64le-linux-gnu` (wrong)

`gcc` produces the correct results for all four.  I suspect this will also fix non-GNU triplets such as musl, but didn't test those.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -80,7 +80,7 @@ Next, set RENPY_DEPS_INSTALL To a \:-separated (\;-separated on Windows)
 list of paths containing the dependencies, and RENPY_CYTHON to the name
 of the cython command::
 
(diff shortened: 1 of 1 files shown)


JeremyRand:
These instructions were last touched in https://github.com/renpy/renpy/pull/2408 , CCing @jackmcbarn in case he has comments on this change.

PR #4021, opened 2022-10-13
Update lifecycle.rst

Gouvernathor:
My revision of 1883815cc58df786257606abf4379c6a7672461b.

The unspecific term "restart" is replaced with "returning to the main menu". Some minor typos are also corrected.


mal:
I'm not sure this really an improvement. Restart, reload, and relaunching mean different and specific things. Replacing a term everywhere with it's definition doesn't seem especially helpful. I'd argue that part of the value of life cycle page should probably be to teach and help creators understand those differences.

Gouvernathor:
There is a simple action called Start, which everybody knows and uses. The fact that the "restart" thing doesn't include it, and is even followed by it in the most general case (unless there is no main menu) is absurd, and the sign that either one should not be named that way.
Now, which should it be ? On the one hand, start is used by one action every game needs to use - save from returning from the main menu label, but it's very unusual - and which is worded that way in the default gui, and on the other hand there's only the full_restarr function which is the least in use among all the ways of "restarting"-as-you-mean-it. That's the choice I see.
We can explain just the same things to creators, I'm just asking that we don't use troublingly similar names for dissimilar actions. If you have a better term we can use it, but "returning to the main menu" seems easy enough to me, and includes a nod to the "return" instruction which may be used to trigger it.

renpytom:
My feeling on this is that the current terminology can be improved, but "returning to the main menu" isn't the right phrasing. There's a point before the main menu, after which the default variables get their values, that is where we return to. I wasn't able to come up with a great name for that point. 

Ideas?

mal:
I think "restart" is fine, but maybe we can include a definitions section to explicitly define what a "restart" vs a "reload" is in the context of the lifecycle. Something like:

Term | Definition
-|-
Restart | Resets the game to just before the main menu is shown for the first time, abandoning any play-through in progress. Triggered by a `return` statement when the call stack is empty, explicitly calling `full_restart`, or using the `MainMenu` action.
Reload | The process begun by `Shift+r` which creates a temporary save, reparses the script, and loads back to the same place in the play-through. _See [auto-reload](#) for more details._

My other thought was that it could potentially be called a "reset", somewhat akin to the Reset buttons on older consoles, but I don't think it delivers enough benefit to muddy the water with things like `full_restart` already existing. Better to just stick to the one term for it (more or less) everywhere, and help readers understand what that means to Ren'Py in the context of the lifecycle.

renpytom:
I don't like "return to the main menu", simply because, for example, one  can write:

```
label main_menu:
    return
```

and  we'll never get back to the main menu. I'm  going to close this PR.

PR #4027, opened 2022-10-17
fix & changes: bypassing exception on `MP.save`, some functions, docs for `MP`

qTich:
This poolrequest includes a workaround of the error when trying to save, and some changes and additions inside `renpy.persistent`, namely they deal with the interaction with objects of the `_MultiPersistent` class.

### DIFF
--- a/renpy/__init__.py
+++ b/renpy/__init__.py
@@ -257,7 +257,7 @@ name_blacklist = {
     "renpy.audio.audio.lock",
     "renpy.audio.audio.periodic_condition",
     "renpy.webloader.queue_lock",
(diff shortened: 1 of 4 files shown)


Andykl:
I think instead of `WeakSet` and `get_MP` you could use `WeakValueDictionary` with `fn` as keys.
Overall, the changes make sense.

mal:
It'd also be nice to keep the convention of `rv` being used to hold and build the return value (rather than `MP_instance`), to keep consistency with the rest of the codebase, but definitely on the nit side of things. 😉

renpytom:
Merged, and I'm going to then change MP_instance to rv.

Issue #4019, opened 2022-10-11
Keeping the voice from interrupting when opening the navigation


Noriverwater:
This is a feature request.

I've noticed that when playing voice, if the menu or history screen is open, the voice breaks. Some people find this less convenient, and I would like to avoid this, just like audio in the music channel. Many VNs also have this feature. Please add a configuration variable to enable this feature.

renpytom:
What do you mean by "breaks". It should stop playing when you enter the menu.

Noriverwater:
Sorry I didn't explain it clearly. The voice will stop when the navigation screen or the history screen is opened. But some people who play my game feedback that this is not a smart design, because they often can not listen to a character's full voice, and have to use the replay button in the quick menu or rollforward to the next dialogue and use the replay button in the history screen (because the history screen does not include the current dialogue).

Gouvernathor:
(talking about the commit) maybe the name or the explanations should be changed a bit, because there's a confusion between the choice-menu, triggered by the `menu:` statement, and the navigation-menu, the one opening in a new context.
We should make clear that this is about the latter rather than the former (iirc).

renpytom:
Agreed.

PR #4029, opened 2022-10-18
Allow the voice to not stop playing when opening the menu

Noriverwater:
Attempt to implement https://github.com/renpy/renpy/issues/4019.


renpytom:
3c8a31cff supersedes this, and uses a preference rather than a config variable. Thank you for posting this.

PR #4037, opened 2022-10-23
fix: verify on exists directory

qTich:
The save directory can be deleted after the game is started, and thus there are no checks for this.


Andykl:
I don't think it make sense. It's pretty unlikely that the save folder will be deleted after the game has been started. The problem could be antivirus software that blocks the folder from being written to, so I think you should do same as #4022 but for the first file write.

PR #4039, opened 2022-10-24
Fix bugs on steam.get_session_ticket

pazkero:
Without these changes imposed by Python3.9, calling get_session_ticket() raises two exceptions.

Testing Done: Yes

First Exception: ticket is not defined
Second Exception: Cannot slice non-integer

Steps to reproduce: Call get_session_ticket().


Gouvernathor:
How does this behave in py2 ?

renpytom:
Should be the same. Merging - thank you.

PR #4043, opened 2022-10-26
incompatible changes link in the update screen.

Neyunse:
I included a link about incompatible changes, this way it is more accessible from Ren'Py.


renpytom:
I don't love this, as it means all the links will need to be retranslated. If we can do it without retranslation, it's fine.

Neyunse:
> I don't love this, as it means all the links will need to be retranslated. If we can do it without retranslation, it's fine.

Oh, I see... Hmm, I can't think of a way for this(for the moment) …. I guess I'll close it, if I find a way, although maybe we could include some icon for both version notes and incompatible notes so no text would be needed for this.

PR #4053, opened 2022-10-26
Pylint - Remove useless global statements

Gouvernathor:
From #3994

### DIFF
--- a/renpy/webloader.py
+++ b/renpy/webloader.py
@@ -196,8 +196,6 @@ def enqueue(relpath, rtype, data):
 
 
 def process_downloaded_resources():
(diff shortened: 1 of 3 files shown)


renpy/webloader.py
```diff
@@ -196,8 +196,6 @@ def enqueue(relpath, rtype, data):
 
 
 def process_downloaded_resources():
-    global queue, to_unlink
```
renpytom, line 199: This turned out to be wrong for queue.

Gouvernathor: image
Strange, it works out for me, in a template case. Where does the error trigger ?

PR #4056, opened 2022-10-26
Pylint - Rephrase some expressions

Gouvernathor:
From #3994
Not transparent to check, yes, but I did check them and they vastly harmonize the code and improve readability


renpytom:
These are a lot of changes to working code, and I'm not sure any of these are improvements.

PR #4057, opened 2022-10-26
Pylint - Use min and max instead of if/else

Gouvernathor:
From #3944
If/else repeats variable loading and wastes computation time, but can be considered easier to read.


renpytom:
Min and max do function calls, outside of Cython. So I don't think this is a good reason to change things.

PR #4083, opened 2022-11-08
Fix subsurface offset when oversampling

Gouvernathor:


Andykl:
If `el / oversample` is not an integer fraction will we see a less-by-one render bug? Perhaps we need to use `subpixel_blit`

Gouvernathor:
If subpixel_blit requires more computation time or memory as I assume it does, maybe it's not a good idea to use it, but the offset could be passed into round to avoid bugs

PR #4093, opened 2022-11-12
Make extension load order reliable

Ayowel:
I know ren'py extensions are considered unused, but I have a few use-cases for Ren'Py augmentations/core changes/non-intrusive features and not being able to re-use code without copy-pasting because I can't rely on the extensions' load order is a pain. Could this be merged to make them more reliable ?


renpytom:
Merged.

PR #4101, opened 2022-11-18
Add a lint blurb to the launcher's splash

Gouvernathor:
I'll translate it when we agree on the wording.


renpytom:
Perhaps "Lint checks your game for potential mistakes, and gives you statistics." 

I don't like the word "code" in documentation and interface, and the exclamation mark seems unwarranted.

PR #4107, opened 2022-11-22
Update place method of render

valery-iwanofu:
Method Render.place() now returns d.place() result

### DIFF
--- a/renpy/display/render.pyx
+++ b/renpy/display/render.pyx
@@ -1554,7 +1554,7 @@ cdef class Render:
 
             render = renpy.display.render.render(d, width, height, st, at)
 
-        d.place(self, x, y, width, height, render, main=main)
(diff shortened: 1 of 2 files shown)


renpytom:
Could you explain why you'd want this?

valery-iwanofu:
> Could you explain why you'd want this?

Because it will be easier to use in implementation of your own containers.
Instead of:
```renpy
    rv = renpy.Render(width, height)
    child_render = renpy.render(child, width, height, st, at)
    child_pos = child.place(rv, x, y, width, height, child_render)
    offsets.append(child_pos)
```
It will be much easier to write:
```renpy
    rv = renpy.Render(width, height)
    child_pos = rv.place(child, x, y)
    offsets.append(child_pos)
```

renpytom:
Makes sense. Can you update the documentation to mention this change?

PR #4123, opened 2022-11-30
fix to_unlink changes size during iteration error

Vadim-Karpenko:
When checking for outdated media in the to_unlink variable, if you click through images fast enough, sometimes it throws an error `RuntimeError: dictionary changed size during iteration` because of new items being pushed to the dictionary. Iterating through the copy of to_unlink prevents this error.

### DIFF
--- a/renpy/webloader.py
+++ b/renpy/webloader.py
@@ -250,8 +250,8 @@ def process_downloaded_resources():
     # Due to search-path dups and derived images (including image-based animations)
     # files can't be removed right after actual load
     ttl = 60  # remove after 1mn - if your animation is longer than that, use a video
-    for fullpath in to_unlink.copy():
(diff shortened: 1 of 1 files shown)


Gouvernathor:
I would advise writing `for fullpath, value in tuple(to_unlink.items())` and then use `value` instead of `to_unlink[fullpath]`. It would speed up execution and avoid redundent accesses.

Vadim-Karpenko:
> for fullpath, value in tuple(to_unlink.items())

Yeah, I agree, it should improve performance a bit, I made a commit based on your suggestion. Furthermore, I think we could also move the time.time() outside of the loop into a variable instead of calling it every time. What do you think?

Gouvernathor:
Yes, I think that's fine too.

Andykl:
LGTM

PR #4126, opened 2022-12-03
web: Generate PWA icons and prepare files

Vadim-Karpenko:
Support of Progressive Web Application (PWA) in RenPy requires the generation of a bunch of icons for a variety of devices.
This code needs a single icon image (in png format) of 512x512 or bigger. It looks for a file called `pwa_icon.png` in the project directory. If such an icon does not exist, it will look for the default one from the web folder (ref: https://github.com/renpy/renpyweb), which is also optional because it requires an additional pull request to the renpyweb repository to add this file. Then this file is scaled down progressively to all required sizes and saves everything into the icons subfolder in the build directory, so we can use it later for actually building PWA in the `index.html` file.

To test this code, you can use this image:
pwa_icon

### DIFF
--- a/launcher/game/web.rpy
+++ b/launcher/game/web.rpy
@@ -351,6 +351,10 @@ init python:
 
         # Replace the project name with the ones in the game
         manifest = manifest.replace("Ren'Py Web Game", p.dump['build']['display_name'])
+        screen_size = p.dump.get("size")
+        # If width are smaller than height, set the orientation to portrait. If not, leave it as is.
+        if screen_size[0] < screen_size[1]:
+            manifest = manifest.replace("landscape-primary", "portrait-primary")
 
         # Write the file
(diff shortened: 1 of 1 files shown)


Gouvernathor:
Maybe there should be a check on whether the exported icons already exist ?

Vadim-Karpenko:
> Maybe there should be a check on whether the exported icons already exist ?

The build folder gets cleared on every build by this code in the `build_web` function:
```
# Clean the folder, then remake it.
if os.path.exists(destination):
    shutil.rmtree(destination)
```
so there's no point in checking if icons are already present.

Vadim-Karpenko:
As [https://github.com/renpy/renpy-build/pull/51#issuecomment-1340123773](https://github.com/renpy/renpy-build/pull/51#issuecomment-1340123773) noticed, we are lacking maskable icons, so I made an update to generate them, by adding empty space to the corners.

PR #4136, opened 2022-12-07
doc: movies are now supported in web since 8.1

qinlili23333:
as renpy 8.1 support movies in web, we should update text description.

since i can only use English and Chinese, only origin text and Chinese translation in this pr, other language translations need further modification.


Gouvernathor:
I'd phrase it as "Movies have limited support". As far as I understand, the Movie displayable still isn't supported, only the cutscene way is.

renpytom:
I like the idea of this, but I think I need to do a larger rewrite, with the many improvements to the web build in the past few releases. (Audio glitches are a lot less common now, too.)

renpytom:
c105989 takes care of this.

PR #4141, opened 2022-12-09
Add colon warning in the noblock lexer error

Gouvernathor:
This is the lazy version of fixing the issue. The parser calls expect_noblock() so many times (52) it's very hard to grasp how it could be segregated more precisely.


renpytom:
I think I prefer this one, as less of a maintenance burden.

Issue #4143, opened 2022-12-09
Actions shouldn't be documented as classes


Gouvernathor:
This is a consequence of #4132 : despite not being explicitly worded in our explanations, all classes not opted-out with `:doc: something function` will be tagged explicitely as a class by the new (fixed) sphinx parser.
![image](https://user-images.githubusercontent.com/44340603/206598995-2bb55ec4-1a8f-4d31-af1a-e08e0cd07086.png)

All Actions (except the Action class itself and possibly specific classes) should be artificially marked as functions, I think, so that creators don't feel invited to subclassing them or to ask why specific ones are real functions.
Possibly other things than actions should get opted-out of that prepending "_class_", like Values.
It should be then checked if the `:class:` references still work after that (they should).

Gouvernathor:
Fixed in c404490

PR #4144, opened 2022-12-09
Classify actions as functions

Gouvernathor:
Due to #4143 

This reserves all `:doc: *action` as classifying actions, and categorizes those as functions regardless of their true nature.
Overrides like `:doc: somethingaction class` still take precedence over this added behavior.

This works well. Existing `:class:` references still work and link them properly just as before.

This does not cover Bar and Input Values however, which are still classified as classes. And some actions like Replay, EndReplay, gui.SetPreference, gui.TogglePreference and StylePreference are not covered, because their include name does not end with "action".
So either we don't care about them and leave them as classes, or we override them specifically with `:doc: something function`, or we rename their include name to end with "action", or we don't care at all and cancel this PR. But overriding them specifically is kinda the worst option in my opinion.

### DIFF
--- a/sphinx/game/doc.py
+++ b/sphinx/game/doc.py
@@ -212,11 +212,16 @@ documented_list = [ ]
 def scan(name, o, prefix="", inclass=False):
 
     if inspect.isclass(o):
-        doc_type = "class"
-    elif not inclass:
-        doc_type = "function"
-    else:
+        if issubclass(o, (renpy.store.Action,
+                          renpy.store.BarValue,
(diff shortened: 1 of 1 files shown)


renpytom:
Could we add more include names to this list?

Gouvernathor:
Yes, we could look for `section in (...,)`, a specific list of hardcoded include names.

Gouvernathor:
I opted to this version, which was tested and seems to work as intended.

PR #4145, opened 2022-12-09
web: auto mute video for autoplay

qinlili23333:
Usually if a game starts with a video, the video is just a series of logos without sound. So why not auto mute the video if no user click instead of show a prompt to request user click? This can provide a better experience.
This change have no impact on in-game videos since in-game videos must be played after at least one click interaction so won't be muted. Only splash videos may be muted for autoplay. If user clicked video in muted status, the sound will be enabled again.

(I have already applied this change to my site and works well, u can test this feature at https://zhijiang.qinlili.bid/play ,which is a game starts with a splash video)

### DIFF
--- a/renpy/common/00webvideo.rpy
+++ b/renpy/common/00webvideo.rpy
@@ -2,7 +2,7 @@ init -1100 python hide:
 
     config.web_video_base = "./game"
 
-    config.web_video_prompt = _("Touch to turn on volume.")
+    config.web_video_prompt = _("Touch to play the video.")
 
     if renpy.emscripten:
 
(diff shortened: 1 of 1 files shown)


renpytom:
Did you try this on iOS? The prompt is mostly because iOS wouldn't play at all without a click.

qinlili23333:
I can't test on iOS since all of my projects use VP9 as video codec which is not support by iOS. 
But according to apple's document this should work.( https://developer.apple.com/documentation/webkit/delivering_video_content_for_safari#3030251 )
Maybe need someone to test it.

qinlili23333:
I think maybe we can try mute and play first, and if still got error, then fallback to the prompt. This can improve most browsers' user experience and keep the compatibility.

Issue #3541, opened 2022-05-20
Improving the console UX


Numerlor:
When using the console I sometimes find myself lacking some QoL features

### Output shortening

Output truncation into the console is shortened unless long console output is enabled with the `long` command; this can be a bit awkward to use as it needs to be set and reset each time if it's not necessary for the entire session. I think a new function that'd go around the shortening could be introduced as an alternative to print.

The configuration for the sizes of what the output is shortened to could also be exposed to the user instead of being hidden away as class variables

### History

The command history is currently relatively rudimentary where multiple instances of the same input are present inside of it multiple times, and editing of previous multi line inputs is not possible.
For the duplicates, the addition to history can be skipped if the entry is already at the end, or maybe even moving existing equal entries to the end as there should be no reason for the user to want it both at the end and someplace earlier

Auto completion from history could also be nice, but I'm not sure how viable this is with a screen

-----------
Looking at the file it could also use some cleanup.

If there's any interest in these, I'd like to implement it

Booplicate:
Can I also suggest better paste in console? Right now it removes `\n` which makes impossible to paste multiline code.

Numerlor:
Am I free to work on this, or does there need to be further discussion around the issue?

Andykl:
In my project I've modified the console to suit my needs, and this includes some suggestions from here (improving the console history and inserting multilines) but also UI redisign so it is VBox(history viewport, input) and I also almost implemented simple TAB-completer. I can publish these changes and the work will continue from there, so we need to finish TAB-completer and add improve output shortening (probably add `pp` command to pretty-print the thing? But it is Py3 only thing) and probably some code cleanup.

If @renpytom is okay with the console changes, I'm willing to review your work.

Gouvernathor:
The part about history and duplicates was adressed in #4150, in our version which is to allow non-successive duplicates (rationale introduced in 4147).

Andykl:
I think after Ren'Py is updated to Python 3.13 we could take over its improved REPL and just wrap Ren'Py screen around it. That will add some cool features like multiline paste and tab-completer.

PR #4147, opened 2022-12-09
Prevent successive identical entries in console history

Gouvernathor:
Duplicates "at all" are not prevented, because I'm not convinced that would be a good thing. The linux vanilla console works like that, preserving order between entries, avoid it being spammed by repetitive similar entries, and incidentally allowing more different entries to be stored.
Related to #3541.

### DIFF
--- a/renpy/common/00console.rpy
+++ b/renpy/common/00console.rpy
@@ -410,6 +410,9 @@ init -1500 python in _console:
             self.lines = lines
 
         def append(self, value):
+            if self and (self[-1] == value):
+                return
+
             super(BoundedList, self).append(value)
 
             while len(self) >= self.size:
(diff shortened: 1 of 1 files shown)


qinlili23333 (approved):

mal:
Mutating the otherwise agnostic `BoundedList` class and comparing heavy `ConsoleHistoryEntry` instances doesn't seem like a good way to achieve this. Instead, changes to [`run`](https://github.com/renpy/renpy/blob/510c216433c6175c9983dbd219a6dcaaecfdfe87/renpy/common/00console.rpy#L675-L676) seem like they could be more fitting if this feature is to go forward.

mal:
Turned out to belong in `interact` rather than `run`, but the draft PR above should provide the desired functionality now. :)

renpytom:
Superseded by #4150 .

PR #4150, opened 2022-12-10
console: Only save command if not the same as last

mal:
Alternate implementation of https://github.com/renpy/renpy/pull/4147 following brief discussion (https://github.com/renpy/renpy/pull/4147#issuecomment-1344956624).


Gouvernathor (approved):
I don't understand how all of that works and what all the variables represent, but I trust that it works on my end.
No objection.

renpytom:
Merged, thank you.

PR #4152, opened 2022-12-13
web: Generate catalog of all media content for caching in PWA

Vadim-Karpenko:
The first step on working on [https://github.com/renpy/renpy-build/issues/53](https://github.com/renpy/renpy-build/issues/53).
This code creates a minified `pwa_catalog.json` file in the destination build that has the following structure:
```
{"files": { "<file_path>": "<hash>", ... }, "version": "your-game-<timestamp>"}
```
`files` contains all paths in the `game` folder, and will allow us to download all files at once, for offline use.
The `version` will let us track if the catalog is updated to the current version of the service worker. If not, we'll create a warning in the console, to let the dev know that it needs to be updated

### DIFF
--- a/launcher/game/web.rpy
+++ b/launcher/game/web.rpy
@@ -33,6 +33,7 @@ init python:
     import pygame_sdl2
     import zipfile
     import re
+    import hashlib
 
     WEB_PATH = None
 
@@ -325,6 +326,38 @@ init python:
         icon72_maskable = renpy.display.pgrender.transform_scale(icon512_maskable, (72, 72))
         pygame_sdl2.image.save(icon72_maskable, os.path.join(icons_dir, 'icon-72x72-maskable.png'), best_compression)
 
+    def get_md5_hash(file_path: str) -> str:
+        """
+        Generates MD5 hash sum of the given file.
+
+        :param file_path: string, The path to the file.
+
+        :return: string, The MD5 hash sum of the file.
+        """
+        # Check if the file exists
+        if not os.path.exists(file_path):
+            raise FileNotFoundError(f"The file {file_path} was not found.")
+
+        # Check if the file is not a directory
+        if os.path.isdir(file_path):
+            raise IsADirectoryError(f"The file {file_path} is a directory.")
+
+        # Check if the file is not empty
+        if os.stat(file_path).st_size == 0:
+            raise ValueError(f"The file {file_path} is empty.")
+
+        # Create a new MD5 hash object
+        md5_hash = hashlib.md5()
+        # Open the file in read byte mode
+        with open(file_path, "rb") as file:
+            # Read the file in 4KB chunks
+            for chunk in iter(lambda: file.read(4096), b""):
+                # Update the hash with the current chunk
+                md5_hash.update(chunk)
+
+        # Return the hash as a string
+        return md5_hash.hexdigest()
+
 
     def generate_files_catalog(destination, version):
         """
@@ -338,14 +371,20 @@ init python:
 
         :return: None
         """
-        catalog = {"files": [], "version": version}
+        catalog = {"files": {}, "version": version}
+        # Walk through the game folder
         for root, dirs, files in os.walk(os.path.join(destination, "game")):
             for file in files:
+                # Get the absolute path of the file
                 file_path = os.path.join(root, file)
+                # Convert it to relative path of the file
                 file_name = os.path.relpath(file_path, destination)
                 # Replace backslashes with forward slashes
                 file_name = file_name.replace("\\", "/")
-                catalog["files"].append(file_name)
+                # Get the MD5 hash of the file
+                file_hash = get_md5_hash(file_path)
+                # Add the file to the catalog
+                catalog["files"][file_hash] = file_name
         with io.open(os.path.join(destination, "pwa_catalog.json"), 'w', encoding='utf-8') as f:
             # Write the JSON file without spaces and new lines, so it's as small as possible
             f.write(json.dumps(catalog, separators=(',', ':')))
--- a/launcher/game/web.rpy
+++ b/launcher/game/web.rpy
@@ -371,7 +371,19 @@ init python:
 
         :return: None
         """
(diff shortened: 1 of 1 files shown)


qinlili23333:
I think file hash should be contained in catalog, this can help for delta update. Maybe MD5 is a good choice, but for files larger than 1M maybe only calculate the first 1M part MD5.

Vadim-Karpenko:
> I think file hash should be contained in catalog, this can help for delta update. Maybe MD5 is a good choice, but for files larger than 1M maybe only calculate the first 1M part MD5.
> 
> And in service worker update process, we can do like this: Step 1: fetch latest catalog Step 2: if exist cached old catalog then compare file's hash and get a list of files need to update, or if it's a new installation all necessary files will add to the list Step 3: download files in the list Or we can put the update process in the first time the game is launched with new service worker, since if it take too long time in the install event of service worker it may get fail. Just like most mobile games, silent update for main component(apk for Android, and service worker for PWA), and trigger resource update in next launch.

Yeah, makes sense. It will decrease the amount of HEAD requests needed to keep everything up-to-date. And I don’t think that media files will be changed after the build is complete (manually by the user), like it can happen to core files (e.g index.html). In most cases it will be rebuilt and hashes will be updated too. 
I’ll prepare the code.

qinlili23333:
>  And I don’t think that media files will be changed after the build is complete (manually by the user), like it can happen to core files (e.g index.html). In most cases it will be rebuilt and hashes will be updated too. 

Maybe we can also provide a standalone tool for user to generate catalog manually after file change. Put catalog build code as a seperate file "build-catalog.py", invoke this file in building and user can also manually invoke this file after file change.

qinlili23333 (approved):

launcher/game/web.rpy
```diff
@@ -338,8 +361,9 @@ init python:
 
         # Use re to slugify the game name, avoiding use of 3rd party libraries
         slugified_name = re.sub(r'\W+', '-', p.dump['build']['display_name']).lower()
+        version = "{}-{}".format(slugified_name, int(time.time()))
         # Replace the default cache name with the game name + current timestamp
-        service_worker = service_worker.replace('renpy-web-game', "{}-{}".format(slugified_name, int(time.time())))
+        service_worker = service_worker.replace('renpy-web-game', version)
```
qinlili23333, line 407: I would like to seperate app name and version in service worker, and seperately replace these two values in build.
in service worker:
```
var APP_PREFIX = '||GAME_NAME||'
var VERSION = '||GAME_VERSION||'
var CACHE_NAME = APP_PREFIX + VERSION;
```
And in build:
```
service_worker = service_worker.replace('||GAME_NAME||', slugified_name)
service_worker = service_worker.replace('||GAME_VERSION||', version)
```
And we need a way to get version of service worker in main webpage. I think we can do like this.
```
if (event.request.url.indexOf("getVersionWorker") > 0) {
    event.respondWith(new Response(VERSION));
    return;
}
```
So we can get version just like this.
```
let currentVersion = await (await fetch("getVersionWorker")).text();
```

qinlili23333 (changes requested):

renpytom:
Having tried the PWA out, I think the caching here is way too aggressive. 

Right now, if I launch a game in the Ren'Py launcher, change the game, rebuild it, and reload the web page, it will get the old content on the first pass through. I have to then hit reload to get access to the latest version of the game. That's unacceptable for development, and probably unacceptable for production - we don't want to get into a scenario where a creator puts out a fix, and then there are still bugs in it. 

My thinking is that the PWA should probably fetch a catalog from the server, first thing it does. If can use this catalog to determine which files can be pulled from the cache, and which can't. If fetching the catalog fails, run entirely out of the cache - otherwise, prefer the server unless we can confirm the cache is up to date.

qinlili23333:
> My thinking is that the PWA should probably fetch a catalog from the server, first thing it does. If can use this catalog to determine which files can be pulled from the cache, and which can't. If fetching the catalog fails, run entirely out of the cache - otherwise, prefer the server unless we can confirm the cache is up to date.

Maybe we can setup a `version.json`, containing latest version and acceptable versions, just like this structure.
```
{
"latest":"1.3.2",
"acceptable":["1.3.0","1.3.1"]
}
```
And if version is latest or an network error occurs, just use cache directly. If version is in acceptable, use cache and do update in background if network allows (these non-essential updates should not take on mobile networks especially when data saver is on, we can detect by `navigator.connection`, and background update should only take effect in next time launch). If version is too old, just do force update before launch game. This checking action should put in webpage instead of service worker, and we should only check once in a period (maybe in 2 hours ?).
Defaultly, in build we only set `latest` to current version and set an empty array in `acceptable`, but developers can also manually set acceptable versions.

This is also why I want to add a `ForceNoCache` query detect in service worker. Requests with `ForceNoCache` key should always fetch from origin server without check cache. e.g. A fetch of `./config.json` will first detect cache, but a fetch of `./config.json?ForceNoCache=1` will directly request server.

launcher/game/web.rpy
Vadim-Karpenko: The version is just a timestamp of when the game was built, not the version of the game, so not sure if we need to separate them

Vadim-Karpenko:
> Having tried the PWA out, I think the caching here is way too aggressive.
> 
> Right now, if I launch a game in the Ren'Py launcher, change the game, rebuild it, and reload the web page, it will get the old content on the first pass through. I have to then hit reload to get access to the latest version of the game. That's unacceptable for development, and probably unacceptable for production - we don't want to get into a scenario where a creator puts out a fix, and then there are still bugs in it.
> 
> My thinking is that the PWA should probably fetch a catalog from the server, first thing it does. If can use this catalog to determine which files can be pulled from the cache, and which can't. If fetching the catalog fails, run entirely out of the cache - otherwise, prefer the server unless we can confirm the cache is up to date.

Spent the entire morning working on this, looks like it was the desired behavior from PWA devs. Tried a lot of things, but ended up reloading a page automatically if an update is found. Not the best solution, so I will look into it more in the coming days

Issue #4153, opened 2022-12-13
Should grids allow underfull by default?


renpytom:
I received feedback on the Ren'Py discord that there are common use cases, like galleries, where it makes sense to allow grids to be underfull.

morganwillcock:
I would say yes, unless there is a significant performance difference by allowing it. 

There is value in the cases where you always need an exact fit, but I don't think those are the most common scenarios. Since being underfull generates an error I think the natural fit is that it is an opt-in assert; even if you removed the check entirely someone could get an equivalent assert by just counting or measuring how many grid items are being used.

Gouvernathor:
If it can be changed by the following code, as wel as on a grid-by-grid basis, I don't think the default value matters that much either way.
```rpy
style default:
    allow_underfull True
```
If we make it the default, we should include this in the incompatible. If not, it could be added to grid's documentation.

morbil:
> If it can be changed by the following code, as wel as on a grid-by-grid basis, I don't think the default value matters that much either way.

I think the default should reflect the most common use cases, which are more likely to be dynamic items. You are also far less likely to run into issues if the default changed from allow to disallow, so it seems like a low-risk change that will help onboard newer devs to better container options.

Andykl:
I agree with @Gouvernathor. And also, Explicit is better than implicit.
In my project I more often use `viewport` with `hbox` with `box_wrap` to create infinite gallery-like layouts and `grid` with explicit empty frames when amount of enities is known.

mal:
If I'm reading the code correctly, then this is a `config.developer` only feature to throw an exception if there are less than the number of expected items in a `grid` or `vpgrid`. Given that, I think setting allow as the default and asking the creators that want that exception thrown when testing (rather than the often more useful `null` padding) to opt-in seems very reasonable.

I'd almost question if the flag really has much value any more, the cases for opting in seem very slim and developer-only, and would likely be better served by the creator implementing their own check on the data source which would have much more context as to the cause of the under-fill.

Additionally after reading the code and reviewing what it was doing, I also spent a bit of time optimising it, skipping calculation of `allow_underfull` and the construction of `Exception` instances unless grids are underfull and developer mode is active. PR for that work can be found in #4154 (currently leaves the default untouched, but could be extended with flipping them once this open question is resolved).

renpytom:
I've made this change, since I think it makes common use cases for grids (like image galleries) require less boilerplate to maker work.

PR #4154, opened 2022-12-13
Optimize allow_underfull for non-developer mode

mal:
Bit of potential optimisation after looking into https://github.com/renpy/renpy/issues/4153.

The main change is that only in the event that the grid is underfull and developer mode is active will it bother to:
- Resolve the `allow_underfull` value,
- Construct of exception instances.

This is very much a same-behaviour refactor, which I appreciate we're not always looking for, so completely understand if it's not desired. :)


renpytom:
This seems find by me, and I'll change the default, too.

Issue #4158, opened 2022-12-16
Web issues


Andykl:
1. This message is reported as an error by a certain percentage of users. I think it makes sense to remove or improve it, and also to write in the documentation what the server owner needs to do to prevent it from occurring.
![image](https://user-images.githubusercontent.com/27142420/207997909-215d2d8d-b037-435d-83dd-3dd59fb124d2.png)
2. When game exists there should be prompt explaining what user need to do. They are:
a. Normal quit -> The game quits. You may close the tab.
b. Error quit -> The game has crashed. You can download the log file and send it to the developers.
c. Game requested reload (currently `renpy.quit(reload=True)` is broken due to subproccess call) -> The game asks for a restart. You need to manually reload the tab.

qinlili23333:
maybe we can use refresh to do the reload in web, just simply call `location.reload()` when request `renpy.quit(reload=True)`

renpytom:
I'll add here, so we don't forget it, that aec68a96a4e856bcd801d261279caed976dc4b11 seems that it breaks loading in games from renpyweb-7. When you go to a game on itch.io, and download the savefiles, it includes the saves for every game, not just the one you were playing. 

I think it makes sense to keep the change, but also look for files of the format:

f'{config.save_directory }/whatever-LT1.save'

And strip config.save_directory before unpacking, so we can unpack saves exported by Ren'Py 7.

renpytom:
renpy.quit(reload=True) doesn't make sense in the web environment.

MiiNiPaa:
I believe renpy.quit(reload=True) is the only documented way to ensure that init statements are re-run and the game is properly restarted, e.g. conditionally loaded modules would be unloaded if conditions are no longer true. (There is utter_restart, but it is not documented).

RenpyOnly:
Links not wokring in web. Do not redirect

RenpyOnly:
Why are you complicating the web version? If you want to kill her just develop

renpytom:
I believe that everything in here has been resolved.

PR #4159, opened 2022-12-16
web: Show prompts to user when game quits

Andykl:
This is second issue of #4158. Probably needs a better wording.

### DIFF
--- a/renpy/bootstrap.py
+++ b/renpy/bootstrap.py
@@ -292,7 +292,13 @@ You may be using a system install of python. Please run {0}.sh,
                 exit_status = e.status
 
                 if e.relaunch:
-                    if hasattr(sys, "renpy_executable"):
+                    if renpy.emscripten:
+                        import emscripten
+                        emscripten.run_script('Module.print("The game asks for a restart. '
+                                              'You need to manually reload the tab.")')
+                        sys.exit(0)
+
+                    elif hasattr(sys, "renpy_executable"):
                         subprocess.Popen([sys.renpy_executable] + sys.argv[1:]) # type: ignore
                     else:
                         if PY2:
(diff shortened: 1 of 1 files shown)


renpy/bootstrap.py
```diff
@@ -306,6 +312,14 @@ def bootstrap(renpy_base):
             except Exception as e:
                 renpy.error.report_exception(e)
 
+        if renpy.emscripten:
+            import emscripten
+            if exit_status == 0:
+                emscripten.run_script('Module.print("The game quits. You may close the tab.")')
```
Gouvernathor, line 318: Maybe these should allow translations, using something like this:
```diff
-                 emscripten.run_script('Module.print("The game quits. You may close the tab.")')
+                 emscripten.run_script('Module.print("{}")'.format(__("The game quits. You may close the tab.")))
```
That way the messages, and them only, could be translated through common.rpym without risk of typos in the emscripten call.

renpy/bootstrap.py
```diff
@@ -292,7 +292,13 @@ def bootstrap(renpy_base):
                 exit_status = e.status
 
                 if e.relaunch:
-                    if hasattr(sys, "renpy_executable"):
+                    if renpy.emscripten:
+                        import emscripten
+                        emscripten.run_script('Module.print("The game asks for a restart. '
```
qinlili23333, line 297:
```
emscripten.run_script('location.reload()')
```
We can just automatically reload the webpage.

renpytom:
I've implemented something similar in renpy/renpy-build@0b1daac , except it's done at the Python level, so it'll catch most ways of exiting. 

In general, a web game should never quit. Web games are similar to mobile games, insofar as quitting is an error.

qinlili23333:
> In general, a web game should never quit. Web games are similar to mobile games, insofar as quitting is an error.

However lots of mobile games on Android have implement quitting actions just like the picture below.
Screenshot_20221219-105425_War Robots

In my view, platforms with a "Back" action feature do need quitting feature. When user click "Back" at homepage, what user expect is the game can quit so they don't need to manually kill it from recent activities. 
Web apps on Android and ChromeOS accept the "Back" action and should work as native app so web game should have quit feature to provide a better PWA user experience.

We can implement quitting under specify conditions:
```
if (history.length == 1 && window.self === window.top) {
    window.close();
};
```
`window.close()` only works when no history to go back and it's not in iframe, so we only need to implement quitting in this condition.

PR #4164, opened 2022-12-18
web: implement reload feature

qinlili23333:
Just use webpage reload to implement `renpy.quit(reload=True)` on web platform, solve part of #4158 
Already tested it works fine.

### DIFF
--- a/renpy/__init__.py
+++ b/renpy/__init__.py
@@ -185,7 +185,8 @@ else:
 arch = os.environ.get("RENPY_PLATFORM", "unknown-unknown-unknown").rpartition("-")[2]
 
 # A flag that's true if we're on a smartphone or tablet-like platform.
-mobile = android or ios or emscripten
+mobile = android or ios
+web = emscripten
 
 # A flag that's set to true if the game directory is bundled inside a mac app.
 macapp = False
@@ -620,6 +621,11 @@ def reload_all():
 
     if mobile:
         raise Exception("Reloading is not supported on mobile platforms.")
+
+    if web:
+        import emscripten
+        emscripten.run_script('window.location.reload()')
+        return
 
     import renpy
 


renpytom:
It doesn't make sense to have this feature on the web platform.

qinlili23333:
> It doesn't make sense to have this feature on the web platform.

It usually occurs in complex multi-language games or games with selective graphic filters. When players need to change language the game needs a reload. Although there are workarounds to change language without a reload, but to ensure everything works as expected the best way is just reload it. Web platform can do reload by just refresh the page very well.

In fact even Android platform can have reload feature, start a new activity and fully destory current activity, and in the new activity launch the game activity again, and reload can work. The only platform cannot carry out a reload is iOS (maybe also exist ways to do it but i'm not familiar with iOS). I think we should implement reload on as many platforms as we can, to ensure better cross-platform compatibility.

Andykl:
I do have such game (both Live2D and lazy-translations are init time addons). And I think it is completly fine to ask the user to reload the app manually. A lot of games even on PC just tell "Restart the app for the settings to take effect". Probably we will change the message even on PC to cross-platform "To toggle [languge/live2d] restart the app. Quit the game now?"

renpytom:
> > It doesn't make sense to have this feature on the web platform.
> 
> It usually occurs in complex multi-language games or games with selective graphic filters. When players need to change language the game needs a reload. Although there are workarounds to change language without a reload, but to ensure everything works as expected the best way is just reload it. Web platform can do reload by just refresh the page very well.

Ren'Py's translation system is designed to work without a reload.

qinlili23333:
> Ren'Py's translation system is designed to work without a reload.

But multi-language feature is not just text translation. In some CG pictures there exist text label, and the CG picture  also provide multi-language versions. For example a CG that one person holds a phone, the content of the phone screen also needs to use the language as user select to ensure the player can recognize the meaning. Some games also use images containing text as UI buttons, like Everlasting Summer, shown in picture below:
image

At current stage, this feature cannot use renpy's translation system. RenPy's translation system can only solve in-game text translation, although it's high-effeciency and easy to use. Developers still need reload for more complex multi-language features.

RenpyOnly:
text _("\nMade with {a=https://www.renpy.org/}Ren'Py{/a} [renpy.version_only]")


even your links do not redirect, you can try, does not depend on the settings


maybe i'm dumb

Issue #3645, opened 2022-06-11
Allow to set a default custom mixer channel


raspberry-soft:
The same way `config.default_music_volume = 0.7` is used to set the default volume of the 'music' mixer, custom mixers don't have a way to be set.

I suggest something like: `config.default_mixer_volume = { "mixer1" : 0.5, "mixer2" : 0.9 }`

Right know I'm using:

```
init python:
    if not persistent.system_sound_volume_set:
        persistent.system_sound_volume_set = True
        preferences.set_volume('system_sound', 0.7)
```

I think having a config variable will look more elegant and mix well with the other default variables.

henvu50:
Any updates on how to properly set the default volume of a custom mixer channel created with register_channel?  From what I've researched, I may have to use the persistent var check; code example in OP.

Gouvernathor:
The provided example uses documented API so it's safe and not _that_ complex to implement, so you're safe doing it that way - though we now use set_mixer instead of set_volume, but if you use it with a tested arbitrary value it's the same.

But I think we should provide a simpler way to do that. I'd be more inclined towards a dict for custom mixers and keeping separate entries for the builtin ones, but if the implementation is easier that way I guess I'm fine with it too.

PR #4165, opened 2022-12-19
Default values for custom mixers

lauwurence:
Related to https://github.com/renpy/renpy/issues/3645

### DIFF
--- a/renpy/common/00defaults.rpy
+++ b/renpy/common/00defaults.rpy
@@ -53,17 +53,13 @@ init -1500 python:
     # If not None, the default value of emphasize_audio.
     config.default_emphasize_audio = False
 
-    # If not None, the default value of set_volume (music)
-    config.default_music_volume = 1.0
+    # For compatibility with old versions.
+    config.default_music_volume = None
+    config.default_sfx_volume = None
+    config.default_voice_volume = None
 
-    # If not None, the default value of set_volume (sfx)
-    config.default_sfx_volume = 1.0
-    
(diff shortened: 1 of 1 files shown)


renpy/common/00defaults.rpy
```diff
@@ -58,6 +58,9 @@ init -1500 python:
 
     # If not None, the default value of set_volume (sfx)
     config.default_sfx_volume = 1.0
+    
+    # The default values of set_volume for custom mixers
+    config.default_mixer_volume = {}
```
Booplicate, line 63: Might make sense to deprecate old mixer configs (default them to `None`). If they are set (non-`None`), add their values to the new `default_mixer_volume`, otherwise just ignore.

via-my:
Speaking of volumes, is it just me or default value of `config.volume_db_range` is set too high with 60?
It feels too sensitive when the volume bars go from 75% to 100%. 
Does it make sense to reduce `config.volume_db_range` to 45? It will make volume adjusting little smoother and leave the intended behavior as it is.

renpy/common/00defaults.rpy
Booplicate: I think you forgot to add the default values to `default_mixer_volume` for those 3 builtin mixers.

via-my: I assume they are not needed because the mixers are set to 1.0 by default:
https://github.com/renpy/renpy/blob/master/renpy/preferences.py#L323

Issue #4139, opened 2022-12-08
Transitions disable screen keybindings


via-my:
In some screens I override the 'game_menu' key for different custom actions. But when a transition occurs, those key statements do not catch the mouse events and the game menu opens instead.

```
screen test():

    text "test screen" align (.5, .5)

    key 'mousedown_1' action With(Dissolve(1.0))
    key 'game_menu' action Notify("RMB pressed")
```

renpytom:
If you want to override the game menu, use config.game_menu_action.

I think it might also make sense to have config.bottom_layers, similar to config.top_layers, to allow a place for screens like these to go - thought keymap blocking screens are a bad idea. 

#4181 is the new bug.

PR #4169, opened 2022-12-20
Predefine the `events` variable in transitions

via-my:
Makes it possible to define whether these transitions should catch or ignore events by default.
Related to https://github.com/renpy/renpy/issues/4139


Andykl:
I think it should be a separate keyword defaults to False. Otherwise `Dissolve(0.5, events=True)` also leaks it into `Dissolve.style.properties`.

via-my:
Oh, I see. Thanks!

renpytom:
I don't see a need for this, especially not #4139. A decent amount of work has gone into coming up with sensible ideas of how events and transitions work, and I think making that overrideable is solving a problem the wrong way.

Issue #4137, opened 2022-12-07
Touchscreen keyboard in 8.1 not korrectly working


RenpyOnly:
Touchscreen keyboard in web version not working korrectly in vertical position smartphone.
The touch is offset relative to the buttons on the screen.
All browsers.

renpytom:
This looks fixed now.

PR #4173, opened 2022-12-20
web: Call prompt using JS on mobiles

Vadim-Karpenko:
When `renpy.input` is called, this code will execute JS (if available) to call the browser's default prompt window for input.
A user will be able to enter text using a keyboard or voice input instead of using the on-screen widget.

@renpytom, I wasn't sure where exactly the on-screen widget is being called in the code, it will be better to move the call of JS in this PR to where you think it will fit, to avoid drawing the mobile widget completely. Right now it shows up for a moment

### DIFF
--- a/renpy/exports.py
+++ b/renpy/exports.py
@@ -932,9 +932,7 @@ def input(prompt, default='', allow=None, exclude='{}', length=None, with_none=N
     rv = None
     if renpy.emscripten:
         # escape quotes in prompt and default for javascript usage
-        prompt = prompt.replace("'", "\'").replace('"', '\"')
-        default = default.replace("'", "\'").replace('"', '\"')
-        rv = emscripten.run_script_string("(function callMobilePrompt() { if (window.ontouchstart !== undefined) { return prompt('%s', '%s'); } return ''})()" % (prompt, default))
+        rv = emscripten.run_script_string("(function callMobilePrompt() { if (window.ontouchstart !== undefined) { return prompt(%r, %r); } return ''})()" % (prompt, default))
 
     if not rv:
         rv = renpy.ui.interact(mouse='prompt', type="input", roll_forward=roll_forward)
--- a/renpy/exports.py
+++ b/renpy/exports.py
@@ -132,6 +132,9 @@ import sys
 import threading
 import fnmatch
 
+if renpy.emscripten:
+    import emscripten
+
 
 # The number of bits in the architecture.
 if sys.maxsize > (2 << 32):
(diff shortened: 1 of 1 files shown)


renpy/exports.py
```diff
@@ -926,7 +929,12 @@ def input(prompt, default='', allow=None, exclude='{}', length=None, with_none=N
     if fixed:
         renpy.ui.saybehavior()
 
-    rv = renpy.ui.interact(mouse='prompt', type="input", roll_forward=roll_forward)
+    rv = None
+    if renpy.emscripten:
+        rv = emscripten.run_script_string("(function callMobilePrompt() { if (window.ontouchstart !== undefined) { return prompt('%s', '%s'); } return ''})()" % (prompt, default))
```
renpytom, line 934: Using %s like this is bad - think about what would happen if prompt or default contained an apostrophe.

Vadim-Karpenko: You are right, I'll try to make it safer so it won't fail in case there's an apostrophe or something else that can fail the script. As for use of %, it might be the only way, `.format` will not work here, because the string contains a lot of `{` and `}` and it fails as a result of it.

Gouvernathor: Maybe `"something(%r, %r)" % (prompt, default)` ? It would use repr and so quote the strings automatically.
That, or .format and doubling the four braces in the string.

renpy/exports.py
```diff
@@ -926,7 +929,13 @@ def input(prompt, default='', allow=None, exclude='{}', length=None, with_none=N
     if fixed:
         renpy.ui.saybehavior()
 
-    rv = renpy.ui.interact(mouse='prompt', type="input", roll_forward=roll_forward)
+    rv = None
+    if renpy.emscripten:
+        # escape quotes in prompt and default for javascript usage
+        rv = emscripten.run_script_string("(function callMobilePrompt() { if (window.ontouchstart !== undefined) { return prompt(%r, %r); } return ''})()" % (prompt, default))
```
qinlili23333, line 935: `function callMobilePrompt()` can be replaced with `()=>`, using arrow functions makes the code shorter.

qinlili23333:
this pr also fix #4137 , as prompt won't change size of web when screen keyboard shows, at least on Android Chrome/Firefox
need someone test on iOS Safari

Vadim-Karpenko:
I'm closing the PR, as @renpytom implemented his version of this PR (1feed24f14f37bc90e47381a615348c673d9b4d9), which includes custom form input that does not block the game.

PR #4199, opened 2023-01-01
Update Spanish translation

Moshibit:


launcher/game/tl/spanish/common.rpy
```diff
@@ -785,3 +785,19 @@ translate spanish strings:
     # renpy/common/00speechbubble.rpy:304
     old "(hide)"
     new "(ocultar)"
+    
+    # renpy/common/00gui.rpy:455
+    old "This save was created on a different device. Maliciously constructed save files can harm your computer. Do you trust this save's creator and everyone who could have changed the file?"
```
Gouvernathor, line 790: These strings are bound to change before the release, so maybe they shouldn't be merged right now

Moshibit: So, shall I close the PR until the time is right? If the strings are not the final ones it would be best to close the PR and do the translation until the next pre-release.

Gouvernathor: As you prefer, we can merge the other strings now, or leave it open until the strings are finalized and merge it all then.

Issue #4149, opened 2022-12-10
Sticky Layers


mal:
**Problem:**

There's currently (so far as I am aware) no good way to apply a single transform to a selection of displayables while having them remain available to attribute changes. i.e. applying a single shader to multiple layeredimage characters and being able to alter their attributes while the shader is running.

**Tentative solution:**

One way this _could_ be done would be to place the layered images on a separate layer, and apply the transform to the layer. 

**Nested problem:**

Unfortunately however attribute changes and show statements always use a tag's default layer irrespective of where it is currently being shown. This means that every show statement must include an `onlayer` clause, and that say attributes are non-functional unless the tag is on its default layer.

**Nested solution: Sticky layers!**

A proposed flag (`config.sticky_layers`?) which would result in the `SceneLists` instance storing the layer each shown tag is placed on, and the `default_layer` function preferring this map to the "true" defaults. The outward facing result would be that the initial show would still require the `onlayer` clause, but any subsequent show statement or say attribute use would receive the "correct" layer automatically, until that tag was hidden, at which point the `SceneLists` instance would discard the tag's entry, and the `default_layer` function (now encountering a "miss") would revert back to the "classic" method of determining the default layer.

---

I would note that needing to create and manage a separate layer for the less than 1% of scenes in a game that need this sort of thing isn't really ideal, so a better solution or options around group transforms - the root problem - would be most welcome, however in the interim the "sticky layers" functionality would help with working around that limitation, and seems to my mind a potentially valid and useful feature in it's own right, hence this write up.

---

Happy to take a crack at this if it's felt to be a valid feature request and worth pursuing. :)

morbil:
Transform properties are inherited by subsequence showing of a specific tag. That includes shaders. Here's an example using assets from "The Question" from the SDK:

```renpy
define s = Character("Sylvie")
define g = Character("Gylvie")

transform signalwarp:
    shader "3pood.signalwarp"

label main_menu:
    return

label start:
    scene bg uni:
        xysize (1920, 1080)
    show sylvie blue normal at left
    show gylvie green normal at right

    s "You've created a new Ren'Py game."

    show sylvie blue normal at signalwarp
    show gylvie green surprised at signalwarp

    s "You've created a new Ren'Py game."

    show sylvie blue smile at right
    g "Once you add a story, pictures, and music, you can release it to the world!"
    show gylvie green giggle at left
    pause

    return
```

And here's how that looks:


[...]

PR #4207, opened 2023-01-03
Make sticky layers always an opt-in for existing games

Gouvernathor:
This allows creators who use custom layers not to have to change their code in any way in order to keep the current behavior.
What was initially discussed in #4149 was to make stickiness a global opt-in, and then to make it an opt-in for all layers except detached layers (which don't exist yet).
This PR re-establishes the behavior that was discussed.

### DIFF
--- a/renpy/exports.py
+++ b/renpy/exports.py
@@ -3951,7 +3951,7 @@ def clear_line_log():
     renpy.game.context().line_log = [ ]
 
 
-def add_layer(layer, above=None, below=None, menu_clear=True, sticky=None):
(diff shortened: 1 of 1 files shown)


mal:
I think the confusion here arose from how the feature changed between inception and implementation. Opt-in made sense in the context of how the feature was originally envisioned which actively restricted showing the same tag on multiple layers at the same time.

However, the current implementation imposes no such restriction, and allows existing features such as say-attributes to "just work" with additional layers, in non-pathological use-cases, out-of-the-box. As such I don't think this default should be changed from it's current state, but do accept that it should probably be addressed in release notes.

PR #4208, opened 2023-01-03
Add the `in` keyword to show statements

Gouvernathor:
The objective is to encourage creators to use stickiness on a show-by-show basis and in a way that reflects in the script.
The rationale is that I don't see the first following example as being harder to understand or to use than the second one - in addition to the additional features it introduces.
```rpy
label start:
    show tag in c # sticky because in
```
```rpy
define config.sticky_layers += ["c"]

label start:
    show tag onlayer c # sticky because c
```

The reason for that is to make renpy script, and excerpts thereof, as unambiguous as possible.
That helps a great deal in the context of support channels, because you don't need to enquire about the state of the config variables or even to think about them, before knowing what a bit of code does.

Prediction was unchanged, I think there's no need for it. The exiting sticky layer implementation doesn't change it either, so 🤷 
Hide and scene-without-displayable also accept ``in`` in lieu of ``onlayer``, but that's not documented.
Just to state the obvious : the existing sticky-by-layer behavior related to the config variable was not altered, **nothing was removed**, even though undocumenting it would be fine by me.

### DIFF
--- a/renpy/ast.py
+++ b/renpy/ast.py
@@ -1312,7 +1312,10 @@ def predict_imspec(imspec, scene=False, atl=None):
     in imspec.
     """
 
-    if len(imspec) == 7:
+    if len(imspec) == 8:
+        name, expression, tag, at_expr_list, layer, _zorder, _behind, _sticky = imspec
+
+    elif len(imspec) == 7:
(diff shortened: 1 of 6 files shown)


renpytom:
I don't think this is needed anymore, with the current approach to sticky layers.

PR #4210, opened 2023-01-04
Make sticky layer behaviour consistent between layers

mal:
**tl;dr**: tags are sticky, screens are not... maybe. see https://github.com/renpy/renpy/pull/4210#issuecomment-1372665361 😛

---

After speaking with @Gouvernathor at length about this feature we reached what is hopefully a fairly sane and creator-friendly balance.

This PR has the feature adopt a take it or leave it stance, with the default being that it's turned on - a default that seeks to ensure that features like say-attributes "just work" and `onlayer` clauses need not be added to every subsequent `show` statement when displaying a tag on a layer not considered to be it's default.

In this implementation tag/screen conflicts are avoided via an explicit carve out for `ScreenDisplayables` which are never treated as sticky, which allows all layers to a share a consistent behaviour without needing to assess whether or not they are sticky.

This change does take the stance that there's no valid use for a mixture of sticky and non-sticky layers.

A summary of a code level changes:
- `config.sticky_layers` is now a compat boolean
- When a tag is hidden, the layer is checked before removing it from the `sticky_tags` dict.
  - This not only gives `hide` the same semantics as `clear`, but also avoids the need to check for screens - since they aren't sticky, and so if a screen is being removed from a layer, that layer could not contain a sticky tag of the same name.
- `sticky_tag` clean up is unconditional in both `hide` and `clear`.
  - The removal is cheap enough to not warrant a check on the grounds of performance, and has the side benefit of ensuring that anything placed there during development while the feature was enabled will still get cleaned up if it is later disabled which should provide a better creator experience if experimenting with the flag across reloads or saves.
- `sticky` argument is removed from `add_layer` as the feature is now an all or nothing proposition.
- Attempt to capture and explain feature in documentation with the expectation that it could be revisited prior to release.

### DIFF
--- a/renpy/config.py
+++ b/renpy/config.py
@@ -169,8 +169,8 @@ top_layers = [ 'top' ]
 # Layers below all other layers, that do not participate in transitions.
 bottom_layers = [ 'bottom' ]
 
-# Layers which will override the default layer for a tag while shown.
-sticky_layers = [ 'master' ]
+# Which layer a showing tag was last displayed on will be used as it's default.
+sticky_layers = True
 
 # True if we want to show overlays during wait statements, or
 # false otherwise.
--- a/renpy/common/00compat.rpy
+++ b/renpy/common/00compat.rpy
@@ -269,7 +269,7 @@ init -1100 python:
             config.context_clear_layers.remove("top")
             config.context_clear_layers.remove("bottom")
 
-            config.sticky_layers.remove("master")
+            config.sticky_layers = False
 
             store._errorhandling._constant = True
             store._gamepad._constant = True
--- a/renpy/display/core.py
+++ b/renpy/display/core.py
@@ -1120,7 +1120,7 @@ class SceneLists(renpy.object.Object):
             self.remove_hide_replaced(layer, key)
             self.at_list[layer][key] = at_list
 
-            if layer in renpy.config.sticky_layers:
+            if renpy.config.sticky_layers and not isinstance(thing, renpy.display.screen.ScreenDisplayable):
                 self.sticky_tags[key] = layer
 
         if key and name:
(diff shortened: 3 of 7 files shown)


renpy/config.py
```diff
@@ -169,8 +169,8 @@
 # Layers below all other layers, that do not participate in transitions.
 bottom_layers = [ 'bottom' ]
 
-# Layers which will override the default layer for a tag while shown.
-sticky_layers = [ 'master' ]
+# Which layer a showing tag was last displayed on will be used as it's default.
```
Gouvernathor, line 172: This should probably end up at the end of the file then

renpytom:
So what set of PRs should I merge? I'd like to know what I should be reviewing.

Gouvernathor:
This one. The two others are contradictory, I'm only keeping them not to lose code until it's settled.

mal:
This is good for review now, I've updated the PR body with a few more details and break down of the changes and rationale.

mal:
@renpytom, prior to merging, a specific point of consideration should probably be whether the `ScreenDisplayable` test is needed and makes sense. As I see it, the reasons for keeping or dropping it are as follows:

Keep (adding a screen will never impact `sticky_tags` dict)
- Avoids potential for screens that share a name with an image "polluting" the `sticky_tags` dict, leading to potentially unintended consequences, while still allowing the feature to remain active.

Drop (`sticky_tags` dict tracks all tags regardless of what they represent)
- Functions making use of the `default_layer` helper retain consistent semantics irrespective of what a tag represents.
- Keeps the concept of a tag consistent regardless of what kind of displayable is tagged.

If dropped, then games that require a screen to share a name with an image would work in the same way as those expecting to use the same image tag on multiple layers at the same time - by toggling the feature off via the compat variable.

I've left the check in for now, both to illustrate where it would belong, and also because I myself am split on which way this should fall, hence this comment.

Gouvernathor:
Dropping the protection would make `show screen eileen`, when an `eileen` image is showing, show the screen on the master layer and replacing the image. That's in my view a massive breach of the separation we're maintaining between screens and images - visible in the distinct `show` and `show screen` statements, and `hide` and `hide screen`.
It would in addition introduce a regression about screen names, which were always documented as a free naming pool (because never documented as being incompatible with image names), which is consistent in that defining a `screen op` then trying `show op` doesn't work. Just as label names and screen names always were independent even though both are called, image names and screen names should remain independent even though both are shown.

PR #4211, opened 2023-01-04
Skip storing the unused public signing key

mal:
The public key currently being stored is never used, as it is derived from the private key on load (which makes sense).


renpytom:
I explicitly changed things to store it. There are cases where you might want to see it, and having a second copy in text is nice.

PR #4222, opened 2023-01-06
Use the regular way of testing py's version

Gouvernathor:
This is the last use of the six module across the repo - other than `__init__.py` importing it. Maybe we can remove it from the requirements ?


renpytom:
We can't remove it. There are games that import it, so it needs to remain around.

PR #4226, opened 2023-01-07
Document name-munging more visually and add warning for f-strings incompatibility

Gouvernathor:
That, and moves it, it was under Keyword but belongs under Name.


Gouvernathor:
I tried using a warning _then_ a note, or two warnings, I don't know which is the bestest.
image

PR #4238, opened 2023-01-10
Disable notify when the quicksave doesn't happen

Gouvernathor:


renpytom:
I took a different approach to this. Rather than trying to determine all the cases in which a quicksave could fail, I added the ability to run an action to FileSave when it succeeds - I think that's less prone to getting out of sync in the future.

PR #4252, opened 2023-01-15
Apply transforms to displayables the proper way

Gouvernathor:
By the way, the `entering` function does nothing with its `sle` parameter and wraps the `new_sle` displayable directly. Shouldn't it work the same way as `leaving`, and wrap its parameter ?

### DIFF
--- a/renpy/display/movetransition.py
+++ b/renpy/display/movetransition.py
@@ -240,13 +240,13 @@ def OldMoveTransition(delay, old_widget=None, new_widget=None, factory=None, ent
             return rv
 
         def entering(sle):
-            new_d = wrap(new_sle)
+            new_d = wrap(sle)
             move = enter_factory(position(new_d), delay, new_d, **offsets(new_d))
 
             if move is None:
                 return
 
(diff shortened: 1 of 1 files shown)


mal:
I think you're right about the `new_sle` vs `sle` thing. The final line of `leaving` seems to make the same mistake with `old_sle` too.

The reason `At` works here is that it ultimately passes `sle` as the `child` keyword arg (rather than positional), which might be a better option to do directly here, i.e. `enter(child=new_d)`.

Gouvernathor:
The `child=` solution wouldn't work for callable transforms such as this https://www.renpy.org/dev-doc/html/trans_trans_python.html, so it would create another incompatibility with other transforms previously accepted.
I'm open to gaining time another way, i.e by importing At in some different way, but not at the risk of breaking stuff.

For the entering/leaving thing I'm not sure what you mean and I don't understand the function enough, so if you want to do it, [this branch](https://github.com/renpy/renpy/tree/movetrans) is goody.

renpytom:
I think the typo fixes in the movetrans branch make sense - can we merge them into a single PR?

PR #4277, opened 2023-01-24
Support for new cursor types

altskop:
In this PR, I added support for two new mouse cursor states:

- `button`, for when the player hovers over an instance of Button or its subclasses
- `mousebuttondown`, which is used when the user left clicks

Additionally, I updated the documentation to expand upon the existing (and the new) states and their usage.

Advantages to using the new states:
- `mousebuttondown` was not available at all previously, and there was no way to convey the mouse click via the cursor.
- `button` provides a streamlined approach to adding a cursor that responds to buttons/imagebuttons and respects styles if they exist

Considerations:
- if `mousebuttondown` isn't present in `cursor_cache`, meaning we don't have a cursor to show for this state, then we won't be checking if the left mouse button is pressed
- `mousebuttondown` currently takes precedence over any other state the cursor can be in. I felt this was appropriate from the UX perspective
- I'm not a big fan of using `"mousebuttondown"` and `"button"` magic strings. Seeing as `"default"` is a magic string right now as well (`core.py` line 3123), I've let it be.
- `mainmenu` and `gamemenu` cursor styles seem to be unused. I was unable to make them appear during testing. This might require further investigation.


renpytom:
Thinking about this, the right response might be to prepend pressed_ to the default style. So we'd compute the mouse, then check if pressed_ exists, and if that's in place, use it. 

I think that's the best compromise for this. Do you want to make the change?

altskop:
To clarify, is "pressed_" going to be a modifier for all cursor styles? (Which would allow users to have a separate cursor image for "pressed_default", "pressed_button", "pressed_insensitive" etc.) Or will we use only one cursor style if "pressed_" is in the mouse_name?

renpytom:
I'm thinking pressed_ would be prepended to whatever cursor type is in use - so pressed_default, pressed_button, etc.

altskop:
I like this approach. Will make the change later this weekend.

renpytom:
Thank you.

altskop:
Changes in this commit:

- `mousebuttondown` key for cursor_styles replaced with `pressed_` prefix. It can be used to define a pressed cursor style in any state (for example, `pressed_button`, `pressed_menu`, etc.). If no specific `pressed_` variant for a state is defined, we will fallback to `pressed_default` (if present in cache). Effectively, `pressed_default` will be used for all pressed states that aren't explicitly defined. 
- Updated `config.rst` and `mouse.rst` with more examples and documentation.

Not a fan of repeating `cursor = anim[self.ticks % len(anim)]` three times in core.py, but the alternative isn't much more readable, so I left it the way it is.

Additionally, right now to define a "pressed_" variant for a cursor state, a corresponding non-pressed cursor has to be defined (i.e. to define "pressed_button" we need to have defined "button" cursor). This is due to the fallback mechanism in `core.rpy L3141` which sets mouse_name to "default" if for example "button" isn't in the cache. I'll address that later.

altskop:
Refactored and moved the logic to `get_mouse_name` method, where I think it fits better. I'm more or less satisfied with how it is now.

By extension, it fixed the restriction I've mentioned above.

renpytom:
Merged, thank you.

PR #4278, opened 2023-01-25
Add renpy.lookup_ast

Gouvernathor:
A simple tool to navigate the AST nodes of a game (the others can be reached through .next and .block on the returned node).
Very handy when navigating the AST, for reachability analyses for example and to understand how it works.
This is about renpy's AST of course, not Python's.

With this, in a default project, launch a new game, open the console and type `renpy.lookup_ast().what`. It should return `"You've created a new Ren'Py game."`.

This is absolutely not meant to be documented.

### DIFF
--- a/renpy/exports.py
+++ b/renpy/exports.py
@@ -3863,6 +3863,32 @@ def set_return_stack(stack):
     renpy.game.context().set_return_stack(stack)
 
 
+def lookup_ast(entry=None):
+    """
+    :undocumented:
+
+    Returns the AST node of the provided label, which can be either a label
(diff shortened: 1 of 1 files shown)


renpytom:
I don't think that this is something we can include, as it returns something undocumented and opaque. If we're going undocumented, just use renpy.game.script.lookup, not something in the renpy namespace.

Gouvernathor:
This is mainly a shortcut since every other time I forget what the full path is (especially since lookup is actually a method and not a function). And it's very hard to find when you don't know exactly where to look, compared to how useful it is.

PR #4279, opened 2023-01-25
Add Movie() support for web

Tey:
This PR allows Movie displayables to be seen on the web platform. Movies are rendered by the browser through a video element, video frames are then copied to a canvas element, converted to TGA and put in a MEMFS file for Renpy to read them. Video files are marked properly in `renpyweb_remote_files.txt` for Renpy to know it should not try to download them. I've set up a [demo project here](https://tey.github.io/renpy/video_demo-1.0-web/).

A few caveats:
* It won't work on web2 because the TGA image format has been disabled when building sdl_image (not the case for web3).
* Video are laggy. On my 7 years old computer, the 1080p videos renders at ~15 fps on Chrome and Firefox, and the 240p at ~24 fps. It seems better on my 2 years old phone though. The FPS info is shown in the console at the end of the movie playback. I doubt this can be improved much more, but a laggy rendering is probably better than no rendering at all?
* There is no sync between audio and video apart from the sync done by the browser. As video frames have to be copied multiple times before being drawn on Renpy canvas, it could create some desync with the audio. I didn't experience a noticeable delay though, and only the latest video frame is presented to Renpy, so even if frames are dropped, there should be no increasing delay between audio and video.
* Movies with a side mask do render properly, but they flicker a bit when the movie is restarted. This could probably be improved by letting the loop be managed by JS, or at least, pre-queue the next movie before the end of the current movie playback.
* Movies that use another movie as the mask will not render properly because both movies are out of sync.
* I've tried on Chrome and Firefox, but cannot try on Apple devices.


Tey:
I forgot to add: the only formats available natively for exporting canvas content into images are JPEG and PNG. On my tests, it added too much delay because of the compression, hence the use of TGA even though it's not done natively by the browser.

renpytom:
This doesn't seem like the right approach. 

I suspect the right approach would be to load the movie into a texture, as described at https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/Tutorial/Animating_textures_in_WebGL (the updateTexture function), and expose that texture to Ren'Py to be drawn. 

But this can wait for 8.2, I think. Movie playback on the web is a so-so feature anyway, for sheer size reasons.

Tey:
How would you expose the texture to Renpy then? If this is through a MEMFS file, I doubt the performances will be any better because the texture would have to be written to [a framebuffer first](https://stackoverflow.com/a/13640310), which is probably the same as copying to the canvas.

renpytom:
I think we could create a texture in Ren'Py, then pass it over to the javascript side to be filled in with the image data using one of the emscripten.run_script functions.

Tey:
I see, I'll try to figure out if this is possible.

qinlili23333:
A good try for this feature.

But in my view I think using multi canvas and native video element is a better choice.
We can seperate the game to two canvas, upper one and lower one, with a video element between these two canvas. So that no need to copy video frames into canvas, no need to concern about video framerates, just let the browser itself complete the composition.
And for mask videos, why don't we just convert origin file and mask file into a new video file with alpha channel? There is already alpha channel support for videos [in modern browsers.](https://kitcross.net/hevc-web-video-alpha-channel/) Since videos are merged before release, no need to think about sync problems.
So finally we should get a three-layers page, from upside to downside:
1. foreground canvas (transparent)
2. video element (transparent)
3. background canvas (non-transparent)

This method can solve all performance and sync problems since all critical processing are all done by browser natively, but we need to rewrite many parts of the engine which is a huge work.

renpytom:
> But in my view I think using multi canvas and native video element is a better choice.

I don't think so. I don't think this could support, for example:

```renpy
show Movie(play="test.webm") at truecenter
pause 1.0
with pixellate
```

Or anything else that takes the movie as an argument. I believe taking the movie into a texture is the best match to Ren'Py's semantics.

qinlili23333:
> I believe taking the movie into a texture is the best match to Ren'Py's semantics.

For compatibility yes taking movie into a texture is better. But since the movie data is transfered across wasm and javascript, from GPU to CPU (hardware decoded frame to texture) and to GPU again(texture rendered in canvas), it's nearly impossible to solve the performance problem, especially as the video framerate and resolution get higher the performance get lower.
The only way to solve the performance problem is avoiding transfer video data from GPU to CPU then to GPU again, which means we should keep processing in GPU only so that CPU will never become a bottleneck and modern GPU can handle these actions easily of course.
 
Another problem is framerate sync problem. Videos may have a different framerate with the game, some video codecs even support variable framerate. For example, a game capped at 30fps using `preferences.gl_framerate` has a 24fps movie, running on a 120hz monitor, there is no way to show this movie as texture in a fixed frame latency. However using native video element with double canvas can achive this: one game frame for each 4 refresh and one video frame for each 5 refresh.


But as texture can solve some compatibility problems, maybe we can try to make movie as a special texture and implement avaliable actions such as transitions in a native way using CSS. We don't need to ensure everything in web version render the same as native clients if this will result bad performance, as performance is more important than full effects in user experience.

Tey:
> But since the movie data is transfered across wasm and javascript, from GPU to CPU (hardware decoded frame to texture) and to GPU again(texture rendered in canvas)

AFAIK, the goal of drawing the video frame into a WebGL texture is to prevent that, as textures are stored in GPU RAM. Renpy already uses OpenGL/WebGL to compose the screen, so the video frame would never leave the GPU RAM (except maybe for videos with mask, that's still unclear to me).

Tey:
Performances are indeed much better when using WebGL textures ([demo here](https://tey.github.io/renpy/video_demo-1.1-web/)).

Movies with masks is not supported yet though. Ren'Py uses SDL surfaces to apply the movie alpha mask, so that would mean copying the frame to CPU RAM. Is there a way to do it directly on textures instead?

renpytom:
We should be able to do this on the GPU, as that's where AlphaMask does its thing. At the very least, a shader should be able to make short work of this. (And it's fine here, because the browser should always be using gl2.)

qinlili23333:
> Performances are indeed much better when using WebGL textures ([demo here](https://tey.github.io/renpy/video_demo-1.1-web/)).
> 
> Movies with masks is not supported yet though. Ren'Py uses SDL surfaces to apply the movie alpha mask, so that would mean copying the frame to CPU RAM. Is there a way to do it directly on textures instead?


Yes now the performace is really better a lot, smooth even under powersave mode.

I prefer to render masks into videos with alpha channel in build, using vp9 for any browser except using hevc for safari on ios. However there is a small problem that only VideoToolbox on apple platform can encode hevc with alpha channel.

Tey:
Movies with mask should work now ([demo here](https://tey.github.io/renpy/video_demo-1.3-web/) with transitions tests added). The added "alpha_mask" shader seems to do the trick, although I don't quite understand why all the color components had to be multiplied by the "alpha" component of the mask texture (it doesn't work otherwise). But I'm far from being an OpenGL expert anyway 😅

`get_movie_texture()` now returns a Render instead of a Texture for web when using movies with masks, but that's okay with the way to return object is used.

Movies that use another movie as the mask are still out of sync. There are some tricks we can use to try improving that in JS like synchronizing `currentTime` between the 2 movies periodically, making sure they start at the same time, grab the mask video frame at the same time as the source video frame and/or let the browser manage the video loop. But this will never be perfect anyway as the HTML Video element provides no way to be frame accurate.

BTW, there might be an issue with the recent changes to Matrix (or it can be my builds which are not okay):
```
renpy-pre.js:119 <frozen importlib._bootstrap>:241: RuntimeWarning: renpy.display.matrix.Matrix size changed, may indicate binary incompatibility. Expected 84 from C header, got 88 from PyObject
renpy-pre.js:119 <frozen importlib._bootstrap>:241: RuntimeWarning: renpy.display.matrix.Matrix2D size changed, may indicate binary incompatibility. Expected 84 from C header, got 88 from PyObject
```

renpytom:
Looking good. I haven't reviewed this yet, but I'm wondering if this could replace the webvideo files entirely. 

You need to multiply colors by alpha because Ren'Py uses premultiplied alpha, inside textures. That's because premultiplied alpha works far better with the scaling done on a GPU.  Say you have two pixels, one transparent (0, 0, 0, 0) and one white (1, 1,  1, 1).  Average those two pixels together, you get (0.5, 0.5, 0.5, 0.5). In premultiplied alpha, that's 50% transparent white, while with normal alpha, that's 50% transparent gray. 

I'd say that non-side-by-side masked movies aren't that important anymore, now that we have side-by-side.

You need to clean out the various renpy/module/gen* directories to fix that error. It happens when a cython-defined class changes.

qinlili23333:
Looks nice, but I think we should also test using Movie() in splash, as the autoplay issue may appear again.

Currently in demo, u used movie_cutscene to show splash video, and since my previuos fix (#4145 ) it works well. 
But if someone use Movie() for splash video, due to no user interaction before play, he may face autoplay issue again.
We should test Movie() as splash, and auto mute if no user interaction is still a good way to fix it.


And another thing is that currently for loop video, this commit just set the video src again and again. This works but is not the best performance. I suggest to use `videoElement.loop=true` for better performance.

mal:
Putting this out there as something that might yield better performance, but it's only hypothetical at this point.

Might it be possible to have the shader do the double sampling on a single texture (`(x / 2, y)` for the video, then `(.5 + x /2, y)` for the mask), and avoid the need to pre-split the texture with subsurface? Not sure if that would be more efficient but might be worth trying. 🤷

Tey:
@renpytom Thanks for the explanations. Do you mean merging "00webvideo.rpy" into "webaudio.py" and "_audio.js", or actually showing cutscenes the same way as movie displayables (through WebGL)?

@qinlili23333 Good points, I'll look into it.

@mal That's [`GL2Model.subsurface()`](https://github.com/renpy/renpy/blob/a49bd0133c8c869272109cb98f0f661d0300c52b/renpy/gl2/gl2model.pyx#L73) which is used here. It only copies the texture reference and resize the mesh, so no new texture is created and it's pretty cheap. That looks pretty close to what you suggest, probably faster as there is no need to do extra operations in the shader.

renpytom:
I'm thinking mostly about showing cutscenes the same way we show movie displayables. 00webvideo.rpy is honestly kind of a hack, since I thought that actually showing videos would be much harder than it proved to be. Now that it's working reasonably well, I don't think we need the hack anymore. 

I agree that GL2Model.subsurface() is likely to be just as fast as a dedicated shader.

qinlili23333:
> I'm thinking mostly about showing cutscenes the same way we show movie displayables. 

Yes I agree that if cutscenes can use the same way the code can be much more simple.

And maybe we can provide a special splash video feature for web, as web version usually takes a long time to load, we can directly show the splash video outside of the game engine, just in pure javascript, as the extension of pre-slpash image. 
I'll try to do this.

PR #4283, opened 2023-01-26
Add new config screenshot_name_callback

Booplicate:
I feel like only being able to use static pattern for screenshot filename isn't enough, there's a few cases where I'd want to dynamically select a name.
One usage example:
```py
import time

def _get_screenshot_name(fn: str) -> str:
    dt_str = time.strftime("%d-%m-%Y %H-%M-%S")
    return f"{dt_str} {fn}"

config.screenshot_name_callback = _get_screenshot_name
config.screenshot_pattern = "%04d.png"
```
which would give something like `01-01-1970 00-00-00 0001.png`.

But there's more: can add current chapter to filename, place screenshots into different folders based on the chapter, etc.

### DIFF
--- a/renpy/common/00keymap.rpy
+++ b/renpy/common/00keymap.rpy
@@ -287,6 +287,9 @@ init -1600 python:
 
         pattern = renpy.store._screenshot_pattern or config.screenshot_pattern
 
+        if config.screenshot_name_callback is not None:
+            pattern = config.screenshot_name_callback(pattern)
+
         # Try to pick a filename.
         i = 1
         while True:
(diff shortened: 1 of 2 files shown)


Booplicate:
@renpytom can you think of another way to add a way to format files more nicely? The goal wasn't to make filenames unpredictable, it's much easier to search thru the screenshots if you know when they were made. If you think such callback provides too much power, can there be added other formatting options? Like date/time?

renpytom:
I guess I don't think the current formatting is not nice. If you want to know the  date and time, that's in the metadata of the file.

PR #4288, opened 2023-01-29
Upgrade the fixed rollback documentation to strip obsolete functions

Gouvernathor:
Looking at the commit diffs separately makes it more clear.
Squash-merge is advised.


Gouvernathor:
I'm not sure what the behavior would be regarding the last commit. Is it necessary to add that last button, otherwise the screen can't be returned from ? If so, is it necessary to manually get the rollforward information and feed it to Return, or would Return() do that by itself ?

PR #4297, opened 2023-01-31
Disable empty keywords in screen parsing - with compat

Gouvernathor:
I don't think this should be the one to be merged, but it's a good example of how a compated syntax change could be implemented.


renpytom:
This is close, though we probably would have used the deferred error system for this. But I think it's something that wasn't used, so I took the other PR.

PR #4298, opened 2023-01-31
Bring the web platform off of Beta

Gouvernathor:
Given the amount of support we gave it in recent times, it's probably safe not to consider it beta anymore.


renpytom:
Not yet, especially not when everything is brand-new.

PR #4299, opened 2023-02-01
Keep non-standard spaces in parsed strings

Gouvernathor:
Instead of replacing any patch of whitespace with a standard one, this replaces it with the first space of the batch. The regex will also match less often, only on 2+ spaces instead of 1+ spaces, I guess this can save some computation time.
The result in any game only using standard whitespace should be exactly the same.

I still believe removing collapsing altogether would be a good thing, but this is a conservative way of keeping it while enabling non-standard single whitespaces.

### DIFF
--- a/renpy/lexer.py
+++ b/renpy/lexer.py
@@ -907,7 +907,7 @@ class Lexer(object):
         if not raw:
 
             # Collapse runs of whitespace into single spaces.
-            s = re.sub(r'\s+', ' ', s)
+            s = re.sub(r'(\s)\s+', r'\1', s)
             s = re.sub(r'\\(u([0-9a-fA-F]{1,4})|.)', dequote, s) # type: ignore
 
         return s
@@ -961,7 +961,6 @@ class Lexer(object):
 
         if not raw:
 
-            # Collapse runs of whitespace into single spaces.
             s = re.sub(r' *\n *', '\n', s)
 
             rv = [ ]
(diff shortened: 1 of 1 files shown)


mal:
The change to `\s\s+` seems to make sense (if keeping the collapsing), but using the grouping results in a strange implementation detail where `\t\r` collapses to `\t` but `\r\t ` collapses to `\r`, which even if documented seems like it would become a source for more confusion rather than less vs consistently outputting a replacement space. Maybe only a targetting a subset of the whitespace character class would be better, such as those defined in `string.whitespace` (` \t\n\r\x0b\x0c`)? 🤷

Gouvernathor:
A `\r` character can't be part of a renpy script at that point, it's been stripped by this when inside strings https://github.com/renpy/renpy/blob/d284791ffb8da4fa4f69879e9bd4dd30fa3346f7/renpy/lexer.py#L417-L419 (and by a similar thing just further up for outside of strings). So it doesn't seem like we need this.

Otherwise, string.whitespace doesn't contain the incriminated \u3000 character (　), so it would lead to runs of those to stop being collapsed, while Tom wants the collapse to stay. It would be inconsistent with how the other kinds of whitespace are handled, and I'd rather have none of them collapsing, or all.

mal:
`\r` was just an example of a whitespace character besides a regular space, the point is still valid.

Gouvernathor:
Oh, in that case it's intended behavior on my end. I think to some extent if you put special spaces in your game, you need to expect it to become a little weird and it's your responsibility. If we start considering special spaces I don't think we should discriminate between them without a particularly sound rationale, and I think one that excludes \u3000 is not.
\r (alias \u000D) is a line break on some platforms when on its own, so it's weird all right, but that's as a very special case as \n. I wouldn't put them in the same bag as \t and other more straightforward whitespace gaps of variable length.

Or we can find another way of choosing which space replaces a run of diverse whitespace characters. I'm open to that potentially, but it would need to be a way that keeps all current collapsings in - i.e turns `\t \u3000` into a single character - since I wouldn't support an only partial repeal of the collapsing behavior. And the more complex that choice is, the slower and the harder to understand for creators.
Or we could consider that any of these is too weird and too different as a resulting behavior and it's better not to change anything at all. 🤷

PR #4307, opened 2023-02-02
Fix error in the tutorial when mutating a constant store

Gouvernathor:
I kept this one in just in case, but the other PR is better.


Gouvernathor:
Obsoleted by ebc1ed7

PR #4315, opened 2023-02-03
Add renpy.is_pixel_opaque

Gouvernathor:
Replaces calls to undocumented method Render.is_pixel_opaque.
renpy.render is preferred to Displayable.render because of its caching effects, saving potentially lots of computation time.


renpytom:
This should probably have explicit args, rather than *args. 

I'm also thinking the argument list should be:

renpy.is_pixel_opaque(d, width, height, st, at, x, y)

PR #4324, opened 2023-02-04
Cutscene and loop for web

Tey:
As discussed in #4279, this PR makes cutscenes render the same way as Movie displayables. I'd be lying if I said there is no performances drop wrt. the previous method which let the browser display the cutscene directly using an HTML Video element on top of Renpy canvas. I've tested with Chrome, Edge, Firefox and Android (Chrome), and the drop is mostly noticeable on Firefox though (which has always been the slowest browser for Renpyweb anyway).

The PR also handles the autoplay issue and let the browser manage the video loop instead of relying on Renpy for that (this is faster so there is no more video pause between the end and the restarting of the video, especially good for sprite videos).

There is a [demo here](https://tey.github.io/renpy/video_demo-1.4-web/) (splash can be set to either cutscene or Movie displayable).


Tey:
@keggorrack Thanks for the feedback. Can you open the [Develop menu](https://support.apple.com/guide/safari/use-the-developer-tools-in-the-develop-menu-sfri20948/mac) on Safari and copy/paste everything that is shown in the console tab please? My guess is something goes wrong when [Renpy checks for browser support](https://github.com/renpy/renpy/blob/d2cb7c82b81973c0d0c8c3ec534906f931822e7b/renpy/audio/audio.py#L954) so it does replace renpysound methods with the web version.

keggorrack:
> @keggorrack Thanks for the feedback. Can you open the [Develop menu](https://support.apple.com/guide/safari/use-the-developer-tools-in-the-develop-menu-sfri20948/mac) on Safari and copy/paste everything that is shown in the console tab please? My guess is something goes wrong when [Renpy checks for browser support](https://github.com/renpy/renpy/blob/d2cb7c82b81973c0d0c8c3ec534906f931822e7b/renpy/audio/audio.py#L954) so it does not replace renpysound methods with the web version.

```
[Log] Downloading engine... (renpy-pre.js, line 119)
[Log] Downloading game data... (renpy-pre.js, line 119)
[Log] Running... (renpy-pre.js, line 344)
[Log] Unpacking... (renpy-pre.js, line 119)
[Log]  (renpy-pre.js, line 344)
[Log] Loading script... (renpy-pre.js, line 119)
[Log] warning: 2 FS.syncfs operations in flight at once, probably just doing extra work (renpy-pre.js, line 145)
[Log] warning: 3 FS.syncfs operations in flight at once, probably just doing extra work (renpy-pre.js, line 145)
[Debug] can_play_types – ["audio/ogg", "audio/mpeg"] (2)
[Log] Can't play – "audio/ogg"
```

Tey:
It looks like "audio/ogg" is not supported, so Renpy does not load webaudio. I can prevent the crash in this case, but that means neither audio nor video will work at all.

@renpytom Is that expected that webaudio does not work on Safari? It does load on epiphany (WebKit based browser for Linux) though.

Tey:
I've updated the PR to prevent the crash in case audio/video playback is not supported, and updated the demo as well.

[Safari does not really support "audio/ogg"](https://caniuse.com/?search=ogg) so that explains why webaudio is not loaded. [This is expected](https://www.renpy.org/doc/html/config.html?highlight=webaudio_required_types#var-config.webaudio_required_types) for Ren'Py, but since video playback depends on webaudio, this also disables it. On Safari, we could use wasm decoding for ogg and webaudio for everything else, but the way webaudio is implemented right now makes that impossible (the webaudio methods erase the original renpysound methods). This is probably beyond the scope of this PR, although without this PR, cutscenes do work on Safari then (but not Movie displayables).

qinlili23333:
> On Safari, we could use wasm decoding for ogg

Maybe we can also use wasm decoding for webm videos on safari. Currently webm videos don't work on safari.

Tey:
I've updated the PR to let audio decoding happen in wasm if the browser does not support the required audio formats, while making it possible to support video decoding from the browser.

Some articles seem to indicate that Safari should support WebM containers with VP8+Vorbis at least, but my tests in browserstack fail for such video files. The only video files supported by Safari seem to be MP4 containers with H.264+MP3/AAC. This is also supported by Firefox and Chrome, so that's probably the best target for Web platform if Safari is to be supported. Ren'Py does not include H.264 or AAC decoders on other platforms though, so I've added a note about that in the documentation.

I've updated [the demo here](https://tey.github.io/renpy/video_demo-1.5-web/) with test cases for audio files playback. Only the MP4 video format is expected to work on Safari.

> Maybe we can also use wasm decoding for webm videos on safari. Currently webm videos don't work on safari.

The performances will be awful, at least as long as threads are not supported on the Web platform.

qinlili23333:
> The performances will be awful, at least as long as threads are not supported on the Web platform.

A Chinese streaming website called Bilibili released HEVC video decoding in wasm two years ago as at that time Chrome has no HEVC support. They set up several workers (usually 8 workers at the same time) to solve performance problem, although the effecient is low.
Here is Bilibili's article about this, but only in Chinese. 
https://www.bilibili.com/read/cv16257864

renpytom:
The restrictions on hosting multithreaded wasm are high enough it's not going to be a good solution for a while still. (And it's hard to see why you'd use HEVC rather than AV1, just because of the patent problems.)

PR #4326, opened 2023-02-05
Include an example of a proper usage of time_warp in the docs of ImageDissolve

Galo223344:
Even though time_warp is mentioned three times in the [transition classes document](https://www.renpy.org/doc/html/transitions.html#transition-classes), it's never shown how to actually use it. Since the idea of using the _warper module is not intuitive, and it's mentioned only once at the beginning of the section, I thought it would be best to at least include it in one of the examples.

### DIFF
--- a/renpy/display/transition.py
+++ b/renpy/display/transition.py
@@ -442,8 +442,8 @@ class ImageDissolve(Transition):
 
     ::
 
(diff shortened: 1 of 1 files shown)


Gouvernathor:
Looks good to me

PR #4336, opened 2023-02-08
Reachability update for UserStatements

Gouvernathor:
TODO:
- check PostUserStatement for possible complications
- check the case where block != "script" but the parsed content still includes renpy statements (the "random" example in the doc, for one)
- change the nomenclature of the passed values, RIR is probably illegible ("flow" ?), and Always is terrible ("innate" ?)
- add examples and more explanation in the doc ?


renpytom:
I just landed a different approach to this (a function that's can report what's reachable). I think it's fundamentally simpler.

PR #4369, opened 2023-02-19
Fix classes signatures in the documentation (again)

### LINKED ISSUE #4317: Signatures of classes not overriding __init__
A lot of actions, mainly, have a mro where no parent except RevertableObject redefine the `__init__` method.
As a result, they end up with RO's bogus signature, when actually they don't accept any parameter. (RO ignores any parameter when not config.developer, for compat reasons I assume)
This is probably a very old issue.

Some examples can be found in [screen_actions](https://www.renpy.org/doc/html/screen_actions.html) : NullAction, HideInterface, RollForward, Screenshot.

Gouvernathor:
The first commit ~~grinds commit points~~ simplifies things : now `object().__init__` is the same as `python_object().__init__`.
The second bans classes not overriding `__init__` (regardless of whether they inherit revobject or not, thanks to the first commit) from getting their init signature read. I'd gladly monkeypatch python_object the same way I had revobject, but the python_object type is sadly not writeable (as written in C). This fixes achievement.Sync's `/, *args, **kwargs` signature, and fixes #4317, for good this time.
The third seeks to remove `self` from signatures such as `(self)` as well. It does so not by string replacements, but by directly removing the self parameter from the signature object. I think this is less prone to errors.


Gouvernathor:
I suggest merging this _**after**_ #4368, so the changes to script.rpy don't get overridden.

PR #4386, opened 2023-02-22
Fix inertial viewport dragging

shawna-p:
This fixes inertial viewport dragging. Namely, it adds an `instantly` parameter to the Adjustment `end_animation` method such that the relevant Adjustment parameters can be reset for the animation to instantly stop where it is. This is used 1) when the user touches down again, so the viewport stops when they press down, and 2) when the drag hits one of the viewport limits (so it can't be dragged any further). You may also consider using it in other circumstances, such as stopping the animation when the viewport scrollbar is grabbed.


renpytom:
Merged, and I'll be adding this to the bar focusing code as well.

PR #4398, opened 2023-02-25
renpy: Make version PEP440-compatible

lanodan:
Starting with 67.2.0 setuptools enforces PEP440 compliance.


renpytom:
Merged, thank you.

PR #4437, opened 2023-03-13
Premature redraws erase pending redraw orders

Gouvernathor:


renpytom:
Fixed by a628b2dba3c596461e956c471545c05961e8c241 , which is I think a better place to document this.

PR #4439, opened 2023-03-14
allow poi the image tag or camera by adding poi_is property

kyouryuukunn:
.. transform-property:: poi_is

    :type: str or None
    :default: None

    This gives the displayable or camera as the point of interest. The camera
    or the displayable being transformed are rotated to face the displayable or
    camera, even if the position of the camera or the displayable is changed.

    If this is not None. this is string giving the displayable tag or camera.
    That format is "tag@layer" or "tag". If layer is not given, default is
    "master".  If tag is "camera", it means the camera

    poi_is has priority over poi.

    Note poi_is isn't updated automatically. so, you should write like below if
    you want it is updated::

        show eileen happy at center:
            poi_is "camera"
            0
            repeat


Gouvernathor:
If the two properties are mutually exclusive, and in addition they don't take the same types why not merge them ?
You have either None, and nothing happens, or a tuple, and the poi behavior triggers, or a string, and the poi_is behavior triggers.

In addition, I would be against reserving the string "camera" since it is a valid image name/tag. Why not `True` as the value for targeting the camera ?

mal:
If this goes forward, then in addition to Gouv's suggestions, when not provided the `default_layer` function should be used to determine the layer rather than hard code `master`.

What happens if a non-existent displayable is referenced, or the displayable is removed from the scene before the thing referencing it (in the repeat scenario)? What happens if the reference is itself? 😬

It looks like the new `if state.poi_is` blocks are copy-paste duplicates, and while I realise the wider `poi` code is not between the perspective and non-perspective branches, perhaps there's a saner way to handle the repeated code block. Haven't thought too much about it, just putting it out there as something to consider.

More generally, I'm not entirely sure how to feel about transforms being contingent on the scene list and other displayables, it definitely feels like quite a departure from the current features, not to mention what compatibility issues it may create in future for a more deterministic ATL.

renpytom:
I like pointing at the camera. That's certainly something we should have. I'm a little skeptical about the utility of pointing at an image, rather than the camera. @kyouryuukunn, do you have an example of where something like: 

```renpy
   show lucy happy:
        point_to "eileen"
```

would be necessary? I can't think of any, but I'm willing to have my mind changed on this. 

I don't think that making transforms contingent on the scene list matters. The property (which is just a string) depends just on time. What happens in the renderer doesn't actually matter that much when it comes to predictability, compared to what happens in ATL, which is where determinism needs to be. 

Regardless of what we do I'd like to rename `poi` to `point_to`. I think the more verbose name is better. I think this property should be allowed to contain:

* None, to disable this behavior.
* A tuple of (x, y, z) coordinates.
* the special value Camera. (I can set up a sentinel value in the store.) 

If we then want to add non-camera pointing, we can do that - but I'd like to know where pointing at a displayable would be useful.

Gouvernathor:
I think there is a hidden risk in the displayable-pointing mechanism, as mal pointed out. The property itself, being a string, is constant, sure, and if the displayable moves it will result in different xyz rotations when called twice on the same value : I understand why _that_'s not a problem (same as for the `shader` property). The issue would be if the displayable is not _found_ to be displayed at all, at the time the renderer executes. And, as mal said, if the reference is the caller itself (which is harder to detect than you may think : displayables in Flatten in Fixed in a screen...). It's not a problem of predictability of the result, but of having _a_ result in any given situation.
If I write `function my_func` or `my_py_transf(child)` in ATL and the function raises an exception, it's my fault, I opened the pandora's box by letting python enter the chat. Same as if I call a shader with a bug in it. But if I set a string to a documented transform property, I think that's less expected to ever fail. It would likely bounce back on us as a bug.

Pointing at a layer (at its center or at its normal vector for example) may save the day, as being more reliable. Depends on whether the use-cases for pointing at a displayable can be approximately satisfied by pointing at a layer.
Careful about detached layers when implementing though.

renpytom:
Let's figure out if image-pointing is useful, first. Then we can worry about the semantics of it. 

We don't avoid useful features just because they might break. It's perfectly find to have the semantic where if a points_to isn't found, it's the same as points_to None.

kyouryuukunn:
@Gouvernathor 
Thank you. simplification. I will rewrite this after #4447 is merged.

OK, I merged poi  and poi_is to point_to and I changed point_to format like below
3-tuple:      (x, y, z)
str:             "tag"
True:          camera for "master" layer
(str, str):     ("tag", "layer")
(True, str):   camera for str "layer"

@mal 
I thought If an image which is not shown is specified , it is acceptable to throw an error.
If the image specified is removed while the image which faces it is being shown, it might be better to store the last position into anywhere and keep the last state.

if itself is given, nothing is shown or it is rotated at odd angles.

@renpytom 
In case of Image to Image, there certainly seems to be little opportunity for use.
It could be done in the same way as Image to camera and camera to Image, so I just added it.

renpytom:
I've merged #4447 

Let's get rid of image-to-image. I think it adds a lot of complexity for something without an obvious use case. (For example, it introduces a novel layer syntax.) I don't really want to add image-to-image as something we have to support unless there's a clear use for it.

kyouryuukunn:
moved to #4449

PR #4443, opened 2023-03-15
Fix TintMatrix to respect alpha

mal:
Per the documentation, the alpha shouldn't ever have been impacted.

This change instead treats the alpha in the supplied RGBA value as the tint opacity rather than applying the tint formula to the alpha channel. The result is the ability to use a colour value of `#FF000080` to apply a 50%-red tint to the target.

<details><summary>Control, before, and after screenshots.</summary>

screenshot0000
screenshot0001
screenshot0002
</details>

<details><summary>Reproduction code.</summary>

```rpy
label main_menu:
    $ _confirm_quit = False
    return

image s = Solid('777', xysize=(320, 240))

label start:
    scene
    show s at Transform(align=(.5, .5))
    'Original'
    show s:
        matrixcolor TintMatrix('ff000080')
    'Test'
```
</details>

I did consider compat for this, but given the current behaviour directly conflicts with the documentation, and also doesn't align with the concept of tinting (when alpha is present) in most image editing software, decided against it.

### DIFF
--- a/renpy/common/00matrixcolor.rpy
+++ b/renpy/common/00matrixcolor.rpy
@@ -167,17 +167,16 @@ init -1500 python:
                 b = oldb + (b - oldb) * done
                 a = olda + (a - olda) * done
 
-            # To properly handle premultiplied alpha, the color channels
-            # have to be multiplied by the alpha channel.
-            r *= a
-            g *= a
-            b *= a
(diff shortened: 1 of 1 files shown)


renpytom:
Makes sense. Merged.

PR #4446, opened 2023-03-16
New action for lists

lauwurence:
I have an idea about a new action that causes variable to be set to the next value in list.
It significantly reduced amount of code in my project and I think it may help others to write their custom preferences screens faster.

Before:
```python
if preferences.graphics== "low":
    action SetVariable('preferences.graphics', "medium")
elif preferences.level == 'medium':
    action SetVariable('preferences.graphics', "high")
else:
    action SetVariable('preferences.graphics', "low")
```

After:
```python
action NextVariable('preferences.graphics', ["low", "medium", "high"])
```

So, what do you think?


Gouvernathor:
The kwargs probably need to be kept off the documented signature, but that's a detail.
This implementation will suffer the same thing as SetLocalVariable does : you can't define the action globally, outside a screen, and use it in several ones. While it makes little sense to do that with SetLocalVariable (due to what it does), it possibly would for this one, and saving actions is something we usually support. See Andy's comment in #3194.

Overall I'm not convinced it's a good idea to add something to what I see as our massive chaotic heap of actions (whereas #3194 ultimately didn't add one to the list). Perhaps this is something good for snippets to be shared and copied by creators from game to game ? We do support creating actions and even subclassing the [Action](https://nightly.renpy.org/doc/screen_python.html#Action) class.

via-my:
Hmm, ok. But it was worth a shot.

renpytom:
I'm going to re-open this, for now, until we can convert it to a feature request. 

I'm sort of thinking what we probably want to do  - post 8.1 - is to break up the actions (Set, Toggle, and call this one Cycle) and how one accesses the data (Variable, ScreenVariable, Field, Dict, etc.) into classes, and then use a factory function to make the appropriate sub classes that double-inherit. 

So I don't think this is something I plan to merge directly, but I do think it's a good idea.

via-my:
Well, it's not really the graphics level but the UI animations complexity. It has a changing description and an animated or static image.

```python
$ uwu = preferences.graphics

use _sc_preferences_template(
    __("UI Animations Complexity"),
    description={
        'low'  : __("simplified interface animations"),
        'full' : __("the most complete interface animations"),
    }.get(uwu,   __("animations disabled")),

    button_image={
        'low'  : 'img_preferences_donut_1',
        'full' : 'img_preferences_donut_2',
    }.get(uwu,   "gui/icons/preferences/donut.webp"),
    button_action=NextVariable('preferences.graphics', [None, 'low', 'full']),
    )
```

mal:
When this gets worked on, it might be nice to have a counterpart to whatever `NextVariable` becomes to cycle the other way, and also an option to enable or disable looping.

PR #4447, opened 2023-03-16
Rename poi to point_to

Gouvernathor:
Includes documentation references.
Also simplifies the code, it should be a small bit quicker and simpler to read now (which are both for the common good in accelerator.pyx).
Discussed in #4439.


renpy/display/accelerator.pyx
```diff
@@ -610,17 +610,17 @@ def transform_render(self, widtho, heighto, st, at):
     else:
         self.reverse = Matrix2D(rxdx, rxdy, rydx, rydy)
 
+    poi = state.point_to
+    if not (isinstance(poi, tuple) and len(poi) == 3):
```
mal, line 614: `None` would raise here but I don't believe that's the intent.

```suggestion
    if poi and not (isinstance(poi, tuple) and len(poi) == 3):
```

PR #4449, opened 2023-03-17
allow point_to to take "tag" or camera

kyouryuukunn:
I rewrite #4439 to base #4447.

    :type: (float, float, float), (str,str), (True, str), str, True or None
    :default: None

    This gives a position to which to point. The camera or the
    displayable being transformed are rotated to face that point,
    even if the position of the camera or the displayable is changed.

    If this is None, no point-of-interest rotation is applied.

    If this is not None, this is a 3-tuple, 2-tuple, True or string. (x, y, z)
    format tuple represents the position of the point of interest. 
    ("tag", "layer") format tuple represents the tag on the layer as the point
    of interest. the string giving the tag represents the tag on the layer
    which is the default layer for that tag as the point of interest.

    True means camera. (True, "layer") format tuple represents the camera for
    "layer" as the point of interest. True is the camera for "master" as the
    point of interest.

    no point-of-interest rotation is applied if displayable given to point_to
    isn't shown.

    Note point_to isn't updated automatically. so, you should write like below if
    you want it is updated::

        # eileen always faces to the camera for master layer.
        show eileen happy at center:
            point_to True
            0
            repeat


kyouryuukunn:
Sorry, I found some problem in code. I will fix

renpytom:
Can you remove the (tag, layer) values? I don't think this should be included in Ren'Py, until there's a more obvious use for it.

kyouryuukunn:
OK, I removed 2-tuple format.

mal:
Following the conversation in the last issue, my expectation was for the `tag` part to go too, at least "until there's a more obvious use for it".

So the supported formats would be `(float, float, float), True, or None` representing a point in space `(x, y, z)`, the player's viewpoint (i.e. camera), or disabled respectively.

mal:
Also wondering if the conversion of `True` to the camera's (x, y, z) ought to be done in a function that could be reused to allow interpolation between `True` and another point in space to work as a creator might expect. It could also help clean up the currently quite sizeable duplication in this PR.

kyouryuukunn:
I thought tag as a syntactic sugar which allow users to not have to keep the position of the image when the camera is pointed at that image.

But, I noticed that the image size wasn't gotten before rendering it, it is required to consider anchor and matrixanchor.
So, I remove "tag" format from point_to.

And I allow interpolate True to (x, y, z) and (x, y, z) to True.

renpytom:
Merged, and then did fbbc8e0e3477e13d71caf836e74f8f8c5a363b09 and cc77af4d9fe348c5a028b56b9388235697dae6cb.

I removed "interpolate True to (x, y, z) and (x, y, z) to True.", as I don't think it was correct when the camera itself was moving.

PR #4454, opened 2023-03-18
Multiline support for input

valery-iwanofu:
In some special cases, multiline for input is required. This PR covers it.

What this PR contains:
- Added `input_next_line` into `keymap`(both `00keymap.rpy` and `keymap.rst` updated)
- Added `input_next_line` event handling in `renpy.display.behaviour.Input`
- Added multiline keyword in `renpy.display.behaviour.Input`, `renpy.exports.input()`, `sldisplayables.py`, `sphinx/keywords.py`, `sphinx/screens.rst`


Gouvernathor:
This needs to be tested, but I like the idea and purpose.
It could be a great improvement to the console.

PR #4473, opened 2023-03-23
Action mixins

Gouvernathor:
Added features :
- the 5 new Cycle actions (via-my's idea), with the reverse and loop parameters
- the 5 new Increment actions (my idea)

And some help from Kassy2048 for the documentation.

<details>
<summary>This is how the doc renders now (collapsed, click to expand, nearly but not up to date)</summary>

image
image
image

</details>

(settled things)
- I did a comprehensive audit of which exception was raised when the data to update isn't available when running the action or calling its get_selected.
Previously, exceptions were raised (and never caught) only in the Set/ToggleField and Set/ToggleVariable actions, in their get_selected methods (which is weird), as well as when the Toggle ones were run (which is normal).
Having a Toggle (or a Cycle) raise at `run`time if the data can't be found is normal, having it raise in the get_selected or get_sensitive methods is probably not, especially since the Dict and ScreenVariable actions do protect their get_selected methods.
Now, the get_selected and get_sensitive methods are protected from all possible data-access exceptions.
- the existing ToggleVariable function, for some reason, has its first parameter named `variable` where all other Set/Toggle(Screen/Local/)Variable call it `name`. The new one has a parameter called `name`, as do the others. I don't think enough people would be passing `variable=` as a keyword argument to SetVariable to deserve a special-case in the class generator mechanism.
- The `kind` parameter. It was taken by SetField and ToggleField but never documented. They help format the error message when the field/variable name is not found. It was not documented, so we don't support it.
- Pickling works fine, I tested it. Legacy Set/ToggleVariable are resurrected into SetToggleField, but they keep working all the same apart from not testing equal (which could have a microscopic impact on performance). Set/ToggleLocalVariable were unpicklable before (raising an exception), so we don't care about those either.


Gouvernathor:
<details>
<summary>Here is how the doc renders now (collapsed, click to expand)</summary>

image
image
image

</details>

lauwurence:
Hmm... looks like something that is hard to quickly understand but I like the idea of ​​the table.
What about making it a quick accessor to the actions?

Here are my five cents.

<details>
  <summary>Documentation</summary>

_

### Data Actions:

......................................................................................
| The table with links to the actions below |
......................................................................................
Some general info like: 
`Set-`, `Toggle-`, `Cycle-` and `Increment-` are the managers and are used as prefix to `-Variable()`, `-ScreenVariable()`, `-LocalVariable()`, `-Field()` and `-Dict()`.

_

### Data Set Actions:
The `Set-` actions simply set the value of the target to the passed `value`. Note that this has nothing to do with ``set``, which is a builtin type in Python.

**• SetVariable**(name, value)
Causes the variable called name to be set to value.
> **Note:** The name argument must be a string, and can be a simple name like "strength", or one with dots separating the variable from fields, like "hero.strength" or "persistent.show_cutscenes".

**• SetScreenVariable**(name, value)
Causes the variable called name associated with the current screen to be set to value.
> **Note:** In a used screen, this action sets the variable in the context of the screen containing the used one(s). To set variables within a used screen, and only in that case, use `-LocalVariable()` instead `-ScreenVariable()`.

**• SetLocalVariable**(name, value)
Causes the variable called name to be set to value in the current local context.
> **Note:** This function is only useful in a screen that has been used by another screen, as it provides a way of setting the value of a variable inside the used screen. In all other cases, `-ScreenVariable()` should be preferred instead `-LocalVariable()`, as it allows more of the screen to be cached.
> 
> For more information, see [Use].
> 
> This must be created in the context that the variable is set in - it can't be passed in from somewhere else.

**• SetField**(object, field, value)
Causes the a field on an object to be set to a given value. object is the object, field is a string giving the name of the field to set, and value is the value to set it to.

**• SetDict**(dict, key, value)
Causes the value of key in dict to be set to value. This also works with lists, where key is the index at which the value will be set.

_

### Data Toggle Actions:
The `Toggle-` actions invert the boolean value of their target, between `true_value` and `false_value`.
>**true_value:**
>   If not None, then this is the `True` value used.
>**false_value:**
 >  If not None, then this is the `False` value used.

**• ToggleVariable**(name, true_value=None, false_value=None)
Toggles the variable called name.
> This action follows the same note as `SetVariable`. < [hyperlink]

**• ToggleScreenVariable**(name, true_value=None, false_value=None)
Toggles the value of the variable called name in the current screen.
> This action follows the same note as `SetScreenVariable`. < [hyperlink]

. . .


</details>

Gouvernathor:
That's possible, but it would make the doc much longer and duplicate explanations (about Managers in this instance), in a page that's already far too long in my view. Giving various actions different doc entries would also require a more complex generation system, or to write it all manually in the rst file which would make *it* long, and hard to navigate internally.
I'm not sure how best to optimize the two

Kassy2048:
> I think the _X_ is not very visible and understandable. I tried "-Variable" instead of "_X_ Variable" but it's not much better. Maybe letting it go altogether, and talking about "the Variable actions", "the Cycle actions" ?

Maybe you can use the Sphinx [abbreviation feature](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-abbr) to make it more visible and clear. Something like `` :abbr:`XVariable (SetVariable, ToggleVariable, CycleVariable, IncrementVariable)` ``?

Gouvernathor:
I did that in cunjunction with dashes, I think that's the best we can make this version - other than changing it more deeply like lauwurence proposed.

renpy/common/00action_data.rpy
```diff
@@ -34,372 +35,333 @@ init -1600 python:
...
+            if self.loop:
+                return True
 
-        def get_selected(self):
             try:
-                rv = self.dict[self.key]
-            except (KeyError, IndexError):
+                value = self.current_value()
+            except (__LookupError, __NoCurrentScreen, __ScreenVariableNameError): # __FieldNotFound ?
                 return False
```
Ayowel, line 296: I can't write as a suggestion because this covers deleted lines.
The test for `self.loop` should occur later to be able to handle cases where the loop is small (0 or 1-sized).

```py
            try:
                rv = self.dict[self.key]
            except (KeyError, IndexError):
                value = self.current_value()
            except (__LookupError, __NoCurrentScreen, __ScreenVariableNameError): # __FieldNotFound ?
                return False

            if self.loop:
                return len(self.values) > 1 or (len(self.values) == 1 and self.values[0] != rv)
```

renpy/common/00action_data.rpy
```diff
@@ -34,372 +35,333 @@ init -1600 python:
...
-        :func:`ToggleScreenVariable` should be preferred, as it allows more
-        of the screen to be cached.
+            values = self.values
+            if value in values:
+                idx = values.index(value)+1
+                if idx >= len(values):
+                    return False
 
-        For more information, see :ref:`sl-use`.
+            return True
```
Ayowel, line 304:
```suggestion
            return len(values) > 0
```

Ayowel (changes requested):
Add support for some edge cases when cycling.
Might also be a good time to add a RemoveFromDict action or update RemoveFromSet so it supports it

renpy/common/00action_data.rpy
Gouvernathor: I changed my mind for the empty sequence case, because it's cheap in execution time.
It will still raise when the action is run of course.

PR #4483, opened 2023-03-25
Cancel long press wait when user stops touching the screen

### LINKED ISSUE #4482: Alternate action always runs on touch variant (web)
On touch devices (at least for the Web platform), when a button has an alternate action and the user clicks on it, both the normal and alternate actions will be executed. The alternate action is only supposed to be executed on long presses, but it also happens on clicks after some delay wrt. the normal action.

My guess is the button [unfocus()](https://github.com/renpy/renpy/blob/d3dbc4074917bb5fede3fdd0c9bf8bac4a9785e1/renpy/display/behavior.py#L998) method is not called when user does not touch the button anymore, so the long press timer is not "canceled". The normal action is executed though, so RenPy does know that the user has stopped touching the button somehow.

This can be reproduces on any Web games by either playing it on a mobile or using Chrome on a desktop and switching to mobile device mode using the DevTools, then reloading the game. Make sure all skip options are enabled, then click on the Skip button in the quick menu. The game will start to skip normally, then a prompt will show up a few seconds later to confirm if the user wants to skip. This is the alternate action of the Skip button being executed.

I'm trying to fix that, but if an obvious/easy solution comes to mind, please chime in.

Kassy2048:
Either choosing to cancel the long press detection when focus is lost was a bad decision, or something is broken in the focus logic for touch devices. Checking for a mouse up event to cancel the long press detection as this PR does is a more obvious way to do it.

I've kept the detection logic based on focus just in case it makes sense for other platforms.

Fix #4482.

### DIFF
--- a/renpy/display/behavior.py
+++ b/renpy/display/behavior.py
@@ -1115,11 +1115,10 @@ class Button(renpy.display.layout.Window):
 
                 renpy.game.interface.timeout(renpy.config.longpress_duration)
 
-            elif ev.type == pygame.MOUSEBUTTONUP and ev.button == 1:
-                self.longpress_start = None
-
             if self.longpress_start is not None:
-                if math.hypot(x - self.longpress_x, y - self.longpress_y) > renpy.config.longpress_radius:
+                if ev.type == pygame.MOUSEBUTTONUP and ev.button == 1:
+                    self.longpress_start = None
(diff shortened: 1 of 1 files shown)


mal:
No opinion on whether or not this is the right fix, but a general observation is that it may be good to move this check inside the `self.longpress_start is not None` conditional immediately below it since there's no point adding this check for the case where it's already `None`.

Kassy2048:
Makes sense👍

PR #4494, opened 2023-03-30
Allow comments in shader variables lists

Gouvernathor:
Just like in any code, C-style (or GLSL-style) single-line comments, starting with //, are allowed.
This applies to the `variables` parameter of renpy.register_shader.


renpytom:
We should also support the /* */-style comments. Or at least detect them and report it's not a supported syntax.

PR #4510, opened 2023-04-05
Fix detection of multiple declarations of a style's parent

Ayowel:
* Fix an issue that caused multiple declarations of a style's parent in a single statement to not trigger an error in `renpy/parser.py`.
* Update `gui/game/screens.rpy` to declare parents with the style content's declaration to improve style readability and avoid undetected multiple-definition of parents in the default project file (e.g.: at the moment, the `choice_button` style inherits from `button`, which is later silently erased in favor of `default`)

### DIFF
--- a/gui/game/screens.rpy
+++ b/gui/game/screens.rpy
@@ -121,16 +121,7 @@ screen say(who, what):
 init python:
     config.character_id_prefixes.append('namebox')
 
-style window is default
-style say_label is default
-style say_dialogue is default
-style say_thought is say_dialogue
-
-style namebox is default
-style namebox_label is say_label
-
-
-style window:
+style window is default:
     xalign 0.5
     xfill True
     yalign gui.textbox_yalign
@@ -138,7 +129,7 @@ style window:
 
     background Image("gui/textbox.png", xalign=0.5, yalign=1.0)
 
-style namebox:
+style namebox is default:
     xpos gui.name_xpos
     xanchor gui.name_xalign
     xsize gui.namebox_width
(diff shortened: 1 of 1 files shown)


renpytom:
I see it the opposite - the parser change is trivial, and I'll accept it right out of the bat. It doesn't touch the default theme, as far as I can tell.

The redoing of the style inheritance - I'm not s ure about it. I'm going to hold this for 8.2 at the earliest.

Ayowel:
> starting with all those using the default theme

It doesn't, because the check only occurs if the parent is re-declared *in the same statement*, which is not the case in the 'old' default file.

> The redoing of the style inheritance - I'm not s ure about it.

I'll rebase the branch later to leave only this change as the other is now merged.

The main point of the file change is to ensure that users get examples that can be used with as many fail-safes as possible.
The current layout might make someone believe that they can set multiple parents, which is not supported and silently dropped if the user does it in multiple statements as the current layout suggests:

```py
style say_thought is say_label # This is silently dropped
style say_thought is say_dialogue # This is the actual parent
```

Furthermore, the current layout shows a separation of content declaration and inheritance, which means that - as a developer - I end up scrolling up and down to figure out the style hierarchy of an item and might end up re-declaring a parent without being aware of it (thus erasing some of the previously-applied attributes). I know how to use `SHIFT` + `I`, but depending on the scene, doing it on the right element can be hard, and systematically having to run the game and retrieve the one place a style is used when doing minor changes is annoying yet still faster than direct file editing in many cases due to this "split" declaration.

Overall, I feel that this change would make the code's behavior more predictable for new users, especially when making changes to it (hence why I only changed this file and not many of the others used in Ren'Py).

renpytom:
I don't think I want this change. I liked the separated inheritance and definition when I first wrote it, and I still like it, stylistically.

PR #4541, opened 2023-04-18
Parse (ignore) typing annotations for define and default

### LINKED ISSUE #3947: [Feature Request] - Add support for type annotations to default and define statements
Hi there, just putting in a request for type annotation support for default/define statements as there's situations in my games where I `default` a variable to `None`, but later down the line, assign it to some complex type.

For example, a dictionary mapping strings to tuples of floats.

Under the current system I'd have to use a comment, which is okay, but it'd be nice if the syntax supported it directly, such as:
`default someVariable: dict[str, tuple[float, float]] | None = None`

Thank you!

Gouvernathor:
The following is now allowed :
```py
define a : str = "Hello World"
default b : str = "Hello World"
# define c : int # fails
# default d : int # fails

init python:
    def adj_eileen(attributes):
        return attributes

define config.adjust_attributes["eileen"] : Callable[tuple, tuple] = adj_eileen
define config.adjust_attributes : Dict[str, Callable[tuple, tuple]] |= {"eileen2" : adj_eileen}
define config.after_default_callbacks : List[Callable] += [print]
```
The annotations are completely ignored, and taking them into account falls solely on the IDE syntax analyzer. I think this is also how it works in Python for non-return and non-parameter annotations ?

This was particularly easy to implement, I'm not sure it's worth it though - if the burden of support falls on the IDE extensions anyway, it might as well use the `typing :` comments.

### DIFF
--- a/renpy/parser.py
+++ b/renpy/parser.py
@@ -963,6 +963,9 @@ def define_statement(l, loc):
     elif l.match(r'\|='):
         operator = "|="
     else:
+        if l.match(":"):
+            l.match(r"[^=\n]*")
+
         l.require('=')
         operator = "="
 
@@ -998,6 +1001,9 @@ def default_statement(l, loc):
     while l.match(r'\.'):
         store = store + "." + name
         name = l.require(l.word)
+
+    if l.match(":"):
+        l.match(r"[^=\n]*")
 
     l.require('=')
     expr = l.rest()


mal:
To me, this seems like potentially opening a hole into which non-programmer creators can fall without intending to, for very limited benefit to a minority (supposition mine) of programmer creators eager for type annotations in RPY files. It also feels a bit strange to add type annotation support to RPY syntax piece-meal, such that the answer to "can I use type annotations in rpy scripts?" is along the lines of "well ... kind of, a bit".

All that aside, I'm not sure if we have "support type annotations in rpy" as a long-term design goal, and perhaps that is the real deciding factor here.

Gouvernathor:
I think I agree. The [Python equivalent](https://peps.python.org/pep-0526/) actually doesn't ignore the annotation value, even at a module level : the evaluated annotation (or the string, if using the 563 `__future__` which is the default in renpy I think) is stored in the `__annotations__` dict.
I don't think we should enable annotating short of updating that dict as well - and if we do, we have to make it cooperate with `renpy.dynamic` variables and called label parameters. So, this would probably be too much work for too little use, at least until proven otherwise.
I'll leave this open, but I don't advise merging without further discussion.

renpytom:
I'm sort of thinking we do want annotations at some point, but it should be a comprehensive thing. At the very least, this is going to need to be supported in the VS code plugin, to have any benefit.

PR #4543, opened 2023-04-19
Reachability tweaks

Gouvernathor:
About the doc change, I'm not sure why the "label of the statement" arg would always be a string.

### DIFF
--- a/renpy/lint.py
+++ b/renpy/lint.py
@@ -854,8 +854,8 @@ def check_unreachables(all_nodes):
     # Unreachable nodes - this set shrinks as nodes become reachable.
     unreachable = set(all_nodes)
 
-    # Weakly reachable nodes - nodes that are reachable, but don't imply
-    # that nodes
+    # Weakly reachable nodes - nodes that are reachable, but don't
+    # make their next reachable.
     weakly_reachable = set()
 
     # The worklist of reachable nodes that haven't been checked yet.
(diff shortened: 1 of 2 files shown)


renpy/lint.py
```diff
@@ -917,12 +925,7 @@ def add_names(names):
         if next in unreachable:
             to_check.add(next)
 
-    unreachable -= weakly_reachable
-
-    locations = sorted(set((node.filename, node.linenumber) for node in unreachable if not isinstance(node, (renpy.ast.Return, renpy.ast.EndTranslate, renpy.ast.Init, renpy.ast.TranslateBlock))))
-    # the auto-generated Return at the end of every file is hard to segregate from the other Return nodes, so we don't check Return nodes
-    # EndTranslate nodes can't be manually created, so it makes no sense to show them to the user in the first place, and EndTranslate nodes from explicit translate blocks are naturally unreachable
-    # Init and TranslateBlock nodes are meant to be unreachable, but we had to check them because if they are reachable, what follows them is too and must not be flagged as unreachable
+    locations = sorted(set((node.filename, node.linenumber) for node in (unreachable - weakly_reachable)))
```
mal, line 928: I believe part of this is what was referred to by "Avoid a set update if the original is not read afterwards", but while it avoids updating `unreachable`, it does so the cost of creating an entirely new set prior to iterating which is slower. Not terribly important either way, just pointing it out.

Gouvernathor: I wasn't sure about it (about which would be slower), and you're probably right.
Lint is not very performance-critical and I think the content and meaning of the variables is more clear this way (the weakly reachable _are_ unreachable, we only decide not to report them), but I'm ok with changing it for performance reasons if you think it can make a difference.

PR #4545, opened 2023-04-20
Add a type check on data saved in persistent

Gouvernathor:
This checks whether the type of the object being pickled into persistent is of a supported type (defined `early` enough), or the object itself if the object is a type.

This probably adds some amount of execution time at loading, and possibly at saving, but I don't know how the check could really be done more quickly than this.
For some reason I did not fully investigate, `renpy.config.developer` is not set to True when init() executes, so the creation of `safe_types` cannot be done on that condition. It would save some time in distributed games, if it could. The computation cannot be made any later, since the only call to persistent.init() happens right at the time when it's too late to define new persistable types.

If the check was done in `safe_deepcopy` as the issue suggests, the creator could see easily which persistent entry is faulty ; however it wouldn't check the type of elements of lists, or attributes of objects.

### DIFF
--- a/renpy/compat/pickle.py
+++ b/renpy/compat/pickle.py
@@ -58,6 +58,8 @@ if PY2:
         else:
             return pickle.dumps(o, PROTOCOL)
 
+    persistent_dumps = dumps
+
 else:
 
     import functools
@@ -92,6 +94,18 @@ else:
 
             return super().find_class(module, name)
 
+    class PersistentPickler(pickle.Pickler):
+        def reducer_override(self, obj):
+            if renpy.persistent.safe_types is not None:
+                t = obj
+                if not isinstance(obj, type):
+                    t = type(obj)
+
+                if t.__module__.startswith("store") and (t not in renpy.persistent.safe_types):
+                    raise TypeError("Cannot save {} : Type {} is not safe for persistent saving.".format(obj, t))
+
+            return NotImplemented # lets normal reducing take place
+
     def load(f):
         up = Unpickler(f, fix_imports=True, encoding="utf-8", errors="surrogateescape")
         return up.load()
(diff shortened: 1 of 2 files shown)


mal:
Perhaps this would be something better monkey-patched in after `early` code has been run if `config.debug` is set to `True`?

```py
if renpy.config.debug:
    allowedtypes = set()
    # populate allowed types

    class DebugPickler(Pickler):
        ...

    def debug_dumps(...):
        ...

    renpy.persistent.dumps = debug_dumps
```

It would give clean separation from the compat module, only incur a cost when activated, and all be encapsulated in one place.

Gouvernathor:
Good idea about getting it all in one place, but I'm less sold on the config.debug toggle. Andy's idea was for this to be a permanent safety for creators when they test their game, not a lint-like one-shot stress test. This because as opposed to code that could fail under particular conditions, things reported here would definitely break the next loading.

Maybe we could have a different approach to persistent security than Andy proposed, and which would be more time and memory efficient : customize the unpickler and make it return None for invalid objects when config.developer is not True. Somehow.

Booplicate:
You could define `safe_types` to an empty `set` right away, `None` adds complexity, but no benefits in this case.

Currently you're adding some overhead, but only getting a nice error message `Type X is not safe for persistent saving`. You'd get a pickle error on next load anyway - something you should do during testing. So technically, you're replacing one error with another.

> customize the unpickler instead (again) and make it return None for invalid objects

While this might save you from a crash in some cases, in most cases you'd get a crash anyway (e.g. `NoneType has no attribute X`). And then the developer would have to guess why some variable became `None` suddenly, instead of a more clear "Couldn't unpickle X because it's not found". So is there really any gain out of it?

The problem of an object not being pickleable is not unique to RenPy - you can encounter it in Python as well. Its root is incorrect logic in code, and I'm not sure it makes sense to try to fix wrong logic for other people. Especially if there's a trade-off that *every* game will have to pay for. The right solution is to do testing and ideally only pickle builtin types (and use something like `to_dict`/`from_tuple` for custom classes).

Gouvernathor:
This is not simply a better error message though, it's about preventing the persistent file from being overwritten by corrupted data. It was also the point with replacing things with None : you lose them, and they cause errors, but at least you keep the rest, the seen labels, the preferences (I think), the console history...

PR #4564, opened 2023-04-23
Tweak and fix bubble doc

Gouvernathor:


renpytom:
I'd probably write better documentation if I did it while awake.

Gouvernathor:
That tends to work better on my end too

PR #4572, opened 2023-04-24
Make it possible to pass rpyc version value as an argument

Andykl:
Main use-case for that is to run `renpy --compile --rpyc-version int(HEAD_COMMIT_SHA, 16)` in Git environment (by git pull/checkout hook or other means), so we have the same rpycs content between members without involving of old-game (I used that approach and it wasn't very convenient).

I don't think it makes old-game obsolete, because for non-techie it is easier to copy and paste folders with rpycs then playing with console.

### DIFF
--- a/renpy/arguments.py
+++ b/renpy/arguments.py
@@ -33,6 +33,7 @@ from renpy.compat import PY2, basestring, bchr, bord, chr, open, pystr, range, r
 import argparse
 import os
 import sys
+import time
 
 import renpy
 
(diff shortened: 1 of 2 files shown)


Andykl:
Note for myself. This could cause `Name (..., ..., 0) is defined twice` after reload because the value is stored globally.

renpytom:
I'm not convinced that this is correct (that is, that it will wind up with the same .rpyc files for different members of the team).

PR #4576, opened 2023-04-25
Allow None transition for ShowMenu

shawna-p:
The current implementation of ShowMenu means that if you have `config.intra_transition` set to anything other than None, and you write an action like `ShowMenu("preferences", _transition=None)` in order to make one of your menu screens not use a transition, it will simply use the intra_transition anyway and ignore the provided one. This fixes that. It also uses the provided transition in the case that the menu is a label rather than a screen.


Gouvernathor:
There will be a small difference of behavior between unpickled actions and new ones, but I think that's ok in this case.
However there may be an issue if the action is called during init time before the config is set, do we have another instance of that in an action constructor ?

shawna-p:
I've updated the code to check for a sentinel instead of `None`. Let me know if that seems like a suitable solution to prevent using the config value before it's been initialized.

PR #4583, opened 2023-04-27
Make renpy.confirm instead of yesno_screen

Gouvernathor:
The screen was renamed some time ago from "yesno_prompt" to "confirm", and `layout.yesno_screen` is kind of a terrible name in every way.
To be better as a function this should probably take _values_ as yes and no parameters which would get returned by the function, instead of actions to be executed. But leaving the global `_return` untouched would be good in that case, and I'm not sure how to do that efficiently.


Gouvernathor:
I was expecting this function to be comparable to renpy.display_menu in that it would stop script execution, but somehow it doesn't ? This should probably be patched in the new function.

PR #4604, opened 2023-05-05
Update revertable types according to Python's datamodel

Gouvernathor:
Follow-up to #4598 using PEP 584's recommendations and [datamodel](https://docs.python.org/3/reference/datamodel.html#emulating-numeric-types) support.

### DIFF
--- a/renpy/revertable.py
+++ b/renpy/revertable.py
@@ -203,6 +203,8 @@ class RevertableList(list):
         @_method_wrapper(method)
         def newmethod(*args, **kwargs):
             l = method(*args, **kwargs) # type: ignore
+            if l == NotImplemented:
+                return l
             return RevertableList(l)
 
         return newmethod
@@ -222,10 +224,9 @@ class RevertableList(list):
             return rv
 
     def __mul__(self, other):
-        if not isinstance(other, int):
-            raise TypeError("can't multiply sequence by non-int of type '{}'.".format(type(other).__name__))
-
-        return RevertableList(list.__mul__(self, other))
+        if isinstance(other, int):
+            return RevertableList(list.__mul__(self, other))
+        return NotImplemented
 
     __rmul__ = __mul__ # type: ignore
 
@@ -335,11 +336,15 @@ class RevertableDict(dict):
 
         # https://peps.python.org/pep-0584 methods
         def __or__(self, other):
+            if not isinstance(other, dict):
+                return NotImplemented
             rv = RevertableDict(self)
             rv.update(other)
             return rv
 
         def __ror__(self, other):
+            if not isinstance(other, dict):
+                return NotImplemented
             rv = RevertableDict(other)
             rv.update(self)
             return rv
--- a/renpy/revertable.py
+++ b/renpy/revertable.py
@@ -203,9 +203,7 @@ class RevertableList(list):
         @_method_wrapper(method)
         def newmethod(*args, **kwargs):
             l = method(*args, **kwargs) # type: ignore
-            if type(l) is list:
-                return RevertableList(l)
-            return l
+            return RevertableList(l)
 
         return newmethod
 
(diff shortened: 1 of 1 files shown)


renpy/revertable.py
```diff
@@ -335,11 +336,15 @@ def has_key(self, key):
...
         # https://peps.python.org/pep-0584 methods
         def __or__(self, other):
+            if not isinstance(other, dict):
+                return NotImplemented
             rv = RevertableDict(self)
             rv.update(other)
             return rv
 
         def __ror__(self, other):
+            if not isinstance(other, dict):
```
mal, line 346: Specifically for `__ror__`, I think that this should be either a direct type or isinstance check of RevertableDict rather than dict.

The impact of such a change would be near non-existant `rpy` files as the RHS `dict` or `{}` would be implicitly RevertableDicts already and so `__or__` would be used, however in `py` files, the distinction between `dict() | RevertableDict() -> dict` and `RevertableDict() | dict() -> RevertableDict` would be useful.

Gouvernathor: You would have `if type(other) is not dict: return Notimplemented` ? I understand the context but I'm unsure about what you'd want to see.

mal: I don't recall if we support (or want to) subclassing of Revertable types, if we do then I'd suggest:
```suggestion
            if not isinstance(other, RevertableDict):
```
but if we're not expecting (or wanting) subtypes then as you noted in the list change, we could direct check:
```suggestion
            if type(other) is not RevertableDict:
```

renpytom:
Is this causing any actual problems? This seems like a risky enough change it should wait until 8.2.

mal:
Without this PR the real-world impact would be that `RevertableDict() | (('a', 'b'),)` (and it's inverse) would be legal, in contrast to `dict() | (('a', 'b'),)` which raises a `TypeError`. That's not ideal, but also not critical IMO because, at least in theory, no sane person should be doing that.

The `type(dict() | RevertableDict()) is RevertableDict` (`__ror__`) issue from my comments is more of an issue, as it means that in pure python code (`py` files) merging a `RevertableDict` into a regular `dict` results in an undesirable type change.

Andykl:
> Without this PR the real-world impact would be that `RevertableDict() | (('a', 'b'),)` (and it's inverse) would be legal, in contrast to `dict() | (('a', 'b'),)` which raises a `TypeError`. That's not ideal, but also not critical IMO because, at least in theory, no sane person should be doing that.
> 
> The `type(dict() | RevertableDict()) is RevertableDict` (`__ror__`) issue from my comments is more of an issue, as it means that in pure python code (`py` files) merging a `RevertableDict` into a regular `dict` results in an undesirable type change.

But it _is_ how Python datamodel works. Interpreter calls `right.__ror__(left)` if `isinstance(right, type(left))` so you can overload it.
IMO, there should be no `python_dict` in rpy and no `RevertableDict` in py, except it is written explicitly.

mal:
In the current nightly:
```py
When x is RevertableDict:
    x                 :  <class 'renpy.revertable.RevertableDict'>
    x | {'c': 'd'}    :  <class 'renpy.revertable.RevertableDict'>
    {'c': 'd'} | x    :  <class 'renpy.revertable.RevertableDict'>
    x | (('c', 'd'),) :  <class 'renpy.revertable.RevertableDict'>  # ought to be TypeError
    (('c', 'd'),) | x :  <class 'renpy.revertable.RevertableDict'>  # ought to be TypeError

When x is dict:
    x                 :  <class 'dict'>
    x | {'c': 'd'}    :  <class 'renpy.revertable.RevertableDict'>  # would ideally be dict
    {'c': 'd'} | x    :  <class 'renpy.revertable.RevertableDict'>
    x | (('c', 'd'),) :  <class 'TypeError'>
    (('c', 'd'),) | x :  <class 'TypeError'>
```

Gouvernathor:
I'm not convinced by Mal's point, in that the benefits of revertability could be seen as worth propagating to other dicts in a context where pure native dicts would be handled. And as Andy points out, it's usually how subclasses behave.
Mal's API resembles set/frozenset, which are not subclasses of one another, but I guess if an advantage of not propagating revertableness (keeping native dicts) can be found it's also an acceptable behavior.
I propose we keep and merge the rest of the changes (= the code as it is), and settle that in another PR.
By the way, if we want to do things that way, maybe we could just not define a ror method ? Wouldn't the behavior be the same ?

For the rest of the changes, yes, Tom, there are consequences for people using subclasses of dict, as well as the weird things Mal showed as being possible.

mal:
After further discussion I've retreated from the opinion that `dict | RevertableDict` should result in a `dict`. 🙁

PR #4617, opened 2023-05-10
Translation of Japanese

nattyan-tv:
I checked the machine-translated Japanese and corrected any words that needed correction or could be better worded.
I have also translated some words that were not translated or were translated incorrectly.

Translated target: `launcher/game/tl/japanese`

If you have any questions or problems, please let me know.


renpytom:
Merged, thank you.

PR #4618, opened 2023-05-10
fix: `web cache preload` as `ToggleField` else set in one choice

qTich:


renpytom:
This needs to be this way, to  ensure clearCache runs. Am I missing something else?

PR #4630, opened 2023-05-13
Testcase

Gouvernathor:


arsym-dev:
I like the changes you've made so far with the `not`/`or`/`and` operators and `eval`. These were very much needed.

I'm interested in continuing the work here. I've been running functional tests on my own projects and have the following suggestions:

## (1) Keysym node
Simulate a keysym event. This includes the keys of `config.keymap`. Right now you can do `type F1`, but keysyms would allow the use of modifiers, custom events, and alternate keybindings.

Examples:
- `keysym "skip"`
- `keysym "help"`
- `keysym "ctrl_K_a"`

## (2) Timeout argument for Until clause
Currently it's possible for an Until clause to never terminate because the left-hand clause keeps creating a new state (eg. `type`, `pause`, or `run`).

The timeout would force a maximum time for execution (by default, the value given by `_test.timeout`), regardless of what's happening on the left clause. Setting it to None would disable timeout and revert it to the current behavior.

Examples:
- `pause 0.5 until "Text" timeout 5.0`
- `keysym "skip" until "Text" timeout 10.0`
- `type "t" until "Overflow" timeout None`

## (3) Displayable clause
This would fill a complementary role to Pattern (or it may extend Pattern). Rather than checking for alt text on the screen, it would look for displayables.

It requires `screen` and/or `id` to be specified. If `screen` is None, it checks the `id` in the current screen. If `id` is None, it only checks if the screen is visible.

Keywords:
- **screen** (str | None)
- **id** (str | None)
- **layer** (str | None)

Nodes that accept Pattern would also accept Displayable.

Examples:
- `click until screen "choice"`
- `click id "button2"` (equivalent to `click screen None id "button2"`)
- `pause 0.5 until screen "choice"`
- `click until screen "grid_menu" id "viewport3" layer "overlay"`

## (4) Pattern / Displayable flags
I'm iffy on these right now, but I think they're important to have eventually.
- **sensitive**
- **insensitive**

Examples:
- `pause 0.1 until id "bar1" insensitive`
- `"Choice 2" sensitive`

## Future
I'm aware of other things that need attention long-term, but I'd decided against making suggestions for them right now. But these things would include:
- Run tests without crashing the entire run if one fails
- Generate statistics at the end (eg. `123 tests | 120 passed | 2 failed | 1 skipped`)
- Create testsuites which would run testcases together
- Set [environment variables](https://www.renpy.org/doc/html/environment_variables.html) for a testcase or testsuite.
- As per #5808, visual changes are the hardest thing to avoid changes in, and also the hardest to test. We can probably use a screenshot comparison tool similar to what [BrowserStack](https://www.browserstack.com/guide/screenshot-testing) and [Godot](https://dev.epicgames.com/documentation/en-us/unreal-engine/screenshot-comparison-tool-in-unreal-engine) do. It may need to take multiple screenshots to test animations (along with a way to force synchronization or tick one frame at a time).
- Simulate controller and touch inputs

But all of this is for another day. 

I'd like to know your thoughts about the above @Gouvernathor

renpytom:
I'd be very interested in merging these changes (and the rest of this branch, once it's ready to leave draft) into Ren'Py.

Gouvernathor:
I like the keysym and timeout, not sure about the proposed syntax for displayable : there should probably be a syntax to look for screens, elements of screens, and images, and also to test whether a given image is above another. Not sure how to express all that in simple syntax, it may become useful to implement parts of it in python only.

I don't plan on working on this much more : it's been roughly 4 years since I wrote most of it, and testing has never been my speciality. I can help explain my code though.

renpytom:
This became part of #6644, and has been merged into Ren'Py.

PR #4691, opened 2023-06-04
Add a warning when creating new projects in renpy 7

Gouvernathor:
Not disabled, because it still makes sense for test purposes or to update default files of an older game (that last part can be done in renpy 8 but it's not obvious that it works and having to update to 8, generate the files and downgrade after is not obvious either).
This can be reworded (I'll add translation when it's fixed), and has been made fix-compatible.
Obviously it's meant for the transition period before the end of renpy7.


renpytom:
I'm thinking we can add this for 7.6.2.

PR #4700, opened 2023-06-06
Fix mouse_move documentation

Gouvernathor:


renpy/common/00action_other.rpy
```diff
@@ -476,7 +476,8 @@ init -1500 python:
         :doc: other_action
 
         Move the mouse pointer to `x`, `y`. If the device does not have a mouse
-        pointer or _preferences.mouse_move is False, this does nothing.
+        pointer or if the :var:`"automatic move" preference <preferences.mouse_move>`
```
Gouvernathor, line 479: Unsure about the quotes-inside-backquotes thing, it should work fine (quotes are not recognized by sphinx/rst) but I didn't test it.

PR #4716, opened 2023-06-09
Repeal the annotations future and expand the `rpy python` statement

Gouvernathor:
Disables the "annotations" `__future__` being activated by default, and enables creators to opt-in file by file just like in Python.

Reverts 09e56040abc0177ab761d4ddf9cf1d88c6595f2e.


mal:
The cure here seems to be far more complicated than the disease. Perhaps a project-wide early set config variable would be better?

Also this all assumes that leaving it does active harm set by default. Is there any issue with leaving it as is until we move to a version of Python where this is an issue, and then removing it completely in that release?

Gouvernathor:
The longer we leave it in, the more projects and user will rely on it and the harder it will hurt when we remove the rug from under them - which will happen sooner or later as Python moves on to a more featureful syntax.
A project-wide config is acceptable for me, but only if it's an opt-_in_ to the annotations future. If it's a project-wide `rpy python annotations`, if you will, and not an opt-in of the normal, non-stringified annotations syntax. Because stringified annotations should always remain an opt-in, as it will always be in Python.

Booplicate:
> which will happen sooner or later as Python moves on to a more featureful syntax

Last I saw nothing was set in stone and the import isn't going away due to how many people are using it. RenPy, like a lot of libraries, benefits from static typing rather than dynamic.

Has something changed? Are they officially deprecating the import? Is it backwards incompatible?

Gouvernathor:
[PEP 649](https://peps.python.org/pep-0649/) supersedes [PEP 563](https://peps.python.org/pep-0563/) (= stringized annotations), and it was confirmed that the 563 syntax, the annotations \_\_future\_\_, will never be the default behavior in any version of Python.
There has been [a poll among python devs](https://discuss.python.org/t/survey-should-pep-649-implementation-use-another-future-import/23338) about whether to make another \_\_future\_\_ for 649 and although it will likely wait for 3.13 and nothing is set in stone about its implementation process, it's very likely there won't be. In that thread and others, there are discussions about whether and when to deprecate the `annotations` future itself, and it seems it will eventually go that way.
It may be in a long time, but I for one don't want to live another Python 2 situation where we have to maintain compatibility for an obsolete Python syntax across two versions of renpy for more than a year because people are relying on it.

649 allows for static typing while enabling local and private typing and so on, things which are not feasible with the 563 syntax.
Vanilla annotations and 649 are compatible seemlessly with the exception of very extreme edge cases (see the PEP for more details). 563 is not compatible with either as it changes the type of all saved annotations (from type objects to strings).

Booplicate:
Thank you! When I was reading about it, 649 couldn't solve the problems caused by 563, which caused a controversy, what's the point of the new PEP if it doesn't solve the core problems. In fact, they couldn't decide what would be the way to go: 649, 563, or something entirely different, so they delayed it for multiple python versions.

In the case 563 gets deprecated, is there a reason to keep it in RenPy even as opt-in?

Gouvernathor:
This PR makes `rpy python` basically synonymous with `from __future__ import`. As long as Python allows it, no reason to forbid it in renpy, I guess. Not to mention, the process of deprecating things in python is very long.
Also, 649 will not be added into Python before 3.13 at the earliest if I understood correctly, and even then there will be the period of time before Ren'Py catches up with it (since creators can't easily upgrade their python). So we need to offer an opt-in for people using static typing, or whatever the advantage of stringized annotations is, and support for people already using it in their projects.

PR #4719, opened 2023-06-10
Skip patching py2 Popen on web

### LINKED ISSUE #4718: Ren'Py Web not loading
After "Build and Open in Browser".
web

```
   File "main.py", line 244, in <module>
     main()
   File "main.py", line 231, in main
     import renpy.bootstrap
   File "./renpy/__init__.py", line 58, in <module>
   File "./renpy/compat/__init__.py", line 186, in <module>
     class Popen(subprocess.Popen):
 AttributeError: 'module' object has no attribute 'Popen'
```

Ren'Py 7.6.1.23060707

mal:
Speculatively fixes #4718.

This is untested.

### DIFF
--- a/renpy/compat/__init__.py
+++ b/renpy/compat/__init__.py
@@ -181,7 +181,7 @@ if PY2:
 
 ################################################################################
 # Chance the default for subprocess.Popen.
-if PY2:
+if PY2 and not renpy.emscripten:
     import subprocess
     class Popen(subprocess.Popen):
         def __init__(self, *args, **kwargs):


Kassy2048:
That's what I've tried initially as well, but the launcher won't start at all with such a change, most likely because `renpy.emscripten` does not exist yet. What does work though is to check if the `Popen` attribute exists in `subprocess`.

But there are at least 2 others issues that prevent the Web version of 7.6 to start afterwards anyway: `pydoc` being imported in [`00console.rpy`](https://github.com/renpy/renpy/blob/6e7d650f515e80731be99ec4ed1649807630746c/renpy/common/00console.rpy#L131) (but not part of the Web package) and [`endInput`](https://github.com/renpy/renpy/blob/6e7d650f515e80731be99ec4ed1649807630746c/renpy/display/behavior.py#L2864) not exisiting on 7.6. I don't really have much time to look into them yet.

PR #4722, opened 2023-06-10
Fix for 7.6 web

### LINKED ISSUE #4718: Ren'Py Web not loading

Kassy2048:
Fix #4718.

On a side note, there are "bad mtime" warnings from Python when importing most of the PYO files on 7.6 Web, which forces Python to recompile the files and increases the load time. I'm not sure if this is specific to 7.6 and/or Web though.

### DIFF
--- a/renpy/common/00console.rpy
+++ b/renpy/common/00console.rpy
@@ -128,7 +128,11 @@ init -1500 python in _console:
     import sys
     import traceback
     import store
-    import pydoc
+    try:
+        import pydoc
+    except ImportError:
+        # Web2 does not have pydoc, help() will fail but game will load at least
+        pass
 
     from reprlib import Repr
     class PrettyRepr(Repr):
--- a/renpy/compat/__init__.py
+++ b/renpy/compat/__init__.py
@@ -183,13 +183,14 @@ if PY2:
 # Chance the default for subprocess.Popen.
 if PY2:
     import subprocess
-    class Popen(subprocess.Popen):
-        def __init__(self, *args, **kwargs):
-            if ("stdout" not in kwargs) and ("stderr" not in kwargs) and ("stdin" not in kwargs):
-                kwargs.setdefault("close_fds", True)
-            super(Popen, self).__init__(*args, **kwargs)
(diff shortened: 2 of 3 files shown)


Kassy2048:
> On a side note, there are "bad mtime" warnings from Python when importing most of the PYO files on 7.6 Web, which forces Python to recompile the files and increases the load time. I'm not sure if this is specific to 7.6 and/or Web though.

That's because the file modification timestamp is [not preserved](https://stackoverflow.com/q/9813243) while extracting from "game.zip", so the PY file mtime does not match the timestamp in the associated PYO/PYC file anymore. It probably affects both Web versions.
That's a different problem than this PR solves though, so I'll try to "fix" that in another PR later.

renpytom:
Good catches. 

I'm not sure it's worth it to fix the mtime issue on 7.6. 8.1 uses a different codepath for extracting game.zip, and I'm thinking the web builds for Ren'Py 7 should only have the most critical fixes.

Kassy2048:
A few notes about the mtime issue:
* I've done some experiments and recompiling the Python scripts does not take that much time anyway (~10% of load time for the tutorial).
* the mtime have a 2 seconds resolution in ZIP files by design, so we can't restore it accurately, and the comparison can fail 50% of the time. The best solution is to use a different archive file format like TGZ instead (supported by Python).
* there is always recompilation on 8.x for Web for now because the launcher compiles the files for Python 3.9 while Web3 uses Python 3.11.
* RPYC files are not recompiled as they depend on a hash of the source file instead of mtime. I'm surprised it works though, as they are compiled/pickled with a different Python version than what the Web uses.

I agree it's not worth it to fix on 7.6. For 8.x, this will have to wait for the launcher to use the same Python version as Web.

PR #4724, opened 2023-06-11
`emscripten.run_script("location.reload();")` for relaunch tab on quit

qTich:


mal:
Related: A similar feature for web was rejected in #4164.

Issue #3666, opened 2022-06-15
Hbox/Vbox should distribute spacing, rather than box size. 


renpytom:
https://lemmasoft.renai.us/forums/viewtopic.php?f=8&t=64805

Note that this should be an alternative to xfill/yfill, perhaps reusing justify.

PR #4746, opened 2023-06-16
Remove trailing space in hbox/xfill and vbox/yfill

mal:
Seeks to address some pain points closely related to #3666 — the use of `hbox` with `xfill` and `vbox` with `yfill`. This is intended as something like a stop-gap measure until such time that the spacing distribution issue can be looked at. It does two small things to hopefully make this a bit better:

1. It zeros the last spacer, shrinking any trailing space to consist only of the indivisible left overs.
2. Allows `first_spacing` to work with `box_reverse`. It'll be used between the penultimate and final items, rather ending up as a trailing space and making the xfill/yfill cases worse.

### DIFF
--- a/renpy/display/layout.py
+++ b/renpy/display/layout.py
@@ -947,6 +947,10 @@ class MultiBox(Container):
 
         spacings = [ first_spacing ] + [ spacing ] * (len(self.children) - 1)
 
+        # Suppress trailing space, as in practicality we want one fewer than
+        # the number of children.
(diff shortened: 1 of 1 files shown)


renpytom:
I don't think that this is something I'd want to merge. It introduces a new behavior, that isn't justified boxes and isn't the current behavior. I don't think that having some sort of new and idiosyncratic behavior is something that's worth adding.

PR #4749, opened 2023-06-17
Add last_say function.

### LINKED ISSUE #4733: Documenting _last_say variables.
There are four variables in question:
- **_last_say_who**
- **_last_say_what**
- **_last_say_args**
- **_last_say_kwargs**

All are undocumented because of their internal use (I'm guessing), but are being used fairly commonly by programmers. The only harm I see in documenting them is that people could decide to set them directly and cause a mess, but hey, they can already do that with plenty of variables.

If you agree, I'll write (Or have @Gouvernathor do so :P) them onto the Store Variables page.

Lezalith:
Resolves #4733

### DIFF
--- a/renpy/exports.py
+++ b/renpy/exports.py
@@ -4270,6 +4270,45 @@ def get_say_image_tag():
     return renpy.store._side_image_attributes[0]
 
 
+class LastSay(tuple):
+    """
+    Object containing info about the last say statement.
+    """
+
+# Returns a LastSay object.
+def last_say():
+    """
+    :doc: other
+
+    Returns a `LastSay` object containing information about the last say
+    statement. This object has the following fields on it:
+
+    ``"who"``
+        A `Character` object of the speaker.
+
+    ``"what"``
+        A string with the dialogue spoken.
+
+    ``"args"``
+        A tuple of arguments passed to the last say statement.
+
+    ``"kwargs"``
+        A dictionary of keyword arguments passed to the last say statement.
+    """
+
+    who = renpy.store._last_say_who
+    what = renpy.store._last_say_what
+    args = renpy.store._last_say_args
+    kwargs = renpy.store._last_say_kwargs
+
+    ls = LastSay((who, what, args, kwargs))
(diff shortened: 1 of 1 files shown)


renpy/exports.py
```diff
@@ -4270,6 +4270,47 @@ def get_say_image_tag():
...
+        A :func:`Character` object of the speaker.
+
+    .. attribute:: what
+        A string with the dialogue spoken. This may be None if dialogue
+        hasn't been shown yet, for example at the start of the game.
+
+    .. attribute:: args
+        A tuple of arguments passed to the last say statement.
+
+    .. attribute:: kwargs
```
Gouvernathor, line 4297: Those cause display issues in the generated docs, they should be replaced by `` `kwargs` ``, `` `args` `` and so on.

mal:
We may wish to consider making this a singleton with properties that return the various values. That would allow (for example) `renpy.last_say.who` to provide `eval_who(renpy.store._last_say_who)` without the need to create an additional object or tempt the user to try and assign that object into a store at which point this becomes a potential save maintenance issue.

<details><summary>tl;dr: Maybe something like this?</summary>

```py
class LastSay(python_object):
    """
    :undocumented:
    Singleton class to act as a public API proxy to internal last_say variables.
    """

    __slots__ = ()

    @property
    def who(self):
        return eval_who(renpy.store._last_say_who)

    @property
    def what(self):
        return renpy.store._last_say_what

    @property
    def args(self):
        return renpy.store._last_say_args

    @property
    def kwargs(self):
        return renpy.store._last_say_kwargs


last_say = LastSay()
```

Also considered having `kwargs` return a copy, but on balance it didn't seem worth it given that who could already be a mutable object.
</details>

renpytom:
We want to create a new object each time. That's because it seems useful to store the object for processing - and it's not like creating an object once per line of dialogue is a big deal.

renpy/exports.py
Gouvernathor: fixed in 9189178

PR #4752, opened 2023-06-18
fix: periodic animate

qTich:
When creating `adjustment` via `default | define adj = ui.adjustment()`, if you wait some time (so that the animation start time is long) and then run the scroll animation and close the screen, then after opening the screen again - since `st` will start its countdown from zero, it causes an exception - `OverflowError: math range error` at the time of calculating `def inertia_warper`

### DIFF
--- a/renpy/display/behavior.py
+++ b/renpy/display/behavior.py
@@ -2037,6 +2037,9 @@ class Adjustment(renpy.object.Object):
         if self.animation_start is None:
             self.animation_start = st
 
+        if st < self.animation_start:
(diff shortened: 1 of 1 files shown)


qTich:
Maybe you shouldn't add a `self.animation_delay`, but with it, I suppose, the animation will end faster.

renpytom:
I think the change is important, but eb0d4948d  is my approach to it - I like the idea that the animation is ended directly, rather than indirectly through time adjustment. 

Thanks for bringing this up.

PR #4753, opened 2023-06-18
Tweak 1a62894

Gouvernathor:
Linkify, fix a dubious link, keep the best phrasing in the two locations, and simplify the code a bit.


Gouvernathor:
I should have made that target fix instead of master... too bad.

PR #4764, opened 2023-06-22
Timed nw tag

Gouvernathor:
The doc change looks very messy but it adds just one example and the same line of explanation as for the p and w tags. I also reordered the tags in incremental order, from simpler to most complex. They didn't seem to be in any particular order to begin with.


Gouvernathor:
Done in fb03b82

PR #4775, opened 2023-06-23
Global transform for the whole display

### LINKED ISSUE #4773: Add the ability to apply transforms to all the display
Wrapping all the layers as a group, rather than each individual layer, because it may not yield the same behavior and improve performances.
There would be some uses to make it context-dependent (like the camera statement) and some others to making it global and impervious to that. There will be losers to that match, so 🤷 maybe the simplest to implement ?

Gouvernathor:
The implem is clanky, that's certainly not the best way of applynig it. It glitches when going from main gameplay to game menu and back.

### DIFF
--- a/renpy/display/core.py
+++ b/renpy/display/core.py
@@ -935,6 +935,9 @@ class SceneLists(renpy.object.Object):
             if i not in self.camera_list:
                 self.camera_list[i] = (None, [ ])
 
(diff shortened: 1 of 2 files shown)


mal:
I think an alternative approach to this landed in `8.2.0` as `config.layer_transforms[None]`, so this can be closed?

renpytom:
That seems right.

PR #4788, opened 2023-06-28
Extra options for mesh manipulation

valery-iwanofu:
This change can prevent copying source geometry/attributes/triangles when passed to mesh functions.

### DIFF
--- a/launcher/game/tl/french/launcher.rpy
+++ b/launcher/game/tl/french/launcher.rpy
@@ -1590,7 +1590,7 @@
 
     # game/gui7.rpy:339
     old "Custom. The GUI is optimized for a 16:9 aspect ratio."
-    new "Personnalisé. Le GUI est optimié pour un ratio de cadre 16:9."
+    new "Personnalisé. Le GUI est optimisé pour un ratio de cadre 16:9."
 
     # game/gui7.rpy:355
     old "WIDTH"


renpytom:
Do you have something motivating this? I'm not sure why we'd want this, and I'd like to make sure this covers more use cases.

valery-iwanofu:
Example from my SpriteBatch, which not done yet, but it already uses the changes from PR:

<details>
  <summary>SpriteBatch</summary>

```python
class SpriteBatch:
    geometry : array.array
    attribute_data: array.array
    index: int
    mesh: Mesh2

    def __init__(self, capacity=1024):
        points = capacity * 2 * 4 # 4 Point2 per rect
        triangles = capacity * 2 # 2 triangles per rect
        geometry = create_geometry_array(points) # capacity length array filled with zeros
        attribute_data = create_attribute_data(points) # array per each vertex for uv and color attributes
        triangle_data = create_triangles_data(triangles) # array where every 2 triangles form a rect
        
        mesh = Mesh2(UV_COLOR_LAYOUT, points, triangles)
        mesh.set_triangle_data(triangle_data) # triangle data will not changed ever

        self.geometry = geometry
        self.attribute_data = attribute_data
        self.index = 0
        self.mesh = mesh

    def begin(self):
        self.index = 0

    def add(self, x, y, width, height, color):
        index = self.index
        next_index = self.index + 1

        x = float(x)
        y = float(y)
        width = float(width)
        height = float(height)
        color = array.array('f', Color(color).rgba)

        self.geometry[index * 2 * 2 * 4: next_index * 2 * 2 * 4] = array.array('f', (
            x, y, # top left
            x + width, y, # top right
            x + width, y + height, # bottom right
            x, y + height # bottom left
        ))
        attr_offset = index * UV_COLOR_LAYOUT.stride * 4 + 2
        attribute_data = self.attribute_data
        for i in range(4):
            j = attr_offset + UV_COLOR_LAYOUT.stride * i
            attribute_data[j: j + 4] = color

        self.index += 1
    
    def end(self):
        mesh = self.mesh
        mesh.set_triangles_size(self.index * 2) # no need to copy triangle_data
        mesh.set_geometry_data(self.geometry, 0, self.index * 4) # copy only part of geometry
        mesh.set_attribute_data(self.attribute_data, 0, self.index * 4) # copy only part of attributes
```
</details>

The main part of it is the `end` method, which wins some FPS:
```python
    def end(self):
        mesh = self.mesh
        mesh.set_triangles_size(self.index * 2) # no need to copy triangle_data
        mesh.set_geometry_data(self.geometry, 0, self.index * 4) # copy only part of geometry
        mesh.set_attribute_data(self.attribute_data, 0, self.index * 4) # copy only part of attributes
```

But after submitting this PR I realized that I have no idea how such a SpriteBatch is supposed to work with multiple textures, without direct access to gl. So at this point I think I sent this PR for nothing.

PR #4794, opened 2023-06-30
Add the screen property directive

### LINKED ISSUE #4774: default_focus and other UI Statement properties are difficult to search
The properties under the sub-header "User Interface Statements" (https://www.renpy.org/dev-doc/html/screens.html#user-interface-statements) are not easily searchable in the docs as-is, and are extremely easy to miss even if you know what you're looking for. I knew there was a focus property for displayables (`default_focus` in fact), but I could not find it by searching "focus" because it is not a linkable subheader.

I also suggest this property be linked in other places where a user might look for it, for example in https://www.renpy.org/dev-doc/html/screen_python.html#renpy.set_focus

Gouvernathor:
image

(the new one is on the left)

The internal issue with this is the massive amount of duplicate screen (element/displayable) properties : "action", "unhovered", "keysym", "modal" and others which are shared between several displayables.
For comparison, style properties only have two pairs of duplicates : "mouse" and "spacing".

To mitigate that we could only linkify the first 8 properties, those applying directly to the screens and not the screen displayables, from "modal" to "roll_forward". Possibly add to those the 12 properties shared by all screen elements, from "at" to "properties".

The reason I didn't just reuse the style property directive is that they would end up in the style property index, and that some may imply they can be put in style definition blocks - and I think that doesn't work.

Fixes #4774

### DIFF
--- a/sphinx/source/renpydoc.py
+++ b/sphinx/source/renpydoc.py
@@ -56,7 +56,6 @@ def parse_var_node(env, sig, signode):
 
 style_seen_ids = set()
 
-
 def parse_style_node(env, sig, signode):
     m = re.match(r'(\S+)(.*)', sig)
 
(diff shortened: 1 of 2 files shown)


renpytom:
What are you trying to accomplish here? It's not totally clear to me why we'd want this change.

PR #4797, opened 2023-06-30
Enable jump/call expression to local labels

Gouvernathor:
Still doesn't extend to Call and Jump, which would be much more complex to do.

### DIFF
--- a/renpy/ast.py
+++ b/renpy/ast.py
@@ -1697,19 +1697,22 @@ class Call(Node):
         'label',
         'arguments',
         'expression',
+        'global_label',
         ]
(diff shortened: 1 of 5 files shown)


renpytom:
I don't like the way this reuses expression to store the global label. That should be another field in the AST classes.

Gouvernathor:
Done. The weird and/or expression is to avoid having None as a global label and then fail the string sum, or storing numerous useless strings for non-expression nodes.

renpytom:
Merged. Please add this to the changelog when you get a chance.

PR #4802, opened 2023-07-01
Turn Fade into ATL supporting transparent colors

Gouvernathor:
Converts Fade into an ATL transition.
Adds the new behavior (described in the docstring) when using a transparent color. Previously the transparency only showed, underneath it, either the grey checkboard in dev mode or plain black in distributed mode.

Compat is maintained in three ways : first the widget feature is maintained (overriding the color but only if no color is passed), then the alpha parameter which was passed to Dissolve whose docstring says it's ignored, now the parameter is truly ignored but kept for compat, and finally the transform object is re-injected in the original module for pickle compatibility.


renpytom:
I don't think this would maintain compatibility with Fade, as a Fade object pickled as renpy.display.transition.Fade can't be loaded after this change. (I'm also not sure what the rationale for changing Fade is.)

PR #4806, opened 2023-07-02
Add Zoomin, Zoomout and Zoominout transition factories

Gouvernathor:
It's not ZoomInOut because the name was already taken, so I used "Camelcase" for all three. I don't think it matters very much.

### DIFF
--- a/renpy/common/00definitions.rpy
+++ b/renpy/common/00definitions.rpy
@@ -363,10 +363,47 @@ init -1400:
     define pushup = PushMove(1.0, "pushup")
     define pushdown = PushMove(1.0, "pushdown")
 
-    # Zoom-based transitions. Legacy - nowadays, these are probably best done with ATL.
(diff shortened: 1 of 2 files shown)


mal:
Given the very limited demand, perhaps this would be better left for consideration post-ATL2 rather than adding more easy to confuse/typo members to the namespace that might have to stick around for decades to come?

Gouvernathor:
I doubt something specific like this (applying a transform to all images in a scene, which only matters in a transition) has its place in ATL2, but until we can be sure of that I agree it's probably better to wait than to pollute the namespace.

renpytom:
I don't want to add this. MoveTransition was frankly a mistake - it works differently than all the other transitions, and isn't really a transition in the first place. We're stuck with it, but I don't want to add more variants of MoveTransition, and I especially don't want to add things based on OldMoveTransition.

PR #4809, opened 2023-07-03
Tweaks to renpy.last_say

Gouvernathor:
Add some safety in the case the object ends up in save data

### DIFF
--- a/renpy/exports.py
+++ b/renpy/exports.py
@@ -4281,13 +4281,25 @@ class LastSay():
     Returned by the last_say function.
     """
 
+    def __init__(self, who, what, args, kwargs):
+        self._who = who
+        self.what = what
+        self.args = args
+        self.kwargs = kwargs
+
+    @property
+    def who(self):
+        return eval_who(self._who)
+
 def last_say():
     """
     :doc: other
 
-    Returns an object containing information about the last say statement. This
-    While this can be called in a say statement, if the say statement is using
-    a normal Character, the information will be about the current say statement.
+    Returns an object containing information about the last say statement.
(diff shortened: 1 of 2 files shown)


renpy/exports.py
```diff
@@ -4281,13 +4287,25 @@ class LastSay():
     Returned by the last_say function.
     """
 
+    def __init__(self, who, what, args, kwargs):
+        self._who = who
+        self.what = what
+        self.args = args
+        self.kwargs = kwargs
```
mal, line 4294: I _think_ you can do this, and at which point `last_say` could simply be an alias to `LastSay`, and since unpickling doesn't go through `__init__` it shouldn't impact loading.
```suggestion
    def __init__(self):
        self._who = renpy.store._last_say_who
        self.what = renpy.store._last_say_what
        self.args = renpy.store._last_say_args
        self.kwargs = renpy.store._last_say_kwargs
```

It may not be worth the trouble though, so I'll leave it to you as an option. 🙂

Gouvernathor: I would do it if it were only me, but I'd rather keep the code vaguely similar to how Lez wrote it. And it could allow creating custom LastSay entries, in general that's a thing I like to be able to do when creating classes.

mal: Fair. In that case I do think it's worth calling `__init__` with positional args from `last_say()`. Since it's a private interface we don't need to worry about changing how we call it in future so they're not really providing any benefit, and arbitrarily passing them as keyword args seems a bit - for lack of a better descriptor - odd. 😛

Issue #4810, opened 2023-07-03
Additions to the layeredimage language


Gouvernathor:
- [ ] Add the `pass` statement, which makes no sense in a finished layeredimage but is very useful when writing it and testing it at the same time
- [ ] Allow a group's prefix to be passed True, in which case it's replaced with the group name. More than 95% of my prefixes are just the group name.
- [ ] Enable writing ATL inside layeredimage. Either using `always atl:` (I don't like `always at transform` or `always transform` since you're defining a displayable image, not a series of abstract transformations), or by also allowing it in attributes. I'll have to think about a good syntax (suggestions open). I'm tending towards an `atl:` which would "return" an image like `show a:` does, and which would be put inside an always, attribute or if block.

renpytom:
>Add the pass statement
Yes.

> Allow a group's prefix to be passed True
No. True isn't going to be much shorter than an actual prefix, it's not taking data from somewhere else, and it's not the default. So it seems like it would make things less readable, for not a large amount of benefit. 

What might make more sense would be some sort of auto-prefix mode, that would automatically supply prefixes to groups. That would be a big savings that could justify this.

>  Enable writing ATL inside layeredimage.

Maybe. I think using ``atl`` as the keyword is a non-starter. I'd suggest that we allow both ``at transform:`` and ``image:``. The former would be used to supply an anonymous transform that wraps the image associated with the attribute/always, and the latter can supply an ATL image that replaces the attribute/always. 

In both cases, it would need to be the last thing on the line, and if it was the last thing on the line, would suppress further parsing. 

```renpy:
always image:
      "eileen happy"
      0.5
      "lucy concerned"
      0.5
      repeat
```

and

```renpy
always:
     image:
           "eileen happy":
           rotate 0
           linear 2.0 rotate 360.0
           repeat
```

Should both be accepted syntaxes.

PR #4812, opened 2023-07-03
Add the pass keyword to the layeredimage language

Gouvernathor:
I made sure every pass is followed by an expected end of line and noblock. I'm not a master of parsing, so I roughly tested the different cases and it seems to work.
Made for #4810.

### DIFF
--- a/renpy/common/00layeredimage.rpy
+++ b/renpy/common/00layeredimage.rpy
@@ -1083,7 +1083,7 @@ python early in layeredimage:
                 ll.expect_eol()
                 ll.advance()
 
-            if ll.keyword('attribute'):
+            elif ll.keyword('attribute'):
 
                 parse_attribute(ll, rv)
                 ll.advance()


renpy/common/00layeredimage.rpy
```diff
@@ -954,12 +956,14 @@ python early in layeredimage:
             ll = l.subblock_lexer()
 
             while ll.advance():
-                if ll.keyword("attribute"):
-                    parse_attribute(ll, rv)
-                    continue
+                if not ll.keyword("pass"):
```
renpytom, line 959: This doesn't seem right. Something like: 

```python
if ll.keyword("pass"):
     ll.expect_eol()
     ll.expect_noblock("pass")
    continue
```

Seems much more correct. (And the same thing should be on line 997.)

renpytom:
Check out the comment I added on how to parse pass correctly.

renpytom:
Merged by mistake, so I'm undoing this for now.

PR #4813, opened 2023-07-03
Update the list of operators in the lexer

Gouvernathor:
[Source](https://docs.python.org/3/reference/lexical_analysis.html#operators)
Yes, these two operators can't be used in PY2, but the first operator in the list, <>, can't be used at all either and it didn't cause any issue, so it doesn't appear that a conditional is needed.

### DIFF
--- a/renpy/lexer.py
+++ b/renpy/lexer.py
@@ -609,6 +609,8 @@ OPERATORS = [
     '/',
     '%',
     '~',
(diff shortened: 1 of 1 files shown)


Booplicate:
Just in case, `<>` can be used at least in py2, it's a deprecated alternative to `!=`.

PR #4814, opened 2023-07-04
Add pass, at transform and image to the layeredimage language

Gouvernathor:
```rpy
layeredimage dummy:
    pass
    always image:
        "eileen happy"
        0.5
        "lucy concerned"
        0.5
        repeat
    always:
        image:
            "eileen happy"
            rotate 0
            linear 2.0 rotate 360.0
            repeat
```
I didn't showcase it just above but you can do the same with attribute, and with if/elif/else but only inside their blocks ; and you can do `at transform:` in all these places plus inside and inline groups, and inside layeredimages directly (meaning everywhere you could use the at clause).

I imagine a parser master could improve over my implementation, I fear it's getting a little spaghettified or at least intricate.

This PR contains #4812 for ease of use, but I'm keeping the other open in case this one's implem is not good enough.

### DIFF
--- a/renpy/common/00layeredimage.rpy
+++ b/renpy/common/00layeredimage.rpy
@@ -2,8 +2,9 @@
 
 python early in layeredimage:
 
-    from store import Transform, ConditionSwitch, Fixed, Null, config, Text, eval, At
     from collections import OrderedDict, defaultdict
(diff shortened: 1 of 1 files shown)


Gouvernathor:
This implementation is not good enough, I'm trying another way as part of a larger rewrite.

PR #4823, opened 2023-07-05
add: `path_to_logdir`

qTich:
This change will explicitly customize the path to the log directory.

### DIFF
--- a/renpy.py
+++ b/renpy.py
@@ -204,13 +204,27 @@ def path_to_saves(gamedir, save_directory=None): # type: (str, str|None) -> str
 # the launcher, usually.)
 def path_to_renpy_base():
     """
-    Returns the absolute path to thew Ren'Py base directory.
+    Returns the absolute path to the Ren'Py base directory.
     """
 
     renpy_base = os.path.dirname(os.path.realpath(sys.argv[0]))
     renpy_base = os.path.abspath(renpy_base)
 
     return renpy_base
+
+def path_to_logdir(basedir):
+    """
+    Returns the absolute path to the log directory.
+    `basedir`
+        The base directory (config.basedir)
+    """
+
+    import renpy # @UnresolvedImport
+
+    if renpy.android:
+        return os.environ['ANDROID_PUBLIC']
+
+    return basedir
 
 ##############################################################################
 
--- a/renpy/bootstrap.py
+++ b/renpy/bootstrap.py
@@ -264,10 +264,7 @@ You may be using a system install of python. Please run {0}.sh,
                 renpy.config.gamedir = gamedir
                 renpy.config.args = [ ] # type: ignore
 
-                if renpy.android:
-                    renpy.config.logdir = os.environ['ANDROID_PUBLIC']
-                else:
-                    renpy.config.logdir = basedir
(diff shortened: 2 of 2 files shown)


mal:
The end result is the same, it just costs an additional function call. 😕

qTich:
I did point out that it gives you the ability to customize the path.
For example for android this is useful because there is privacy and I can specify the directory to which the user has access. aka `Download/com.app/`

Gouvernathor:
It doesn't give you the ability to customize anything since the function has no side-effect, it merely enables you to know what the path to the logs is, provided you input the correct basedir (why, by the way ?).

qTich:
> It doesn't give you the ability to customize anything since the function has no side-effect, it merely enables you to know what the path to the logs is, provided you input the correct basedir (why, by the way ?).

I don't quite understand. The `log.txt` and `traceback.txt` write to `config.logdir`.
And `config.basedir`, because it is the root directory relative to the project.

Gouvernathor:
Yes. The existence of that function doesn't change the value of config.logdir.
If your goal is to change the value of config.logdir, this commit doesn't achieve that.
If your goal is to get, in-game, the path to the logs, you can either read config.logdir or `os.environ['ANDROID_PUBLIC'] if renpy.android else config.basedir`.
Either way, this function has no point.

qTich:
> Yes. The existence of that function doesn't change the value of config.logdir. If your goal is to change the value of config.logdir, this commit doesn't achieve that. If your goal is to get, in-game, the path to the logs, you can either read config.logdir or `os.environ['ANDROID_PUBLIC'] if renpy.android else config.basedir`. Either way, this function has no point.

You're confusing me because there are identical methods inside the same `renpy.py`.
The beauty of this method is that I can manually customize this path without worrying about it when I update the engine version.
I don't see why this is a pretext, because the initialization happens inside `bootstrap` and there is no other way to do this for us.

Gouvernathor:
> I can manually customize this path

Do you believe that after this change, putting in your game, for example,
```rpy
python early:
    renpy.renpy.path_to_logdir("some path I choose")
```
...it would change config.logdir ? Because it wouldn't.
In what way do you think this function helps you customize the path ?

qTich:
> > I can manually customize this path
> 
> Do you believe that after this change, putting in your game, for example,
> 
> ```python
> python early:
>     renpy.renpy.path_to_logdir("some path I choose")
> ```
> 
> ...it would change config.logdir ? Because it wouldn't. In what way do you think this function helps you customize the path ?

Is it illegal to customize `renpy.py`? I voiced that there is no other way but to customize directly, but this option would make things easier than changing `bootstrap` every time.

Gouvernathor:
You're free to change whatever file you want on your downloaded copy of renpy and distribute it, nothing of the sort is illegal.
Since it's possible anyway, I don't think there's a sufficient reason to add this function.

renpytom:
This makes sense to me. There are several functions in renpy.py that are meant to be edited - not by an end user, but by someone making a Linux distribution that includes a game, that has specific thoughts as to where different output goes. Customizing the logdir makes sense.

PR #4826, opened 2023-07-07
Unable to choose directory in preferences

rajchauhan28:


rajchauhan28:
This is from the renpy-8.2.0.23070601+nightly-sdk

raising error while changing default directory or making new first project
Screenshot_20230707_201802

Screenshot_20230707_200013

Gouvernathor:
The faulty commit is 4284d450692dffe7635be866fd686b9e8577fdb3, so I think the fix will involve a little more changes. Thanks for the report, though.

Gouvernathor:
Fixed in c02ccf8dcd15273bafe6eb6a2f9fa48bb0b8c749.

PR #4841, opened 2023-07-13
let a viewport/vpgrid respect the force_step argument of a used ui.adjustment

the66F95:
the ui.adjustment.step / 2160 calculation might also use a config var instead of the hardcoded value


renpytom:
Where does the 2160 come from?

the66F95:
it's just a constant i use to have the snap animation to the nearest "step" timed to my liking and still scale with the `step` argument of the `ui.adjustment()`.
if think the value of `config.viewport_inertia_time_constant` is too high to use here too .
maybe add a `config.viewport_snap_time_constant` or something like that.

PR #4843, opened 2023-07-13
Fix absolute.slots to __slots__

Gouvernathor:
My rewrite of the change propagated the error into the fix branch, when it was originally contained in master.
I tested a save-load cycle on an instance of absolute before and after this very fix, it passed without problem. But if someone added arbitrary attributes on an absolute instance, that will now fail.


renpytom:
Looks good, merged.

PR #4849, opened 2023-07-18
Add Inputvalue.disable_on_enter

Gouvernathor:
A nice feature that was in a cookbook and is so straightforward that it deserves to be builtin.

### DIFF
--- a/renpy/common/00inputvalues.rpy
+++ b/renpy/common/00inputvalues.rpy
@@ -86,6 +86,7 @@ init -1510 python:
         default = True
         editable = True
         returnable = False
+        disable_on_enter = False
 
         def get_text(self):
             raise NotImplementedError
@@ -96,8 +97,9 @@ init -1510 python:
         def enter(self):
             if self.returnable:
                 return self.get_text()
-            else:
-                return None
+            elif self.disable_on_enter:
+                renpy.run(self.Disable())
+                raise renpy.IgnoreEvent
 
         def Enable(self):
             if self.editable:
@@ -123,11 +125,12 @@ init -1510 python:
         common features of the documented input value classes.
         """
 
-        equality_fields = ("default", "returnable")
+        equality_fields = ("default", "returnable", "disable_on_enter")
 
-        def __init__(self, default=True, returnable=False):
+        def __init__(self, default=True, returnable=False, disable_on_enter=False):
             self.default = default
             self.returnable = returnable
+            self.disable_on_enter = disable_on_enter
 
     class ScreenVariableInputValue(__GenericInputValue):
         """
@@ -287,6 +290,9 @@ init -1510 python hide:
         `returnable`
             If true, the value of this input will be returned when the
             user presses enter.
+        `disable_on_enter`
+            If true, and if `returnable` is not true, pressing enter
+            will disable this input.
         """)
 
         for ivalue in (ScreenVariableInputValue, FieldInputValue, VariableInputValue, DictInputValue, LocalVariableInputValue):
--- a/sphinx/source/changelog.rst
+++ b/sphinx/source/changelog.rst
@@ -124,26 +124,29 @@ displayed. Clicking on the filename and line will open
 the file in the default text editor, at the given line,
 if possible.
 
-Data Actions
-------------
+Actions and Screen Values
+-------------------------
 
-The :ref:`data-actions` are now presented and explained in a more
-condensed manner. These actions have been reimplemented using a data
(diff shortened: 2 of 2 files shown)


mal:
I don't think this makes sense on the base input value. It seems too niche to see widespread use, while also being too simple for most cases that want to manipulate the editable state of inputs, such anyone really wanting this functionality is probably going to have to define their own class anyway to do the extra bits.

Perhaps there could be some limited value to separate ActionInputValue (or something) class which could take an action to perform on enter, with a special case for using it's own Disable, but even then, allowing the key press to propagate and using `keysym` with a `button` is probably a better choice for cross-platform accessibility.

mal:
This isn't about facilitating the disabling of the input, that is already possible with the existing Disable action attached to the InputValue. It's about disabling the field once enter is pressed, which in isolation isn't terribly useful unless you've done a bunch of other groundwork to allow re-enabling it, or some other way of triggering progression and processing of the value provided.

shawna-p:
Once again, I strongly disagree with your idea that this is not "terribly useful". I am saying this as someone who has worked on dozens of games and always needs this kind of functionality for usable mobile input, and for screens with multiple input values. It isn't relevant that there is an alternative way to do this with the existing Disable action - the question is, can we make this easier to accomplish, and the answer is, yes, by adding an incredibly simple argument which won't affect any existing input values unless they opt into it. 

This functionality is already built into the game in the `file_slots` screen with a clear example on how to set up and use this sort of input - the issue is that this functionality is exclusive to the `FilePageNameInputValue` class by default. The setup involves the following:

```renpy
default my_input = VariableInputValue("something")
button:
  key_events True
  action my_input.Toggle()
  input value my_input
``` 

If I use `key` to disable the input when Enter is pressed and I have multiple input values on the screen which could be active, I have to disable every one of them, when it would be simpler to just have each input be in charge of disabling itself.

I feel you're approaching this from the wrong angle - the question has never been "can Ren'Py already do this", to which the answer is clearly yes, it can. The question is if we can make it easier for devs to add this functionality into their games, and the answer to that is also yes. I've given several reasons why I think it's a useful feature to boot.

mal:
I'm not saying this has no use, and that we can't do better in the realm on input management, we absolutely can, however I do feel like this feature as presented falls in a middle ground of scratching an itch indicative of a more general issue without really leading towards a solution of that wider problem.

For example, to the point about multiple inputs specifically, we know at the engine level that only one input can be active at once, so perhaps something to look at as part of a broader solve would be in ensuring that is the case so it's harder to get into the state of having multiple active inputs in the first place (i.e. when one is enabled, others can't be), which would benefit not only this issue, but the wider ecosystem.

Please don't take this as my being against improving the situation, I'm not. I also deal with this issue when building games, I'm just not convinced that this particular addition moves us towards a better system overall and am wary of expanding a public API for what may be a short term solve but that will need to be supported for the next decade and beyond.

Gouvernathor:
Your proposition of taking another Action as a parameter, to be run when enter is pressed (or as a parameter to a subclass of InputValue, whatever), is more generalized than this, sure, but it also doesn't solve the issue in any simple way. You can't pass a value's Disable action to the value itself at the moment of its own creation, you have to first define the value in a screen variable, then at best set an enter_action attribute to the Disable attribute of the same value. That's a two-liner and that's pretty ugly in screens, in addition to messing with screen optimization. And it's not that much simpler or evident for your typical creator than Fen's initial solution of a subclass.
If we stick to an improvement where what happens at enter is passed by parameter, self.Disable can only be a special-case triggered by a flag value. So, your solution could be a supplement to this one, but I don't think it would have much practical usability on its own.

That being said, I didn't understand everything in that matter of multiple inputs active at the same time but instead of disabling just itself, the input value could run the DisableAllInputValues action. If we assume that only one input is active at any given time, the result is the same - only the one active input, catching the enter keystroke, would get disabled. And we can branch back to your proposition of rather taking an action as a parameter - though I think adding a little syntax sugar on that one wouldn't hurt.

mal:
I did explicitly note that using it's own disable would need to be special cased ...

It's a moot point anyway as that was more an example of a broader approach than it was any real advocation for such a thing (the problems extend beyond that).

I've laid out my opinion: very specifically that this aims to treat a symptom - disabling an input in a multi-input scenario - rather than the cause - that Ren'Py was designed with a view towards there being only one input on screen, and multi-inputs being sufficiently advanced that creators should handle that in their entirety - and that if we want to do a better job making multi-inputs more accessible it's much wider topic than this, so I'll leave it there.

Remember that it is just my opinion and thankfully we don't all have to agree, that's why we have Tom. 😛

Issue #4865, opened 2023-07-23
Document and expand renpy.get_translation_identifier() 


raspberry-soft:
`renpy.get_translation_identifier()`  does not appear on the doc translation page. I think it should be added that it works from the console. Also, it would be super handy if it it automatically copy the id to the clipboard.

Gouvernathor:
It is already documented : https://nightly.renpy.org/doc/translation.html#renpy.get_translation_identifier
It is used internally so it should not mess with the clipboard, and special-casing it from the console would be weird - either something is a console command, or it is a function, but we can't have it both ways.

There could be a `copy` command to the console with similar syntax with `watch`, and which would copy the stringized version of the value of the parameter. There could also be an overlay screen, much like the `F2` one, with a single text button to copy the current translation id - or we could add that button the the existing F2 screen.

mal:
> There could also be an overlay screen, much like the F2 one, with a single text button to copy the current translation id - or we could add that button the the existing F2 screen.

Already building it. 😉

PR #4866, opened 2023-07-24
Tool to view and copy translation identifiers

mal:
Inspired by #4865.

Adds a screen (toggled via the developer menu) which displays the current translation identifier and, on click, copies the value to the clipboard.


renpytom:
Looks good. Thanks!

PR #4887, opened 2023-08-02
better implementation of `AlphaMask`

Elckarow:
-use `DISSOLVE` as render operation rather than `IMAGEDISSOLVE`. that way, no need to create and render a `Fixed` to get a third texture (since only two are needed)
-use its own shader `renpy.alphamask`. i know `renpy.alpha_mask` is also a shader, but it's only used for `Movie`s. maybe it could be renamed to `renpy.movie_mask`?
-new parameter `alpha_over` (not sure about the name tho). a float that ranges from 0.0 to 1.0. at 0.0, nothing changes. at 1.0, the behavior is swapped (the child is opaque where the mask is transparent). allows for more customization (for instance, 1.0 creates a "reversed" alphamask).

### DIFF
--- a/renpy/common/_shaders.rpym
+++ b/renpy/common/_shaders.rpym
@@ -143,6 +143,7 @@ init python:
         gl_FragColor = texture2D(tex0, v_tex_coord.xy, -1.0);
     """)
 
+    # used by Movie
     renpy.register_shader("renpy.alpha_mask", variables="""
         uniform sampler2D tex0;
         uniform sampler2D tex1;
@@ -157,6 +158,7 @@ init python:
         gl_FragColor = vec4(src.r * mask.r, src.g * mask.r, src.b * mask.r, mask.r);
     """)
 
+    # used by AlphaMask
     renpy.register_shader("renpy.alphamask", variables="""
         uniform float u_alpha_over;
         uniform sampler2D tex0;
--- a/renpy/common/_shaders.rpym
+++ b/renpy/common/_shaders.rpym
@@ -157,6 +157,25 @@ init python:
         gl_FragColor = vec4(src.r * mask.r, src.g * mask.r, src.b * mask.r, mask.r);
     """)
 
+    renpy.register_shader("renpy.alphamask", variables="""
+        uniform float u_alpha_over;
+        uniform sampler2D tex0;
+        uniform sampler2D tex1;
+        attribute vec2 a_tex_coord;
+        varying vec2 v_tex_coord;
+    """, vertex_200="""
+        v_tex_coord = a_tex_coord;
+    """, fragment_500="""
+        vec4 source = texture2D(tex0, v_tex_coord.xy);
+        vec4 mask = texture2D(tex1, v_tex_coord.xy);
+
+        float alpha = source.a * mask.a;
+        float reversed_alpha = 1.0 - alpha;
+        float final_alpha = mix(alpha, reversed_alpha, u_alpha_over);
+
+        gl_FragColor = vec4(source.rgb * final_alpha, final_alpha);
+    """)
+
 init python hide:
     from operator import mul
 
--- a/renpy/display/layout.py
+++ b/renpy/display/layout.py
@@ -2229,7 +2229,8 @@ class AlphaMask(Container):
     size of the AlphaMask is the size of `child`.
 
     `alpha_over` is a float that ranges from 0.0 to 1.0.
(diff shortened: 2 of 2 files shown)


renpy/display/layout.py
```diff
@@ -2228,55 +2228,50 @@ class AlphaMask(Container):
     The `child` and `mask` parameters may be arbitrary displayables. The
     size of the AlphaMask is the size of `child`.
 
+    `alpha_over` is a float that ranges from 0.0 to 1.0.
+    At `0.0`, nothing changes. At `1.0`, the behavior is swapped (`child` is opaque where `mask` is transparent).
```
Gouvernathor, line 2232: "Nothing changes" is inaccurate : with 0.0, the alphamask still does its job and the child gets masked. Passing 0.0 to [InvertMatrix](https://www.renpy.org/dev-doc/html/matrixcolor.html#InvertMatrix) is a situation where we can actually say nothing changes.

Also, I think it should be introduced and documented as a boolean defaulting to False and which inverts the mask, and then you maybe say you can pass a float to tweak things more specifically (but it seems like a weird idea to pass anything other than 1 or 0 so it could stay undocumented as far as I'm concerned).

renpy/common/_shaders.rpym
```diff
@@ -157,6 +157,25 @@ init python:
         gl_FragColor = vec4(src.r * mask.r, src.g * mask.r, src.b * mask.r, mask.r);
     """)
 
+    renpy.register_shader("renpy.alphamask", variables="""
```
Gouvernathor, line 160: Maybe add a comment before this and the previous one, saying which is for display.layout.AlphaMask and which is for movies, so that the code is more understandable for posterity.

renpy/display/layout.py
Elckarow: i meant that the behavior stays the same

Elckarow: > And I'm also not sure about the name. invert_mask ?
i was thinking of `invert_alpha` but i guess the meaning is the same

Elckarow: > but it seems like a weird idea to pass anything other than 1 or 0

one such use case would be some sort of animated displayable where the value is interpolated. if you're just passing a static value, then yes, not a very bright idea since only `0.0` and `1.0` have a good looking effect

mal:
What advantage is there to this shader being more complex than a copy-paste of the existing `renpy.alpha_mask` shader, but leveraging the `.a` swizzle rather than `.r`? An inverted alpha mask is more straight-forwardly achievable by flipping the arguments, and doesn't a partial alpha mask result in unwanted zero-color artifacts?

Elckarow:
> What advantage is there to this shader being more complex than a copy-paste of the existing `renpy.alpha_mask` shader
the main idea behind this update is the `alpha_over` (or `invert_mask` or whatever), not really the use of `.a` instead of `.r`. if it was just that, i would've let governathor (with whom i discussed this update before opening the pr) do the commits (along with the use of `DISSOLVE` instead of `IMAGEDISSOLVE`).

> An inverted alpha mask is more straight-forwardly achievable by flipping the arguments
then the displayable that gets masked is different. not the same as the masking being inverted.

> doesn't a partial alpha mask result in unwanted zero-color artifacts?
haven't encountered anything odd with my testings

Elckarow:
that works as well. do we go with that?

renpytom:
I think I'd rather go with the shader-based approach, rather than the blend func one. The reason is the blend func one forces particular blend functions, but blend functions can be changed by the creator.

Elckarow:
> but blend functions can be changed by the creator.

i mean, shaders can be overriden... i'll let you guys discuss things before making any final changes

mal:
I'll try and find some time this weekend to look a bit more at this. There was one part especially that made me hesitate to turn the blend version into a PR which was related to how it interacts with Ren'Py's automatic cropping of transparent edge pixels - a feature which also creates issues with various other gl2 things, some shaders among them.

FWIW, the blend version only uses the `uguu` constants, and avoids creator defined function mapping but, if the shader version is the way to go, then I think there are some small improvements we can make there before we ship it. I'll do some testing and circle back.

mal:
Given Tom's preference, and the limitations I was running into when `config.optimize_texture_bounds = True`, I took a pass at an optimised version of the shader and cut down the displayable code a little in #4924.

mal:
@Elckarow any objections to closing this in favour of #4924?

Elckarow:
None. go for it 👍

PR #4889, opened 2023-08-03
Fix for [!c] text substitutions

Tristimdorion:
Fixed [!c] text substitution to ignore formatting tags like {color} when capitalizing the substitution, the existing code assumes there are no tags in the substitution, like color / font or size and only capitalizes the first character in the string.

### DIFF
--- a/renpy/common/00library.rpy
+++ b/renpy/common/00library.rpy
@@ -304,24 +304,6 @@ init -1700 python:
 
 
     ##########################################################################
-    # Name-only say statements.
-
-    # This character is copied when a name-only say statement is called.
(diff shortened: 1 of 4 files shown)


mal:
This isn't safe to merge. Various inputs result in a`StopIteration` being thrown. If we go down this road it probably needs to parse the tags properly, not use a regex.

Gouvernathor:
Don't we already have a method to strip tags from a string ? [`renpy.filter_text_tags(s, allow=())`](https://nightly.renpy.org/doc/text.html#renpy.filter_text_tags)

PR #4891, opened 2023-08-04
Edgecase with OneDrive vault

michaelsstuff:
Edgecase issue:

When lunching a renpy project from a OneDrive Vault `os.path.dirname(os.path.realpath(sys.argv[0]))` returns the Name of the virtualized file system Volume, but not the actual path. 
Changed it to `os.path.dirname(os.path.abspath(__file__))`, which in my limited test seems to work.


renpytom:
Merged, thanks.

Issue #4695, opened 2023-06-05
`if_changed` breaks `tight`


avonder:
```
init:
    $ renpy.music.register_channel("ambient", "sfx", loop=True, tight=True)

"Line 1"

play ambient amb01

"Line 2"

play ambient amb01 if_changed

"Line 3"

stop ambient fadeout 6.0

"Line 4"
```

In this situation, after you wait for a loop to occur on line 3, `tight` gets ignored and the sound is cut off upon reaching the end on line 4.
_(Test with a ~4-second clip for convenience)_

PR #4914, opened 2023-08-15
Revert removal of `loop_only` to fix `if_changed` loop seam issue

### LINKED ISSUE #4872: New `if_changed`/`loop_only` behaviour in `music.play()` incorrectly restarts looping music tracks (8.1.1/Nightly)
**AUGUST 14th UPDATE.** Please read the comments and TL;DR for updates on this issue. I'll provide a new TL;DR below and update the title with a better one to match. The existing reproducible example is still a relevant example of the issue.

TL;DR: Calling `play` on a track that is already looping with `if_changed` set to `True` will queue the same track as what is currently playing due to the removal of `loop_only` in commit `b0b096b`. This track will play from the **very beginning once** the current track reaches it's end, **ignoring any loop points defined in the track**. This leads to an abrupt interruption and restart to the track if it was made with a specific `loop` loop point in mind, even if that loop point is included in the track's file name.

Comments in code as well as documentation mention that this will "queue up an additional loop" to the currently playing music. If you have a set `loop` loop point that you are expecting the track to jump back to, this is ignored and the track will abruptly start again at the beginning (due to it being queued a second time).

----- 

Original body, describes discovery and recreating the issue:

After updating Ren'Py to 8.1.1 from 8.0.3, we noticed that our main menu theme (`config.main_menu_music`) would loop correctly for a few minutes, loop incorrectly once, then resume looping correctly afterwards. On this incorrect loop, the song starts again from the very beginning once reaching the end of the track as opposed to jumping directly to the loop point roughly 22 seconds into the track as it would on a successful loop. This issue does not occur on 8.0.3 or any other previously tested versions of Ren'Py (as far back as 6.99). We have observed this issue on Windows 10 and Mac.

When trying to recreate this issue, we discovered that the incorrect loop would only occur in situations where the main menu was entered from `splashscreen`, which in our project includes a call to start the main menu music during the game's intro sequence using `$ renpy.music.play(config.main_menu_music)`. The incorrect loop would not occur when entering the main menu from other parts of the game such as the pause menu or the ending of a scene.

To isolate and confirm the issue, we created a blank project on Ren'Py 8.1.1 where the only additions were our existing main menu theme file and a splashscreen label that contains `$ renpy.music.play(config.main_menu_music)` (value assigned in options.rpy). We were able to recreate the issue in this blank project, and I've included it as a .zip below.

[mainmenumusictest.zip](https://github.com/renpy/renpy/files/12175915/mainmenumusictest.zip)

Recreation steps: 

1. Unzip provided project
2. Start project in Ren'Py 8.1.1
3. Idle on main menu until track loops
4. Observe the music jumping back to the beginning of the track, not the loop point
5. Continue to idle, observing that it loops correctly forever after the one incorrect loop has occurred.
6. (Optional) Start a new game, exit to main menu, observe that the incorrect loop never occurs.

Additional investigation showed that reverting the files included in commit `b0b096b` to their previous revisions causes the incorrect loop to not occur, both in the test blank project and in our full game. This lines up with us not encountering this issue before updating to Ren'Py 8.1.1.

mithost:
Fixes #4872 by restoring the `loop_only` parameter of `enqueue()` in audio.py.

After commit `b0b096b`, calling `play` on a music track with the `if_changed` parameter would *always* add it to the queue, even if the same track was already currently playing on loop. This newly queued 'loop' would always begin at the very start of the track, **ignoring any user-defined `loop` point parameters** that were added by the user to allow the track to seamlessly loop. This caused the looping track to abruptly 'restart' the next time it would normally loop whenever a player visited a label that wanted that track to play if it wasn't already looping (for example, the 'market' example in Ren'Py's audio documentation page for `if_changed`).

This change restores the `loop_only` parameter and functionality to `enqueue()` and `play()` to avoid a track from being added to the queue and potentially abruptly restarting the track if it is currently already playing on loop. To attempt to preserve the fix the change `b0b096b` accomplished for #4695, the position/logic of the `c.dequeue()` calls in `music.play()` has not been reverted to it's previous behaviour.

Documentation wise, the Audio documentation page mentions that `if_changed` queues 'an additional loop' of the currently playing track when this situation is reached, which is only an accurate description if the currently looping track does not have a user defined `loop` point (or if the loop point was 0.0). A more accurate description is that the track is queued to play again from the very beginning after the current loop is complete, discarding any user-defined `loop` points. The issue mentioned above is reproducible using the code present in the `if_changed` section of the audio page, provided the 'market' track has a non-zero `loop` parameter.

### DIFF
--- a/renpy/audio/audio.py
+++ b/renpy/audio/audio.py
@@ -668,7 +668,7 @@ class Channel(object):
             renpysound.dequeue(self.number, True)
             renpysound.stop(self.number)
 
-    def enqueue(self, filenames, loop=True, synchro_start=False, fadein=0, tight=None, relative_volume=1.0):
+    def enqueue(self, filenames, loop=True, synchro_start=False, fadein=0, tight=None, loop_only=False, relative_volume=1.0):
 
         with lock:
 
@@ -676,20 +676,22 @@ class Channel(object):
                 filename, _, _ = self.split_filename(filename, False)
                 renpy.game.persistent._seen_audio[str(filename)] = True # type: ignore
 
-            if tight is None:
-                tight = self.tight
+            if not loop_only:
 
-            self.keep_queue += 1
-
-            for filename in filenames:
-                qe = QueueEntry(filename, fadein, tight, False, relative_volume)
-                self.queue.append(qe)
-
-                # Only fade the first thing in.
-                fadein = 0
-
-            self.wait_stop = synchro_start
-            self.synchro_start = synchro_start
+                if tight is None:
+                    tight = self.tight
+    
+                self.keep_queue += 1
+    
+                for filename in filenames:
+                    qe = QueueEntry(filename, fadein, tight, False, relative_volume)
+                    self.queue.append(qe)
+    
+                    # Only fade the first thing in.
+                    fadein = 0
+    
+                self.wait_stop = synchro_start
+                self.synchro_start = synchro_start
 
             if loop:
                 self.loop = list(filenames)
--- a/renpy/audio/music.py
+++ b/renpy/audio/music.py
@@ -116,11 +116,13 @@ def play(filenames, channel="music", loop=None, fadeout=None, synchro_start=Fals
 
             if if_changed and c.get_playing() in filenames:
                 fadein = 0
+                loop_only = loop_is_filenames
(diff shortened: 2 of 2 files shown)


renpytom:
Merged, thank you for the fix.

PR #4921, opened 2023-08-18
action Continue()

### LINKED ISSUE #4901: Add action Continue()
I was thinking that renpy could natively have a resume button where it was last left off, for example. If the player saved a game manually, or if it was saved automatically the game would allow to show in the main menu a continue button.

an example would be:
```
screen navigation ():

    vbox:
        style_prefix "navigation"

        xpos gui.navigation_xpos
        yalign 0.5

        spacing gui.navigation_spacing

       if renpy.can_continue():
             textbutton _ ("continue") action Continue()

        if main_menu:

            textbutton _ ("Start Game") action Jump ("restart")

        else:

           textbutton _ ("History") action ShowMenu ("history")

           textbutton _ ("Save") action ShowMenu ("save")

        
        textbutton _ ("Load game") action ShowMenu ("load")

        textbutton _ ("Options") action ShowMenu ("preferences")
```
Examples: 

Action Continue(save="auto")
   save: determines where the last save will be loaded from. by default it will take the automatic game saves (although it can take the manual saves, I thought it would be better to take the automatic ones).

renpy.can_continue() Check if there are saves that can be continued.

Neyunse:
Closes #4901 

* Tested in version 8.1.1 with The Question

### DIFF
--- a/module/generate_styles.py
+++ b/module/generate_styles.py
@@ -195,6 +195,7 @@ style_properties = sorted_dict(
     rest_indent=None,
     right_margin=None,
     right_padding=None,
+    ruby_line_leading=None,
     ruby_style=None,
     shaper=None,
     size=None,
--- a/renpy/common/00style.rpy
+++ b/renpy/common/00style.rpy
@@ -165,6 +165,7 @@ init -1800:
         slow_cps None
         slow_cps_multiplier 1.0
         slow_abortable False
+        ruby_line_leading 0
         ruby_style style.ruby_text
         altruby_style style.altruby_text
         # hyperlink_functions is set in 00defaults.rpy
--- a/renpy/common/_errorhandling.rpym
+++ b/renpy/common/_errorhandling.rpym
@@ -81,6 +81,7 @@ init label _errorhandling:
         line_leading 0
         line_overlap_split 0
         layout "tex"
+        ruby_line_leading 0
         subtitle_width 0.9
         slow_cps None
         slow_cps_multiplier 1.0
--- a/renpy/text/text.py
+++ b/renpy/text/text.py
@@ -759,7 +759,7 @@ class Layout(object):
 
             # Figure out the line height, line spacing, and the y coordinate of each
             # glyph.
-            l, y = textsupport.place_vertical(par_glyphs, y, self.scale_int(style.line_spacing), self.scale_int(style.line_leading))
+            l, y = textsupport.place_vertical(par_glyphs, y, self.scale_int(style.line_spacing), self.scale_int(style.line_leading), self.scale_int(style.ruby_line_leading))
             lines.extend(l)
 
             # Figure out the indent of the next paragraph.
--- a/renpy/text/textsupport.pyx
+++ b/renpy/text/textsupport.pyx
@@ -674,7 +674,7 @@ def place_horizontal(list glyphs, float start_x, float first_indent, float rest_
 
     return maxx
 
-def place_vertical(list glyphs, int y, int spacing, int leading):
+def place_vertical(list glyphs, int y, int spacing, int leading, int ruby_line_leading):
     """
     Vertically places the non-ruby glyphs. Returns a list of line end heights,
     and the y-value for the top of the next line.
@@ -685,6 +685,8 @@ def place_vertical(list glyphs, int y, int spacing, int leading):
     cdef int pos, sol, len_glyphs, i
     cdef int ascent, line_spacing
     cdef bint end_line
+    cdef bint has_ruby
+    cdef int line_leading
 
     if not glyphs:
         return [ ], y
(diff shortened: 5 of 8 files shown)


renpy/common/00action_menu.rpy
```diff
@@ -133,6 +133,44 @@ init -1500 python:
...
+        `regexp`
+            If present, will be used in `renpy.newest_slot`. 
+            By default this searches for all manually saves.
+        """
+
+        def __init__(self, regexp=r"\d+"):
+            self.regexp = regexp
+        def __call__(self):
+            if not self.get_sensitive():
+                return
```
KiloOscarSix, line 155: I don't think sensitive should be added here as devs can add `sensitive` property to the widgets

Gouvernathor (commented):
I think the default value for regexp should be None, so that it does not filter in any way by default. Auto and quick saves are just as valid in the typical case - and your doc entry is inexact since quick saves _are_ manual saves.

I'm not against defensive programming in the `__call__` and prevent the call from failure, but in that case the `store.main_menu` test (in which the store prefix is not necessary) shouldn't be enough to reject the call. In fact, I don't think that's a valid reason to make the action insensitive - it should be left to the creators.
And the get_sensitive method should avoid returning None.

All in all, I'm not sold on this action being necessary, but if these changes are made, the implementation would be fine by me.

Neyunse:
@Gouvernathor can you check if is ok now?

Gouvernathor:
I liked the test for _in_replay and the presence of get_sensitive, but it's also fine like this.

Neyunse:
@Gouvernathor i am seeing that when i delete all the games, the button loads a non-existent game. i think it is the "_reload" one. should this be the case?

Gouvernathor:
That's the result of doing a reload with Shift+R. It should not happen to end users, only creators, but it's probably better to get rid of it if possible.

Neyunse:
@Gouvernathor There, it's done. Thanks for the corrections

Gouvernathor:
The issue with it is that a creator using Reload will have the option disabled _even if_ there are other valid saves just after the _reload one.
I suggest having `r"(?!_reload).+"` as the default regexp instead (and testing `is None` in the call method), that will match anything not starting with `_reload`.

Neyunse:
@Gouvernathor i added `r"(?!_reload).+"`  as the default regexp

Org-gith:
Maybe the action could have an argument to open a confirm screen or not (for example a bit like MainMenu(confirm=False/True)). The confirm message could be for example:
"Do you want to load your last save?" 
Yes > load your last save
No > open the normal load screen

Don't think in a main menu, you would ever want to have the load button as well if you plan on using this as a button, this seems redundant and takes space for nothing, a confirm would streamline it imo.

Gouvernathor:
I think a Continue button will be expected to just load the latest save without confirmation, at least in the main menu / general case. Also, a "Continue" button isn't easy to understand as being the way to go to the Load screen provided you answer "no" to a confirm. I wouldn't know how a creator would name that button in an intelligible way.
I think Continue should behave just like [FileLoad](https://www.renpy.org/doc/html/screen_actions.html#FileLoad) when it comes to confirming.

Also, I think it would be simple for creators to implement the version you proposed using the Confirm action and this version of Continue, while the reverse is not true.

Last point, this may be good to add to The Question or to the default GUI. But let's do that in another PR if this one is merged.

Org-gith:
I mean in the case I was explaining if you want to just load your last save without confirmation, you would do Continue(confirm=False) if that makes sense?
I think it's bad to have a continue and load button on a main menu screen, that doesn't seem well streamlined. But if people want to do this it would still be possible. Also I was not considering this would be the default at all in the new gui. So this would be something put together by the creator regardless.
I don't think if you have only one button that is continue on your main menu, saying yes to it will load to a previous save, saying no would open the load menu. I don't think what else could you read here but again I don't plan to use it, just saying.

Neyunse:
@Gouvernathor @Org-gith I implemented confirm message, but it's false by default. Since I feel that it can be annoying for the player to be shown a message all the time, the purpose of this is to create a small shortcut in the main menu for the player to load his save faster. Of course, it's up to the developer to implement it or not.

Org-gith:
"Are you sure you want to continue where you left off?" seems slightly better I think, otherwise it seems to work well and I see that it can be annoying in some cases, definitely off by default works well.

mal:
This feels like it's being significantly over-cooked.

A "Continue" action should be analogous to `Start`. A confirmation option makes no sense since, "continuing" while you're in a game already make no sense, so it'd only really be used in the main menu, and prompting there has no value.

renpy/common/00layout.rpy
```diff
@@ -452,6 +452,7 @@ init -1400 python hide:
     layout.LOADING = "Loading will lose unsaved progress.\nAre you sure you want to do this?"
     layout.QUIT = "Are you sure you want to quit?"
     layout.MAIN_MENU = "Are you sure you want to return to the main menu?\nThis will lose unsaved progress."
+    layout.CONTINUE = "Are you sure you want to continue where you leave it?"
```
Mikurox, line 455:
```suggestion
    layout.CONTINUE = "Are you sure you want to continue where you left off?"
```

Mikurox (commented):
Just a small text suggestion as it "sounds" better to me.

PR #4922, opened 2023-08-18
Enhanced version of "[!c]" interpolation that finds the first printable word and capitalizes it.

Tristimdorion:
Tested with various combinations of tags and printable text and with uni-code characters.
I think an actual lex parser would be overkill for this 'simple' function.

### DIFF
--- a/renpy/substitutions.py
+++ b/renpy/substitutions.py
@@ -30,9 +30,10 @@ from renpy.compat import PY2, basestring, bchr, bord, chr, open, pystr, range, r
 import renpy
 import string
 import os
+import re
 
 update_translations = "RENPY_UPDATE_TRANSLATIONS" in os.environ
-
+find_display_tags_regex = re.compile(r"([{[[].*?[]}])")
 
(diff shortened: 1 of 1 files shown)


mal:
As previously noted, regex isn't a good solution for this. We already have centralised code for tag parsing which should be used for things like this so that if it changes in future this feature doesn't get left behind.

With that said, I still think it would be better to advocate that creators keep their styling (tags) and content (text) separate, especially in variables targeted for interpolation as it sets them up much better for any future translation efforts they may wish to persue, in addition to (in general) being good practice.

Tristimdorion:
I understand what you are saying, but since capitalization is unrelated to the tags, it would be strange to completely parse all tags just to find the first printable word and capitalize it. This solution just throws out the non-printable text and capitalizes the first printable word, as intended by the description of the feature.

Booplicate:
Parsing and removed tags is related. You're reinventing the wheel with the regex. There's already a function to do what you want, by not using it, you're adding extra burden for little to none benefit.

PR #4946, opened 2023-08-31
Update 00inputvalues.rpy

the66F95:
restores the previous behavior to be able to deactivate InputValue objects via "Enter".
for example, the current behavior can be seen in any project using the button that wraps the FilePageNameInputValue object.


renpytom:
Merged, thanks.

PR #4950, opened 2023-09-03
Add Quick SaveLoad in Keymap

Otorhin:


renpytom:
Do you have any standard for f5 and f6 being save and load keys?

Otorhin:
> 

i dont think so, but in many games f5 is a hotkey for quicksave, and f6 is a hotkey for quickload

renpytom:
I don't think I want to add these bindings right now. (I do want to revisit the bindings, probably getting rid of the ctrl binding that we have now.)

PR #4956, opened 2023-09-06
incompatible.rst: Fix typo Whine -> While

brunoais:


Gouvernathor:
It was better handled in the fix branch, but thanks for the tip.

brunoais:
Ups. Sorry for the mistake.

Gouvernathor:
Oh it's fine. It's not easy to manage two branches in parallel. And reporting an error is not a mistake :)

PR #4971, opened 2023-09-13
Tweaks to hbfont

Gouvernathor:
The [casefold](https://docs.python.org/3/library/stdtypes.html#str.casefold) method is preferred to lower for case-insensitive comparisons.
Break looks faster than continue in that for loop (there's another similar instance lower down the file but it's not as clear that the situation is the same, I didn't change it).


Gouvernathor:
casefold doesn't work on py2

PR #4972, opened 2023-09-14
Drag & Drop fixes and expansions

shawna-p:
This PR fixes a few things with the Drag & Drop system and adds some new features.

**Fixes:**
- `start_x` and `start_y` are correctly relative to the top-left corner of the drag, not the grab point
- `hovered`, `unhovered`, and `tooltip` arguments are documented
- `grab_x` and `grab_y` are documented
- `last_drop` is documented
- Fixed a spelling mistake in the RST file

**Features:**
- `clicked` will pass the Drag object which was clicked to the callback. This is in a try/except block for backwards compatibility with the previous option that did not pass the Drag object. This is in line with other callbacks passing the Drag object - I suggest a better name for a property which does not pass a Drag be `action` to match buttons, if desired.
- `grab_x` and `grab_y` are not cleared before the callbacks which occur when a drag is released (so they can be checked)
- Snap improvements:
  - `snap` takes an optional warper, which is used to animate the snapping animation
  - A special `snapping` property is set to True when the drag is in the middle of a snapping animation
  - A `snapped` callback can be provided to a Drag. It is called with several relevant arguments when a snap animation ends, in particular so users can better manage behaviour around interrupted snaps or whether drags/drops should be available for dragging/dropping while snapping.
- `drag_joined` will take either a direct `Drag` object or a `drag_name` of a Drag within the same DragGroup

Overall these features are intended to improve the flexibility of the Drag & Drop system for users by providing them with more tools to customize drag behaviour.


renpy/display/dragdrop.py
```diff
@@ -146,17 +146,26 @@ class Drag(renpy.display.displayable.Displayable, renpy.revertable.RevertableObj
...
         be necessary to increase :var:`config.longpress_duration` if
         this triggers to early on mobile platforms.
 
+    `hovered`
+        An action to run when the drag gains focus.
+
+    `unhovered`
+        An action to run when the drag loses focus.
+
+    `tooltip`
```
Gouvernathor, line 166: This (and possibly other additions) need to be added to :args: at the top of the docstring.
I suggest also using ``:scpref:`tooltip` `` as a link to the property.

Also, as written it seems like these get documented as parameters but I don't know they actually are.

renpy/display/dragdrop.py
```diff
@@ -146,17 +146,26 @@ class Drag(renpy.display.displayable.Displayable, renpy.revertable.RevertableObj
         None.
 
     `clicked`
-        A callback this is called, with no arguments, when the Drag is
-        clicked without being moved. A droppable can also be focused
-        and clicked.  If the callback returns a value other than None,
-        that value is returned as the result of the interaction.
+        A callback that is called when the Drag is clicked without being moved.
+        It is called with one argument, the Drag being clicked on. A droppable
```
Gouvernathor, line 150: To make it future-proof, maybe tell people that it should ignore unknown positional and keyword parameters ?

mal:
> - `drag_joined` will take either a direct `Drag` object or a `drag_name` of a Drag within the same DragGroup

What's the use case for this? The `drag_joined` is already passed the drag in question which it can use to formulate the response.

shawna-p:
> > * `drag_joined` will take either a direct `Drag` object or a `drag_name` of a Drag within the same DragGroup
> 
> What's the use case for this? The `drag_joined` is already passed the drag in question which it can use to formulate the response.

Sure, but the `drag_group` property isn't documented so there isn't a clear way on how to get other drags if you used screen language to declare them. Consider the following:

```renpy
for i in range(5):
  drag:
    drag_name "card{}".format(i)
    xpos 100*i
    drag_joined join_cards
```

It's not clear how in the `join_cards` function I'm supposed to figure out the Drag objects of `card0` to `card4` when all I'll have to work with is the Drag for whichever card is being dragged. Besides which, I think it's simply a convenience to be able to refer to drags by their name rather than having to make your own method of fetching `drag.drag_group.get_child_by_name("card1")` for each drag. It isn't impossible to do, but it's inconvenient, unintuitive, and hidden, and it doesn't have to be. The `drag_name` already acts as an ID of sorts so its use to refer to specific drags is intuitive here.

mal:
I think documenting `drag_group` would be a better solution here, since then when you write your custom `drag_joined` you can give it whichever behaviour you want (including use of `self.drag_group.get_child_by_name` as you outline) as well as choosing how to handle cases such as a drag not existing, drag_group being `None` etc, since different creators will likely want different behaviours in those cases.

PR #4973, opened 2023-09-14
Add ScrollViewport action

altskop:
This PR introduces a new Screen Action, `ScrollViewport`. 

Unlike the existing `Scroll` action, which only offers relative scrolling in one dimension (X or Y), `ScrollViewport` provides the flexibility to scroll to precise coordinates (which could either be relative or absolute).

For example, in a use case where you have a viewport that has to scroll to the center upon a button press, `Scroll` action is insufficient. It lacks control over both dimensions and requires the developer to set the scroll amount, which often is not a calculation that can be done without knowing the viewport's current adjustment values.

With the introduction of `ScrollViewport`, it is as easy as this:
```
action ScrollViewport("test_viewport", (0.5, 0.5), 0.5)
```
This action will cause the viewport with the id `test_viewport` to scroll to its center in 0.5s.

Considerations:
- `Scroll` action could be further enhanced by allowing providing specific value to scroll to, vs the amount of pixels/pages to scroll by. It would still lack control over both dimensions, but a set of two `Scrolls` would achieve the same effect as the `ScrollViewport` action. Let me know if you'd prefer this option instead.

### DIFF
--- a/renpy/common/00action_other.rpy
+++ b/renpy/common/00action_other.rpy
@@ -710,6 +710,70 @@ init -1500 python:
 
 
     @renpy.pure
+    class SetViewportPosition(Action, DictEquality):
+        """
+        :doc: other_action
+
+        Causes a Viewport to scroll to a specific position.
+
+        `id`
+            The id of the viewport in the current screen.
+
+        `position`
+            x and y coordinates for the viewport to scroll to.
+
+        `delay`
+            If non-zero, the scroll will be animated for this many seconds.
+        """
+
+        delay = 0.0
+
+        def __init__(self, id, position, delay=0.0):
+            self.id = id
+            self.position = position
+            self.delay = delay
+
+        def __call__(self):
+
+            d = renpy.get_widget(None, self.id)
+
+            if d is None:
+                raise Exception("There is no displayable with the id {}.".format(self.id))
+
+            if not isinstance(d, Viewport):
+                raise Exception("The displayable with the id {} is not a viewport. SetViewportPosition only takes Viewports as arguments.".format(self.id))
+
+            if not isinstance(self.position, tuple):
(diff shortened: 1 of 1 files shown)


renpy/common/00action_other.rpy
```diff
@@ -709,6 +709,70 @@ init -1500 python:
...
+                    xyadjustment_change.append(pos - adjustment[i].value)
+
+            if self.delay == 0.0:
+                # if there is no delay, call change and sum the x and y adjustment changes to arrive at the final position
+                xadjustment.change(xadjustment.value + xyadjustment_change[0])
+                yadjustment.change(yadjustment.value + xyadjustment_change[1])
+            else:
+                # otherwise, animate the x and y adjustment changes
+                xadjustment.animate(xyadjustment_change[0], self.delay, _warper.ease)
+                yadjustment.animate(xyadjustment_change[1], self.delay, _warper.ease)
```
mal, line 773: This section can be simplified significantly by leveraging changes merged earlier in the 8,2,0 cycle.
```suggestion
            px, py  = self.position
            x = absolute.compute_raw(px, d.xadjustment.range)
            y = absolute.compute_raw(py, d.yadjustment.range)

            if self.delay:
                d.xadjustment.animate(x - d.xadjustment.value, self.delay, _warper.ease)
                d.yadjustment.animate(y - d.yadjustment.value, self.delay, _warper.ease)
            else:
                d.xadjustment.change(x)
                d.yadjustment.change(y)
```
```

mal (commented):
My first impression was that this probably belonged in the existing `Scroll` action, but after seeing how geared towards relative movement that action is I became less convinced.

Either way, I've included a code suggestion to simplify and take advantage of previous changes in the current release cycle.

PR #4977, opened 2023-09-18
Or-augmented default statement

### LINKED ISSUE #4976: Better integrate dicts with defaults
A common issue when you use a dict variable in renpy is that when you open a save where the variable was not set, it is created all right, but when you add keys to the default dict and load a save where those keys aren't there, it doesn't add them.
In that situation you have to either add some code to the after_load label or add a `config.after_default_callbacks`, which is not very handy.
A good solution when keys are strings is to create the different key-value pairs as separate defaulted variables in a specific named store, but in that case you lose the ability to iterate over them.
For context, define supports augmented (some) assignments like `define var |= x` but default doesn't.
The default behavior should not be changed, first because of compat, second because removing a dict key can be meaningful in other uses, third because it would make it depend on the type of the value being defaulted (once evaluated) and that sounds terrible.

I think adding `default var |= val` could be good. What it would do is :
- if the variable is not defined, it is set to that value
- if the variable does exist, it is updated with the value

That means it would not break the normal behavior of the | operator and work with sets and frozensets as well, even though they're not likely to be used very much we don't exclude a valid part of Python.
I don't think adding `+=` would be good since as opposed to update/union it does not stabilize when applied several times : the result of a sum/concat is always different with the base value.

Gouvernathor:
close #4976 

An inconvenience is that the | operation is done in the reverse order of the one readable : when `var` is already defined, we do `var = val | var` and not `var |= val` as it would appear. It could reveal to be a thorn later, I'll leave that to the judgement of others.

### DIFF
--- a/renpy/ast.py
+++ b/renpy/ast.py
@@ -2645,20 +2645,20 @@ class Default(Node):
         if self.varname in defaults_set:
             if start and renpy.config.developer:
                 raise Exception("{}.{} is being given a default a second time.".format(self.store, self.varname))
-            return
 
-        # do the variable shadowing if not in this case, for compatibility reasons
-        if start and (renpy.config.developer is True):
-            fullname = '.'.join((self.store, self.varname))
-            if fullname in renpy.python.store_dicts:
-                raise Exception("{} is being given a default, but a store with that name already exists.".format(fullname))
+        else:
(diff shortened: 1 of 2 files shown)


renpytom:
This particular version is rejected in favor a .setdefault equivalent.

Issue #5009, opened 2023-09-27
Interpolation functions


renpytom:
When interpolating, the object returned should be checked to see if it's callable. If it is, then the function should be called, and the result of the call should be formatted and interpolated.

mal:
Interpolating the result of `__call__` rather than the result of `__str__` would be surprising and break games that already expect/rely on it, so if pursued this feature would need to differentiate such cases. Perhaps a new `[custom_func!f]` suffix.

Issue #5006, opened 2023-09-26
Optionally storing custom text tag states for retroactive viewing via _history_list


feltcute:
I am involved in a project that is trying to make use of custom text tags to cleanly present text tags of game variables that are highly mutable per each interval of dialogue.

Using [text interpolation] requires directly referring to variables that can otherwise be rather lengthy to tell a writer to type (e.g. `player.inventory.money`), and other times uses code inside the text tag functions to conveniently determine and format its final value when it was called.

It is possible to alias using `money = player.inventory.money`, but I'd rather keep it out of global space if it could be helped, especially for things built to be as little characters as possible for shorthand referral. It is also more sensitive to when and where a variable for the shorthand is declared, whereas using custom text tags was found to be far more flexible in the treatment of declaring variables.

The only shortcoming to custom text tags is that it always reflects the latest result in history. Consider the following scenario:
1) '300' is the text tag, making use of custom text tags to make it bold and and colored gold out of the box as well. 
2) Represented as the short and concise `{money}`, fetching the value of `player.inventory.money`
```
"You have **300** coins? That's too little." (it was 40 at the time)
...
"You have **300** coins?! That's a lot!" (it was 900 at the time)
...
"You now only have **300** coins left!" (present time)
```

This behavior is preferable by default for things like the player's name often times, but less-so when it is an intentionally mutable value.

I'd like to optionally store the custom text tag state for the functional sake of when displayed via _history_list in the history screen so that it always reflects what its value was at the time, rather than what its latest value currently is.

renpytom:
This isn't what custom text tags are for. 

If you'd like this, you can figure something out with https://www.renpy.org/doc/html/config.html#var-config.history_callbacks, but if you want interpolation, use interpolation and not text tags, as that's the supported way of doing this.

Gouvernathor:
I'm surprised that the custom text tags doc doesn't mandate the functions to be pure (= returning the same value for the same input and without side-effect). It most definitely should.
The text tag way is not a good idea to compute dynamic values depending on the state of the game because it can be called at any time during prediction, which can yield index or attribute errors in typical cases if the data is not always correctly set. I also once tried to have a text tag trigger a sound effect : it was a terrible idea.

feltcute:
I was in serious pursuit of trying to gauge feasibility of figuring something out with history_callbacks, though that would be moot if the functionality is restricted.

[...]

PR #5011, opened 2023-09-27
Use flag to govern callable conversion in interpolation

mal:
Follow up/proof-of-concept to https://github.com/renpy/renpy/issues/5009 and 4d8d065c893d434abf75f9ec2bd5f6cd44a3ecb2.

Advantages over the original implementation as I see it:
- Makes it clear which interpolations will call functions, and which won't, when reading scripts.
- Is purely an additive feature, so no incompat with previous versions.
- Allows a game to use both modes, rather than having a binary config flag.
- Can surface exceptions in developer mode as there's no ambiguity over when a callable should be used/not used.

Other notes:
- `f` was selected as the flag due to its obvious link to "function" but could potentially be something else.

### DIFF
--- a/renpy/common/00compat.rpy
+++ b/renpy/common/00compat.rpy
@@ -299,7 +299,6 @@ init -1100 python:
             config.simple_box_reverse = True
             build.itch_channels = list(build.itch_channels.items())
             style.default.shaper = "freetype"
-            config.interpolate_calls_functions = False
 
     # The version of Ren'Py this script is intended for, or
     # None if it's intended for the current version.
--- a/renpy/config.py
+++ b/renpy/config.py
@@ -1401,9 +1401,6 @@ simple_box_reverse = False
 # A map from font name to the hinting for the font.
 font_hinting = { None : "auto" }
 
-# Should interpolated functions be called for a value?
-interpolate_calls_functions = True
-
 del os
 del collections
 
--- a/renpy/substitutions.py
+++ b/renpy/substitutions.py
@@ -178,8 +178,16 @@ class Formatter(string.Formatter):
         if not conversion:
             raise ValueError("Conversion specifier can't be empty.")
 
-        if set(conversion) - set("rstqulci!"):
-            raise ValueError("Unknown symbols in conversion specifier, this must use only the \"rstqulci\".")
+        if set(conversion) - set("frstqulci!"):
+            raise ValueError("Unknown symbols in conversion specifier, this must use only the \"frstqulci\".")
+
+        if "f" in conversion:
+            try:
+                value = value()
+                conversion = conversion.replace("f", "")
+            except Exception:
+                if renpy.config.developer:
+                    raise
 
         if "r" in conversion:
             value = repr(value)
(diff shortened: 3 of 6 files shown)


Gouvernathor (approved):

feltcute:
Would it be possible to have support for both with the initial implementation of https://github.com/renpy/renpy/commit/4d8d065c893d434abf75f9ec2bd5f6cd44a3ecb2 off by default?

I can see why explicitly stating !f is better for standard use cases and agree it should be the case by default, but the project I am working on would benefit from being able to plainly and implicitly do [InterpolatedFunctionHere] without the extra overhead of sometimes having to remember to add !f to certain ones.

mal:
For your use case (presented in #5006), which I'd summarise as "easy to remember aliases for script writers", I wouldn't use this feature, I think something like this would likely fit your problem better, and would work right now in 8.1.3:
<details><summary>Code sample</summary>

```rpy
init python:
    class Alias:
        def __init__(self, fn):
            self.fn = fn
        def __str__(self):
            return str(self.fn())

define money = Alias(lambda: player.inventory.money)
define level = Alias(lambda: player.level)

label start:
    e "I have exactly [money] cents and am level [level]."
    $ player.inventory.money -= 100
    e "Eeep! I now only have [money]!"
    $ player.level += 1
    e "... At least I leveled up to [level] though."
    return
```
</details>

renpytom:
I think biting the bullet and moving to f-strings is the way to go here, where: 

```renpy
"Hello, [function()]
```

is what we have. With the option of using Mal's Alias class above if you really want to avoid the parens.

feltcute:
> For your use case (presented in #5006), which I'd summarise as "easy to remember aliases for script writers", I wouldn't use this feature, I think something like this would likely fit your problem better, and would work right now in 8.1.3:
> Code sample
> 
> ```python
> init python:
>     class Alias:
>         def __init__(self, fn):
>             self.fn = fn
>         def __str__(self):
>             return str(self.fn())
> 
> define money = Alias(lambda: player.inventory.money)
> define level = Alias(lambda: player.level)
> 
> label start:
>     e "I have exactly [money] cents and am level [level]."
>     $ player.inventory.money -= 100
>     e "Eeep! I now only have [money]!"
>     $ player.level += 1
>     e "... At least I leveled up to [level] though."
>     return
> ```

Your summary is accurate, and that does seem like a well fitting solution, I'll give it a whirl particularly considering it's something I can make use of now without using nightly. Much appreciated.

> I think biting the bullet and moving to f-strings is the way to go here, where:
> 
> ```renpy
> "Hello, [function()]
> ```
> 
> is what we have. With the option of using Mal's Alias class above if you really want to avoid the parens.

Would this allow for providing arguments as well? I have a usable enough work around in mind for the `storedColor` example with optional color declaration per #5006 regardless, though arguments through this behavior would be the best fit.

A thanks to Mal, Gouvernathor, and Renpytom for exploring this all so promptly over the last while.

mal:
Superceded by 5023.

PR #5055, opened 2023-10-14
add the `LocalVariableValue` action

Elckarow:
_not sure why this wasn't a thing before_

the sphinx doc isn't written since _i do not know what the widenekoflap `inc/value` is_


Elckarow:
oh, that's what the `:doc:` directive is for. (i think. idk)

renpytom:
Might be obsoleted by #5064, though.

PR #5066, opened 2023-10-21
move achievements to .py file

KiloOscarSix:
With the introduction of `_ren.py` files, I've been utilising Python tooling throughout my projects however one particular pain point has been the achievements module which is defined in an `.rpy` file. 
This PR aims to solve that by moving the definitions to a dedicated `.py` file while maintaining backwards compatibility by keeping the `.rpy` and importing everything into the old namespace/store.

The primary goal is to enhance developer experience by enabling them to easily access the achievements module within their own `_ren.py` code using the statement `from renpy import achievement`. This change paves the way for incorporating current and future Python tooling for the achievements module.

However, it's important to note that this approach may not be the most optimal one. While it has been tested and is functional in my projects, there might be potential issues with pickle that require further testing.

Your feedback and testing are highly appreciated and I hope there is the potential to improve Python compatibility within renpy and the `_ren.py` feature


renpytom:
I don't think this is something I'm likely to merge anytime soon. It might make sense to move the achievements into a _ren.py file, but I don't want it grossly moved like this.

PR #5082, opened 2023-10-30
Make the doc URLs local when it's available

Gouvernathor:
Generalize the usefulness of the existing actions, and use that in a link that was web-based even though it could afford to be local.

The schinese translation was using another website than the main doc, but with a wrong URL. I fixed it and then generalized it, but I think we can apply that to japanese and tchinese as well since they seem to have fairly stable [translated docs](https://doc.renpy.cn/).


renpytom:
I don't think I want this as is. The web documentation is a bit better than the local documentation - it has a few megabytes of pictures we don't ship with Ren'Py. 

What would be fine would be a toggle that lets the user pick between local and remote documentation.

Gouvernathor:
I disagree that the web version of the docs is that much better to be linked in the launcher. There are very few images on a very limited number of pages, and the likeliness that a creator will want to access a page with images is much lower than the likeliness that he uses renpy on a train with no internet. And there are people who for some reason stick to an older version of renpy and for whom the local documentation is more accurate than the one up to date. Also, the only place where this PR turns a web link into a local link is to the skins page of the documentation, which contains no images.

So, I made the local choice the default.

And I didn't test the code yet.

PR #5083, opened 2023-10-30
Allow more default lint options in the future

Gouvernathor:
Fix a short-sighted implementation. I consider this a bug, even though it's not one strictly speaking until we decide to add an opt-out lint option.

On PY2 the update if will probably trigger more often than it should since keys() returns a list, list equality is order-dependent, and dict ordering is arbitrary. Too bad.

Not yet tested.

### DIFF
--- a/launcher/game/front_page.rpy
+++ b/launcher/game/front_page.rpy
@@ -259,7 +259,7 @@ label lint:
         interface.processing(_("Checking script for potential problems..."))
         lint_fn = project.current.temp_filename("lint.txt")
 
-        project.current.launch([ 'lint', lint_fn, ] + list(persistent.lint_options), wait=True)
+        project.current.launch([ 'lint', lint_fn, ] + list(filter(persistent.lint_options.get, persistent.lint_options)), wait=True)
 
         e = renpy.editor.editor
         e.begin(True)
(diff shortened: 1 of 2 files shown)


renpytom:
This doesn't look good, as it will re-enable lint options if the old ones change. (And also the Py2 problem.)

Gouvernathor:
Later updates override prior updates, so the last call to update which passes the persistent read from disk, overrides the True value of any default option already present in persistent.
The PY specificity is not a big problem considering it just adds a few operations at launcher boot.

Gouvernathor:
I checked, the options persist and are not overridden, on py2 and py3.

mal:
The list of options to display in the interface, and the loop to display them I understand.

The changing of the persistent I do not. Lint defaults are set and managed by lint, not the launcher interface. The interface just helps a creator select which flags they'd like to apply when clicking the button.

If lint's default behaviour (i.e. what lint does without additional flags) is to do something and we decide that maybe some creators would like it to not then a new flag would be (for example) `--no-stats`. We can add option to the launcher interface without issue and creators can toggle it as needed. Think of `persistent.lint_options` as a list (set) of flags to add to lint when running it from the launcher, changing it to a dict seems to overly complicate things and will result in two tiers of "default" lint behaviour which doesn't seem desirable.

PR #5092, opened 2023-11-01
Better manage lint toggles V2

Gouvernathor:
From Mal's idea in #5083.
I had objections, but they were overcome by InvertSelected and the fact that we don't need a separate and complete list of options now : they can be written manually in the screen.

### DIFF
--- a/launcher/game/preferences.rpy
+++ b/launcher/game/preferences.rpy
@@ -1,4 +1,4 @@
-﻿# Copyright 2004-2023 Tom Rothamel <pytom@bishoujo.us>
+# Copyright 2004-2023 Tom Rothamel <pytom@bishoujo.us>
 #
 # Permission is hereby granted, free of charge, to any person
 # obtaining a copy of this software and associated documentation files
@@ -344,7 +344,7 @@ screen preferences():
 
                             add SPACER
 
-                            textbutton _("Check Script (Lint)") action Jump("lint")
+                            textbutton _("Check Script (Lint)") action Call("lint")
 
 
     textbutton _("Return") action Jump("front_page") style "l_left_button"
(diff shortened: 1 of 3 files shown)


launcher/game/preferences.rpy
```diff
@@ -334,19 +332,19 @@ screen preferences():
 
                             add HALF_SPACER
 
-                            textbutton _("Orphan translations"):
+                            textbutton _("Check for orphan/obsolete translations"):
                                 style "l_checkbox"
-                                action ToggleSetMembership(persistent.lint_options, "--orphan-tl")
-                            textbutton _("Parameters overriding builtin names"):
+                                action InvertSelected(ToggleSetMembership(persistent.lint_options, "--no-orphan-tl"))
+                            textbutton _("Check for parameters overriding builtin names"):
```
mal, line 338: "Shadowing" is the correct term for this, overriding is something different, and it would be the builtins themselves, not their names in either case.

launcher/game/preferences.rpy
```diff
@@ -334,19 +332,19 @@ screen preferences():
...
                                 action ToggleSetMembership(persistent.lint_options, "--builtins-parameters")
-                            textbutton _("Word count and character count for speaking characters"):
+                            textbutton _("Print word and character counts for speaking characters"):
                                 style "l_checkbox"
                                 action ToggleSetMembership(persistent.lint_options, "--words-char-count")
 
                             add SPACER
 
-                            textbutton _("Check Script (Lint)") action Jump("lint")
+                            textbutton _("Check Script (Lint)") action Call("lint")
```
mal, line 347: I don't think this belongs here. This is the global options page, and it's a project specific action.

mal (commented):
A few notes on wording, and a disagreement on the lint action living on the lint options tab. More intended for Tom during review, since I know we're already touched on these aspects in Discord.

PR #5093, opened 2023-11-01
Allow linting unclosed tags

Gouvernathor:
I don't think it's a good addition, but the code is there.


renpytom:
I think this is fine, as long as it's disabled by default.

PR #5099, opened 2023-11-05
Standalone parameterinfo

Gouvernathor:
Rewrite the ParameterInfo class, see #4403 for a rationale, but this version is independent of the inspect module and is py2 compatible.
Incidentally offers a very pretty repr for anything with a signature (I only changed the screens repr, since labels can't be viewed in the console and ATL already has a customized repr).

The direct accesses to Signature.parameters by other parts of renpy have been checked.
The `positional_only` and `keyword_only` attributes had an internal use only.
The other former attributes (`positional`, `extrakw`, `extrapos`) are still offered as accessors, for compatibility :
the last uses of `positional` are in atl.py, and the last use of `extrapos` and `extrakw` are in the `transform` statement parser (which needs reform in any case, that's #4799 which will need do be redone if this is merged first, and that's why I added a comment there).
The only use of the legacy constructor, which is also offered as a classmethod for compatibility, is in atl.py, in the same area which still uses the `positional` accessor and which generated invalid signatures - which are now fixed using the "long-time nonsense" I wrote in legacy_params.

The code looks (and is) longer, but a large part of it is copied from the inspect module which means it can be more easily maintained.
The pickle policy is that due to the new name of the class, pickles of this new version are purposefully impossible to resurrect in the old version, leading (in part) to a mandatory rpyc recompilation. That means that new saves containing ATL transforms, and saved screens (though that would be weird), cannot be opened in old versions of renpy.


Gouvernathor:
Todo : test the merging of 4799, and maybe remove some of the legacy accessors.

mal:
I think this would benefit from a clear list of the advantages (and disadvantages) for Ren'Py developers, creators, and players; and ideally goals of this specific PR. I've done my best to piece it together based on the descriptions and rationales offered and the previously linked issues, and it's still not clear to me. So far my take away has been:

1. It's meant to help make code maintainable and reduce duplication.

    But there's more code, not less, and more duplication (just of the stdlib instead).

2. It comes with a performance degradation.

    35% faster instantiation is nice, but +30% for every call is clearly not worth that trade.

3. It provides prettier repr output for screens, and a few other things.

    While nice, this doesn't seem enough to justify this.

I have to believe there is more and better reasoning for it, but it's been a slog to try and find this much hence this request for clarification of the "why" rather than the "what".

Gouvernathor:
The python parameters and calls syntax is inherently very complex, so if we want to emutale it, we need to have complex code at some point or other. The previous system, built upon the old parameter rules (with no * and no /, among others), was patched over time, making it less and less understandable and maintainable. Noticably, it contained a lot of _data_ duplication (not to be confused with _code_ duplication), which generally causes problem and actually does pose a big problem with ATL generating corrupt signature objects. I mentioned that in more details in the previous, closed PR.
That's why I wanted to redo it from the bottom up.

To help with this complexity, there is a standard python implementation that's very well maintained, that is part of the stdlib (it's the inspect module), and every possible future change on the parameter/call syntax in python will have to take it into account. So, a) we can confidently rely on that code's correctness, b) if in the future the rules change, we will be able to copy or adapt the changes of the inspect module to our system, provided we keep our code sufficiently similar to the stdlib implem since the stdlib will definitely be updated.
If we want to maintain or debug the existing implem, we have to first understand the current code, and I assure you it's a chore given how nothing is named like the python documentation describes it (and that is because our system was created under the former parameter/calls rules). Now, with this PR, we have a working system to compare ours with : if we see a bug, we can test whether the bug reproduces using the inspect module, and that helps a great deal for locating and solving it.
I believe that's worth having a somewhat longer code, though I would also note that a) it's not _that_ long, b) there are the legacy and pickling methods to take into account, which take some room, c) it sets the stage for new features which aren't available in ATL now but will be, so more features requires somewhat more code.

The deal is baseically this : there are a lot of things to solve about ATL parameters and calls, which needs reimplementing. If you want to do that based upon the existing signature class, then be my guest, I'm throwing the towel, and see you in 2 years.
If you have specific bottom-lines and requirements for the reimplement (performance-wise for example), I'm open to that.

About that performance thing, I'll have to redo the meaurements since the code has changed a bit since, but my measurement method was probably not very indicative considering I only tested one set of arguments being passed to one signature, under both systems. We can work together building a reliable measuring system, if we consider performance to be a blocking point, having reliable data will be necessary.

PR #5115, opened 2023-11-12
Differentiate between adv and nvl menus for window auto hide

### LINKED ISSUE #5089: NVL choice menu with nvl_menu: unwanted fading before and after choice
This is the same issue as this one from 2020: https://lemmasoft.renai.us/forums/viewtopic.php?p=527890

On a new project, I use the NVL narrator with `define narrator = nvl_narrator` in script.rpy.

If I add a choice menu, it will appear alone on the screen, with the NVL box hidden. There, the fading defined in the default `config.window_hide_transition` and `config.window_show_transition` does make sense because we switch from the NVL box to this fullscreen, dramatic choice.

If, however, I follow https://www.renpy.org/doc/html/nvl_mode.html and set define menu = nvl_menu so I can display the choice menu right below the last piece of text (Choose-Your-Own-Adventure / IF-style), the NVL box will *still* try to disappear, before reappearing right away, together with the choice.

Renpy 8 1 3 - NLV preserved text with menu causes fading

I can avoid this by using

```rpy
init python:
    config.window_hide_transition = None
    config.window_show_transition = None
```

but then I will lose transitions I may want outside choice menus. A cleaner solution would be to remove the hide+show again NVL box entirely (and after the choice, the unnecessary re-show NVL box).

Repro code: add this to script.rpy

```rpy
# Use NVL narrator by default
define narrator = nvl_narrator

# Display the menus immediately after the current page of NVL-mode text (IF-style)
define menu = nvl_menu
```

Kassy2048:
Use "menu-nvl" instead of "menu" for `config.window_auto_hide` when showing a nvl menu, as this kind of menus don't need to hide the dialogue window by default (contrary to adv menu).

Fix #5089.

### DIFF
--- a/renpy/ast.py
+++ b/renpy/ast.py
@@ -1883,15 +1883,17 @@ class Menu(Node):
 
         next_node(self.next)
 
-        if self.has_caption or renpy.config.choice_empty_window:
-            statement_name("menu-with-caption")
-        else:
-            statement_name("menu")
-
         if self.arguments is not None:
             args, kwargs = self.arguments.evaluate()
         else:
             args = kwargs = None
+
+        if self.has_caption or renpy.config.choice_empty_window:
+            statement_name("menu-with-caption")
+        elif kwargs is not None and kwargs.get('nvl') is True:
+            statement_name("menu-nvl")
+        else:
+            statement_name("menu")
 
         choices = [ ]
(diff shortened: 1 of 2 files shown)


renpytom:
Looks good. Thank you.

PR #5129, opened 2023-11-18
doc: Clarify renpy.pause's parameters

Gouvernathor:
It was unclear whether the pause ends when the time is expired AND the prediction is done, or if it ends when the time is expired OR the prediction is done.
From the issue the feature was proposed in, it's pretty clear that's an AND, and the passed time is a minimum rather than a maximum.
Now, what happens when the player clicks/skips, and what happens with timeless pauses ? I don't know, and I think it should be documented.

Also moves the hard parameter last, as it's not advised to use it.


renpytom:
It's actually an OR. 

The general thinking is you'd do renpy.pause(prediction=True), at which point the game would pause until either the condition was true, or the user clicked. It would also be possible do do renpy.pause(prediction=True, hard=True) to force prediction to finish.

PR #5170, opened 2023-12-12
fix: black texture on OS `Windows`

qTich:
In `gl2` mode on OS `Windows` on low-powered devices it is possible to have a side-effect in the form of a blank - black texture.

---

If PBO in its current form behaves better than before on Android mobile devices,
then in my opinion I would sacrifice it on Windows devices,
unless of course there is no other way to improve the situation.


renpytom:
I've done d0fd529ad45d06c6511825cc047c4502b929dc33 , which reverts PBO behavior to what it's been since June 2022 - disabled on angle and emscripten, enabled elsewhere. 

I'd need a case that this is causing problems on non-obsolete hardware to make further changes.

PR #5209, opened 2023-12-26
Danish translation of the launcher and The Question

symegac:
I translated the Ren'Py Launcher and The Question into Danish.
I also reordered the languages and balanced the two columns in the languages section of the The Question's preferences menu.


renpytom:
Any chance I can get the launcher (and The Question) if it's ready? I'd like this to be there for 8.2, which will be soon.

symegac:
Yes, of course. I'll mark this PR as ready for review. I'll add the translation of the tutorial in a future PR instead of here.

PR #5210, opened 2023-12-26
fix: loading sequence

qTich:
Loading the file from the archive inside `config.searchpath` on Android will be ignored, as the file is most likely also contained in `apk`.

A better solution would be to target `os.stat`, but I guess that's not currently available.

### DIFF
--- a/renpy/loader.py
+++ b/renpy/loader.py
@@ -526,26 +526,6 @@ def load_from_filesystem(name):
 file_open_callbacks.append(load_from_filesystem)
 
 
-def load_from_apk(name):
-    """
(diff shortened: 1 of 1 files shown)


renpytom:
What is this trying to do? It's not obvious to me.

qTich:
When distributing archives supplementing the game, for example - images of higher quality,
if they will be added using the archive with `config.searchpath`, they will be ignored inside the archive,
because the order of loading from inside `apks` has a higher priority.

renpytom:
Sure, I'll accept this.

Issue #5102, opened 2023-11-07
Renpy calls and ATL roadmap


Gouvernathor:
When an ATL transform is called, three things happen : the inner scope is updated (as with any call in renpy ot python), the parameters are curryfied and prepared for a potential next call, and the child of the transform is updated. All three of these are buggy.
Some parameters are not managed and treated as they should : pos-only are treated as pos-or-kw, required kw-only are not handled and cause exceptions, and variadic parameters are rejected at parse time. This affects the inner scope update and the correct currying of parameters for the next call.
The rules for setting the child are not consistent, resulting in the very function made to do it (At) failing in some edge cases, not to mention when you call the transform directly (not everyone uses At).

- [ ] Reform the ParameterInfo class into one with or around inspect.Signature interface. This is necessary to settle the following evolutions on solid ground.
- [ ] Fix ATL calls and currying, which comes into subtasks.
  - [ ] Fix and enable non-classic parameters (which means rewriting that part of the code)
    - [ ] Fix positional-only parameters to be just that (instead of treating them as pos-or-kw)
    - [ ] Fix/Enable keyword-only required parameters
    - [ ] Enable var-pos and var-kw parameters (\*args and \*\*kwargs)
    - [ ] Fix the "long-time nonsense" call of ParameterInfo, which yield inconsistent ParameterInfo objects
  - [ ] Clarify and improve how the ATL child is managed
    - [ ] Fix `At` not setting the child sometimes
    - [ ] clarify when the child is set depending on the signature (the child parameter, the old_widget parameter, and the case with no positional parameter)
- [ ] Allow the child to be accessed by the `child` local variable in ATL, regardless of the signature (a pure improvement, but it will help if taken into account in the previous steps)

As a sidenote, a difficulty when currying ATL is that in the resulting signature, the default values of the passed parameters are actual, evaluated values, whereas a "raw" ATL transform has evaluable expressions as default values of the optional parameters. Properly managing that will probably require two Signature classes, or two Parameter classes, or both.

renpytom:
@Gouvernathor Can you taker your new pull requests out of draft status, so I can merge them? (Also, a short description other than the title would help.)

Gouvernathor:
They are not tested and need to be, I will do it eventually.

Issue #4405, opened 2023-02-28
Pass the child to the ATL scope


Gouvernathor:
```rpy
transform lucy_jump_scare(child):
    child      # Show the original child.
    pause 5
    "lucy mad" # Jump scare.
    pause .2
    child      # Go back to the original child.
```
This is taken from the doc, about the special child parameter. However, there's no way of doing that cleanly in the show-with-atl statement:
```rpy
v "But..."
show eileen vhappy:
    2.
    "eileen screamer"
    .1
    "eileen vhappy"
e "Hmm..."
show eileen sad
```
In this example, if the player clicks too fast, the change to sad is overwritten by the ATL to screamer then to vhappy.
What I propose is this:
```rpy
show eileen vhappy:
    2.
    "eileen screamer"
    .1
    child
```
The `child` variable inside the ATL scope gets set to its child.
The possible interfering with the child parts of #4371 will need to be clarified, but in my view it should be done with a `context.context.setdefault("child", child)`, so that this behavior kicks in only in cases when it would have raised an exception.

Gouvernathor:
That would be a bug magnet.
The thing enabling the example I gave above is the fact that the `child` variable is evaluated at call time rather that the time where the statement is executed (2.1 seconds afterwards). But when `show eileen sad` is executed, the value of `child` is not evaluated again, so it's exactly the same as if the transform still read `"eileen vhappy"` instead of `child`.
Forcing a sort of special-case value to be evaluated at statement time rather than ATL apply time would likely introduce more bugs than it's worth.

The answer to this use case is multipurpose ATL transforms, and the use of the ATL partial.

Issue #4371, opened 2023-02-20
Required ATL transform's keyword-only parameter can't be curryfied


Gouvernathor:
```rpy
transform al(child, b, *, c, d=5):
    pass
```
`al(b=5)` works, and returns a (sort-of) curryfied version of al (sort-of because `al(b=5)(b=3)` works)
`al(d=6)` works and does (sort-of) the same
`al(c=5)` fails, for no apparent reason.

This is because the related section of atl.py was written when defaultless (=required) keyword-only arguments weren't a thing, and it wasn't updated ~~in 13 years~~ when Andy upgraded the parameters syntax to py3 standards.

This should/will be fixed when upgrading the parameters system to using the inspect module.

Also, atl doesn't take *args or **kwargs (the parser checks against it), and I'm not sure why.

Gouvernathor:
Unbound required keyword-only parameters are also forgotten when currying the transform : `al(b=7)` has an equivalent signature of `(child, b=7, d=6)`. Not only has the star been omitted (d is now positional-or-keyword, instead of keyword-only), but c has simply vanished.

Gouvernathor:
In addition to this, defaulted positional-or-keyword parameters are turned into keyword-only parameters for subsequent calls.
This means that for a signature such as `transform t(a, b, c=8)`, `t(1, 2, 3)` and `t(1, 2)(c=3)` will work but `t(1, 2)(3)` will not.

Gouvernathor:
Maybe this deserves a separate issue, because it has to do with `child` is managed.
```rpy
transform h(a, child, b=5):
    pass
```
`h(child=Null()).child` is None. The Null() has been passed to the `child` variable in the ATL block's namespace, but it was not set to be the `child` attribute of the resulting transform (it or a duplicated version, whatever).
One could argue that it makes sense, if the creator added a `child` parameter, to trust them with it hands-off. I'd note that when calling `h(1, Null())` the attribute does get set, so it's still a discrepancy, but ok. What's actually a problem, is that it breaks At.
`At(Null(), h)` returns that same _child-less_ transform. `chil != At(chil, t).child` should be an invariant, and here that invariant is broken.

PR #5237, opened 2024-01-02
Atl partial's overhaul and fix

Gouvernathor:
Adresses most of the main points of #5102, and #4371.
The parameters that get fixed are still forbidden though, as are var-pos and var-kw parameters. That will be for another PR. But the implementation that goes with handling those is already here, commented-out for performance.
The solution to #4405 is there too but also commented out. That change will also go in separately.


Gouvernathor:
This removes the last external call to ParameterInfo.legacy and paves the way to nuking it or maybe merge it with the unpickling mechanism (reducing the number of methods dedicated to that process).

Gouvernathor:
The (minor) behavior changes are not documented in the main doc, only in incompat. If that's all right, it will be part of #5238 to catch it all in one go.

PR #5246, opened 2024-01-06
feature: allow archives to be loaded from within `APK`

qTich:


renpytom:
Why would we want to load an archive from within another archive?

qTich:
> Why would we want to load an archive from within another archive?

I originally wanted writing a custom archive type to have a role for android as well,
but I guess at the current moment it significantly degrades performance.

PR #5252, opened 2024-01-08
Fix polar anchor and some ATL docs improvements

### LINKED ISSUE #5162: Broken ATL & Transform
16e50a978ee1f58735158c11f5ad5bbedb1e7aba - Broke some operations at the Transform and ATL level:

```py
File "PP\renpy\atl.py", line 684, in execute
action, arg, pause = block.execute(trans, timebase, self.atl_state, events)
File "game/modules/ATL.rpym", line 166, in execute
File "PP\renpy\atl.py", line 1415, in execute
startanchorangle = trans.state.anchorangle
File "PP\renpy\display\transform.py", line 392, in get_anchorangle
angle, radius = self.cartesian_to_polar_anchor(xanchor, yanchor)
File "PP\renpy\display\transform.py", line 292, in cartesian_to_polar_anchor
dx = x - xaround
File "PP\renpy\atl.py", line 109, in __sub__
return self + -other
TypeError: unsupported operand type(s) for +: 'position' and 'float'
```
If we uncomment isinstance (int, float) in `__add__`, we will get a different error:

```py
File "PP/renpy/atl.py", line 1415, in execute
startanchorangle = trans.state.anchorangle
File "PP/renpy/display/transform.py", line 392, in get_anchorangle
angle, radius = self.cartesian_to_polar_anchor(xanchor, yanchor)
File "PP/renpy/display/transform.py", line 295, in cartesian_to_polar_anchor
radius = math.hypot(dx, dy)
TypeError: must be real number, not position
```
Likewise, when entering a name in the tutorial:

```py
File "PP/renpy/display/transform.py", line 926, in render
return RenderTransform(self).render(width, height, st, at)
File "accelerator.pyx", line 989, in renpy.display.accelerator.RenderTransform.render
File "accelerator.pyx", line 328, in renpy.display.accelerator.RenderTransform.cropping
File "accelerator.pyx", line 125, in renpy.display.accelerator.relative
TypeError: int() argument must be a string, a bytes-like object or a number, not 'position'

While running game code:
File "game/script.rpy", line 186, in script call
call expression tutorial.label from _call_expression
File "game/tutorial_quickstart.rpy", line 599, in script
e "That's done with Python, and especially the renpy.input function. The first line of this example prompts the player for some texts, and sticks it in the name variable."
TypeError: int() argument must be a string, a bytes-like object or a number, not 'position'

```

Gouvernathor:
Roughly fixes #5162.
A jolly cocktail meant to be cherrypickable.

### DIFF
--- a/renpy/atl.py
+++ b/renpy/atl.py
@@ -1431,7 +1431,7 @@ class Interpolation(Statement):
             revdir = self.revolution
             circles = self.circles
 
-            if (revdir or ((has_angle or has_radius) and renpy.config.automatic_polar_motion)) and (newts.xaround is not None):
+            if revdir or ((has_angle or has_radius) and renpy.config.automatic_polar_motion) or has_anchorangle or has_anchorradius:
 
                 # Remove various irrelevant motions.
                 for i in [ 'xpos', 'ypos',
@@ -1445,8 +1445,8 @@ class Interpolation(Statement):
                 if revdir is not None:
 
                     # Ensure we rotate around the new point.
-                    trans.state.xaround = newts.xaround
-                    trans.state.yaround = newts.yaround
+                    trans.state.xaround = newts.xaround or .0
+                    trans.state.yaround = newts.yaround or .0
                     trans.state.xanchoraround = newts.xanchoraround
                     trans.state.yanchoraround = newts.yanchoraround
 
--- a/renpy/atl.py
+++ b/renpy/atl.py
@@ -1494,11 +1494,6 @@ class Interpolation(Statement):
                         startanchorangle_absolute += circles * 360
                         startanchorangle_relative += circles * 360
 
-                    has_radius = True
-                    has_angle = True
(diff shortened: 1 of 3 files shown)


mal:
Minor personal opinion, but I find that [`2e79eb4`](https://github.com/renpy/renpy/pull/5252/commits/2e79eb49cea17f000dffd3254475313d9db8471a) hurts readability more than it helps since the sole function of those methods is to act as getters for the x/ycenter property, and their relation to x/ypos is incidental, not reflective of the API.

renpytom:
I don't think removing the duplicate getters hurts readability - if anything, I think it reinforces "this isn't really storing this data."

Gouvernathor:
That's not exactly what mal was pointing at, but I think renaming the functions for what they do, rather than what they are used for, makes the code more consistent after that change.
But also, I just noticed that xalign and yalign don't have those getter/setter properties :thinking:

PR #5318, opened 2024-01-28
Lint : Check reachable labels with required parameters

### LINKED ISSUE #5166: Feature request: Debug: improve error message when entering label with parameters due to missing return in previous label
**Context**

**Example**

```rpy
    # return

label say_goodbye(duration):
    # return

    "[text]"
    pause duration
    return
```

**Proposal**

**Alternatives**

Gouvernathor:


mal:
It may not be pretty, but this is entirely legal and functional:
```rpy
label a(foo=1):
    "Hello [foo]"

label reentry(foo=1):
    "I'm conversing with [foo]."
    return

label alternate_greeting(foo=1):
    "Bonjour [foo]!"
    call reentry(foo)
    return
```

renpytom:
I'd say even if it's legal, it's probably more likely to be an error than not.

Gouvernathor:
My test code would be this :
```rpy
label entry:
label fine:
label finetoo():
label also(e=5):
label problem(f):
    pass
```
the `entry` label is not reachable by other means than jumping to it (assuming it's the start of the file). fine, finetoo and also have no required parameter, so passing them "from the top" is fine (even though weird in the case of `also`), and only `problem` is reported by lint.

mal:
Fair enough, I'm just always wary of lint accidentally becoming an opinionated "we don't think you want to do this" (weird label structuring) tool vs a "these are actual problems" (using attributes that don't exist) tool.

Gouvernathor:
I agree, I aimed for the second one.

Note that this is a prototype to illustrate how the base issue can be solved and help consideration of that issue, I don't strongly support adding that check or that particular implementation.

Gouvernathor:
The implem isn't workable due to changes in the lint implem

hsandt:
That's too bad we can't implement the alternative Lint fix for https://github.com/renpy/renpy/issues/5166, but I'm okay with my original proposal of simply improving the error message with a runtime check (may be easier to detect when flow is trying to enter a label from the label just above missing a `return`/`jump` as opposed to using `call`).

About developers who may want to exploit label leaking to next one, I was thinking about enabling some "Strict flow" mode for the Linter (see my 2nd post on the linked issue), however if we'd need a new working implementation to test it anyway.

PR #5319, opened 2024-01-29
Type consolidation and improvements

mal:
Recommend commit-by-commit review using commit messages as your guide.

Headlines:
- Moved `absolute`, `position`, and `DualAngle` types to new `renpy.types` module, along with various coercion functions.
- Optimised creation of `position` type negating need for `position.from_any`.
- Added support for `==` and `!=` operators to `position` to aid transform change detection code.
- Removal of `position_or_none` in favour of allowing `position` to gracefully handle `None` itself.
- Moved `DualAngle.from_any` to module level `dualangle` and optimise type checks.
- Moved `absolute.compute_raw` to module level, optimise type checks, and rename to `pixels`.

Commit messages have a little more detail, but there's anything you feel requires more explanation give me a shout. :)


renpytom:
My first reaction is that I really do not like the renaming of absolute.compute_raw to pixels. To me, the name "pixels" refers to an array of pixels, and not a position. 200709d63b062a62682487a12a279f448cf72d33 should go away - it seems more confusing to me. (Or perhaps become absolute with a second parameter.)

renpytom:
eb05edaf1f82571737f84691cd80c7b40fbc7653 - I like the idea of this, but I think it should probably be renpy.display.types, since it only has display-related types. 

aeb4586b5655f53b08f1e588bafdb2a8cf3cc948 - This seems wrong. If someone was to write: 

```
position(absolute=0.5)
```` 

They'd get position with relative set to 0.5, which is very confusing. 

eae1b6d9e59d2302bcfe5507f03a1e8cd91f42f1 - Fine.

16be0444082a4137fadf47e27c0c99cc42c7a016 - Fine.
 
cba2c2934f7f163a08e65494a619e4edfd63cadb - Fine. 

c15f39cb1235aa09bd0dbc0f3595c2af8f464223 - Seems unsafe, because of aeb4586b5655f53b08f1e588bafdb2a8cf3cc948. 

09490b7b5ec5c9dcaf6130cc08b812de40485946 - As we discussed on dev, having a type's constructor return an instance of a different type is very confusing. I like position_or_none, since it matches the way I think about positions.

8d82878632c6e734546a637e79f5a26e39cd1cc0 - Fine. 

200709d63b062a62682487a12a279f448cf72d33 - As discussed, I really don't like the name pixels. I do like absolute, though I'm not a huge fan of absolute.compute_raw as a name. 

I'm wondering if it makes sense to replace absolute.compute_raw with absolute.compute, and speed absolute.compute up so we can just used it everywhere. 

26551fe4476ec454d48181376ea292028132dfc1 - Fine. 

ba1bf446d378c840fde08c5c4a17d57b02cf0645 - Fine, but we may not want to do this everywhere, only in cases like this where it's really obvious what's absolute and what's relative.

mal:
> https://github.com/renpy/renpy/commit/aeb4586b5655f53b08f1e588bafdb2a8cf3cc948 - This seems wrong. If someone was to write:
> ```
> position(absolute=0.5)
> ```
> They'd get position with relative set to 0.5, which is very confusing.

This is addressed by making the first argument of `position` positional and thus required, eliminating this possibility. The public API docs remain unchanged, and still document that two positional parameters is the only supported signature.

> https://github.com/renpy/renpy/commit/c15f39cb1235aa09bd0dbc0f3595c2af8f464223 - Seems unsafe, because of https://github.com/renpy/renpy/commit/aeb4586b5655f53b08f1e588bafdb2a8cf3cc948.

Covered by the adjustment above.

> https://github.com/renpy/renpy/commit/09490b7b5ec5c9dcaf6130cc08b812de40485946 - As we discussed on dev, having a type's constructor return an instance of a different type is very confusing. I like position_or_none, since it matches the way I think about positions.

This has been replaced with https://github.com/renpy/renpy/pull/5319/commits/6a570b4abd51aad8ccf9f2b5eb476d1c84cab4ed, which seeks to simplify the ATL property type spec to what is functionally useful (see commit message for more details on why the `*_or_none` pseudo-types weren't really used/very useful.

> https://github.com/renpy/renpy/commit/200709d63b062a62682487a12a279f448cf72d33 - As discussed, I really don't like the name pixels. I do like absolute, though I'm not a huge fan of absolute.compute_raw as a name.
>
> I'm wondering if it makes sense to replace absolute.compute_raw with absolute.compute, and speed absolute.compute up so we can just used it everywhere.

This has been done, moving to and optimising `absolute.compute` instead.

> https://github.com/renpy/renpy/commit/ba1bf446d378c840fde08c5c4a17d57b02cf0645 - Fine, but we may not want to do this everywhere, only in cases like this where it's really obvious what's absolute and what's relative.

The calling sites affected by this use variable names making it clear which is which, so I think this is fine as long as we're conscious of the same concern when it comes to any future uses.

---

Additional changes include:

- https://github.com/renpy/renpy/pull/5319/commits/d82f9dab5c35133e3f2dd3034adb653c4fbf95f2
  Merged `DualAngle` and `dualangle` resulting in fewer imports, simplification of calling code that no longer needs to worry about which capitalisation variant is required, and mirrors the behaviour of the position type - such that groking one is groking both.

- https://github.com/renpy/renpy/pull/5319/commits/1e0b9bbdbba22ba7480808d718d1150a855ac433
  Properly support pickling of objects using `__new__` by adding `__getnewargs__`, and adding unpickling compat for 8.2 types.

---

I don't love the pickle compat, but after trying multiple approaches this was the one that worked, was the most sane, and didn't add confusion to other parts of the repo. I have the beginnings of an idea for general improvements to inter-version pickle compat, however they are out of scope for this PR.

mal:
This area of code has and is still being changed significantly, especially with the more recent typing push. There are a few ideas in here worth keeping, so I've pulled them out and refiled them in #6700, the remainder are mostly already done in other ways, or nothing things we wish or are worth pursuing.

Issue #3669, opened 2022-06-16
[Feature request] Fading slow_cps


Kuroonehalf:
I've seen this in a few visual novels, most commonly in ones made in Unity, where the implementation of slow_cps, instead of revealing one character at a time, fades them in progressively, as if it was an AlphaDissolve. I find that this has a very smooth polished look, and it'd be cool if Ren'py devs could have the option to use it.

Here's an example of how it looks: https://youtu.be/4eZqtiVMj3c?t=34
![CORPSE_FACTORY_-_Act_1_Playthrough_Part_1_ PC _4K__2022_06_16 (2)](https://user-images.githubusercontent.com/12227163/174123576-ae3dad25-6142-4431-b421-ae9fab0edd11.jpg)

I imagine you can probably hack together an actual AlphaDissolve animation for this, but it wouldn't be flexible. You could only make it work on something of fixed size. Ideally we'd have an implementation that would work on any piece of text like we do with our current slow_cps.

Gouvernathor:
Similar (although not identical) to #52

kyouryuukunn:
If any transitions can be used for each character, it would be good because it would expand the production that can be used for the characters.

kyouryuukunn:
This is available by FancyText.
https://github.com/yukinogatari/Ren-Py-FancyText

Issue #52, opened 2012-12-31
Please add letter-by-letter text fade


Jonathan-X:
The ability for Ren'py to fade in text for every letter as it is displayed is a highly requested feature that would improve the look and feel of games. Better yet, the ability to apply an ATL transition to each letter as it appears.

renpytom:
This has been implemented for a few versions.

PR #5329, opened 2024-02-01
add config.text_effects

kyouryuukunn:
This allow creators to customize text showing effect. This referred to FancyText.
 https://github.com/yukinogatari/Ren-Py-FancyText
Sample

Related issues
https://github.com/renpy/renpy/issues/52
https://github.com/renpy/renpy/issues/3669

I added one config and three text style properties.

.. var:: config.text_effects = { }

    A dictionary mapping text effect strings to a function.
    When :propref:`slow_effect` is not None, that strings are looked up in this
    dictionary. These functions takes three arguments, current time, time when
    the letter begins to be shown and :propref:`slow_effect_delay` then return
    a function that takes two arguments the letter's render and render the
    letter is drawn on.

    For example::

        init python:
            def letter_by_letter_fade(st, gt, duration):
                def _effect(surf, render):
                    render.absolute_blit(
                        surf,
                        (0, 0)
                    )

                    render.add_shader("renpy.alpha")
                    render.add_uniform("u_renpy_over", 1.0)
                    complete = (st - gt) / duration
                    render.add_uniform("u_renpy_alpha", complete)

                return _effect

        define config.text_effects = {
            ("letter_by_letter_fade", letter_by_letter_fade),
        }

    This fades one character at a time.

.. style-property:: slow_effect None or string

    This gives effect which is applied to a letter. If None, the default,
    Letters are displayed one by one like a typewriter. If not None, This
    should be the string giving one of the :var:`config.text_effects`.

    :propref:`slow_effect_delay` should be also set simultaneously.

.. style-properties:: slow_effect_delay float

   This is the duration seconds which :propref:`slow_effect` takes to perform.
   default is 0.0.

.. style-properties:: always_effect boolean

   If false, the default, :propref:`slow_effect` is only applied to slow texts
   for :propref: `slow_effect_delay` seconds after a letter is shown. Otherwise,
   it is applied always.


renpytom:
I like the idea behind this, but this is not an interface I want permanently in Ren'Py. Creating a render per character is just a bit insane. 

I'll revisit this, and look into seeing if we can do something with shaders, which should be much faster.

PR #5354, opened 2024-02-13
Turn language_tailor (in text/text.py) into a calling function

DipeshAggarwal:
`language_tailor` is mentioned in https://www.renpy.org/doc/html/changelog6.html#improved-japanese-support and in https://www.renpy.org/doc/html/text.html#non-english-languages but a doc entry for it's not generated, presumably because it's defined in a .pyx file.

Porviding a calling function named `language_tailor` in text.py should now properly generate the documentation associated with this function.

I consider this an exposed/external function because it is referenced in the documentation.


DipeshAggarwal:
I think this should target `master` instead of the `fix` branch. Closing this and will create one that correctly targets the `master` branch.

PR #5355, opened 2024-02-13
Stop sphinx from evaluating default argument values of function.

DipeshAggarwal:
Ensure that the default argument values of functions will be not evaluated on generating document.

Currently in https://www.renpy.org/doc/html/translation.html#renpy.translate_string the signature of the Default variable is shown, which for a newcomer, wouldn't make much sense.


renpytom:
This doesn't look like a good PR, given that it merges in master. You want to either make the change on fix, or make it on master and use git cherrypick to pull it into fix. 

Also, I don't think that this PR matters, as Ren'Py doesn't use autodoc.

PR #5356, opened 2024-02-13
 Turn language_tailor (in text/text.py) into a calling function

DipeshAggarwal:
`language_tailor` is documented in https://www.renpy.org/doc/html/changelog6.html#improved-japanese-support and in https://www.renpy.org/doc/html/text.html#non-english-languages but a doc entry for it's not generated, presumably because it's defined in pyx file.

Providing a calling function named `language_tailor` in text.py should now properly generate the documentation associated with this function.

I consider this an exposed/external function because it is referenced in the documentation.

### DIFF
--- a/renpy/text/text.py
+++ b/renpy/text/text.py
@@ -2578,7 +2578,25 @@ class Text(renpy.display.displayable.Displayable):
         return new_tokens, displayables
 
 
-language_tailor = textsupport.language_tailor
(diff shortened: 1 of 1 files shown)


renpytom:
Real problem, but this isn't the right solution for it. The problem was actually an issue with the documentation generator, which I just fixed. Thanks for pointing it out.

PR #5390, opened 2024-03-05
Update changelog.rst and template_projects.rst

DipeshAggarwal:
`changelog.rst`: Fix link to the vertical property and remove unncessary ` after renpy.fetch

`template_projects.rst`: config.build_name doesn't exist. Replaced it with build.name


DipeshAggarwal:
Oh and I have a question. Documentation that is autogenerated from the .py files, should I edit and push to `fix` or should I edit and push to `master/main`?

Gouvernathor:
If the thing to fix exists in `fix`, fix it there.

PR #5394, opened 2024-03-06
fix: do not apply what_suffix if dialogue will be extended

### LINKED ISSUE #5370: What_suffix should not be applied if dialogue will be extended.

JellyBlack:
fix #5370


Gouvernathor:
I disagree with the implementation (which does not treat prefix and suffix symmetrically), and also with the base idea but I will write that in the issue.

Jaybe-Games:
not compatible with current iteration of `characters.py`

PR #5401, opened 2024-03-08
Allow ren'py 8 to load .rpyc files using old-style screens.

CensoredUsername:
Currently ren'py 8 cannot even load .rpyc files that contain the compiled result of screen language version 1. I've validated this with several files from ren'py 6.17, which was the last version using old-style screens.

Ren'py 8 still has all the code needed to support these files, but as the ast module of python has changed between 2.7 and 3.9 it has become impossible to even load these files as the unpickling process would error due to unknown classes. This means nearly any ren'py 6.17 game or before cannot be ported over without a recompile.

This commit fixes this by intercepting the construction of these classes using pre-existing mechanisms in renpy.compat.pickle. When not needed this adds only a single equality check to the critical path of unpickling files.

I've tested this by running the 6.17 version of the tutorial in 8.2.1 (with some unrelated changes to other files to get it to run). It was able to now run it out of rpyc files without issues related to the screens.

### DIFF
--- a/renpy/compat/pickle.py
+++ b/renpy/compat/pickle.py
@@ -62,6 +62,7 @@ else:
 
     import functools
     import datetime
+    import ast
 
     def make_datetime(cls, *args, **kwargs):
         """
@@ -90,6 +91,9 @@ else:
                 elif name == "datetime":
                     return self.datetime
 
+            if module == "_ast" and name in REWRITE_NODES:
+                return REWRITE_NODES[name]
+
             return super().find_class(module, name)
 
     def load(f):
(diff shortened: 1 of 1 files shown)


renpy/compat/pickle.py
```diff
@@ -90,6 +91,11 @@ def find_class(self, module, name):
                 elif name == "datetime":
                     return self.datetime
 
+            if module == "_ast":
```
Gouvernathor, line 94:
```suggestion
            if module == "_ast" and name in REWRITE_NODES:
                return REWRITE_NODES[name]
```
Saving a function call can be important at pickle time.

Also, if this find_class is to grow this way, maybe we should make it rely on a `{modulename : {classname : wrapper}}` dict which would make the lookup faster than sequential equality tests.

CensoredUsername: >Saving a function call (two, even) can be important at pickle time.

Fair enough. I focussed more on the early bailout on "_ast" as most important perf item, as it should be extremely rare to enter the branch there anyway. I tried to avoid the double dict lookup in favour of the extra function call, but it seems I forgot this is python we're dealing with. That said, I don't really think it matters, as this isn't really the slowest part of the code if we hit the _ast branch.

If the amount of modules affected increases, it'll indeed be worth to move away from the sequential equality checks. I'd still keep to the `{modulename : {classname : wrapper}}` approach, as these bypasses should be rare enough that early bailout is almost always preferable.

The current approach trades an extra bytecode hit for a bit less indirection right now with the two equality checks. But regarding performance, the `_ast` path should only ever be when old-style screenlang support is hit. Nothing else in ren'py uses it, and as storing ast's is simply incredibly inefficient use of pickling to begin with, I doubt that will change. Storing source / bytecode is practically always preferable.

At which point we're already going to hit a significantly worse perf drawback with the use of the `NodeTransformer` that crawls the entire `_ast` node tree again, sadly that one was unavoidable. Normally you can coax the pickle machinery into doing some more fun things, but all ast nodes have an explicit `__reduce__` implementation that passes no arguments at construction time. Thus I could only get it to spit out a different node type unconditionally. And sadly `_ast.Name` needs to be split conditionally.

Booplicate: Does it make sense to drag this to renpy 8 when you generally have to rewrite quite a bit to migrate from such an old version as renpy 6? Is there really a case where you can't recompile? Especially if you have to refactor your code for newer python/renpy. I thought this module was more of a persistent/saves compatibility thingie for the end-users.

CensoredUsername: Beats me. The 8.0.0 release notes mention renpy 8 being used to run unmodified ren'py games from 2006, and old-style screen language is from 2014 so it felt weird to me when I discovered that it couldn't do this. What confused me mostly was that the engine had all the code ready to still support these games, except for that it wasn't reachable at all, as trying to load the rpyc file itself would fail due to some very complex interactions between python versions. I realized there was a way to work around that that wasn't too terrible so I had a go at it.

I honestly don't care much if it makes it in, but figured it'd at least be useful to offer.

renpytom:
I need to get my mind around how it works, but I'm leaning towards merging this in. (Or something similar that only kicks in in certain cases, like if it can detect an old file, or a file that won't load on the first try.)

CensoredUsername:
As far as I know, old style screen language was a thing up to and including 6.17, and during 6.99 the new rpyc v2 format was introduced. So it could conceivably only be enabled when a rpyc v1 file is detected. That'd require some changes in `renpy.script`, but I'm not quite sure which part of that file is public API. Alternatively, a just retrying if it fails could indeed be used.

The problem we're tackling here is that the format of the python ast has changed between 2.7 and 3.9. The following nodes were removed:

```python
ast.TryExcept
ast.TryFinally
ast.Exec
ast.Repr
ast.Num
ast.Str
ast.Print
```

And the following were changed:

```python
ast.arguments
ast.keyword
ast.Call
ast.Raise
ast.With
ast.Module
```

From my understanding of the old screenlang module, out of these it only used `ast.Module, ast.Call, ast.keyword, ast.Num and ast.Str` The rest of the nodes are (part of) unused statement nodes. (I just realized this isn't completely correct, technically user code that gets interpolated into it could use `ast.arguments` and `ast.Repr` nodes via lambdas and the backtick syntax. I'd say it is extremely rare that those would be used, but I can update that).

Furthermore some things were changed in the AST, namely that True/False/None aren't just ast.Name nodes, but they're keywords that get parsed to an ast.Constant.

When unpickling this without intervention, the unpickler would simply throw errors due to either these classes not existing in the current `ast` module, or later on `compile()` would error due to expected attributes not being set.

As for how it works: So in python 2, items in the `ast` module define a `__reduce__` implementation that instructs how they will be pickled. Said reduce implementation essentially tells the pickle machinery "to recreate me, just call my constructor without arguments, and then set the value of these attributes to this".

```python
>>> ast.Str("test").__reduce__()
(<class '_ast.Str'>, (), {'s': 'test'})
```

This results in the pickle machinery generating the following instructions (under protocol 2, as ren'py uses).

```python
>>> pickletools.dis(pickletools.optimize(pickle.dumps(ast.Str("test"), 2)))
    0: \x80 PROTO      2
    2: c    GLOBAL     '_ast Str' # this does stack.append(Unpickler.find_class("_ast", "Str")). default impl just does the equivalent of from _ast import Str
   12: )    EMPTY_TUPLE # puts an empty tuple on the pickle stack
   13: R    REDUCE # does args = stack.pop(), stack[-1] = stack[-1](*args), so basically calls _ast.Str() with no args
   14: }    EMPTY_DICT # construct an empty dict on the stack
   15: U    SHORT_BINSTRING 's' # puts a string on the stack
   18: U    SHORT_BINSTRING 'test' # idem
   24: s    SETITEM # does value = stack.pop(), key=stack.pop(), stack[-1][key] = value
   25: b    BUILD # does state = stack.pop(). Then either does stack[-1].__setstate__(state), or stack[-1].__dict__.update(state)
   26: .    STOP
```

The trick used here is in the `BUILD` opcode. If the object it's called on does not have a __setstate__ method, it'll just call `obj.__dict__.update(state)`. Which is what happens for most normal objects.

But what we do here, is use the subclassable `Unpickler.find_class` method to return an object that is a subclass of an `ast` node, with an added `__setstate__` method. By setting it it to be a subclass of the node it should be translated into, we solve the issue of the node type not existing, and as `compile` treats subclasses of ast nodes the same as the base nodes, it'll compile no problem. And then the `__setstate__` method takes care of translation between the source node's attributes.

Finally, when we edit the root `ast.Module` node, we end up running an `ast.NodeTransformer` over the generated ast tree to do any remaining transforms (in this case, changing `ast.Name` nodes containing `True/False/None` into `ast.Constant` nodes. We couldn't do this earlier as we had to control the node type at `REDUCE` time, while we only got the attributes at `BUILD` time.

CensoredUsername:
Updated: added additional compatibility for `ast.arguments`, `ast.Param` and `ast.Repr`, which can appear in user-supplied code that gets interpolated into the screens via either `lambda` expressions or ``` `repr` ``` expressions. Validated that this worked with some custom test cases.

mal:
Is it possible that this results in the wrapper classes making it into saves/regenerated rpyc files, because that seems like something we'd want to avoid if this goes forward, otherwise we'd potentially be setting ourselves up for more future pain.

Gouvernathor:
Maybe that risk could be averted by making the getstate/reduce of these new classes raise exceptions. And doing test runs on some of these old games to see what comes out of it.

CensoredUsername:
Good idea. There should be no chance of these node making it into save files, but it'd be a hassle if they made it into regenerated RPYC files.

Adding a `__reduce__` method that just references the superclass should do the trick. alternatively, giving them a matching name, and then setting their `__module__` attribute to "ast" causes the pre-existing `__reduce__` impl to do the work. That approach has my preference personally.

That way they'll be pickled completely the same as the ast nodes they're subclassing. And for the few classes that we do not replace, there's no danger of re-hitting the modifications, because we only intercept classes declared from the `_ast` module (which is what ast classes report they're from in Python 2), while in Python 3 ast classes say they're from the `ast` module.

CensoredUsername:
Updated it, ended up having to go for the `__reduce__` override method as pickle would complain about lying about the class' location otherwise.

renpy/compat/pickle.py
```diff
@@ -104,3 +108,182 @@ def dump(o, f, highest=False):
...
+    # screenlang support records raw python ast nodes into the rpyc data, making these
+    # impossible to load normally. This function can be called for node name `name`
+    # It will return None if the node type does not require any changes, or return
+    # a wrapper class with a custom __setstate__ implementation that will cause the pickle
+    # machinery to do the right thing. There are some things that cannot be handled by this
+    # (as the type of the node to be emitted is not fixed at that point), so those
+    # are handled by a NodeTransformer in the ast.Module replacement.
+    # Note: this isn't a complete python 2 -> python 3 ast conversion, we just convert
+    # what is needed to still support old-style screens (Ren'py 6.17 and below)
+    def ast_rewrite(name):
```
Gouvernathor, line 122: This function is obsolete, as it stands.

CensoredUsername: I've removed it.

renpytom:
Merged, thank you for the contribution.

PR #5423, opened 2024-03-16
Add to web.rst the details where to find the generated web files

minger0:


renpytom:
Merged and then cherry-picked to the fix branch, so it'll be up on renpy.org tomorrow. Thank you for the contribution.

PR #5427, opened 2024-03-21
add: `duration` parameter & allow filter to be set to `None` as can be done directly.

qTich:


renpytom:
This is fine, but in general it probably makes sense until something I'm working on is stable for a couple of days before pull requests. Just because there are things I haven't gotten to yet.

PR #5450, opened 2024-04-03
Better None pickling under py2

Gouvernathor:
Avoid taking a compatibility dependency to a function that useless

### DIFF
--- a/renpy/python.py
+++ b/renpy/python.py
@@ -1286,4 +1286,4 @@ copyreg.pickle(types.ModuleType, module_pickle)
 # Allow weakrefs to be pickled, with the reference being broken during
 # unpickling.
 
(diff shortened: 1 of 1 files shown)


renpytom:
I don't think this is in any way better than a dedicated function next to the code that uses it.

mal:
Hmm, I could see a slight advantage in that saves made in 8.2.2/7.7.2 could still be loaded in 8.2.1/7.7.1 and previous, whereas with a dedicated function not originally present they could not, and that in future when dropping py2 support there would be no need to retain this function when switching back to using NoneType.

renpytom:
@mal Loading saves in older versions isn't a concern, and a single function like this isn't a problem. 

The major issue with this is that it will print output, which is a no-go.

PR #5486, opened 2024-04-28
Update existing Italian translations and add missing ones

andrigamerita:
Hello again everyone, this time I decided I would contribute something to Ren'Py!
My initial idea was to work on translating both of the example novels included with the SDK, since I was surprised to see that they aren't available in my native language, Italian. However, after starting to work on them, I noticed how many strings from the engine were not properly translated: a few were missing here and there, some had a few inconsistencies or slightly bad word forms, and quite a big chunk were of really low quality due to being machine-translated. So, here I got to work:

* [x] Updating and improving the Italian translation for the engine template and launcher
    * [x] Checking and revising all translations marked as automatic (then unmarking them)
    * [x] Slightly revising the existing manual translations (correcting inconsistent capitalization and unnatural word forms)
    * [x] Creating translations for entirely missing strings
* ~[ ] Translating example and tutorial novels to Italian~ (will be moved to distinct PRs)
    ~* [ ] The Question~
    ~* [ ] Tutorial (this will probably take a while, but I think it's a good chance to study Ren'Py better)~

You might have noticed that I've sent this as a draft PR. What I've already checked on the list is effectively completed, but I thought of making a single pull request for everything since that's probably easier to manage. If that's okay with you, I will first complete everything on the to-do list, and then mark this request as completed, so that you can then merge; otherwise, I will split what is still missing at time of writing into separate PRs. If you have any comments about my contribution, or if you think there are things to fix in it, please let me know as soon as you see fit!

### DIFF
--- a/launcher/game/tl/italian/common.rpy
+++ b/launcher/game/tl/italian/common.rpy
@@ -183,7 +183,7 @@ translate italian strings:
 
     # 00gui.rpy:232
     old "Are you sure you want to return to the main menu?\nThis will lose unsaved progress."
-    new "Confermi di voler tornare al menù principale?\nQuesto causerà la perdita di progressi non salvati."
+    new "Confermi di voler tornare al menu principale?\nQuesto causerà la perdita di progressi non salvati."
 
     # 00gui.rpy:233
     old "Are you sure you want to end the replay?"
@@ -203,7 +203,7 @@ translate italian strings:
 
     # 00keymap.rpy:250
     old "Saved screenshot as %s."
-    new "Screenshot salvato come %s."
+    new "Cattura di schermaa salvata come %s."
 
     # 00library.rpy:142
     old "Self-voicing disabled."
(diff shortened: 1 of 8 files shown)


renpytom:
Thank you! 

You're right, newer strings were machine-translated, to keep the translation from falling completely out of date. So I appreciate the work. 

I'd suggest breaking it up into 3 PRs - engine/launcher, and then the two games. That's because people and projects will benefit immediately from the engine translation. There's no need to wait for the others to be finished.

andrigamerita:
> I'd suggest breaking it up into 3 PRs - engine/launcher, and then the two games. That's because people and projects will benefit immediately from the engine translation. There's no need to wait for the others to be finished.

Yeah, that makes sense. I''ll strikethrough the incomplete points on this list then, take the PR out of the draft status, and send those other two PRs when I've started and finished working on them!

PR #5488, opened 2024-04-29
Add root folder for Web archives

Kassy2048:
The current ZIP archive structure looks like that:
* game/
* game.zip
* index.html
* ...

This PR changes it that way:
* MyGame-1.0-web/
  * game/
  * game.zip
  * index.html
  * ...

The ZIP archive for other platforms always contains a single root folder where the game files are put, so this PR makes the format more consistent for Web. This also prevents the need to create a new folder manually where to extract the ZIP content on the Web server.

### DIFF
--- a/launcher/game/web.rpy
+++ b/launcher/game/web.rpy
@@ -482,8 +482,9 @@ init python:
                 zip_targets.append(os.path.join(dn, file))
 
         with zipfile.ZipFile(destination + ".zip", 'w') as zf:
+            parent_path = os.path.dirname(os.path.abspath(destination))
             for i, target in enumerate(zip_targets):
-                zf.write(target, os.path.relpath(target, destination))
+                zf.write(target, os.path.relpath(target, parent_path))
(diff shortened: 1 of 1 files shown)


mal:
When deploying a game to a server, I think it's much more typical to unzip directly into a location corresponding to the URL the game should eventually have. Unzipping into a named and versioned game folder, from which the entire contents then need to be moved for it to be correctly placed correctly in a server's directory hierarchy doesn't seem particularly advantageous in my opinion. It becomes another hoop to jump through for a would-be human deployer or automated deployment script.

renpytom:
I concur with @mal on this one, and I think that itch.io requires the current format, or at least did in the past. So I don't see a great advantage to changing.

Kassy2048:
On my server, I generally extract the games in a common folder then create a symlink into the public server hierarchy where I want the game to appear. Also, this is rather uncommon to create an archive of a folder from command line without having the folder as its root (e.g., you tar the folder directly not its content).

But if this is an itch.io requirement, I agree this is better to keep it that way.

Issue #5510, opened 2024-05-14
Error when issuing permission android


N0Fanru:
I tried to issue permission for android:

```
label splashscreen:
    if renpy.variant("android"):
        if renpy.request_permission("android.permission.MANAGE_EXTERNAL_STORAGE"):
            pause 0.01
    return
```

I used this before and it worked, but now when I build the apk file and run it on Android or an Android emulator, the following error appears:
```
I'm sorry, but an uncaught exception occurred.

While running game code:
  File "game/script.rpy", line 30, in <module>
AttributeError: 'NoneType' object has no attribute 'SDL_AndroidRequestPermission'

-- Full Traceback ------------------------------------------------------------

Full traceback:
  File "00start.rpyc", line 241, in script call
  File "script.rpyc", line 30, in script
  File "renpy/ast.py", line 1842, in execute
  File "renpy/python.py", line 1209, in py_eval
  File "renpy/python.py", line 1202, in py_eval_bytecode
  File "game/script.rpy", line 30, in <module>
  File "renpy/exports.py", line 4666, in request_permission
AttributeError: 'NoneType' object has no attribute 'SDL_AndroidRequestPermission'
```

Now I'm using Ren'Py 8.2.1

I also tried using it in a new project to make sure nothing was causing conflicts

renpytom:
Fixed by @OleSTEEP .

PR #5519, opened 2024-05-16
Fix permission requesting on Android platform

OleSTEEP:
On android platform renpy.request_permission produce an exception: 
`AttributeError: 'NoneType' object has no attribute 'SDL_AndroidRequestPermission'`

See: #5510

It happens because `renpy.exports.get_sdl_dll()` function return None on android platform. Function tries to find a native `librenpython.so` in `os.path.dirname(sys.executable) + "/"`, but this path on Android is point to `/data/user/0/{package.name}/files`, which does not contain this file. 

So I added just `librenpython.so` in libs list, just like in the old versions where there is no this problem and function started to return what it had to.

### DIFF
--- a/renpy/exports.py
+++ b/renpy/exports.py
@@ -4531,7 +4531,7 @@ def get_sdl_dll():
 
         lib = os.path.dirname(sys.executable) + "/"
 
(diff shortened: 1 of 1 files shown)


renpytom:
Merged, thank you.

PR #5531, opened 2024-05-21
call `renpy.const` for stores marked with `_constant = True`

Elckarow:
since constant stores aren't supposed to change, let's use that for screen optimization


mal:
At least as I understand it, calling `renpy.const` with the module names wouldn't be beneficial anyway as the whole name (as it would appear in screens) needs to be declared. i.e. `renpy.global('renpy.foo')` isn't going to result in `renpy.foo.bar` being treated as `const`.

Elckarow:
> `renpy.const('renpy.foo')` isn't going to result in `renpy.foo.bar` being treated as const.
it does. i did do some tests before opening the pr

renpytom:
I don't think this will work with how Ren'Py defines _constant, so I'm closing this PR.

PR #5558, opened 2024-06-08
fix: live2d fadeout

qTich:
In the previous case, we will proceed from `0.0` to `1.0`

### DIFF
--- a/renpy/gl2/live2dmotion.py
+++ b/renpy/gl2/live2dmotion.py
@@ -210,7 +210,7 @@ class Motion(object):
                 factor = min(factor, st / fadein)
 
             if st > self.duration - fadeout:
-                factor = min(factor, 1.0 - (st - (self.duration - fadeout)) / fadeout)
+                factor = min(factor, 1.0 - (st - self.duration) / fadeout)
 
             if fade_st is not None:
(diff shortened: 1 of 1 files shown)


renpytom:
I don't think this is correct, after a couple of tries looking at it. Plotting the old and new functions out, using matplotlib, I see: 

image

Just intuitively, (st - self.duration) will always be negative, and so subtractintg that from 1.0 will always be bigger than factor - assuming factor is between 0.0 and 1.0. 

Am I missing something? 

This does proceed from 


```python
import matplotlib.pyplot as plt

duration = 100
fadeout = 10

x = list(range(duration))
y1 = [ ]
y2 = [ ]
for st in x:
    y1.append(min(1.0, 1.0 - (st - (duration - fadeout)) / fadeout))
    y2.append(min(1.0, 1.0 - (st - duration) / fadeout))

plt.plot(x, y1, label="old")
plt.plot(x, y2, label="new")
plt.legend(loc='lower left')
plt.show()
```

qTich:
I made a mistake. 

I didn't notice that the project used a wrapper over `Live2D` that ignored `fadein` and `fadeout` in `Motion`.

PR #5563, opened 2024-06-12
Allow mousewheel to paginate save/load screen

luejerry:
Allow use of mousewheel to navigate between pages in the default save/load screen.

Not sure if there is a deliberate reason for omitting it, but as a player it has always bothered me that I couldn't scroll through my savegames using the mousewheel.


renpytom:
Merged, thank you.

PR #5605, opened 2024-06-30
Be more lenient with '@' in image names

levicratic:
Currently, the '@' character in image names must be placed at the end, followed by a float value. It will raise an exception otherwise.

I'd like to be able to do the following:
- Name a directory `xyz@2` and have all the images inside it be oversampled
- ~~Name an asset or directory of assets after the handle of an artist who created them without throwing an error~~

Notes:
- The most specific specifier should win, so `assets@2/image@2.5.png` should have an oversample of 2.5.
- ~~I realize naming an asset after an artist's handle is a niche request, so I understand if you'd rather not have this to prevent confusion.~~

***

Examples of old behavior, which work the same / are unchanged:
```
images/assets/image.png -> no oversampling
images/assets/image@2.png -> oversampling of 2, and so on
images/assets/image@3,2.png -> 3 (I'm not sure what the purpose of allowing commas is)
```

Examples of new behavior, which previously threw exceptions:
```
images/assets@2/image.png -> 2
images/assets@2/image@2.5.png -> 2.5
images/assets@2/image@3,2.png -> 3
```

### DIFF
--- a/renpy/display/im.py
+++ b/renpy/display/im.py
@@ -690,7 +690,7 @@ class Image(ImageBase):
 
         if "@" in filename:
             base = filename.rpartition(".")[0]
-            extras = base.partition("@")[2].split(",")
+            extras = base.rpartition("@")[2].partition("/")[0].split(",")
 
             for i in extras:
                 try:


renpytom:
I'd rather not have the @artist syntax - that doesn't feel necessary to me.  But I do like the rest of it. 

The reason for allowing commas is that we could add other types of information after the @, if required.

Gouvernathor:
If I understand your code correctly, only the top-most directory name is taken into account. If I have `images/characters/elena@5` I think it should work just as well.
Also (and kind of in the same vein) I think the various factors found along the path should be multiplied, instead of having the right one replace the left one(s). It makes more sense to me at least, but maybe I'm mistaken.

levicratic:
> If I understand your code correctly, only the top-most directory name is taken into account.

`base.rpartition("@")[2].partition("/")[0]` grabs the right-most value, then cuts off any trailing `/path...`, if present:
```
images/characters/elena@5 -> 5
images/characters@4/elena@5 -> 5
images@3/characters@4/elena@5 -> 5
images@3/characters@4/elena -> 4
images@3/characters/elena -> 3
```
Perhaps it deserves a comment or being split into two lines.

> I think the various factors found along the path should be multiplied, instead of having the right one replace the left one(s).

If multiplying values is more intuitive, something like this might be in order:
```python
    ...
    # e.g. base = "images/bg@2,3/lair@1.5,2"
    extras = [i.partition("/")[0] for i in base.split("@")[1:]]

    oversample = 1
    for i in extras: # extras = ["2,3", "1.5,2"]
        try:
            oversample *= float(i.split(",")[0]) # 2 * 1.5 = 3
        except Exception:
            raise Exception("Unknown image modifier %r in %r." % (i, filename))
    properties.setdefault('oversample', oversample)
```
However, if I have a directory called `images/bg@2`, and I add a new image that needs an oversample of 2.5, it's not immediately obvious to me that I have to name it `images/bg@2/lair@1.25.png`. You could allow things like `@*2` with:
```python
    ...
    for i in extras:
        arg = i.partition(",")[0]
        try:
            if arg.startswith("*"):
                oversample *= float(arg.lstrip("*"))
            else:
                oversample = float(arg)
        ...
```
So that for `images/bg@2/lair@2.5.png`, 2.5 would overwrite the 2, but for `images/bg@2/lair@*1.25.png`, they would be multiplied.

renpytom:
I'm fine with the current version, thanks for the contribution.

PR #5626, opened 2024-07-10
Fix video playing if config.web_video_base URL is cross-origin

agejevasv:
When `config.web_video_base` is set with an absolute URL to another origin, modern browsers won't load the video with an error:

> The HTMLMediaElement passed to createMediaElementSource has a cross-origin resource, the node will output silence.

For CORS to work:
1. The server serving the resources should have a `access-control-allow-origin: *` header.
2. The `video` element should have `crossorigin="anonymous"` [attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin).

This PR adds such an attribute to the `video`.


renpytom:
Thanks for the fix.

PR #5654, opened 2024-07-27
store the total size for grids

Elckarow:
and vpgrids

the point of this is for creators making cdds not have to cv paste the entire `render` method to get the full size.

not sure where it should be documented (or if it should be documented at all) so im keeping this as a draft for now.

### DIFF
--- a/renpy/display/layout.py
+++ b/renpy/display/layout.py
@@ -435,6 +435,9 @@ class Grid(Container):
 
     allow_underfull = None
 
+    total_width = 0
+    total_height = 0
+
     def __init__(self, cols, rows, padding=None,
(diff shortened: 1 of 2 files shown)


renpytom:
What do you see as the use of this? CDDs shouldn't be using the internals of Ren'Py-provided displayables.

Elckarow:
for vpgrids? you could use this to render a background behind the grid, one that follows the adjustment values
image

for grids? no idea, but i figured i'd also add it for them

renpytom:
This would make game  code dependent on the implementation of a displayable, and that's not something I want. So I'm rejecting this.

PR #5676, opened 2024-08-06
Fully compat transform pos-only parameters (at the parser level)

Gouvernathor:
This requires the atl_pos_only config to be set very early in order to work. It's probably not actually the case, since the 00compat code does not execute in an early block. So it may require to either add an early treatment for that config variable in the 00compat file, or to directly check the early_script_version config variable in the parser, which is dirty.


mal:
Isn't conditional parsing a horrible foot-gun because changing the config variable isn't going to (nor should it) force recompilation?

Gouvernathor:
Right.

Issue #5682, opened 2024-08-08
Relative Python module import errors with Python 3.12


onitake:
The Debian package of Ren'Py executes in the system's default Python interpreter, which is Python 3.12 in Debian trixie and sid at this time.

With Python 3.12, Ren'Py fails to start due to ModuleImportErrors caused by relative imports in .rpy code:
```
    File "launcher/game/gui7.rpy", line 31, in <module>
      import gui7
 ModuleNotFoundError: No module named 'gui7'
```

When logging `sys.path` just before the import statement, it is as follows: 
```python
sys.path=['/usr/share/games/renpy', '/usr/share/games/renpy', '/usr/lib/python312.zip', '/usr/lib/python3.12', '/usr/lib/python3.12/lib-dynload', '/usr/local/lib/python3.12/dist-packages', '/usr/lib/python3/dist-packages', '/usr/lib/python3.12/dist-packages', '/usr/share/games/renpy']
```

The Python code is definitely installed, and the issue can be fixed by changing all submodule-relative imports to installation-relative ones.

For example:
```python
    #import gui7
    from launcher.game import gui7
    #from gui7 import translate_define, translate_copy, translate_code
    from launcher.game.gui7 import translate_define, translate_copy, translate_code
```

Debian bug: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1077202

lordmatt:
I'm also seeing this error

PR #5686, opened 2024-08-08
Support find_spec in custom module loader

onitake:
This is a quick and possibly incomplete fix for relative import errors from .rpy scripts to .py scripts in the same module folder.

I only replaced those imports that were causing startup problems for the launcher.
Loading an existing project and launching it worked without problem, however.

Fixes: #5682

### DIFF
--- a/launcher/game/distribute.rpy
+++ b/launcher/game/distribute.rpy
@@ -35,7 +35,7 @@ init python in distribute:
     import store.updater as updater
     import store as store
 
-    from change_icon import change_icons
+    from launcher.game.change_icon import change_icons
 
     import sys
     import os
--- a/launcher/game/distribute.rpy
+++ b/launcher/game/distribute.rpy
@@ -35,7 +35,7 @@ init python in distribute:
     import store.updater as updater
     import store as store
 
-    from launcher.game.change_icon import change_icons
+    from change_icon import change_icons
 
     import sys
     import os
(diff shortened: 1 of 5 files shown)


renpytom:
This isn't the right answer, as these shouldn't be imported through the Python import mechanism, but should be imported using RenpyImporter in loader.py. If the files aren't loading, something is going wrong there.

onitake:
Ok, I'll see if I can hunt down the root of the problem...

onitake:
I might have found the culprit: https://docs.python.org/3/library/sys.html#sys.meta_path
> **Changed in version 3.12:** Removed the fallback that looked for a find_module() method if a [meta_path](https://docs.python.org/3/library/sys.html#sys.meta_path) entry didn’t have a [find_spec()](https://docs.python.org/3/library/importlib.html#importlib.abc.MetaPathFinder.find_spec) method.

onitake:
All right... here's a 2nd attempt at fixing this.
I added a `find_spec` method, which is actually supported since Python 3.4.
This is a bit more complicated, because it requires you to return a spec instead of a loader. Wrapping the loader in a `spec_from_loader()` seems to do the trick.

renpytom:
I don't think can go in until 8.3 is out, as we still need to suppoort Python 2. I'll add this early in 8.4.

Issue #5681, opened 2024-08-08
Compile errors with FFmpeg 7


onitake:
Some APIs were changed in FFmpeg 7, which leads to Ren'Py throwing build errors:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1077537

The (autogenerated) list of API changes can be found here: https://git.videolan.org/?p=ffmpeg.git;a=blob_plain;f=doc/APIchanges;hb=n7.0
Unfortunately it doesn't clearly list what was removed, but previous deprecations may help fix the issues.

Some examples:
```
ffmedia.c:108:9: error: passing argument 6 of ‘avio_alloc_context’ from incompatible pointer type [-Wincompatible-pointer-types]
/usr/include/x86_64-linux-gnu/libavformat/avio.h:404:25: note: expected ‘int (*)(void *, const uint8_t *, int)’ {aka ‘int (*)(void *, const unsigned char *, int)’} but argument is of type ‘int (*)(void *, uint8_t *, int)’ {aka ‘int (*)(void *, unsigned char *, int)’}
```

```
ffmedia.c:693:30: error: ‘AVFrame’ has no member named ‘channel_layout’; did you mean ‘ch_layout’?
  693 |             converted_frame->channel_layout = AV_CH_LAYOUT_STEREO;
```

```
ffmedia.c:697:74: error: implicit declaration of function ‘av_get_default_channel_layout’ [-Wimplicit-function-declaration]
  697 |                                 ms->audio_decode_frame->channel_layout = av_get_default_channel_layout(ms->audio_decode_frame->channels);
ffmedia.c:697:126: error: ‘AVFrame’ has no member named ‘channels’
  697 |                                 ms->audio_decode_frame->channel_layout = av_get_default_channel_layout(ms->audio_decode_frame->channels);
```

PR #5687, opened 2024-08-08
Replace deprecated APIs when compiling against newer FFmpeg

onitake:
This PR makes Ren'Py compatible with newer FFmpeg versions. A lot of deprecated APIs were removed in FFmpeg 7, leading to compile errors.
Old API compatibility is kept by means of compile-time version checks.

Fixes: #5681


renpytom:
T?his needs to wait for us to drop Windows 7 support, which will happen in the 8.4 release cycle.

PR #5701, opened 2024-08-12
doc: typo correction

Neyunse:
chile -> child


Gouvernathor:
That's not a typo, it's to show what behavior differs between having a parameter named or not named "child". 🇨🇱

PR #5709, opened 2024-08-16
Update options.rpy in ukrainian

Veydzher:
Changed strings.


Veydzher:
ah, I can't understand why it comes with 3 additional commits I submited before. Is it must to be like that?

Andykl:
It's likely that you have not rebased your master branch after last pull request so git automatically mereged current state to your local master branch. It does not matter for the merge.

Veydzher:
Would 'Sync Fork' feature fix it?

Andykl:
Yeah, I think before you add new commits to your master branch you need to sync fork.

PR #5722, opened 2024-08-21
Fix type error when style attributes are None

k-blo:
# Fix type error

`scale_int` may return `None` but `textsupport.place_vertical` expects `int` type. This PR ensures we pass `int=0` to `textsupport.place_vertical` instead of None.

## how to reproduce the error

Have a style with `ruby_line_leading = None`, which is allowed.

When starting the game, this will result in:

```File "renpy-8.2.3-sdk/renpy/text/text.py", line 766, in __init__
    l, y = textsupport.place_vertical(par_glyphs, y, line_spacing_scaled, line_leading_scaled, ruby_line_leading_scaled)
  File "textsupport.pyx", line 677, in renpy.text.textsupport.place_vertical
TypeError: an integer is required
```

### DIFF
--- a/renpy/text/text.py
+++ b/renpy/text/text.py
@@ -827,7 +827,13 @@ class Layout(object):
 
             # Figure out the line height, line spacing, and the y coordinate of each
             # glyph.
-            l, y = textsupport.place_vertical(par_glyphs, y, self.scale_int(style.line_spacing), self.scale_int(style.line_leading), self.scale_int(style.ruby_line_leading))
+            l, y = textsupport.place_vertical(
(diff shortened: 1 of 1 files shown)


Andykl:
According to [documentation](https://renpy.org/doc/html/style_properties.html#style-property-ruby_line_leading) `ruby_line_leading` is `int`, so I am not sure it is allowed.

renpytom:
The ruby_line_leading is set to 0 in the default style, so the only way to make this None is to assign None to it. I don't think this warrants a change.

PR #5728, opened 2024-08-24
Fix syncfs call while a rename is in progress

### LINKED ISSUE #5725: `Warning: write error: FS error` happening on autosave in browser builds on Ren'Py 8.3.0
Upon hitting a choice, quitting to main menu, or prompting an autosave in any other way, games built in Ren'Py 8.3.0 targeting the web will provide a generic "FS error". 
Screenshot 2024-08-23 at 3 05 06 AM (1)

Attached is the [log.txt](https://github.com/user-attachments/files/16724985/log.5.txt) from this game. 

The example I used for this screenshot is unchanged from generated default, minus the inclusion of a background and Eileen. The `script.rpy` for the example is below:

```renpy
define e = Character("Eileen")

label start:
    scene bg room
    show eileen happy

    e "You've created a new Ren'Py game."
    e "Once you add a story, pictures, and music, you can release it to the world!"
    e "Got that?"

    menu: 
        e "Got that?{fast}{done}"

        "Yep! I got it!":
            e "Great to hear!"
        "No... could you explain from the start?":
            e "Alrighty then..."
            jump start 
    # This ends the game.

    return
```

A live example of a game with this issue is my current project, [Pesterquest Rewritten](https://pesterquest.com/).

Kassy2048:
`FS.syncfs()` is asynchronous, so if the FS changes while this function hasn't complete, it can fail. This is what happens in #5725: each call to `save()`, `cycle_saves()` and `rename_save()` trigger multiple calls to `FS.syncfs()`, so when the final renaming happens, the FS state has changed since the previous calls to `FS.syncfs()`, so they fail and the error messages show up.

Calling `FS.syncfs()` is expensive, so it should only be called once at the end of `autosave_thread_function()` anyway. So this is what this commit does.
That also removes the apparition of the "warning: X FS.syncfs operations in flight at once, probably just doing extra work" messages for autosaves. They still show up for quicksaves though, but that's probably a smaller problem.

Fix #5725.

### DIFF
--- a/renpy/loadsave.py
+++ b/renpy/loadsave.py
@@ -377,7 +377,7 @@ class SaveRecord(object):
         self.first_filename = filename
 
 
-def save(slotname, extra_info='', mutate_flag=False, include_screenshot=True):
+def save(slotname, extra_info='', mutate_flag=False, include_screenshot=True, sync=True):
     """
     :doc: loadsave
     :args: (filename, extra_info='')
@@ -457,7 +457,7 @@ def save(slotname, extra_info='', mutate_flag=False, include_screenshot=True):
     json = json_dumps(json)
 
     sr = SaveRecord(screenshot, extra_info, json, logf.getvalue())
-    location.save(slotname, sr)
+    location.save(slotname, sr, sync=sync)
 
     location.scan()
     clear_slot(slotname)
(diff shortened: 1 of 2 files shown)


renpytom:
I took this idea and implemented it in a different way, in 3b1161c2a4e6396f2602c3199d8b8917a7081a4c.

Kassy2048:
Makes sense, the less code changes the better 👍

PR #5729, opened 2024-08-24
Added type annotations

abstractee:
Added type annotations to functions for consistency

### DIFF
--- a/renpy.py
+++ b/renpy.py
@@ -32,7 +32,7 @@ import warnings
 
 # Functions to be customized by distributors. ################################
 
-def path_to_gamedir(basedir, name):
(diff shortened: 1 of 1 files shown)


renpytom:
My feeling is that this should probably wait for a month or so, at which point changes to Ren'Py 7.8 will slow down enough we can start adding Python 3 style type annotations to Ren'Py.

abstractee:
I just saw that a couple of functions in the renpy.py file already had them and thought of including the others. But waiting to include the entire engine may be for the better.

Gouvernathor:
These annotations are done through comments, not through the new specific annotation syntax.

PR #5733, opened 2024-08-25
[Web] Only call syncfs once from MultiLocation functions

Kassy2048:
With Ren'Py default config, multiple `MultiLocation` functions call a related function on 2 different instances of `FileLocation` (one for "savedir" and one for "sync"). Some of these `FileLocation`'s functions call `synfc()`, so the warning message "2 FS.syncfs operations in flight at once, probably just doing extra work" shows up. This commit prevents that.

My main goal was to fix the problem for `MultiLocation.save_persistent()`, but since it appears for other functions, it probably does not harm to fix these as well (using a context for concision).

On a side note, emscripten 3.1.61 (05/31/24) introduces [an auto mode to persist the FS data](https://github.com/emscripten-core/emscripten/pull/21938) so that it is not needed anymore to call `syncfs()`.

### DIFF
--- a/renpy/common/00action_file.rpy
+++ b/renpy/common/00action_file.rpy
@@ -408,16 +408,11 @@ init -1500 python:
                     layout.yesno_screen(layout.OVERWRITE_SAVE, FileSave(self.name, False, False, self.page, cycle=self.cycle, slot=self.slot, action=self.action))
                     return
 
-            try:
-                renpy.savelocation.pause_syncfs()
-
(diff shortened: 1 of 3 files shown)


renpytom:
I like this, about to merge. 

If you could target PRs like this to the fix branch, I'd appreciate it. Makes my life easier that way.

Kassy2048:
I hesitated to target the fix branch actually, but didn't dare in the end😅 I will next time.

PR #5734, opened 2024-08-26
set and show the various skipping states

the66F95:
i simply don't understand what's wrong with the ability to change slow to fast skipping and vice versa, hence my changes. however to get the default `get_selected` method of `class Skip()` to work also for fast skipping i had to add the alternate action to the `is_selected` method of `class Button()` (also added to `is_sensitive`). i'm not sure if problems may arise because of it, but shouldn't a valid alternate action also lead to a selected / sensitive state? if that's not desired a simple `return config.skipping` in the `get_selected` method of `class Skip()` should also work.

### DIFF
--- a/gui/game/screens.rpy
+++ b/gui/game/screens.rpy
@@ -1216,11 +1216,15 @@ screen skip_indicator():
         hbox:
             spacing gui.scale(6)
 
-            text _("Skipping")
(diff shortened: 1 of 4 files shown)


renpytom:
Fast skipping shouldn't cause interactions, and so this seems useless.

PR #5740, opened 2024-08-27
Skip seen text based on translation identifiers (#5628)

### LINKED ISSUE #5628: Skip seen text based on translation identifiers
Hi,

Some games get updates with regenerated/reworked .rpyc files, which currently breaks seen texts (due to changed internal statements identifiers, which include a random part). This in turns breaks the Skip feature, which makes it inconvenient to start a new game, replay old saves, or identify new content.

However, I notice that the F2 progress screen still correctly references the number of seen dialogues. When replaying previous savegames, the F2 progress screen will often correctly stay as 0 new seen texts, though those texts could not be skipped and had to be manually passed with space/enter (at which point they get skippable again). This F2 screen appears to rely on the text's translation identifier, which is not random and more stable/deterministic than the internal statement identifier.

So the F2 progress screen knows what texts were read and we could add the same logic to skip seen texts in a more robust way.

Would it be sensible to implement this?

Beuc:
- Skip node if the current translation_identifier was previously seen: this allows the player to skip through texts that were seen in a previous playthrough, even when the developer mistakenly deletes or regenerates .rpyc files while preparing a game update, which changes node identifiers and breaks the list of previously seen nodes.

  Limitation: only applies to translatable content, so not to e.g. image displays.

  This is still a solid QoL improvement for the player, who needn't remember each and every path they might have already gone through in the previous game update/"episode" months/years ago.

- Translations are now considered seen *after* we progress to the next node (as with seen nodes), not when we execute them; this avoids skipping all texts unconditionally in the above, and also de-duplicates the counting code.

- FIXME: ctx.translation_identifier is not reset so it can be used by the above; this probably should be fixed but currently allows discussing this first proof-of-concept.

Fixes #5628

What do you think of this approach? :)

### DIFF
--- a/renpy/ast.py
+++ b/renpy/ast.py
@@ -2475,11 +2475,6 @@ class Translate(Node):
             next_node(self.next)
             raise Exception("Translation nodes cannot be run directly.")
 
-        if self.identifier not in renpy.game.persistent._seen_translates: # type: ignore
-            renpy.game.persistent._seen_translates.add(self.identifier) # type: ignore
-            renpy.game.seen_translates_count += 1
-            renpy.game.new_translates_count += 1
-
         next_node(self.lookup())
 
         renpy.game.context().translate_identifier = self.identifier
@@ -2555,11 +2550,6 @@ class TranslateSay(Say):
 
         if self.language is None:
 
-            if self.identifier not in renpy.game.persistent._seen_translates: # type: ignore
-                renpy.game.persistent._seen_translates.add(self.identifier) # type: ignore
-                renpy.game.seen_translates_count += 1
-                renpy.game.new_translates_count += 1
-
             # Potentially, jump to a translation.
             node = self.lookup()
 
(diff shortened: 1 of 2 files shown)


mal:
Translations aren't 1-to-1 with "seen" lines though, right? Since the number of nodes, and by extension number of interactions in each block can differ?

Beuc:
> Translations aren't 1-to-1 with "seen" lines though, right? Since the number of nodes, and by extension number of interactions in each block can differ?

That is true but can you explain in what user scenario this relates to this issue?

Andykl:
Your fix will mark both Say statements as seen in case of
```
translate french id:
    "I am"
    extend "a non 1-to-1 translate block."
```
Which is fine in my opinion, I believe that most of the time people change languages to None to circumvent an exception in translation, or if they just mess with the game, so such change is okay in my book, but this is change in behavior that should be considered.

Issue #3239, opened 2022-01-18
Add Renpy CLI to docs


Geoffrotism:
Would love to see the renpy CLI documented on the documentation. Was trying to get some automated builds for our project and set up releases to integrate with github and couldnt find anything on how to create a renpy release except from the GUI.

PR #5749, opened 2024-08-29
CLI Documentation

arsym-dev:
Addresses #3239

Updated the command line documentation to match what's present in 8.4.0.

To make it easier to copy/paste commands for different operating systems, I've added [Sphinx Tabs](https://github.com/executablebooks/sphinx-tabs) as a requirement and created a dark theme for it. It can be installed with:
``` bash
pip install -U sphinx-tabs
```

NOTE: I wasn't able to test the `gui_images` and `director` commands myself, but the documentation should be accurate.


renpytom:
Can you please retarget this to the fix branch? Unless I'm missing something, I suspect we'd benefit from this there.

(And wow, this one fell between the cracks - apologies for that.)

PR #5751, opened 2024-08-29
Doc on how to write docs

arsym-dev:
Documentation demonstrating how to use docstrings, commonly used elements, and Ren'Py-specific directives and roles.

As I was improving the documentation, I found myself struggling to understand how to create references to the different indices (eg. style-property, screen-property, function, etc). Hopefully by having them all in one place, it's easier to find the element you need with an example and for new editors to get involved.

### DIFF
--- a/sphinx/source/documentation.rst
+++ b/sphinx/source/documentation.rst
@@ -1,0 +1,684 @@
+=============
+Documentation
+=============
+
+Ren'Py's documentation is built with `Sphinx <https://www.sphinx-doc.org>`_.
+Documentation is written in reStructuredText files found in :file:`sphinx/source`,
+and generated documentation found in function :ref:`docstrings <docstring>`
+scattered throughout the code.
+
+When editing the engine, it's important to keep the documentation
+up to date.
+
+This page covers the most common directives, roles, and fields used,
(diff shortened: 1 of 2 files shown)


renpytom:
I need to review this more, but there does need to be a prominent warning that documenting new things needs to be discussed. That's because the Ren'Py API needs to be something we support forever, while internals aren't that.

arsym-dev:
Needs to be discussed in what sense? Would it not be discussed within the PR by default? It may be that you're thinking of a contributor documenting completely new stuff that still doesn't have a stable API, whereas I was thinking of adding docs to existing code.

If this does get expanded to a proper guideline, it would be nice to include tips to:
- Consider end users of all skill levels
- Create examples that sufficiently demonstrate the point without over-complicating it

The intent with this was simply to be a quick reference, not a guideline. But it can definitely be changed to whatever is best for the project.

renpytom:
I was just thinking out loud here, as I can add the warning. I'll review this more properly once 8.3.1 is out, and probably write that disclaimer myself. My big worry is that I don't want people sending me PRs that document functions I don't want to document, as it would ossify Ren'Py's internals.

Issue #5350, opened 2024-02-11
add the abilty to use raw `ATLTransform`s-


Elckarow:
-instead of needing to apply it to a displayable

for use cases where the atltransform is the displayable itself (like a shader that draws something)

```
show expression transform:
    atl code

# and

fixed:
    add transform:
        atl code
```

instead of having to create a `Fixed` or `Solid` and applying the transform to it

PR #5768, opened 2024-09-06
Implement 'add image' screen language statement.

Andykl:
I tried different approaches several times, but all of them was too hacky or too big. 
This implementation still steal some logic it probably should not.

This adds
```
add image [as var]:
    atl block
```

### DIFF
--- a/renpy/sl2/slast.py
+++ b/renpy/sl2/slast.py
@@ -1377,6 +1377,106 @@ class SLDisplayable(SLBlock):
             i.dump_const(prefix + "  ")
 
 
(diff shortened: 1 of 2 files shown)


renpytom:
How is this different than "add <x> at transform:" ?

Andykl:
Same reasons as for anonymous transforms. I would say 5% of my images are defined near the screen and used only inside one screen.

Elckarow:
> How is this different than "add at transform:" ?

it's pretty much what i described in #5350

PR #5772, opened 2024-09-09
no seen labels, images, audio in replays

the66F95:
i don't think labels, images and audio files should be marked as seen in replays. basically all other variables are local to that context.
for instance developer would be able now to temporarily unlock content of a replay gallery for higher tier patrons without actually changing the seen status of a label. to preserve the old behavior one can set the new config var to True.

### DIFF
--- a/renpy/audio/audio.py
+++ b/renpy/audio/audio.py
@@ -715,8 +715,9 @@ class Channel(object):
         with lock:
 
             for filename in filenames:
-                filename, _, _ = self.split_filename(filename, False)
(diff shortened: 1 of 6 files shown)


renpytom:
I think I disagree with this - if you've seen something, you've see something. I'm not a huge fan of the engine lying about that.

the66F95:
then how about preserving the old behavior as default but adding to option to exclude seen events in replays?

renpytom:
It's a clean enough change that I'll let it in as a default-off option. I'll merge this tonight when I rewrite some documentation.

PR #5775, opened 2024-09-09
Fix size tag preventing outer tags from closing

levicratic:
This removes the second `push()` when handling the size tag, which was preventing the outermost tag from properly closing. Examples of the behavior this should fix:

```rpy
label start:

    "{b}Bold, {size=*2}large and bold,{/size}{/b} still bold."

    "{i}{u}{b}Styled, {size=*2}large and styled{/size}{/b}{/u}{/i}, still italic."

    "{size=*1.5}1.5x large, {size=*2}3x large,{/size}{/size} still 1.5x large."
```

### DIFF
--- a/renpy/text/text.py
+++ b/renpy/text/text.py
@@ -1406,7 +1406,8 @@ class Layout(object):
                     if value[0] in "+-":
                         push().size += int(value)
                     elif value[0] == "*":
-                        push().size = int(float(value[1:]) * push().size)
+                        ts = push()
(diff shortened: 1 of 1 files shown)


Andykl:
Looks good, but I am not sure how to merge it to `fix` branch. This should also be in fix releases I believe.

Gouvernathor:
The ideal would be a cherry-pick of the single commit on top of the fix branch, but it can be done after a merge to master, supposing that code area is not very different between fix and master.

PR #5809, opened 2024-09-25
fix bug that adds 1 second to each successive parallel build

sunrisesarsaparilla:
When experimenting with parallelizing all game builds performed via the GUI, every build took +1 second longer than the previous one, until RenPy was restarted. This change fixes that bug.


renpytom:
Merged, thank you.

PR #5817, opened 2024-10-01
add: pass `value` into `released` action of `Bar`

qTich:


renpytom:
This is an incompatible change.

Gouvernathor:
I'm not sure there's a precedent for actions being called with arguments (not talking about the action type or constructor, but the action itself).

qTich:
> I'm not sure there's a precedent for actions being called with arguments (not talking about the action type or constructor, but the action itself). The solution is probably to look for the value using the bar's id or something, that's what I would do in JS/web anyway.

`PR` has nothing to do with the WEB.

The `changed` method is called with a parameter, but it is not quite possible to determine where exactly the parameter was changed from (from the user or inside the `periodic` method, for example).
The `released` method guarantees that the value was exactly passed from the user when the user pressed it.

renpytom:
I don't like the idea of an action taking an argument. 

What's the point of this? I can sort of see some use, but this probably belongs in the BarValue api, not changing the release action.

PR #5837, opened 2024-10-11
android: Install-time asset pack support.

### LINKED ISSUE #5010: Android app hangs indefinitely at launch after an install or update
Several google play reviews mention an issue in which the game stays in the loading screen and never actually launches. The loading bar never fills up and remains empty.

In the few times I was able to reproduce the issue, the workaround I found was deleting Google Play application's cache. When the game launches again, it actually installs additional app packages (my app is a bit less then 300 MBs) and then the load bar fills up and the game does load properly.

My latest update was rejected by Google Play's system with the following reason:

```
Issue found: Violation of Broken Functionality policy
Your app contains content that isn’t compliant with the Broken Functionality policy.
- App installs, but doesn't load
```

I wasn't able to get the app freeze this time, but I'm adding the logcat in case it provides some information: 
[https://drive.google.com/file/d/1HQGth0aCo1ROd1yAzYyRgA-gIgK2GMMr/view?usp=drive_link](https://drive.google.com/file/d/1HQGth0aCo1ROd1yAzYyRgA-gIgK2GMMr/view?usp=drive_link)

avonder:
Install-time asset packs are served as APKs and can be processed by existing APK routines.
This satisfies Google's expectation of games being in a playable state without any fast-follow packs present.

To be applied with  https://github.com/renpy/renpy-build/pull/141.

Fixes #5010.

### DIFF
--- a/renpy/loader.py
+++ b/renpy/loader.py
@@ -73,17 +73,27 @@ def get_path(fn):
 
 apks = [ ]
 game_apks = [ ]
+split_apks = [ ]
 
 if renpy.android:
     import android.apk # type: ignore
 
+    packs = [os.environ[i] for i in ["ANDROID_PACK_FF" + str(j+1) for j in range(4)] if i in os.environ and os.environ[i].endswith(".apk")]
+
     if renpy.config.renpy_base == renpy.config.basedir:
-        # Read the game data from the APK.
+        # Read the game data from the APKs.
 
         apks.append(android.apk.APK(prefix='assets/x-game/'))
-        game_apks.append(apks[0])
+        game_apks.append(apks[-1])
+        for i in packs:
+            apks.append(android.apk.APK(apk=i, prefix='assets/game/'))
+            game_apks.append(apks[-1])
+            split_apks.append(apks[-1])
 
     apks.append(android.apk.APK(prefix='assets/x-renpy/x-common/'))
+    for i in packs:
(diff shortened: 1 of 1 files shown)


renpytom:
Does this do anything to prevent google from compressing the asset pack contents? Ren'Py expects that media files are uncompressed in the APK - otherwise, a much slower path is required.

avonder:
> [The asset file will be uncompressed, unless `bundletool` has been explicitly configured to compress the asset pack.](https://developer.android.com/reference/com/google/android/play/core/assetpacks/AssetLocation.html)

The asset files delivered during testing are indeed uncompressed.

```
Archive:  split_ff1.apk
 Length   Method    Size  Cmpr    Date    Time   CRC-32   Name
--------  ------  ------- ---- ---------- ----- --------  ----
    1584  Defl:N      597  62% 1981-01-01 01:01 9c664794  AndroidManifest.xml
      16  Stored       16   0% 1981-01-01 01:01 c7862ba2  assets/00_pack.txt
    2200  Stored     2200   0% 1981-01-01 01:01 02ad25e1  assets/renpy/common/000atl.rpyc
    2344  Stored     2344   0% 1981-01-01 01:01 58931824  assets/renpy/common/000namespaces.rpyc
...
```

renpytom:
That's good information. 

Is there any good documentation for this? Everything seemed to be scattered all over the place.

avonder:
Sorry, there probably isn't. The documentation is just that abstruse.

PR #5864, opened 2024-10-24
Fix an obvious undefined variable error

craigds:
Fixes `Uncaught ReferenceError: p is not defined`


renpytom:
Good catch, thank you for the fix.

PR #5870, opened 2024-10-28
transitions in screen actions

the66F95:
various screen Actions() now use their _layer/layer arguments in renpy.transition() to actually have transitions on these layers.

### DIFF
--- a/renpy/common/00action_control.rpy
+++ b/renpy/common/00action_control.rpy
@@ -131,7 +131,7 @@ init -1500 python:
             renpy.show_screen(self.screen, *self.args, **self.kwargs)
 
             if self.transition is not None:
-                renpy.transition(self.transition)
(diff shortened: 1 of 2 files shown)


renpytom:
I don't want to merge it. 

It's a fairly substantial behavior change, and I think we generally want global transitions, not per-layer ones.

the66F95:
so you provide the option to show/hide a screen on another layer but not doing it with a transition.
🤔 i don't quite get it, but ok.

Andykl:
You can achieve your desired goal with passing `transition={layer: transition}` and/or write wrapper that does that in your project.

In my expierence `onlayer` for both images and screens is a trap, because if I see X and search for its `layer/config.tag_layer` I want to be sure X is shown on that layer and one can copy screen/image definition for different layer. So I would like to see less `layer` usages, not more.

the66F95:
ok.

PR #5877, opened 2024-10-30
developer menu fix for f-strings in expressions

the66F95:
using f-strings as expressions in show/scene statements results in "unknown text tag" errors when using the "image attributes" tab in the developer menu.
if my changes are too much or i didn't catch all possible displayables, add at least the `q` conversion flag to the original `[name]` substitutions to prevent the original error.

### DIFF
--- a/renpy/common/_developer/developer.rpym
+++ b/renpy/common/_developer/developer.rpym
@@ -94,6 +94,24 @@ screen _developer:
 
     key "developer" action Hide("_developer")
 
+init python:
+
+    def get_tag_attributes(tag, l):
+        d = renpy.display.scenelists.scene_lists().get_displayable_by_tag(l, tag)
+
+        if tag in d.child.__dict__.get("name", []):
+
+            attributes = " ".join(renpy.get_attributes(tag, layer=l))
+
+        else:
+            tag = "Expression: {}".format(tag)
+
+            attributes = " ".join(d.child.__dict__.get("name", []))
+
+            if attributes:
+                attributes = "Resolved: {}".format(attributes)
+
+        return tag, attributes
 
 screen _image_attributes():
     layer config.interface_layer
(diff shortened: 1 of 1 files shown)


renpytom:
What's the problem that this is trying to solve? 

I'm almost certain this shouldn't be looking into the displayable child's dictionary like this, but I'm not even clear what your'e using f-strings for in the first place.

the66F95:
```py
label main_menu:
    return

label start:

    $ tag = "bg"
    $ attributes = ["washington",]

    scene expression f"{' '.join([tag] + attributes)}"

    "using an f-string produces an \"unknown text tag\" error"

    $ renpy.call_in_new_context("_image_attributes")

    scene expression "{}".format(' '.join([tag] + attributes))

    "using the old .format() counterpart produces an \"empty text tag\" or \"unknown text tag\" error"

    $ renpy.call_in_new_context("_image_attributes")

    return
```

the66F95:
using `text _("    [name!q] [attributes] (hidden)")` and `text _("    [name!q] [attributes]")` would already solve the issue.
the rest is just my attempt to show also the name of the actual displayed image.

renpytom:
I'm not accepting this, since I think that it makes the image attributes mislead. The actual tag is a fairly nonsensical expression, and what you're displaying isn't the actual image tag in attributes. 

I've added !q to prevent crashes in the fix branch. 

In the 8.4 branch, I've changes the behavior of show expression to not use the expression as a tag - instead, it's generating actual tags of the form _show_expression_0, _show_expression_1, etc.

renpytom:
I've changed this further so that: 

```renpy
show expression "bg washington" 
```

is equivalent to:

```renpy
show bg washington
```

Which makes the actual attributes available.

the66F95:
how about wrapping the counter loop in a conditional block?
```py
    if expression is not None:
        expression = renpy.python.py_eval(expression)

        if not renpy.config.old_show_expression:

            if isinstance(expression, str):
                name = expression
            else:

                if not tag:
                    counter = 0

                    while True:

                        tag = "_show_expression_%d" % counter
                        if not renpy.exports.showing(tag, layer):
                            break

                        counter += 1

                name = tag

        expression = renpy.easy.displayable(expression)
```
otherwise you omit a possibly present tag from an `as` clause .

PR #5885, opened 2024-11-05
use the idle child as mask per style property

### LINKED ISSUE #5862: feature request: focus_mask 'idle'
i've noticed when I set focus_mask as displayable it's usually idle image of itself. so i guess it would be nice to do 'idle' attribute, so it can be styled for every buttons

the66F95:
should resolve #5862
the 1st rendered idle child is stored and later used as mask. it will not work with an animated idle child. however, i think using `focus_mask "idle_mask"` wouldn't make much sense in such a case.

### DIFF
--- a/renpy/display/behavior.py
+++ b/renpy/display/behavior.py
@@ -913,6 +913,8 @@ class Button(renpy.display.layout.Window):
     # This locks the displayable against further change.
     locked = False
 
+    idle_mask = None
+
     def __init__(self, child=None, style='button', clicked=None,
                  hovered=None, unhovered=None, action=None, role=None,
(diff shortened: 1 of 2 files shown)


renpytom:
This looks wrong to me, starting with Renders not being something that should be stored like this.

Gouvernathor:
I think this feature opens a lot of problematic questions with edge cases. For instance, making the idle survive while not active seems contradictory with the idea of having an animation when the idle disappears or a transition between the outgoing state and the incoming one. And that is a feature that is quite often requested by creators.

the66F95:
i'm closing this for now because it may solve the problem mentioned, but does not work more generally.

PR #5899, opened 2024-11-10
steam: fix get_session_ticket

Le09:
The definition of GetAuthSessionTicket expects an identity parameter:
https://github.com/renpy/renpy-build/blob/0681b1bdb8836f4ded742491158234beb90d9059/steamapi/steamapi.py#L5083

This simple fix seems to work as expected.

### DIFF
--- a/renpy/common/00steam.rpy
+++ b/renpy/common/00steam.rpy
@@ -424,7 +424,7 @@ init -1499 python in _renpysteam:
         return get_csteam_id() & 0xffffffff
 
 
-    def get_session_ticket():
+    def get_session_ticket(identity=None):
         """
         :doc: steam_user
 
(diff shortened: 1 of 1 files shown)


renpytom:
Thanks for the  fix and the first contribution.

Le09:
Thank you for the quick merge,  and  most of all for all your hard work on renpy!

Le09:
@renpytom Sorry for the bother; I need this patch to be able to distribute our game on Steam. I cannot find a guide on how to do it. Should I just build this repo starting from the release tag + this commit? (I want to target Steam from a Linux x64 system)

qTich:
> @renpytom Sorry for the bother; I need this patch to be able to distribute our game on Steam. I cannot find a guide on how to do it. Should I just build this repo starting from the release tag + this commit? (I want to target Steam from a Linux x64 system)

This commit contains no changes to the compiled files,
so you can replace this logic locally.
Or use one of the nightly versions - https://nightly.renpy.org/

Le09:
@qTich Thank you very much for your quick reply. I had modified the file locally, but that didn't get the distribution build work. I removed 00steam.rpyc before rebuild and that solved it. Thanks again!

Issue #5934, opened 2024-11-21
#5922 broke `dismiss` with nested draggable viewport 


qTich:
Modified example from documentation:

```renpy
default difficulty = "Easy"

screen select_difficulty():
    frame:
        align (.5, .3)
        padding (20, 20)

        has vbox

        textbutton "Difficulty: [difficulty]":
            action CaptureFocus("diff_drop")

        textbutton "Done":
            action Return()

    if GetFocusRect("diff_drop"):
        dismiss action ClearFocus("diff_drop") modal True

        nearrect:
            focus "diff_drop"

            frame:
                has viewport:
                    draggable True

                has vbox
                
                textbutton "Easy" action SetVariable("difficulty", "Easy"), ClearFocus("diff_drop")
                textbutton "Medium" action SetVariable("difficulty", "Medium"), ClearFocus("diff_drop")
                textbutton "Hard" action SetVariable("difficulty", "Hard"), ClearFocus("diff_drop")
                textbutton "Nightmare" action SetVariable("difficulty", "Nightmare"), ClearFocus("diff_drop")
```

Issue #5173, opened 2023-12-12
Draggable viewport (e.g. NVL screen) prevents focusing menu choice children purely via keyboard navigation


hsandt:
This is probably related to other draggable viewport focus issues such as https://github.com/renpy/renpy/issues/892 and https://github.com/renpy/renpy/issues/2167

My game uses NVL and I enabled a number of utility features including draggable viewport (which is not very useful on PC but great on mobile to drag text in the rare cases where it got too long, instead of dragging a small scrollbar). However, it has a side effect of preventing normal focus. I fixed one part of it by adding custom input to advance the story when user presses Space, Enter or mouse clicks from any context - but I still have an issue when I offer NVL menu choices to the player:

**Pressing keyboard up/down keys does not focus the first/last menu item, I need to move the mouse to hover at least one choice to enable key navigation (and probably the same for gamepad).**

**Repro**

1. Download the MWE at the bottom, or copy these snippets to a brand new project:

```rpy
# scripts.rpy

init python:
    # To allow player to advance with all these inputs
    # despite draggable viewport catching focus and mouse events
    config.keymap['dismiss_unfocused'].append('mouseup_1')
    config.keymap['dismiss_unfocused'].append('K_SPACE')
    config.keymap['dismiss_unfocused'].append('K_RETURN')
    config.keymap['dismiss_unfocused'].append('K_KP_ENTER')

label start:
    nvl_narrator "Hello"

    menu(nvl=True):
        "Choice A":
            nvl_narrator "A"
        "Choice B":
            nvl_narrator "B"

# screens.rpy

screen nvl(dialogue, items=None):
    # CUSTOMIZED: surround window with draggable viewport
    viewport:
        # If you comment this out, the bug disappears
        draggable True

        # make sure to indent everything below until the end of screen nvl
        window:
            style "nvl_window"
```

2. Play the game until the choice. I recommend to advance with space, without touching the mouse, or if you prefer clicking, make sure not to keep the mouse away from the choices.

![Bug repro - Draggable NVL viewport prevents pure key nav](https://github.com/renpy/renpy/assets/1067772/85a5975a-45e1-4110-8be2-19b5dd02ae92)

3. Press keyboard up/down

Expected: select first/last items
Actual: nothing is selected

**MWE**

[Repro bug Draggable viewport prevents NVL choice key navigation.zip](https://github.com/renpy/renpy/files/13650892/Repro.bug.Draggable.viewport.prevents.NVL.choice.key.navigation.zip)

[...]

PR #5936, opened 2024-11-22
Again viewport focus

the66F95:
#5934 should be fixed while preserving the functionality of nested viewports and focusing children via keyboard.

### DIFF
--- a/renpy/display/focus.py
+++ b/renpy/display/focus.py
@@ -632,6 +632,9 @@ def mouse_handler(ev, x, y, default=False):
     except renpy.display.layout.IgnoreLayers:
         new_focus = None
 
+    if new_focus and isinstance(new_focus.widget, renpy.display.viewport.Viewport) and new_focus.widget.draggable and not new_focus.widget.arrowkeys:
+        new_focus = None
+
     if new_focus is None:
         new_focus = global_focus
 
--- a/renpy/display/focus.py
+++ b/renpy/display/focus.py
@@ -632,13 +632,21 @@ def mouse_handler(ev, x, y, default=False):
     except renpy.display.layout.IgnoreLayers:
         new_focus = None
 
-    if new_focus and isinstance(new_focus.widget, renpy.display.viewport.Viewport):
-        new_focus = None
-
     if new_focus is None:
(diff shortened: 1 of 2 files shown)


the66F95:
l.635 ff. of focus.py
```py
    if new_focus and isinstance(new_focus.widget, renpy.display.viewport.Viewport):
        new_focus = None
```
may need to be removed to make the entire viewport focusable for the mouse.
currently a dismiss displayable "under" the viewport is clickable "through" the viewports empty space.

renpytom:
I rewrote this, and posted it as 9b290e07f. 

The one thing I don't get is the motivation for https://github.com/renpy/renpy/pull/5936/files#diff-5061088c6e29319dcf2d1f9d8438e9d4299c8f842b1952e037f505212564a57dR638 . That doesn't seem to do much, as it prevented the viewport from being focued, but that shouldn't matter. 

I'm not sure if that's what you meant in the comments above.

the66F95:
after your changes i use now
```py
def mouse_handler(ev, x, y, default=False):
    """
    Handle mouse events, to see if they change the focus.

    `ev`
        If ev is not None, this function checks to see if it is a mouse event.
    """

    global pending_focus_type

    if grab:
        return

    if ev is not None:
        if ev.type not in (pygame.MOUSEMOTION, pygame.MOUSEBUTTONUP, pygame.MOUSEBUTTONDOWN):
            return
        else:
            pending_focus_type = "mouse"

    try:
        new_focus = renpy.display.render.focus_at_point(x, y)
    except renpy.display.layout.IgnoreLayers:
        new_focus = None

    if new_focus is None:
        new_focus = global_focus

    if new_focus and isinstance(new_focus.widget, renpy.display.viewport.Viewport) and new_focus.widget.draggable and not new_focus.widget.arrowkeys:
        new_focus = None

    return change_focus(new_focus, default=default)
```
otherwise the initial #5173 still persists.
you can't focus the viewports children via keyboard while the viewport is focused by the mouse.

renpytom:
Another slightly different fix, in 2da8beecd7e3b2eefd67db1072f259a740e9738b, but I think this should work now.

PR #5942, opened 2024-11-23
Live2D.blend_opacity wasn't usable

farzher:
looks like you forgot to put this code here?

i've been using `live2d.blend_parameter` in my `update_function`, but `live2d.blend_opacity` threw an error saying it didn't exist until i added this code. now it exists and works fine


renpytom:
Merged, thanks for the  fix.

PR #5970, opened 2024-12-01
Update dragdrop.py

### LINKED ISSUE #5968: Modals within a drag can impede its movement
Somewhere between `8.2.3` and `8.3.3` a drag being actively moved by the player started to be interrupted by modal windows within itself. The problem persists in `8.3.4.24112906+nightly`.

I've included a test case below where there is a clear difference in behaviour between the two versions can be observed.

The expected behaviour is that while the mouse is depressed and the action of dragging is ongoing, it wouldn't be interrupted by a modal widget (at least within itself).

The actual result is that if the mouse moves with sufficient speed so as to enter the the bounds of the modal, movement is immediately halted. If the mouse is released within the bounds of the modal, then moved back outside, the drag movement resumes (despite mouse not being depressed) and can only be stopped by clicking within the non-modal area of the drag, or moving the cursor beyond the bounds of the game window.

<details><summary><b>Test case</b></summary>

```rpy
label main_menu:
    $ _confirm_quit = False
    return


label start:
    call screen test()
    return


screen test():
    drag:
        align (.5, .5)
        xysize (400, 225)

        has window
        background '#555518'

        has side 't c'

        window:
            padding (5, 5)
            text 'Grab here and drag down quickly'

        window:
            background Solid('185555')
            modal True
```
</details>

the66F95:
should fix #5968 (not extensively tested)

### DIFF
--- a/renpy/display/dragdrop.py
+++ b/renpy/display/dragdrop.py
@@ -739,9 +739,12 @@ class Drag(renpy.display.displayable.Displayable, renpy.revertable.RevertableObj
 
     def event(self, ev, x, y, st):
 
-        rv = self.child.event(ev, x, y, st)
-        if rv is not None:
-            return rv
+        grabbed = (renpy.display.focus.get_grab() is self)
+
+        if not grabbed:
+            rv = self.child.event(ev, x, y, st)
+            if rv is not None:
+                return rv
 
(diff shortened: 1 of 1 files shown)


renpytom:
Pretty much the same fix I'd been working on.

PR #6000, opened 2024-12-10
Bar property update & addition

shawna-p:
This PR does the following:

1) Updates `thumb_offset` to optionally take a tuple instead of an integer. If provided a tuple, the first number is used to offset the left/top of the bar, and the second number is used for the right/bottom. This is helpful for bars whose thumb is asymmetrical, such as a half-circle. In particular, it avoids issues with semi-transparent image overlap (in general or when using ATL or transitions). It is backwards-compatible with the regular integer value.

https://github.com/user-attachments/assets/293e33dc-9331-47b3-916c-a96264cff8d0

https://github.com/user-attachments/assets/94614917-c43f-4d2c-af47-dba9d229f613

2) Adds a `thumb_align` property to bars. This solves an incredibly common issue which occurs when the bar's thumb is larger than the bar images. Most people want the center and thumb of the bar to be aligned. This provides a simple in-engine solution, rather than forcing players to adjust their bar images in an image editor.

Default behavior (and the behavior of `thumb_align 0.0`, which is the default property so it's backwards-compatible:
image

The results of `thumb_align 0.5`, which aligns the thumb and bar images.
image

I have only added `thumb_align` as a screen property, not as a style one, though I'd be happy if it were added as a style property. Feel free to let me know if there are any particular changes you'd like made, or to adjust the PR before it goes into the engine.

### DIFF
--- a/renpy/display/behavior.py
+++ b/renpy/display/behavior.py
@@ -2289,6 +2289,8 @@ class Bar(renpy.display.displayable.Displayable):
         self.adjustment = adjustment
         self.focusable = True
 
+        self.thumb_align = properties.pop("thumb_align", 0.0)
+
         # These are set when we are first rendered.
         self.thumb_dim = 0
         self.height = 0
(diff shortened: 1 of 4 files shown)


renpytom:
Merged, and thank you. This was a great thing to have for #6000 .

PR #6006, opened 2024-12-15
Modernize `renpy.__init__`.

Andykl:
This PR removes `renpy.update_path` function and implements it by meta_path finder, so it automatically works for any package that need it now or in future. Also I did some small typing related changes and added some annotations.


renpytom:
I like this, but I need to figure out if it's necessary. I'm currently working on modifying Ren'Py's setup.py to allow for editable installs of Ren'Py, which may (or may not) render this unnecessary.

renpytom:
This turned out to be necessary after all. (Or at least, the least bad option.)

PR #6008, opened 2024-12-16
Split `renpy.ast` into submodules and add types.

Andykl:
I have went through the ast module and added type annotations statement slots and for `Node` interface.

I also splitted it into 3 files, because of some further ideas about `Node` class, which will add too many lines for already big `ast` file.

I kinda don't like that PyCode and PyExpr are in separate file, so I will move it to `node.py` (or maybe `renpy.python` is actually a best place for it?). And I think `location` file can be moved in `ast` folder.

### DIFF
--- a/renpy/ast/__init__.py
+++ b/renpy/ast/__init__.py
@@ -1,0 +1,116 @@
+# Copyright 2004-2024 Tom Rothamel <pytom@bishoujo.us>
+#
+# Permission is hereby granted, free of charge, to any person
+# obtaining a copy of this software and associated documentation files
+# (the "Software"), to deal in the Software without restriction,
+# including without limitation the rights to use, copy, modify, merge,
+# publish, distribute, sublicense, and/or sell copies of the Software,
+# and to permit persons to whom the Software is furnished to do so,
+# subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+# LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+# OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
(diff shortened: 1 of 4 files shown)


renpytom:
I agree on moving location, but I kind of think that Node should be in with the other ast nodes.

renpytom:
So, what are your ideas about the node class? 

Right now, I can see moving PyCode and PyExec out (perhaps into python.py), but I'm not sure about the rest - and I'm really not sure about Node becoming much bigger.

Andykl:
The idea is that most of the time, most nodes do not need some of their slots, or slot value is its default value. E.g. 90% of `TranslateSay` in my project fills only 8 out of 20 its slots, and others are default values. Same applies to pretty much any node subclass.

The idea is to do something like
```py
class _NodeUnlockedData:
    """
    Pretty much a dict with restricted set of keys it can use.
    This class is used when Node being constructed in parse.
    """

class _NodeLockedData:
    """
    Class that construct it's subclass with __slots__ that are actually used by a node
    and are not the same as default.
    """

class NodeAttribute:
    """
    Descriptor for Node subclasses that defines possible slot, and has it's default
    value.
    """

class Node(Location):
    slots = ["__data", "name"] 
    __data: _NodeUnlockedData | _NodeLockedData

    next =  NodeAttribute(default=None)

    def __enter__(self):
        self.__data = _NodeUnlockedData(self.__data.as_dict())
        return self

    def __exit__(self):
        self.__data = _NodeLockedData(self.__data.as_dict())
```

With this approach, `Return` is defined as
```py
class Return(Node):
    __slots__ = []
    expression = NodeAttribute(default=None)
```

The drawback of this approach is that it is a bit crypric, and that to change or init node, it should be done with `with`
```py
with renpy.ast.Return(loc) as rv:
    rv.expression = parsed_expression
```
but on the other hand, it will eliminate the need to use `getattr(node, "atl", None)` because default values are defined in slot descriptor.

This idea is based on simillar problem I fixed a while back, with our (and renpy) CDS, where parsed value was 90% redundant memory wise, and my approach saved like 20% of whole ast memory footprint, but it was mostly due to change from `dict` to this slotted `LockedData`.

renpytom:
I don't think this is something I'm likely to merge in, ever, as it stands - it's too complicated, and too much of a maintenance burden. And once it's been done, we'll be stuck with it forever, assuming it's what gets saved into the .rpyc format files. 

I especially don't like the indirection of each of the slots through a second object - just having that second object is going to add 24 bytes of storage, which removes a lot of the savings for most object. 

I'd really be interested in knowing the statistics of your  game, and how much memory is being taken up by each kind of AST node. I can see putting some time into reducing the AST memory, but this  really feels like a blunt solution that makes Ren'Py ugly while not solving the problem well

(One idea I'm having is doing something on the C side to support optional slots. But there are other options that might also make sense, and in the end I wonder if that the right solution for large games is to just allow portions of the AST to be loaded and unloaded as the game progresses.)

Andykl:
After implementing this, it proves to help some types, but hurt others, being marginally better in total, so this not worth it, unless written in C(ython).

PR #5881, opened 2024-11-02
First bit of PEP 657 in RenPy

Andykl:
This PR improves how RenPy reports parse errors in Python code (this can later be expanded to RenPy code).

Before:
image

After:
image

---
Parse errors screen now automatically highlights error part of parse error if its message has more that one caret, e.g.
```
I'm sorry, but errors were detected in your script. Please correct the
errors listed below, and try again.


File "game/script.rpy", line 41: did you forget parentheses around the comprehension target?
    [x, y for x, y in zip([1, 2, 3], [3, 2, 1])]
     ^^^^

File "game/script.rpy", line 45: Generator expression must be parenthesized
    bar(1, i for i in range(3), 2)
           ^^^^^^^^^^^^^^^^^^^

File "game/script.rpy", line 48: cannot assign to True
    True = False
    ^^^^

File "game/script.rpy", line 51: expected '('
    def foo:
           ^

File "game/script.rpy", line 55: expected '('
    def foo
           ^

Ren'Py Version: Ren'Py 8.4.0.0+unofficial
Sun Nov  3 00:37:15 2024
```
---
This PR adds 3 backward incompatible changes in RenPy internal code:
1. Removes `renpy.lexer.get_line_text` which can be replaced with `linecache.getline`, which can be added back.
2. Changes order of parameters of `renpy.lexer.ParseError` so it is in line with `SyntaxError`, can be changed back.
3. `renpy.display.error.report_parse_errors` and `screen _parse_errors` now takes list of error messages instead of joined string, so it know where one error ends and the other starts.

PR #6015, opened 2024-12-19
Tokenized parser

Andykl:
This is draft for incoming tokenized parser. Tokenizer file and changes in Lexer are ready to review, but I need to fix `scriptedit` and merge with changes in #5881.


Andykl:
This introduces following syntax incompatibilities:
1. Tokenizer does not allow backtick strings.
2. Tokenizer does not handle single-qouted strings not closed on the same physical line, e.g. it was used in couple of places in tutorial
```
    e "While the slide transitions slide in the new scene, the
       slideaways slide out the old scene."
```
3. There is weird case of tokens like `00e0` which can be interpreted as a name or as a float, and if someone expected it to be a name, it will fail. Maybe it is better to interpret them as names, because change to number is pretty strightforward.
4. `\f` is dissalowed the same way as `\t` because python's tokenizer does not like it.
5. Spaceship operator is no longer valid.
6. Probably some other with syntaxes I am not aware people use.

Other incompatibilities I am aware of:
1. There are some differences how old and new logical lines handle line continuation.
2. There is no way to construct a `Lexer` the old way, as it expect a list of `Line` objects now.
3. `Lexer.subblock` now is a boolean property. Constructing old value from lines would be expensive and I assume in a wild they are boolean checks from before `Lexer.has_subblock` was introduced.
4. Most of old field are read-only properties.
5. `Lexer.pos` support write, but it snaps the position to the start of token if was set to space character.
7. Return value of `Lexer.checkpoint` has different elements.
8. `list_logical_lines`, `group_logical_lines` and `ren_py_to_rpy` are removed.
In general there should be no incompatibilities in public interface.

Also I am not sure, how to correctly fix pre 7.3 user statements `renpy.statements.parse`.

renpytom:
> Tokenizer does not allow backtick strings.

This isn't acceptable.

> Tokenizer does not handle single-qouted strings not closed on the same physical line, e.g. it was used in couple of places in tutorial
    e "While the slide transitions slide in the new scene, the
       slideaways slide out the old scene."

This isn't acceptable.

We also need to support pre 7.3 parse statements. 

In generally, a new tokenizer should not be changing Ren'Py's syntax - it's not worth it if it does.

Andykl:
Well, because there is a regression in python tokenizer which no longer can parse `0_attr` (because it expect digit after underscore), and such syntax I believe is pretty common, even if we were to accept syntax changes above, to fix that issue it will requiere changing the Python parser itself, i.e. in renpy-build, which make this idea too big.

So, I think this idea is on hold for pretty long time, not until RenPy 9, or never. Which is a bummer, because tokenization is ~30% of parse time, and right now reloading script in big games is a waiting game.

Profiling shows that this idea can reduce loading script by 10% in my project, so I will continue to work with this in our fork of RenPy, so if we ever decide to merge it in RenPy, it will live here.

PR #6025, opened 2024-12-22
Launcher Folders

HorBro38:
This adds the following:

- Support for folders in the launcher for Projects, Libraries and Templates.  Folders only appear to the user if they have libraries or templates, by default they will not appear.
- Adds a new project type, "Libraries".  Uses the same JSON file method as Templates.
- Cleans up the "clear" launcher style and integrates it into the standard light mode.
- Adds preferences for showing/hiding the default projects (The Question and Tutorial) as well as Templates (This was already in the code, just unused).
- Adds one 646 byte image for the folder icon.

### DIFF
--- a/launcher/game/front_page.rpy
+++ b/launcher/game/front_page.rpy
@@ -102,39 +102,82 @@ screen front_page:
 # This is used by front_page to display the list of known projects on the screen.
 screen front_page_project_list:
 
+    $ folders = project.manager.folders
     $ projects = project.manager.projects
     $ templates = project.manager.templates
 
     vbox:
 
-        if templates and persistent.show_templates:
-
-            for p in templates:
-
-                textbutton _("[p.name!q] (template)"):
-                    action project.Select(p)
-                    alt _("Select project [text].")
-                    style "l_list"
-
-            null height 12
-
         if projects:
-
+            text "Projects"
             for p in projects:
 
-                textbutton "[p.name!q]":
+                textbutton _(f"[p.display_name]" if p.display_name else "[p.name!q]"):
                     action project.Select(p)
                     alt _("Select project [text].")
                     style "l_list"
 
-            null height 12
+            null height 6
 
-        textbutton _("Tutorial") action project.SelectTutorial() style "l_list" alt _("Select project [text].")
-        textbutton _("The Question") action project.Select("the_question") style "l_list" alt _("Select project [text].")
+        if folders:
+            for pf in folders:
+
+                button:
+                    if pf.hidden:
+                        hover_background REVERSE_HOVER
+                        xfill True
+                        hbox:
+                            spacing 5
+                            add "folder_closed"
+                            text pf.name.capitalize() color IDLE hover_color REVERSE_TEXT bold True yalign 0.5
+
+                    else:
+                        hover_background REVERSE_IDLE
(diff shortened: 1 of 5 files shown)


renpytom:
What problem does this PR solve? 

I'm not completely clear who would be using this. Library is a new project type, which doesn't exist. And a user might have a couple of templates. So I'm not sure what makes adding this to the launcher worth it.

HorBro38:
Sure Tom, we were talking about this in #dev earlier.

Folders are useful for those of us who have multiple libraries and templates and want to be able to store them in the project folder without seeing them at all times.  The idea is that this is invisible to users who don't do use them, but for those of us who regularly build extra tools for Ren'py this would be extremely useful for organizational purposes.  Anyone who maintains multiple extra libraries quickly finds their projects list fills up.

I asked a few developers on the discord if such a tool would be useful and so far the reaction was very positive, right now many of us are moving projects in and out of our project folders or maintaining multiple versions of Ren'py.  The idea is to not have to do that, and folders allow for this extra organization.

renpytom:
Could you make it read the folder to use out of the project.json file, rather than simply having the Project/Library/Template distinction? My feeling is that this would be much more useful if it was an arbitrary system, rather than one that  forces a specific organization on every user.

HorBro38:
> Could you make it read the folder to use out of the project.json file, rather than simply having the Project/Library/Template distinction? My feeling is that this would be much more useful if it was an arbitrary system, rather than one that forces a specific organization on every user.

I re-wrote it to allow for the use of any name; and now it supports any number of folders.  Thanks for the suggestion; it honestly let me condense and clean up the code.  I've got it set so if you are only use regular projects (either declared as such or not at all) and hidden projects it won't show the folders; but once you add a template; library or anything else it will start using folders to organize them all.  It also culls unused folders if there are no longer projects that need them.

renpytom:
Why not have it kicks in if there's a folder define, rather than having anything to do with project types? 

Also, it looks like there's a weird revert of the OpenDirectory and Edit changes.

HorBro38:
I've modified this to use a new JSON Data type, "folder".  I've also made it so that anyone can use the 'display_name' JSON as well, as it didn't appear to be working correctly.  I've also set it so the Tutorial projects are treated more as normal projects, and can be shown/hidden.  Let me know what you think.

RE: OpenDirectory/Edit changes; can you point me to what you are talking about?

Thanks for the input, apologizes for the delay in getting back to this - I hurt my hand earlier in the month and it delayed me getting to work on this.

HorBro38:
@renpytom - just making sure you saw I updated this.  Feel free to reach out to me on Discord as well.

HorBro38:
Updated this with new logic (thanks to @Luficerx) that uses the actual folders in your project directly to generate the folders in the launcher.  For sanity reasons it will only go one level deep, and ignores folders that contain no sub folders with projects.

Luficerx:
@renpytom Kinda forgot about this PR, One of the issues with this implementation is that after moving the project to a subfolder, You still can't create a project with the same name because the way the projects are stored in the launcher, They all go on a list and renpy checks if the project exists or not inside that list, To fix that it would require changing how the projects are saved which add's a lot of complexity

Just pointing this out in case of future issues raised due that

PR #6034, opened 2024-12-25
Don't overrule focus override for max focus

shawna-p:
This fixes an issue wherein the user wants to set the focus on a screen, but the request is overridden by Ren'Py's default_focus system forcing the largest default_focus to be focused instead.


shawna-p:
Actually I'll put this on the fix branch

PR #6052, opened 2025-01-02
Audit uses of py2to3 compatibility functions.

Andykl:
Changes in this PR commit by commit
1. Removes uses of `future.utils` functions. Now RenPy codebase does not depend on this package.
2. Removes all uses of `basestring`. Most of the time it already raises some obscure exception. In some cases I raise explicit error with simple error message, in other cases decode from bytes. If both bytes and str are valid I replaced with explicit `(str, bytes)`.
3. Removed the rest uses of PY2 constant, and updated all places I have found that had some cooment like `update when python 3 is here`.
4. Replaced uses of `unicode` with `str`, and `bchr` with construction of bytes from list of integers.

I think after this PR it is safe to replace
```
from __future__ import division, absolute_import, with_statement, print_function, unicode_literals # type: ignore
```
with
```
from __future__ import annotations
```
And force string annotations everywhere, because PEP 649 in 3.14 will do it anyway, so if there is dependency on runtime evaluation of type hints, we get the error early.

### DIFF
--- a/renpy/display/swdraw.py
+++ b/renpy/display/swdraw.py
@@ -306,13 +306,11 @@ def draw_special(what, dest, x, y):
 
         ramplen = what.operation_parameter
 
-        from itertools import repeat
+        ramp = b"\x00" * 256
 
-        ramp = bytes([
-            *repeat(0, 256),
-            *(255 * i // ramplen for i in range(ramplen)),
-            *repeat(255, 256),
-        ])
+        ramp += bytes([255 * i // ramplen for i in range(ramplen)])
+
+        ramp += b"\xff" * 256
 
         step = int(what.operation_complete * (256 + ramplen))
         ramp = ramp[step:step + 256]
--- a/renpy/display/swdraw.py
+++ b/renpy/display/swdraw.py
@@ -306,12 +306,13 @@ def draw_special(what, dest, x, y):
 
         ramplen = what.operation_parameter
 
-        ramp = b"\x00" * 256
+        from itertools import repeat
 
-        for i in range(0, ramplen):
-            ramp += bchr(255 * i // ramplen)
-
-        ramp += b"\xff" * 256
+        ramp = bytes([
+            *repeat(0, 256),
+            *(255 * i // ramplen for i in range(ramplen)),
+            *repeat(255, 256),
+        ])
 
         step = int(what.operation_complete * (256 + ramplen))
         ramp = ramp[step:step + 256]
--- a/renpy/easy.py
+++ b/renpy/easy.py
@@ -21,7 +21,8 @@
 
 """Functions that make the user's life easier."""
 
-from __future__ import annotations
+from __future__ import division, absolute_import, with_statement, print_function, unicode_literals
+from renpy.compat import PY2, basestring, bchr, bord, chr, open, pystr, range, round, str, tobytes, unicode # *
 
 from typing import Any
 
@@ -55,13 +56,56 @@ def lookup_displayable_prefix(d):
     return displayable(fn(arg))
 
 
-def displayable_or_none(
-    d: Any,
-    scope: dict[str, Any] | None = None,
-    dynamic: bool = True) -> renpy.display.displayable.Displayable | None:
+def displayable_or_none(d, scope=None, dynamic=True): # type: (Any, dict|None, bool) -> renpy.display.displayable.Displayable|None
+
+    if isinstance(d, renpy.display.displayable.Displayable):
+        return d
 
     if d is None:
         return d
+
+    if isinstance(d, str):
+        if not d:
+            raise Exception("An empty string cannot be used as a displayable.")
+        elif ("[" in d) and renpy.config.dynamic_images and dynamic:
+            return renpy.display.image.DynamicImage(d, scope=scope)
+
+        rv = lookup_displayable_prefix(d)
+
+        if rv is not None:
+            return rv
+        elif d[0] == '#':
+            return renpy.store.Solid(d)
+        elif "." in d:
+            return renpy.store.Image(d)
+        else:
+            return renpy.store.ImageReference(tuple(d.split()))
+
+    if isinstance(d, Color):
+        return renpy.store.Solid(d) # type: ignore
+
+    if isinstance(d, list):
+        return renpy.display.image.DynamicImage(d, scope=scope) # type: ignore
+
+    # We assume the user knows what he's doing in this case.
+    if hasattr(d, '_duplicate'):
+        return d
+
+    if d is True or d is False:
+        return d
+
+    raise Exception("Not a displayable: %r" % (d,))
+
+
+def displayable(d, scope=None): # type(d, dict|None=None) -> renpy.display.displayable.Displayable|None
+    """
+    :doc: udd_utility
+    :name: renpy.displayable
+
+    This takes `d`, which may be a displayable object or a string. If it's
+    a string, it converts that string into a displayable using the usual
+    rules.
+    """
 
     if isinstance(d, renpy.display.displayable.Displayable):
         return d
@@ -69,7 +113,7 @@ def displayable_or_none(
     if isinstance(d, str):
         if not d:
             raise Exception("An empty string cannot be used as a displayable.")
-        elif ("[" in d) and renpy.config.dynamic_images and dynamic:
+        elif ("[" in d) and renpy.config.dynamic_images:
             return renpy.display.image.DynamicImage(d, scope=scope)
 
         rv = lookup_displayable_prefix(d)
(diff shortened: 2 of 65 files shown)


renpy/display/swdraw.py
```diff
@@ -306,12 +306,13 @@ def draw_special(what, dest, x, y):
...
 
-        for i in range(0, ramplen):
-            ramp += bchr(255 * i // ramplen)
-
-        ramp += b"\xff" * 256
+        ramp = bytes([
+            *repeat(0, 256),
+            *(255 * i // ramplen for i in range(ramplen)),
+            *repeat(255, 256),
+        ])
```
mal, line 315: Aside from needing a separate import, this seems actively harder to read now. Perhaps instead only replace the `for` loop with `ramp += bytes(255 * i // ramplen for i in range(ramplen))`?

mal (commented):
A few specific queries, and a couple of general ones:

- In the cases where `e` in the exception being handled, do we need the extra stack frame from `raise e from None` vs simply `raise`?
- `basestring` appears to be currently defined as `(str,)`, so I'm unclear what the motivation is to handle `bytes` explicitly in so many scenarios.

Andykl:
1. `if we raise without from None it will add extra block of `during handling the above exception another exception occured`.
2. No, in `renpy.compat` basestring is `(bytes, str)`.

renpy/easy.py
```diff
@@ -107,43 +109,18 @@ def displayable(d, scope=None): # type(d, dict|None=None) -> renpy.display.displ
     rules.
     """
 
-    if isinstance(d, renpy.display.displayable.Displayable):
-        return d
```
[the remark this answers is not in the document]
Andykl: In 3.12 I believe there was a big work to make Python-to-Python calls several times faster, unless exception occured, but I will revert both changes, no not pollute the PR with unreleated things.

mal:
Hmm, on 2. it seems this may have been an [unintentional change](https://github.com/renpy/renpy/commit/d3c827fc55f01c4cc8e8c8ef2815dc06b22bbc09#diff-b84086e9141da610d1ed8811f3a972149ec46dc6285b4b7d52f310565098ecb8L117) when `future.utils.string_types` was replaced with the `(str, bytes)` literal.

In 8.3.4 `renpy.compat.basestring` in the console yeilds `(<class 'str'>,)`.

renpytom:
The unintentional change seems likely.

Andykl:
Yeah, it seems so. Anyway, now compat may exist to only export useful(?) things for RenPy games, but not for RenPy base, and this change will not affect games, because it was not released if I inderstand correctly.

mal:
On 1. I've just confirmed that it doesn't add the extra block:

<details><summary>Test code</summary>

```rpy
init python:
    import traceback

    def a():
        raise Exception('foo')

    def _from_none():
        try:
            a()
        except Exception as e:
            raise e from None

    def _implicit():
        try:
            a()
        except Exception:
            raise

    def test(f):
        try:
            print(f'## RUNNING {f.__name__}')
            f()
        except Exception:
            traceback.print_exc()

    test(_from_none)
    print()
    test(_implicit)
```
</details>

<details><summary>Output</summary>

```
## RUNNING _from_none
  File "game/src/script.rpy", line 118, in test
  File "game/src/script.rpy", line 107, in _from_none
    c, _, n = a.partition('_')
  File "game/src/script.rpy", line 105, in _from_none
  File "game/src/script.rpy", line 101, in a
    init python hide:
Exception: foo

## RUNNING _implicit
  File "game/src/script.rpy", line 118, in test
  File "game/src/script.rpy", line 111, in _implicit
  File "game/src/script.rpy", line 101, in a
    init python hide:
Exception: foo
```
</details>

Andykl:
Well, this is not the same input I think.
```py
def foo():
    return 1 / 0

def handle(e):
    raise Exception("Can't") from e

def main():
    try:
        foo()
    except Exception as e:
        try:
            handle(e)
        except Exception:
            raise e

main()
```

```
  File "C:\Users\andy_\t.py", line 12, in main
    handle(e)
  File "C:\Users\andy_\t.py", line 5, in handle
    raise Exception("Can't") from e
Exception: Can't

During handling of the above exception, another exception occurred:

  File "C:\Users\andy_\t.py", line 16, in <module>
    main()
  File "C:\Users\andy_\t.py", line 14, in main
    raise e
  File "C:\Users\andy_\t.py", line 9, in main
    foo()
  File "C:\Users\andy_\t.py", line 2, in foo
    return 1 / 0
           ~~^~~
ZeroDivisionError: division by zero

```

mal:
Yup, that's why I was only asking about "the cases where `e` is the exception being handled".

mal:
No, see the difference in trace from the output in my comment above.

Andykl:
Yes, if this is bare `raise` in the exception handler but not in this case. If we do `raise` in handler it will chain exception with the exception of `exception handler, i.e.

```
  File "C:\Users\andy_\t.py", line 9, in main
    foo()
  File "C:\Users\andy_\t.py", line 2, in foo
    return 1 / 0
           ~~^~~
ZeroDivisionError: division by zero

The above exception was the direct cause of the following exception:

  File "C:\Users\andy_\t.py", line 16, in <module>
    main()
  File "C:\Users\andy_\t.py", line 12, in main
    handle(e)
  File "C:\Users\andy_\t.py", line 5, in handle
    raise Exception("Can't") from e
Exception: Can't
```
Note different order and chain message.

Andykl:
Huh. But not with such code
```
def foo():
    return 1 / 0

def handle(e):
    raise Exception("Can't") from e

def main():
    try:
        foo()
    except Exception as e:
        try:
            handle(e)
        except Exception:
            pass

        raise

main()
```

I was thinking that exception in handler completly erase handler above.

renpy/loadsave.py
```diff
@@ -409,25 +411,23 @@ def save(slotname, extra_info='', mutate_flag=False, include_screenshot=True):
...
         if bad is None:
-            reraise(t, e, tb)
+            raise e from None
 
         if e.args:
-            e.args = (e.args[0] + ' (perhaps {})'.format(bad),) + e.args[1:]
+            e.args = (e.args[0] + f' (perhaps {bad})', *e.args[1:])
 
-        reraise(t, e, tb)
+        raise e from None
```
mal, line 430: It was these that drew my initial comment, as all these (but for the nested one on line `422`) could be `raise`.

renpytom:
Yeah, I think the right move is to remove those unnecessary (my fault) reraises, and then use raise e from None for the remaining one.

renpytom:
I'm also thinking we don't want ``from __future__ import annotations`` . It'll be going away eventually, and the PEP 649 behavior is different. I don't think we should opt into a dead end.

renpy/loadsave.py
Andykl: Done.

Andykl:
> I'm also thinking we don't want `from __future__ import annotations` . It'll be going away eventually, and the PEP 649 behavior is different. I don't think we should opt into a dead end.

Fair enough. So, then we can drop both lines after this is merged.

mal:
Given the false premise under which the various `bytes` changes were added, they should probably be removed prior to merging also.

Andykl:
Done.

renpy/loadsave.py
```diff
@@ -409,25 +411,15 @@ def save(slotname, extra_info='', mutate_flag=False, include_screenshot=True):
...
-        if bad is None:
-            reraise(t, e, tb)
-
-        if e.args:
-            e.args = (e.args[0] + ' (perhaps {})'.format(bad),) + e.args[1:]
+    except Exception as e:
+        if not mutate_flag and e.args:
+            try:
+                bad = find_bad_reduction(roots, renpy.game.log)
+                e.args = (e.args[0] + f' (perhaps {bad})', *e.args[1:])
```
mal, line 418: I think you'd need a conditional here to get equivalence with the original, otherwise you could get "perhaps (None)":
```suggestion
                if bad := find_bad_reduction(roots, renpy.game.log):
                    e.args = (e.args[0] + f' (perhaps {bad})', *e.args[1:])
```

renpy/python.py
```diff
@@ -1202,7 +1202,10 @@ def py_eval_bytecode(bytecode, globals=None, locals=None): # @ReservedAssignment
 
 
 def py_eval(code, globals=None, locals=None): # @ReservedAssignment
-    if isinstance(code, basestring):
+    if isinstance(code, bytes):
+        code = code.decode("utf-8", "python_strict")
```
mal, line 1206: I don't think there's much of a case for sending bytes here is there? Implicitly decoding any that do (most likely accidentally) seems like it would only end up hindering debugging.

mal (commented):
Last few bits and pieces I think. :)

renpy/execution.py
```diff
@@ -613,25 +611,31 @@ def run(self, node=None):
...
-                            if renpy.display.error.report_exception(short, full, traceback_fn):
-                                raise
-                    except renpy.game.CONTROL_EXCEPTIONS as ce:
-                        raise ce
+                        if not handled and not renpy.display.error.report_exception(
+                            short,
+                            full,
+                            traceback_fn
+                        ):
+                            handled = True
```
mal, line 629:
```suggestion
                        if not handled:
                            handled = not renpy.display.error.report_exception(short, full, traceback_fn)
```

renpy/display/swdraw.py
```diff
@@ -308,8 +308,7 @@ def draw_special(what, dest, x, y):
 
         ramp = b"\x00" * 256
 
-        for i in range(0, ramplen):
-            ramp += bchr(255 * i // ramplen)
+        ramp += bytes([255 * i // ramplen for i in range(ramplen)])
```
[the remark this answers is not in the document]
Andykl: I am not sure this code even can be hit, it is sw renderer, so I will leave it as is.

Andykl:
Addressed review above.

mal (approved):
LGTM 🎉

renpytom:
Looks good to me too. Reading it over one last time, then merging. 

One note is that we can't remove all of the compat stuff. Variables like PY2  need to show up in minstore, as these variables can be used by games. 

But that's for the  future. @Andykl , thanks for putting in all this work.

PR #6053, opened 2025-01-03
Update lint to use py_compile function

mal:
Lint still uses a function removed when the [bytecode cache was improved](https://github.com/renpy/renpy/commit/fe02e12192d74f448f3b49b1193ef4a9ea676623).


Andykl:
Good catch.

PR #6057, opened 2025-01-04
Do not kill Renders that cache the last video frame on Web

### LINKED ISSUE #6050: Shader error in web version
Good afternoon. Tell me, are there ways to solve this problem? Occurs at random times in different games

screenshot-porngameshub_com-2025_01_02-10_08_20

Kassy2048:
When the next video frame is not ready, the previous one is returned by `get_movie_texture_web()`. Normally, this is just a GLTexture instance, but when using a mask, that frame is a Render instance instead. The problem is that Render instances get killed soon after they have been drawn, so `get_movie_texture_web()` returns a dead instance when re-using the previous frame which causes a crash because of missing uniforms when drawing the screen.

Ideally, we should prevent the Render from being killed as long as it is part of `renpy.display.video.texture`, but I'm not sure how to do that (maybe visiting that dict in [`mark_sweep()`](https://github.com/renpy/renpy/blob/77aaeaa51445a0f0d3eddd26741a23e8b7876b0e/renpy/display/render.pyx#L493)?). At least with this PR, there is no more crash, but maybe it introduces flickering (I did not see any though).

Fix #6050

### DIFF
--- a/renpy/display/render.pyx
+++ b/renpy/display/render.pyx
@@ -501,6 +501,7 @@ def mark_sweep():
     cdef list worklist
     cdef int i
     cdef Render r, j
+    cdef object o
 
     worklist = [ ]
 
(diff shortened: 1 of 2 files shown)


Kassy2048:
I'm trying to move the changes to `mark_sweep()`, so changed this PR to a draft.

Kassy2048:
That was easy enough. Not sure the extra dependency of `renpy.display.render` to `renpy.display.video` is okay though.

PR #6058, opened 2025-01-04
Suggestions for cslots

mal:
I'll preface this by noting that my Cython and C knowledge is pretty low on the ground, so these are changes that seem to make _logical_ sense, but others will have to determine whether they make _practical_ sense.

- Make it easy to continue using the experimental tests while the cslots feature is still in development.
- Remove `default_int_value` as it appears to be a duplicate value store, and isn't strictly needed in the one place it is used.
- Allow `IntegerSlot` descriptors to re-open a compressed object automatically, matching behaviour of the generic `Slot` descriptor.

### DIFF
--- a/experimental/cslots/cslots.pyx
+++ b/experimental/cslots/cslots.pyx
@@ -1,0 +1,1 @@
+../../renpy/cslots.pyx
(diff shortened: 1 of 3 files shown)


renpytom:
I undid 627c6c6. The duplicated variable is there to stop the value from ping-ponging from integer to object repeatedly.

PR #6062, opened 2025-01-05
Document cb_args cb_ prefix behavior

### LINKED ISSUE #6039: Documentation missing reference to Character Callback args
The docs still use `functools.partial` to apply arguments to callbacks when the character class has the ability to add callback args with keyword arguments prefixed with `cb_`. It is easier to understand and is consistent with how other parts of renpy function in conjunction with the Character object.

GiovanH:
Resolves #6039.

This replaces the documentation that describes manually currying arguments to callback functions using `functools.partial` with an updated example using the Ren'py builtin `cb_args` prefix system.


renpytom:
Thanks for the commit. 

If you can, pure-documentation updates should be targeted to the fix branch, which gets pushed up to www.renpy.org as part of the nightly build process. I often cherry-pick documentation onto the fix branch, but targeting it saves me a step. 

Thanks again!

Issue #6060, opened 2025-01-04
Can't access `u_virtual_size` uniform


paranormalized-encryptid:
For some reason I have never been able to access the `u_virtual_size` uniform because Ren'Py always throws this error:
```python
AttributeError: 'renpy.gl2.gl2draw.GL2Draw' object has no attribute 'virtual_size'
```

mal:
Could you provide a stack trace for the error, and ideally a small bit of sample code to reproduce the problem? The only code that should be accessing that property is internal.

mal:
Never mind, have managed to reproduce.

Issue #6061, opened 2025-01-04
Make screen projection matrix available to OpenGL shaders?


paranormalized-encryptid:
The `u_transform` matrix appears to be multiplied by a screen projection matrix before being passed to the shader. Would it be possible to pass in the second matrix separately so we can manually undo its changes to `u_transform`? I think this would make it easier to use `u_transform` in color operations, whereas right now it does odd things like inconsistently flip everything upside-down and apply unknown scaling factors that are hard to correct for.

renpytom:
What are you trying to do, exactly? u_transform is meant for geometry, not for color adjustments.

paranormalized-encryptid:
Transformation of bump map normals. It looks and works great except for the aforementioned problem, which becomes particularly apparent with zooming (scaling), and I'm having to manually flip the y axis.

renpytom:
Wouldn't the bump map usually use the texture coordinates, not the screen ones?

paranormalized-encryptid:
Translation/offsets aren't applied to the normals, but we have to correct for the rotation and scaling elements of the transform by taking the transpose of the inverse of the 3x3 submatrix, or the normals will end up pointing in the wrong direction.

renpytom:
Question - which matrices would you like to have? The current u_transform bakes in a lot - it project through multiple steps.

My guess is that we should break it up, with u_transform going from the displayables coordinates to Ren'Py's coordinates, and then a second matrix going from that to virtual coordinates. But that's just a thought.

paranormalized-encryptid:
Could do! I just need to be able to obtain the initial transform separately, however that happens, though breaking it up into steps like that is intuitive and could keep things simpler. Are there any other matrices we need to think about?

renpytom:
I don't think you're being precise about what you think the "initial transform" is, but I'd be down with adding another matrix that projects the u_transformed coordinates back to the virtual coordinate space.

paranormalized-encryptid:
Well, it's difficult to know exactly since `u_transform` is filtered through multiple operations that I'm not privy to, but I'm agreeing that virtual coordinates are probably what we want, thank you. At some point, some scaling factors `(2/w, -2/h)` are applied that need to be corrected for, where `w` and `h` look somewhat similar to the dimensions of `u_viewport` - that's all I know.

renpytom:
> virtual coordinates are probably what we want

Could you describe, in your own words, what you want? Not in terms of the matrices, but what coordinates you'd like to find.

[...]

Issue #5642, opened 2024-07-20
Android 14; openGL built-in math functions bug when calculating with v_position variable


GitHub-Saga:
I'm trying to create an animated shader ripple effect, with the ripples center determined by mouse position.
The shader works correctly on the PC Build (Win10 64-bit) and on my old phone running on Android 13.
But on my new phone running on Android 14, the shader calculations "freeze" when you reach a certain distance from the center. 
With "freeze" I mean the last calculated value after the certain distance remains the same for the rest of the shader.

 
```
renpy.register_shader("example.ripp", 

    variables="""
        uniform vec2 u_mPos;
        uniform sampler2D tex0;
        uniform vec2 res0;
        attribute vec4 a_position;
        varying vec4 v_position;
        attribute vec2 a_tex_coord;
        varying vec2 v_tex_coord;
        uniform float u_time;
        uniform vec4 u_viewport;
        uniform float u_t0;
        """,
    vertex_300="""
        v_tex_coord = a_tex_coord;
        v_position = a_position;
        """,
    fragment_300="""

        vec2 cPos;
        float cLength;
        vec2 uv;
        float A;

        cPos = u_mPos;
        A=1e-1;
        cLength = distance(v_position.xy, cPos);
        /*cLength = length(v_position.xy - cPos);*/

        float phi = cLength*5e-1 - 0.0*(u_time-u_t0)*100.0;
        uv =  v_tex_coord + vec2(0.0, A*sin(cLength)); 

        gl_FragColor = texture2D(tex0, uv, 0.0);
        """
    )
```

I have tried multiple different ways of calculations to reach the proper effect on Android 14, but to no avail. 
The code works as expected when you replace v_position with gl_FragCoord, but gl_FragCoord comes with alot of other issues (positioning, viewport size, window scaling, etc.). I have been able to deduce that the bug arises from the combination of v_position and the built-in math functions in openGL. I'm using Renpy 8.2.3. Ripples should fill the full viewport as it does on my PC and Android 13 phone, but ripples are limited to a certain area from the center point on my Android 14 phone, outside this area the rest of the texture just receives a constant offset.

renpytom:
Do you have something to force Ren'Py to draw the screen? By default, it may not.

GitHub-Saga:
Force Ren'Py to draw the screen? I don't quite understand what you mean.

After defining the shader, I define the transform and then add images at that transform.

```
init python:
    g0=time.time()
    renpy.register_shader("example.ripp", ...
[...]

PR #6063, opened 2025-01-05
gl2: Fix virtual_size access when creating uniforms

mal:
Fixes https://github.com/renpy/renpy/issues/6060.

Presumably this never worked since it was added speculatively to maybe help with #5642. Given that no one's wanted it until now (at least not enough to notice it didn't work), and if the motivating case for #6060 is better solved by the solution to #6061, then it's possible that removal might be a better option.


renpytom:
35c4ba4a4 seems like a better fix to me, as it would improve other potential places where this was used.

PR #6064, opened 2025-01-06
Linter option to show character names

### LINKED ISSUE #5998: implement name in character statics (Lint)
since it is difficult to get the character name from the Lint, how about using a tuple list in config?

https://github.com/renpy/renpy/blob/62810b94aba7b32024880062b4cdf813625ee723/renpy/lint.py#L883C1-L904C1

```
define config.lint_character_prefix_to_name = [
    ("e", "Eileen"),
] # default []

```

```
def get_character_name(prefix):
        for p, name in renpy.config.lint_character_prefix_to_name:
            if p == prefix:
                return name
        return prefix 

def report_character_stats(charastats):
    """
    Returns a list of character stat lines.
    """

    rv = [ "", "Character Statistics (for default language):", ] # type: list[str|list[str]]

    bullets = [ ]

    for char in sorted(charastats, key=lambda char: charastats[char].tuple(), reverse=True):
        count = charastats[char]
        bullets.append(
            " * " + get_character_name(char)
            + " has " + humanize(count.blocks) + (" block " if count.blocks == 1 else " blocks ") + "of dialogue, "
            + "containing " + humanize(count.words) + " words and "
            + humanize(count.characters) + " characters."
        )

    rv.append(bullets)

    return rv
```

GiovanH:
Resolves #5998.

Optionally uses characters' `name` attribute as keys in the statistics dictionary, instead of the literal token string.


GiovanH:
Fixed merge conflict, but force-pushed to keep git history clean. (charstats changed from a counter to a defaultdict)

PR #6075, opened 2025-01-11
use a dummy function for rendering at init time

Elckarow:
this removes the need for an `if` check inside `renpy.render`


renpytom:
This doesn't seem like a good idea to me.  If checks are cheap, and rebinding cython functions like this is fraught.

PR #6105, opened 2025-01-23
Update `nearrect` documentation

nullvoid8:
https://www.renpy.org/doc/html/screens.html#nearrect is missing documentation for the properties `preferred_side` and `invert_offsets` added in #5114 . Additionally, `class Nearrect` is missing documentation for the `focus` keyword.

This PR also contains a first stab at updating the layout description to account for the fact it can now run in either axis as appropriate.

Apologies if the documentation contains any build errors, I currently have no way to build it.

### DIFF
--- a/renpy/display/layout.py
+++ b/renpy/display/layout.py
@@ -2416,11 +2416,18 @@ class NearRect(Container):
 
     `rect`
         The rectangle to place the child near.
+    
+    `focus`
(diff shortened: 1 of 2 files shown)


nullvoid8:
Whoops, just noticed the big bit at the top of Readme telling me to base this on the `fix` branch

renpytom:
Thank you.

PR #6107, opened 2025-01-23
Limit the number of cython processes ran in parallel

Kassy2048:
Running all the cython tasks in parallel at once is a bit violent. On my Docker container for instance, it makes some processes be killed before they complete because of resources exhaustion.

Using a ThreadPoolExecutor instance to limit the number of parallel tasks guarantees there will be at least as much processes at the number of CPU plus 4 (up to 32) so that should be as fast as before without risking to reach the host resources limit.

### DIFF
--- a/scripts/setuplib.py
+++ b/scripts/setuplib.py
@@ -32,6 +32,7 @@ import warnings
 import pathlib
 import platform
 import subprocess
+from concurrent.futures import ThreadPoolExecutor
 
 import setuptools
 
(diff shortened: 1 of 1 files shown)


scripts/setuplib.py
```diff
@@ -302,24 +303,18 @@ def generate_all_cython():
...
-            t = threading.Thread(target=generate_cython, args=args)
-            t.start()
-            threads.append(t)
-
-    for t in threads:
-        t.join()
+    else:
+        with ThreadPoolExecutor() as executor:
+            for args in generate_cython_queue:
+                executor.submit(generate_cython, *args)
```
mal, line 315: Not certain, but do you not need a call to `shutdown` to await the threads to match the previous `join` behaviour?

Something like (untested):
```suggestion

            executor.shutdown(wait=True)
```

Kassy2048: No, shutdown() is called automatically when leaving the "with" block:

> You can avoid having to call this method explicitly if you use the [with](https://docs.python.org/3/reference/compound_stmts.html#with) statement, which will shutdown the [Executor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor) (waiting as if [Executor.shutdown()](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.shutdown) were called with wait set to True):

https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.shutdown

mal: Ah, nice!

PR #6109, opened 2025-01-24
Tokenized parser

Andykl:
Resurrection of #6015.
Now this should have no syntax changes, and following incompatiable changes:
1. `Lexer.subblock` now is a boolean property. Constructing old value from lines would be expensive and I assume in a wild they are boolean checks from before Lexer.has_subblock was introduced.
2. Some old fields of `Lexer` are read-only properties.
3. Return value of Lexer.checkpoint has different elements.
4. list_logical_lines, group_logical_lines and ren_py_to_rpy are removed.
All other things should work as before.

This is ready for merge, but `_Tokenizer` class and it's regexes need to be rewriten in Cython because it takes 50% of parse time.


renpytom:
The unmerged branch associated with this pull request has been renamed to [work/andy_kl/tokenized-parser](https://github.com/renpy/renpy/tree/work%2Fandy_kl%2Ftokenized-parser) at commit 7f303a5fbf6f10764f60f97165848ea034cb3db0.

Andykl:
I deleted that branch. I think my changes in `lexersupport` made this obsolete.

PR #6111, opened 2025-01-25
mark imported function / variable as `pure` / `const` in the importing store

Elckarow:
when the function / variable has been marked as `pure` / `const`

the idea is that
```rpy

init python in my_store:
    @renpy.pure
    def my_pure_function(): pass

    my_const = 1
    renpy.const("my_store.my_const")

init python in my_other_store:
    from store.my_store import my_pure_function, my_const
```
doesn't mark `my_other_store.my_pure_function` as `pure`, nor `my_other_store.my_const` as `const`

you could do that manually, of course, but it's not practical when you have a lot of things to import

PR is a draft because of 3 things:

- [x] how do we get the name of the module we're importing from?
- [x] how do we know if the thing being imported is marked? (since we're just parsing a tree)
- [x] this doesn't handle `*` imports

for n°1, when compiling an `init python`, we could set a variable to the store's name, which we could grab in visit_ImportFrom

for n°2, one way of doing it would be to add a tuple of (fullname, after_import_fullname) to a list somewhere, and then loop over these tuples after every `init python` has executed, to check if `fullname` is a constant value.

i'll let you guys decide this

### DIFF
--- a/renpy/pyanalysis.py
+++ b/renpy/pyanalysis.py
@@ -168,6 +168,29 @@ def pure(fn):
 
     return fn
 
+def import_from(from_module, in_module, *names):
+    """
+    This function is called after each `from from_module import name` statement,
+    to make sure that if `from_module.name` is a const / pure value, `in_module.name` is marked const / pure as well.
+    """
+    if from_module.startswith("store."):
+        from_module = from_module[6:]
+    
+    if in_module.startswith("store."):
+        in_module = in_module[6:]
+    
+    for name in names:
+        fullname = f"{from_module}.{name}"
+
+        if fullname in not_constants:
+            continue
+
+        fullname_after_import = f"{in_module}.{name}"
+
+        if fullname in pure_functions:
+            pure(fullname_after_import)
+        elif fullname in constants:
+            const(fullname_after_import)
 
 class Control(object):
     """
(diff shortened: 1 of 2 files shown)


Andykl:
I think the correct way to do so is to add `renpy.pyanalysis.import_from(from_module, in_module, *names)` function, and in ast after each import create a code that calls this function with correct arguments. I.e. add call like
```
init python in my_other_store:
    from store.my_store import my_pure_function, my_const
    renpy.pyanalysis.import_from("store.my_store", __name__, "my_pure_function", "my_const")
```
And
`*` imports are bad for variety of reasons, a lot of tools just bail out for them, so I think it is okay if we don't support it.

I think it is a good idea to have, because it is a gotcha as it is right now, but if it is possible to only convert top-level imports, and not enter functions it would be great, because it make no difference to call it again inside functions.

Elckarow:
from my tests, this seems to work

Andykl:
1. I am not sure how it works for class scope. I think it should also skip this when inside class.
```
init python in sub:
    class Foo:
        from store.foo import const_foo
```
2. Could you add a comment above that `ast` construct than explain what code we construct.

Elckarow:
done

Andykl:
LGTM.

PR #6115, opened 2025-01-26
don't sort everything in `libs` / `mods` by filename

Elckarow:
this allows specifying which directory will have its files sorted by filename, rather than full pathname, by writing those directories' name in `libs.txt` / `mods.txt`

by default, files in `libs/` and `mods/` are sorted by filename

the point of this is that some packages might not care about being sorted by filename. they just want to be loaded before the game files

_im not really sure how i should document this_

### DIFF
--- a/renpy/script.py
+++ b/renpy/script.py
@@ -343,8 +343,25 @@ class Script(object):
 
         self.common_script_files.sort(key=lambda item: ((item[0] or ""), (item[1] or "")))
 
-        has_libs = renpy.loader.loadable("libs/libs.txt")
-        has_mods = renpy.loader.loadable("mods/mods.txt")
+        try:
+            with renpy.loader.load("libs/libs.txt") as f:
+                libs_order_by_filename = [s.decode("utf-8") for s in f.readlines()]
+                has_libs = True
+        except Exception:
+            has_libs = False
+            libs_order_by_filename = [ ]
+        
+        libs_order_by_filename.append("")
+
+        try:
+            with renpy.loader.load("mods/mods.txt") as f:
+                mods_order_by_filename = [s.decode("utf-8") for s in f.readlines()]
+                has_mods = True
+        except Exception:
+            has_mods = False
+            mods_order_by_filename = [ ]
+        
+        mods_order_by_filename.append("")
 
         def game_key(item):
             fn = item[0] or ""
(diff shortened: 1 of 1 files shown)


renpytom:
They have to be  sorted by something, why not by filename?

Elckarow:
the package im developing breaks with this behavior. i just want it to be loaded before the game files, while still maintaining the same load order inside it

yes i could just put it outside of `libs/`, slap a `01` in the folder name, and call it a day, but that defeats the purpose of having a `libs/` folder in the first place

Elckarow:
sorry for +1ing this, but i would really like an explanation as to why we have to load scripts in a different order in /libs and /mods specifically, compared to the rest of the engine

Elckarow:
say you have `mypackage/aaa/before.rpy` and `mypackage/bbb/after.rpy`

inside any regular folder, the load order is:
- `mypackage/aaa/before.rpy`
- `mypackage/bbb/after.rpy`

but inside `/libs` and `/mods`, since scripts are sorted by filename, the result is
- `libs/mypackage/bbb/after.rpy` is loaded first, since the file begins with an `a`
- `libs/mypackage/aaa/before.rpy` is loaded second

renpytom:
Yes. That's the point of libs, so you can have bbb/01before.rpy, and have it load before aaa/02after.rpy. It's designed to make it so that package names can be independent of load order - so that a package named zebra-statement can define a new statement that loads before any other code, for example.

mal:
If I'm understanding correctly, I think the issue is that with the example `mypackage`, one might, not unreasonably, expect the entire path from that point on to be considered when ordering rather that solely relying on the basename, i.e.

**Current**
```
ignore         sort
libs/abc/bbb/  after.rpy
libs/xyz/bbb/  after.rpy
libs/abc/aaa/  before.rpy
libs/xyz/aaa/  before.rpy
```

**Expected**
```
ignore     sort
libs/abc/  aaa/before.rpy
libs/xyz/  aaa/before.rpy
libs/abc/  bbb/after.rpy
libs/xyz/  bbb/after.rpy
```

Elckarow:
> I think lib authors will expect files in their folder to be sorted by filename, i.e. if I develop lib `foo-statement` and somebody else develop `bar-statement` I don't have any expectations on what of the two will be loaded first because libs should be agnostic to each other and it is responsibility of the creator to figure out if they confilct, but inside my lib I expect files to be sorted as if they were put in `game` folder.

exactly

PR #6127, opened 2025-02-01
Handle newest_mtime initialized as None while determining newest game save

dm1sh:
Found logical error that prevented one ancient VN from viewing manual saves. Fix is plain and simple.


Andykl:
Seems like missed compare with None.

PR #6152, opened 2025-02-12
Add right_to_left and bottom_to_top Grid properties.

### LINKED ISSUE #6007: [Feature request] More grid properties
Hello. This is my first request, and at the risk of sounding like an idiot because maybe it already exists and I've been unable to find it, here it is!
By default, a grid deploys to the right and then down.

I would like to ask if it could be possible to add terms to be put inside a grid to control in which direction it deploys.
The current solution I'm using is this:
```python
    grid 3 2:
        at transform:
            yzoom -1
        for i, num in enumerate(whatever):
            textbutton "[num]":
                action NullAction()
                at transform:
                    yzoom -1
```
I'm thinking it would be nice to have a prettier/simpler alternative.

I'd like to request additional properties like so:
**"deploy_right_up" :** the grid elements will appear first as a row to the top-right of the anchor, then another row above it.
**"deploy_left_up" :** the grid elements will appear first as a row to the top-left of the anchor, then another row above it.
**"deploy_left_down" :** the grid elements will appear first as a row to the bottom-left of the anchor, then another row below it.

Lezalith:
These properties combined with transpose allow for 8 different ways to fill a grid. The way grid is rendered had to be changed slightly.  
  
Tested for all types of grid (square, more-rows rectangle, more-columns rectangle).  
  
Behind-the-scenes of the grid was changed. It now generates a gridmap (dict of {child : (col, row)}), and children are placed according to it, rather than always being filled in row-major order. Offsets are also sorted based on gridmap.  
This shouldn't break anything in existing code.   
  
Closes #6007

### DIFF
--- a/renpy/display/layout.py
+++ b/renpy/display/layout.py
@@ -488,7 +488,8 @@ class Grid(Container):
         top_margin = compute_raw(self.style.top_margin, height)
         bottom_margin = compute_raw(self.style.bottom_margin, height)
 
-        # Generates a gridmap dict. Key: child, Value: (col, row) tuple of intended position
+        # Generates a gridmap list containing (col, row) tuples of intended position
+        # for each child.
         gridmap = self.generate_gridmap()
 
         # Now, start the actual rendering.
@@ -501,9 +502,8 @@ class Grid(Container):
         if self.style.yfill:
             renheight = (height - (self.rows - 1) * yspacing - top_margin - bottom_margin) // self.rows
 
-        # Key: child, Value: render instance
-        renders = {child: render(child, renwidth, renheight, st, at) for child in gridmap.keys()}
-        sizes = [render.get_size() for render in renders.values()] # Order doesn't matter
+        renders = [render(child, renwidth, renheight, st, at) for child in self.children]
+        sizes = [r.get_size() for r in renders]
 
         cwidth = 0
         cheight = 0
@@ -523,24 +523,19 @@ class Grid(Container):
 
         rv = renpy.display.render.Render(width, height)
 
-        # Key: child, Value: (x, y) tuple of position offsets
-        offsets_dict = {}
+        # List of offsets in the same order as self.children
+        self.offsets = []
 
-        for child in gridmap.keys():
-            col, row = gridmap[child]
+        for i, child in enumerate(self.children):
+            col, row = gridmap[i]
 
-            surf = renders[child]
+            surf = renders[i]
 
             xpos = col * (cwidth + xspacing) + left_margin
             ypos = row * (cheight + yspacing) + top_margin
 
             offset = child.place(rv, xpos, ypos, cwidth, cheight, surf)
-            offsets_dict[child] = offset
-
-        # Offsets sorted to match the initial order of children
-        self.offsets = []
-        for child in gridmap.keys():
-            self.offsets.append(offsets_dict[child])
+            self.offsets.append(offset)
 
         return rv
 
@@ -586,7 +581,7 @@ class Grid(Container):
         else:
             intended_order = [(col, row) for row in row_order for col in col_order]
         
-        return {self.children[i]: pos for i, pos in enumerate(intended_order)}
+        return intended_order
 
 class IgnoreLayers(Exception):
     """
--- a/renpy/display/layout.py
+++ b/renpy/display/layout.py
@@ -432,6 +432,8 @@ class Grid(Container):
                  transpose=False,
                  style='grid',
                  allow_underfull=None,
+                 right_to_left=False,
+                 bottom_to_top=False,
                  **properties):
         """
         @param cols: The number of columns in this widget.
@@ -442,6 +444,12 @@ class Grid(Container):
 
         @params allow_underfull: Controls if grid may be underfull.
         If None - uses config.allow_underfull_grids.
+
+        @params right_to_left: True if cells should be filled
+        right to left.
+
+        @params bottom_to_top: True if cells should be filled
+        bottom to top.
         """
 
         if padding is not None:
@@ -457,6 +465,9 @@ class Grid(Container):
 
         self.transpose = transpose
         self.allow_underfull = allow_underfull
+
+        self.right_to_left = right_to_left
+        self.bottom_to_top = bottom_to_top
 
     def render(self, width, height, st, at):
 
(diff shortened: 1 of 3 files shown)


Lezalith:
Changed gridmap to be a list of (col, row) tuples instead of a dict, and changed the rest accordingly.   
The for loop that renders children now iterates over children in their original order, eliminating the need to sort offsets in any way.

Andykl:
Nitpick. You could just use `range` and `reversed` of range, so you don't create 2 to 4 lists and then throw it away.
Overall I like the idea. It could be useful in case grid represents some text, or other data that can change fill order for different languages.

mal:
Taking the logic you've worked out and implemented, this change can be simplified significantly:

Commit transform this PR into simpler version:
https://github.com/renpy/renpy/commit/dfa98b72df99e84edf44de63895476062e34af82

Combined diff against Ren'Py master (to see how the overall diff is simplified):
https://github.com/renpy/renpy/compare/master...mal:lezalith_grid_props

Ideally I'd like to roll all this into a single commit to reduce churn in the git history, but for now just presenting it.

Lezalith:
> Taking the logic you've worked out and implemented, this change can be simplified significantly:
> 
> Commit to transform this PR into simpler version: [dfa98b7](https://github.com/renpy/renpy/commit/dfa98b72df99e84edf44de63895476062e34af82)
> 
> Combined diff against Ren'Py master (to see how the overall diff is simplified): [master...mal:lezalith\_grid\_props](https://github.com/renpy/renpy/compare/master...mal:lezalith_grid_props)
> 
> Ideally I'd like to roll all this into a single commit to reduce churn in the git history, but for now just presenting it.

Are you sure the offsets will be sorted correctly this way? It's similar to my first draft but what came out of it was a weird bug where the offsets of cases `transpose True right_to_left True` and `transpose True bottom_to_top True` were swapped.   
  
Definitely agree with squashing this into a single commit btw.

mal:
You're correct the first version I posted had neglected offsets as you describe, but I believe the (now) updated version should correctly order them as expected. 🤞

renpy/display/layout.py
```diff
@@ -432,6 +432,8 @@ def __init__(self, cols, rows, padding=None,
                  transpose=False,
                  style='grid',
                  allow_underfull=None,
+                 right_to_left=False,
```
mal, line 435: Now that we're in py3 we can stop the accidental creep of new positional args by making any new ones kwarg-only.
```suggestion
                 *,
                 right_to_left=False,
```

renpy/display/layout.py
```diff
@@ -477,31 +488,22 @@ def render(self, width, height, st, at):
         top_margin = compute_raw(self.style.top_margin, height)
         bottom_margin = compute_raw(self.style.bottom_margin, height)
```
mal, line 490: Worth keeping these.
```suggestion

        # For convenience and speed.
        cols = self.cols
        rows = self.rows

```

renpy/display/layout.py
```diff
@@ -477,31 +488,22 @@ def render(self, width, height, st, at):
...
+        gridmap = self.generate_gridmap()
 
         # Now, start the actual rendering.
 
         renwidth = width
         renheight = height
 
         if self.style.xfill:
-            renwidth = (width - (cols - 1) * xspacing - left_margin - right_margin) // cols
+            renwidth = (width - (self.cols - 1) * xspacing - left_margin - right_margin) // self.cols
```
mal, line 501:
```suggestion
            renwidth = (width - (cols - 1) * xspacing - left_margin - right_margin) // cols
```

renpy/display/layout.py
```diff
@@ -477,31 +488,22 @@ def render(self, width, height, st, at):
...
         if self.style.yfill:
-            renheight = (height - (rows - 1) * yspacing - top_margin - bottom_margin) // rows
+            renheight = (height - (self.rows - 1) * yspacing - top_margin - bottom_margin) // self.rows
```
mal, line 503:
```suggestion
            renheight = (height - (rows - 1) * yspacing - top_margin - bottom_margin) // rows
```

renpy/display/layout.py
```diff
@@ -516,32 +518,24 @@ def render(self, width, height, st, at):
         if self.style.yfill:
             cheight = renheight
 
-        width = cwidth * cols + xspacing * (cols - 1) + left_margin + right_margin
-        height = cheight * rows + yspacing * (rows - 1) + top_margin + bottom_margin
+        width = cwidth * self.cols + xspacing * (self.cols - 1) + left_margin + right_margin
+        height = cheight * self.rows + yspacing * (self.rows - 1) + top_margin + bottom_margin
```
mal, line 522:
```suggestion
        width = cwidth * cols + xspacing * (cols - 1) + left_margin + right_margin
        height = cheight * rows + yspacing * (rows - 1) + top_margin + bottom_margin
```

renpy/display/layout.py
```diff
@@ -573,6 +567,21 @@ def per_interact(self):
         for _ in range(delta):
             self.add(null)
 
+    def generate_gridmap(self):
+        row_order = list(range(self.rows))
```
mal, line 571: No need to convert the range to a list here.
```suggestion
        row_order = range(self.rows)
```

renpy/display/layout.py
```diff
@@ -573,6 +567,21 @@ def per_interact(self):
...
+        if self.bottom_to_top:
+            row_order = row_order[::-1]
+
+        col_order = list(range(self.cols))
```
mal, line 575: Or here.
```suggestion
        col_order = range(self.cols)
```

renpy/display/layout.py
```diff
@@ -573,6 +567,21 @@ def per_interact(self):
...
+        if self.right_to_left:
+            col_order = col_order[::-1]
+
+        if self.transpose:
+            intended_order = [(col, row) for col in col_order for row in row_order]
+        else:
+            intended_order = [(col, row) for row in row_order for col in col_order]
+        
+        return intended_order
```
mal, line 584: Whitespace fixes.
```suggestion

        return intended_order

```

mal (commented):
I've retracted my original larger suggestion, as ultimately I don't think it ended up being a cleaner approach than the one already being proposed.

I've added a few much more minor suggestions as part of this review which, if you're happy with, I believe GitHub will allow to be collected into a single commit automatically somehow.

Lezalith:
> I've retracted my original larger suggestion, as ultimately I don't think it ended up being a cleaner approach than the one already being proposed.
> 
> I've added a few much more minor suggestions as part of this review which, if you're happy with, I believe GitHub will allow to be collected into a single commit using the "Add suggestion to batch" button in the files tab of this PR.

It did indeed! And I learned something new today. I believe it's ready for merge now.

PR #6159, opened 2025-02-16
Fix miner typos

jsfehler:


renpytom:
You're doing the thing with the names on purpose, aren't you?

PR #6164, opened 2025-02-17
Add distribution function support to particle system

jsfehler:
This PR adds support for using distribution functions in the SnowBlossom class and the Particles system in general.

Currently, distribution is hard-coded to be linear. This change preserves that behaviour as the default but adds the `distribution` argument to allow for custom behaviour. Two new distribution functions are added to provide built-in alternatives to a linear distribution.

The import from `renpy.display.particle.distribution` is awkward and I'm open to suggestions. I thought it better to move the distribution function code into a separate directory rather than bloat a single file.

The following script will showcase the differences between each distribution:

```
image sb_linear = SnowBlossom(
    "images/snow.png",
    count=1000,
    border=100,
    xspeed=(0, 0),
    yspeed=(400, 600),
    start=0,
    fast=False,
    horizontal=False,
    distribution=renpy.display.particle.distribution.linear
)

image sb_gaussian = SnowBlossom(
    "images/snow.png",
    count=1000,
    border=100,
    xspeed=(0, 0),
    yspeed=(400, 600),
    start=0,
    fast=False,
    horizontal=False,
    distribution=renpy.display.particle.distribution.gaussian
)

image sb_arcsine = SnowBlossom(
    "images/snow.png",
    count=1000,
    border=100,
    xspeed=(0, 0),
    yspeed=(400, 600),
    start=0,
    fast=False,
    horizontal=False,
    distribution=renpy.display.particle.distribution.arcsine
)

label start:

    show sb_linear

    "Linear Distribution"

    hide sb_linear

    show sb_gaussian

    "Gaussian Distribution"

    hide sb_gaussian

    show sb_arcsine

    "Arcsine Distribution"

    return
```


renpytom:
Changes I'd like: 

1. I don't think this is important enough to be worth a whole package, especially when the module contains four small functions. Merge this back into particle.py. 

2. I'd suggest the distributon argument  to SnowBlossomFactory take either a string or a function. The string can be one of "linear", "gaussian", or "arcsine", or it can be a function.  There are few cases (renpy.audio being the exception) where asking people to reach into renpy modules is a good idea. 

3. SnowBlossomFactory should have the distribution field defaulting to the linear function, similar to the way rotate defaults to False. This will help with backwards compatibility.

jsfehler:
> Changes I'd like:
> 
> 1. I don't think this is important enough to be worth a whole package, especially when the module contains four small functions. Merge this back into particle.py.
> 2. I'd suggest the distributon argument  to SnowBlossomFactory take either a string or a function. The string can be one of "linear", "gaussian", or "arcsine", or it can be a function.  There are few cases (renpy.audio being the exception) where asking people to reach into renpy modules is a good idea.
> 3. SnowBlossomFactory should have the distribution field defaulting to the linear function, similar to the way rotate defaults to False. This will help with backwards compatibility.

Done.

PR #6195, opened 2025-03-02
Support returning an error code when a problem is encountered during linting

Ayowel:
Hi,

Just noticed that our CI had been logging linting problems for a while without failing, thus not prompting us to fix/investigate further.
This PR allows to keep the current differentiation between errors and problems but allows to opt in to more strict linting rules by returning an error code even if only problems occured.

This is quite important for us as Ren'Py's linting output is quite verbose even if no error occurs (a huge amount of game characters makes it so that problem logs are not even visible on the screen), is not very parseable, and the amount of deprecated stuff that is only reported as a problem and not an error will end up biting us hard if not taken care of as it gets deprecated.

Thanks in advance,

### DIFF
--- a/renpy/lint.py
+++ b/renpy/lint.py
@@ -125,6 +125,8 @@ def problem_listing(header, problems):
             for line, message in file_problems:
                 print("    * line {:>5d} {}".format(line, message))
 
+    global error_reported
+    error_reported = True
 
 # Tries to evaluate an expression, announcing an error if it fails.
 def try_eval(where, expr, additional=None):


mal:
Suggest this be done via a value to given to `--error-code` so you'd have:
```sh
lint --error-code        #=> implicit value "error" giving current behaviour.
lint --error-code error  #=> explicit value "error" giving current behaviour.
lint --error-code warn   #=> explicit value "warn" causing warnings (problems) to give an exit code > 1.
```

That way we don't need a new flag, and we're better set up for any future expansion in this space (for instance a hypothetical future change to produce an exit code for errors and deprecations, but not other warning types).

Ayowel:
> ```shell
> lint --error-code        #=> implicit value "error" giving current behaviour.
> lint --error-code error  #=> explicit value "error" giving current behaviour.
> lint --error-code warn   #=> explicit value "warn" causing warnings (problems) to give a non-zero exit code.
> ```

I'm not a huge fan of options with optional parameters for clarity reasons (when multiple parameters are provided, it's not always clear which ones are option values and which ones are regular parameters without intimate knowledge of the command), and implementing without making the parameter optional would be backward-incompatible. I can implement like this (or just make the new option parameterized to be able to use it differently later), however I'd like a maintainer to weigh in first.

renpytom:
Lint never returns 100% problems - everything it returns is a warning.

So it probably makes sense to change the other kinds of warnings to trigger the new error code when the flag is given.

Ayowel:
> when the flag is given

Just so we're clear, you're talking about the current `--error-code` flag, not this PR's `--problem-as-error` (which then becomes useless and should just not be added) ?

Ayowel:
Updated to also return an error code if a problem occurs without additional options/variables

Ayowel:
@renpytom I think this can be merged, or is there additional changes required ?

Issue #6188, opened 2025-02-26
RTL text rendering missing fallback


TDCMC:
The RTL Arabic text should have a fallback for the base glyphs if the "Contextual form" doesn't exist in a font, but it doesn't. So if, for example, the "Isolated form" for a letter doesn't exist in the selected font like the following highlighted blocks:

![Image](https://github.com/user-attachments/assets/94dc3dad-cf2b-4bb4-9821-b9ab561149ae)

It should fallback to using the base glyphs highlighted below:

![Image](https://github.com/user-attachments/assets/315dc3b6-3507-4e39-adec-fe1d4c893a97)

But instead, it shows a "not defined" rectangle character.
Interestingly, if I reverse a word in LTR mode, this problem doesn't exist. But when writing a word in RTL mode, it shows up.

renpytom:
Ren'Py doesn't have a great way to determine if a glyph is missing in a font. It just renders the text, and the placeholder glyph (the not-defined rectangle) shows up it a character is missing. 

I think it probably makes sense to use fonts with the letterforms in the font itself, rather than having Ren'Py guess what to use. (Both because determining the glyph is missing would be hard, and also because I don't feel I understand Arabic text rendering well enough to do this myself.)

TDCMC:
Unfortunately, using fonts that are implemented in a compatible manner is easier said than done, since there only exist a fraction of options compared to languages that use the "latin" script. If you know where the text rendering is implemented, I can take a look (Though I don't know if I'd be able to fix it myself). Problem is that in theory this problem should be fixed in a handful of lines of code by just adding some if statements that check if that glyph exists in the font or not. If you are willing to put time, I can also guide the fixing instead of trying (praying to be able) to do it myself.

renpytom:
Hm... I could put together a way of mapping one character to another, after RTL but before the rest of text processing. Would that give you what you need? 

What I'm seeing is happening is that U+0671 (Arabic Letter Alef Wasla) is getting mapped to U+FB50 (Arabic Letter Alef Wasla Isolated Form). Does it make  sens to just ignore the isolated form glyphs, and turn them back into the underlying letter? If so, do you have a table of the mappings  to use?

TDCMC:
[...]

PR #6201, opened 2025-03-05
Unmap arabic isolated forms before passing them to the shaper

TDCMC:
This fixes the unconditional mapping of the isolated forms done by fribidi (https://github.com/renpy/renpy/issues/6188) and lets the shaper (e.g. Harfbuzz) do so based on the font used by the text. This is NOT a clean, efficient, or even a correct way of fixing it. But it should do the job until someone who knows more than me decides to reimplement the usage of the bidirectional algorithm in a way that doesn't do the mapping in the first place.

### DIFF
--- a/renpy/text/extras.py
+++ b/renpy/text/extras.py
@@ -450,3 +450,93 @@ def thaic90(s):
         rv.append(c)
 
     return u''.join(rv)
+
+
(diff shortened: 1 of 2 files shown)


renpytom:
Closing by request.

Issue #6202, opened 2025-03-07
Glyph misalignment when using a Nastaliq font


TDCMC:
When using a nastaliq font (Where text isn't linear), arabic text gets vertically misaligned:

![Image](https://github.com/user-attachments/assets/1fa4c418-aeaa-4491-adef-3e5bdf416268)

This is what it should look like:

![Image](https://github.com/user-attachments/assets/bbd6beba-a583-4895-ae13-90e07f18dec8)

The way the glyphs are aligned is using "Cursive marks":

![Image](https://github.com/user-attachments/assets/d5a7c98a-c7c5-4683-8986-8b1add03b80a)

The glyph substitution works, but text is misaligned. And I think it might partly be because Arabic text is reversed when rendering and as a result, the "entry" of a glyph is aligned with the "exit" of the succeeding glyph, rather than the preceding one.
Nastaliq fonts aren't used that much, because they are calligraphy. But any non-linear handwritten font should have this problem.
The problem also doesn't seem to just be the reversed text. Because in the example above, the "entry" of the previous glyph doesn't line up with the "exit" of the next glyph.

Also you mentioned that the max height of the text is determined by the longest glyph. I'm not sure what that means exactly, but every font (as far as I know) has some metrics for OS/2 that determine the max ascent and descent. So at least having an option to use those should suffice.

Overall, this issue isn't that big of a priority and only interferes with my fan-translation of Doki Doki Literature Club.

TDCMC:
Nevermind, I changed the y_offset to negative and it worked. Ignore everything I said.

PR #6203, opened 2025-03-07
Fix glyph offsets for HarfBuzz.

TDCMC:
Fixes the issue described in (https://github.com/renpy/renpy/issues/6202)


TDCMC:
Actually this pull request is quite problematic since it reverses the advancing and removed the y advance of vertical text. I'll do a proper pr later.

PR #6216, opened 2025-03-13
Type hint most export files

ksar-fd:
Version 2 of this. I'd borrowed a friend's laptop to make the previous PR, but had forgotten to set myself as author 😅 

Anyway I'll be adding to this bit by bit as I keep finding things that need annotating


renpytom:
I'm looking through this, and I'm seeing a lot of reformatting files to styles I find ugly, like wrapping lines too short, and adding things like: 

https://github.com/renpy/renpy/pull/6216/files#diff-31a458f21cbb822ea0442ed6ce25a4c8236cd1db96a3b03b274681996e42387fR58

Which isn't something I want to see in Ren'Py. 

(I'm considering using black to reformat Ren'Py, but with much longer line lengths, and probably with the preview style.)

ksar-fd:
I'm using black to format the files. Should I configure it within this pr so it follows your style more closely?

ksar-fd:
> https://github.com/renpy/renpy/pull/6216/files#diff-31a458f21cbb822ea0442ed6ce25a4c8236cd1db96a3b03b274681996e42387fR58
> 

Also if you mean where it says
```py
_ = call()
```
I added that to shut up basedpyright's linter. I figured out a better way that doesn't disrupt the code, namely `reportUnusedCallResult `. I can fix it right away too

renpytom:
Yes on using black, though I need to figure out what I want the line length to be. I've been playing around with 120. 

I'm leaning on doing a massive reformat of the Python code towards the end of the current release cycle, after I've merged all of the pending PRs.  So this would wait for that to be merged. (Hopefully I'll be switching into release mode after I finish the current work on speeding up rendering.) 

I'll also need to audit these, so making it clear is going to be important.

ksar-fd:
Ok, I did as much as I could. I'm not terribly familiar with the codebase, so some of these are my best guess, but I tried to make it as coherent as possible

ksar-fd:
Any updates abt this?

renpytom:
Scheduled to go in when I get some time to review it.

renpytom:
Question about this - did you write the types yourself, was this generated with a tool, or something else? 

Part of this is trying to figure out the right way to to deal with some of the formatting changes that are introduced in this PR. They make it a bit hard to review, and also I don't love some of the choices the formatter made. So I'm trying to figure out the right way to deal with that?

ksar-fd:
Yeah, I annotated the types myself, using my best guess based on how the function/class/variable is used.

I apologise for the formatting. I use  [black](https://github.com/psf/black) for automatic formatting, which I imagine isn't ideal since it's very opinionated and allows for little configuration. On the other hand, [ruff](https://docs.astral.sh/ruff/) allows for fine grained configuring, so it would be more ideal for what you're trying to do.

renpytom:
Were you able to run this, especially on Python 3.12? I'm noticing a number of errors, ranging from undefined types being used to syntax errors, like: 

https://github.com/renpy/renpy/pull/6216/files#diff-220e3cc26db85b053ce4e42e2b8a45d825782497cb198d12a9f50aef3f5f48adR320

ksar-fd:
Yeah, I think it would be best if I split this up in a bunch of smaller PRs, so it's easier to manage on both ends

renpytom:
I'd suggest waiting on that. We're putting together some guidelines as to how Ren'Py typing should work, as we need a typing scheme that can last for a long time. That's going to be required for typing work to continue.

ksar-fd:
Alrighty

> https://github.com/renpy/renpy/pull/6216/files#diff-220e3cc26db85b053ce4e42e2b8a45d825782497cb198d12a9f50aef3f5f48adR320

And I'll test the changes and roll out fixes for typos like these

PR #6224, opened 2025-03-16
Create mesh render using post-padding size

### LINKED ISSUE #6223: Transform's mesh_pad property can result in clipping
```rpy
show expression Solid('ff73', xysize=(320, 240)) at Transform(align=(0.5, 0.5), mesh=True, mesh_pad=(50,50,50,50)), Transform(mesh=True)
```

mal:
Attempts to fix #6223.

This may be incomplete as I'm uncertain if the change in size needs to be pushed into `self.width`, `self.height`, `transform.child_size` and/or `transform.render_size` as this seems to occur only if not using `perspective`.

See
https://github.com/renpy/renpy/blob/1e0b44a1721377bd4b4b3059ce2eb50142a5c254/renpy/display/accelerator.pyx#L1007-L1014
vs
https://github.com/renpy/renpy/blob/1e0b44a1721377bd4b4b3059ce2eb50142a5c254/renpy/display/accelerator.pyx#L1068-L1078

Filed against `master` as it's untested, potentially incomplete, and pre-releases are already rolling, but can rebase if appropriate.

### DIFF
--- a/renpy/display/accelerator.pyx
+++ b/renpy/display/accelerator.pyx
@@ -207,9 +207,6 @@ cdef class RenderTransform:
         Handles the mesh, mesh_pad, and blur properties.
         """
 
-        # The render we're going to return.
-        mr = Render(cr.width, cr.height)
-
         mesh = self.state.mesh
         blur = self.state.blur
         mesh_pad = self.state.mesh_pad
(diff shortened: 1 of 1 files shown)


mal:
Closing for now as I think this is only the tip of a bigger problem.

The `mesh_pad` property is already of above average difficulty to use, as the top and left values cause a shift in position. The size change makes this worse as it further breaks positioning in even harder to adjust for ways. For example an `xanchor` of `0.6` will be different depending on `mesh_pad`, which is non-obvious and rather painful if you're just temporarily wanting to add a blur to sprite.

Ideally `mesh_pad` would be reworked in such a way that it expands the area given to shaders, without a creator having to worry about how it impacts positioning, however that's a bigger job ­— assuming it's possible at all.

PR #6250, opened 2025-04-02
Add callback for when persistent file is loaded

mal:
A callback is needed for this as a `persistent` file can be loaded at any time by the save-scanning thread. The most common example of when this would occur is immediate after using Ren'Py Sync to retrieve remote save files, but can also occur if a player drops a newer `persistent` file into one of the various save locations observed by the scan thread.

The callback should be set "early", and has been added to `renpy.ast.EARLY_CONFIG` and documented accordingly. This is so it is already in-place by the time `renpy.persistent.init()` is called in `renpy.main.main`.

In addition `persistent._version` is set to the current version of the game, if is not already set. This felt like a good addition which seeks to parallel the behaviour of `store._version` defined in normal save files, which can be useful for debugging and data migrations. Note that an explicit `is None` check is used as the default version string is `""`.

### DIFF
--- a/renpy/ast.py
+++ b/renpy/ast.py
@@ -71,6 +71,7 @@ EARLY_CONFIG = {
     "defer_tl_scripts",
     "munge_in_strings",
     "interface_layer",
+    "persistent_callback",
 }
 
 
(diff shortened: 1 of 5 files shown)


renpytom:
Can you add something to the changelog about this? Other than that, looks good.

mal:
Added it to changelog in the features section.

PR #6252, opened 2025-04-04
Reduce number of times persistent files are loaded during boot

mal:
More detailed rationale in the respective commits, but here's the headlines:

- Right now a typical boot performs six persistent data load operations (user x3, game x2, sync x1), even though most of this data is ignored due to `mtime` checks.
- The first commit reuses `FileLocation` objects to reduce this to four (user x2, game x1, sync x1).
- Second commit uses the established `mtime` behaviour to cut this down to between one and three (user x1, game x1, sync x1), by skipping game and sync loads if they don't have a newer `mtime`. Anecdotally two was most common, then three, then one. This was due to the write order of files (and thus mtime), and prompted the next commit.
- Finally the third commit reverses the order of `FileLocation`s when saving persistent data. This ensures the user location will always have the newest `mtime` (short of outside interference). The end result is that a typical boot will only need to load a single persistent file (the one from the user location), but will still do the correct thing with other locations when needed.

### DIFF
--- a/renpy/persistent.py
+++ b/renpy/persistent.py
@@ -260,6 +260,8 @@ def init():
     disk, so that we can configure the savelocation system.
     """
 
+    global persistent_mtime
+
     if renpy.config.early_developer:
         init_debug_pickler()
 
@@ -272,6 +274,8 @@ def init():
 
     if persistent is None:
         persistent = Persistent()
+    else:
+        persistent_mtime = os.path.getmtime(filename)
 
     # Create the backup of the persistent data.
     for k, v in persistent.__dict__.items():
--- a/renpy/savelocation.py
+++ b/renpy/savelocation.py
@@ -694,17 +694,26 @@ def init():
 
     location = MultiLocation()
 
+    # Reuse locations when possible.
+    if current := renpy.loadsave.location:
+        reusable = {fl.directory: fl for fl in current.locations}
+    else:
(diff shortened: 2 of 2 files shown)


renpytom:
> Second commit uses the established mtime behaviour to cut this down to between one and three (user x1, game x1, sync x1), by skipping game and sync loads if they don't have a newer mtime. Anecdotally two was most common, then three, then one. This was due to the write order of files (and thus mtime), and prompted the next commit.

I don't think this is correct. 

Take the case where the player: 

* Disconnects from the internet on computer A. 
* Plays on computer B, and that get's put on Steam Sync. 
* Plays on computer A, still disconnected.
* Reconnects computer A, and starts the game. 

I believe in that case, the steam sync data will be out of date, and cause a problem. For correctness's sake, I'd like to load and merge all three persistent files on each game start.

mal:
To the best of my understanding, current operation is to ignore anything older than has already been seen, which is behaviour this PR sought to retain. The relevant condition can be found in `renpy.persistent.update`:

https://github.com/renpy/renpy/blob/dff95f273e77d2b24dff9765ddca229789147876/renpy/persistent.py#L450-L451

This results in the data from game, and sync files being discarded if they're older than the user file when update is called.

If you're proposing changing that behaviour, then I think I'd prefer to tackle it in a separate PR after this.

Issue #6261, opened 2025-04-09
Investigate potential mipmap/texture scaling related regression


mal:
The change in https://github.com/renpy/renpy/commit/dff95f273e77d2b24dff9765ddca229789147876 appears to have had some performance impact in certain scenarios.

Observations:

- Last known good: `8.4.0.25040601+nightly` (~60fps).
- First known bad: `8.4.0.25040604+nightly` (~40fps).
- Flipping the values of `renpy.gl2.gl2uniform.default_texture_scaling` immediately restores 60fps.
- Setting `mipmap True` in the `default` style does not appear to have any noticeable impact.
- Using `linear` as the default, rather than `linear_mipmap_nearest`, also seems to allivate this.
  However of the six scaling options available, only `linear` appears to improve things.
- Using a global transform to change texture scaling at the uppermost level also appears to prevent this:
  ```
  config.layer_transforms[None] = [Transform(gl_texture_scaling='linear')]
  ```

PR #6263, opened 2025-04-10
Remove implicit override of texture_scaling

mal:
Potential fix for https://github.com/renpy/renpy/issues/6261.

The scaling filters of a texture are now left for the texture itself to determine, unless explicitly overridden. This allows the MIN filter to be based on whether mipmaps are available.

You can see the existing code for setting the scaling filters per texture here:
https://github.com/renpy/renpy/blob/f0cb787964544a08c5fd665631047a991b9397e2/renpy/gl2/gl2texture.pyx#L604-L607

Previously, even if a texture had realised that it should use `GL_LINEAR` rather than `GL_LINEAR_MIPMAP_NEAREST` due a lack of mipmaps, the default would override that in the `Sampler2DSetter`.


mal:
Revised the approach to ensure that the defaults determined by the texture are correctly restored in cases where it's no longer overridden by the context. This is now ready for review.

Issue #6192, opened 2025-02-27
"gui" variables not translated everytime


Bibifoc95:
I use a **translate python** block to translate gui variables (in my case for a font change).

```
define gui.ft_james_fajardo = "James Fajardo.ttf"

translate french python:
    gui.ft_james_fajardo = "Hakim Ghazali.ttf"
```

1/ I run Renpy
2/ I switch language to French in Preferences (or the game is already in French)
3/ I start a New game => Everything's fine, the font is translated
4/ I return to the main menu (or the game just ends)
5/ I immediately restart a new game => The font isn't translated anymore

It's even worse while loading an existing game. Everytime I load a game, the font isn't translated.

Tested in 8.3.4 and last nightly fix.

Here is a small example project:
[game.zip](https://github.com/user-attachments/files/19016812/game.zip)

Bibifoc95:
I don't know if the fixe is done? I tried to apply it on a 8.3.4. It works well when a new game is started. But there are 2 problems with loaded game:
1/ gui variables translation are still not applied
2/ I get an error each time a dialogue line is displayed (only when I load a game made with a language different from the default one - English in my case):

> I'm sorry, but an uncaught exception occurred.
> 
> While running game code:
>   File "game/tl/french/script.rpy", line 7, in script
>     "Before the translated sentence."
> AttributeError: 'StoreModule' object has no attribute '_history_list'
> 
> -- Full Traceback ------------------------------------------------------------
> 
> Full traceback:
>   File "game/tl/french/script.rpy", line 7, in script
>     "Before the translated sentence."
>   File "E:\oracle\linux\disk3\_Renpy\renpy-8.3.4-sdk\renpy\ast.py", line 2579, in execute
>     Say.execute(self)
>   File "E:\oracle\linux\disk3\_Renpy\renpy-8.3.4-sdk\renpy\ast.py", line 621, in execute
>     renpy.exports.say(who, what, *args, **kwargs)
>   File "E:\oracle\linux\disk3\_Renpy\renpy-8.3.4-sdk\renpy\exports\sayexports.py", line 132, in say
>     who(what, *args, **kwargs)
>   File "E:\oracle\linux\disk3\_Renpy\renpy-8.3.4-sdk\renpy\character.py", line 1453, in __call__
>     self.add_history("current", who, what, multiple=multiple)
>   File "E:\oracle\linux\disk3\_Renpy\renpy-8.3.4-sdk\renpy\character.py", line 1543, in add_history
[...]

PR #6266, opened 2025-04-12
Support for direct style access in screens

### LINKED ISSUE #6265: Fully support direct style access in screens
**Test case**
```rpy
screen foo():
    window:
        background style.qux.background

style qux:
    background '#300'
```

**Previous Behaviour**
Prior to 45352847ab94b8e41592740aac9ba6913d4f7eef this code would work fine in a single-language scenario. However, it would result in a crash if a language change was attempted:
```
Exception: Style 'qux' does not exist.
```

**Current behaviour**
Following 45352847ab94b8e41592740aac9ba6913d4f7eef the game will crash on launch with the same exception as above.

**Proposed behaviour**
Simply that we support this and guard against the multi-language pitfall present in previous versions.

Looking at this from a creator perspective, expecting this use case to work doesn't seem unreasonable, and any alternative would be both less obvious and require them to scour the documentation to find out why it doesn't work.

mal:
Fixes #6265.

The primary change is that it skips screen preparation during the first call to `renpy.style.rebuild` in `renpy.translation.change_language`. This is done because otherwise, having just reset styles, they're not available for direct access in screens, causing the exception seen in the motivating issue. The justification for this is that later in `change_language` we `rebuild` styles again, this time after having installed all `deferred_styles`, and this call we allow to prepare screens. Even without the broader change to behaviour, only preparing screens in the later call feels like it could be a nice change.

The second change needed is to undo a small part of https://github.com/renpy/renpy/commit/45352847ab94b8e41592740aac9ba6913d4f7eef, restoring the call to `store._init_language` in `renpy.translation.init_translation`. This ensures styles are fully initialised when screens a prepared a little later in `renpy.main.run`.

~~Finally, in order to avoid doing the same work twice on boot, adding the `force` argument to `change_language` in `store._init_language` has been reverted. This has a side benefit of eliminating the click lag seen in games with many screens and/or styles when starting a new game, loading a save, or launching a replay. That lag was introduced by the commit referenced in the paragraph above.~~

While testing this I went back to the issue that motivated 45352847ab94b8e41592740aac9ba6913d4f7eef, https://github.com/renpy/renpy/issues/6192, and used that report to help inform the test game I created below. Best I can determine, the result of this change adds the support for direct style access in screens without introducing a regression related to that issue, and ~~clears up the click lag explained in the paragraph above~~.

Here is the test game for this feature, it covers both traditional and direct style use, as well as the use of `gui` variables and language specific overrides to define them. In the default (`None`) language, you see a red shape followed by a yellow shape, while in the other (`foo`) language you see a blue shape followed by magenta.

Scenarios tested:
Test | Expect | `None` | `"foo"`
:-|:-|-:|-:
Finish game, start new | retain language | ✔️ | ✔️
Save game, start new, switch, load | use new language | ✔️ | ✔️

<details><summary>Test Game</summary>

```rpy
define gui.abc = "#f775"
define gui.xyz = "#ff75"

translate foo python:
    gui.abc = "#7ff5"
    gui.xyz = "#f7f5"

screen test1():
    window:
        align (.5, .4) xysize (300, 200)
        style 's1'
        text 'Test1' align (.5, .5) size 40
        text 'Red' align (.1, .9)
        text 'Blue' align (.9, .9)

    dismiss action Return()

    textbutton 'Default' action Language(None) align (.25, .6)
    textbutton 'Other' action Language('foo') align (.75, .6)

screen test2():
    window:
        align (.5, .4) xysize (300, 200)
        background style.s2.background
        text 'Test2' align (.5, .5) size 40
        text 'Yellow' align (.1, .9)
        text 'Magenta' align (.9, .9)

    dismiss action Return()

    textbutton 'Default' action Language(None) align (.25, .6)
    textbutton 'Other' action Language('foo') align (.75, .6)

style s1:
    background gui.abc

style s2:
    background gui.xyz

label start:
    call screen test1()
    pause 0
    call screen test2()
    return
```
</details>

### DIFF
--- a/renpy/style.pyx
+++ b/renpy/style.pyx
@@ -804,7 +804,7 @@ def rebuild(prepare_screens=True):
 
     renpy.display.screen.prepared = False
 
-    if not renpy.game.context().init_phase:
+    if prepare_screens and not renpy.game.context().init_phase:
         renpy.display.screen.prepare_screens()
 
     renpy.exports.restart_interaction()
--- a/renpy/translation/__init__.py
+++ b/renpy/translation/__init__.py
@@ -696,6 +696,8 @@ def init_translation():
 
     load_all_rpts()
 
+    renpy.store._init_language()
+
 
 old_language = "language never set"
 
@@ -779,7 +781,7 @@ def change_language(language, force=False):
     tl = renpy.game.script.translator
 
     renpy.style.restore(style_backup) # @UndefinedVariable
-    renpy.style.rebuild() # @UndefinedVariable
+    renpy.style.rebuild(False) # @UndefinedVariable
 
     for i in renpy.config.translate_clean_stores:
         renpy.python.clean_store(i)


mal:
Closing a favour a new effort directed at reducing the click lag issue, rather than direct style access — which turned out to be a not so great idea as [explained](https://github.com/renpy/renpy/issues/6265#issuecomment-2799065056) in the original issue.

PR #6274, opened 2025-04-14
Initial implementation of PEP 657

Andykl:
This PR adds support for [Fine Grained Error Locations](https://peps.python.org/pep-0657/) in Ren'Py error handling.
That allow to render different color in exception screen and precice carets in traceback file.
In this PR I adopted some code from Python 3.14 traceback module, adding handling of Ren'Py `_ren.py` extra indent and munged names, and also added ExceptionPrintContext abstract class that actually emits output to abstract printer and implemented it for console attached to the process, traceback.txt file and exception screen.
file
console
screen

Things to fix before it can be merged:
1. Handle non-ASCII text in lines.
2. Handle wide chars in console and text file.
3. Handle SyntaxErrors in new system.
4. Update Ren'Py console to use new system.
5. Test it thoroughly.

### DIFF
--- a/renpy/bootstrap.py
+++ b/renpy/bootstrap.py
@@ -19,15 +19,11 @@
 # OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 # WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-from __future__ import division, absolute_import, with_statement, print_function, unicode_literals
-from renpy.compat import PY2, basestring, bchr, bord, chr, open, pystr, range, round, str, tobytes, unicode # *
-
-from typing import Optional
-
 import os
 import sys
 import subprocess
 import io
+import textwrap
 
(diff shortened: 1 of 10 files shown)


Andykl:
This is ready for review. This should keep formatting almost the same but add carets/coloring in 4 places user can see it:
1. `traceback.txt` file.
2. Console output
3. Exception screen
4. Console

So for the code
```py
#game/libs/lib_ren.py
"""renpy
init python in my_lib:
"""

def __internal_function(f, *args, **kwargs):
    return f(*args, **kwargs)

def non_ascii_функция(f, *args, **kwargs):
    return __internal_function(f, *args, **kwargs)

def cjk_函数(f, *args, **kwargs):
    return non_ascii_функция(f, *args, **kwargs)

def call_me(f, *args, **kwargs):
    return cjk_函数(f, *args, **kwargs)
# script.rpy
init python:
    def f(a, b):
        return a / b

label start:
    ...
    $ a = my_lib.call_me(f, 1, 0)
```
this produces following output
image
image
image
image

Issue #6249, opened 2025-04-02
Marks on ligatures are not displayed correctly


TDCMC:
Ligatures that are supposed to have marks on them (For example Arabic Lam and Arabic Alef) have more than one anchors, one for the first letter, one for the second, and so on. Only the first anchor gets used and every other one gets completely ignored. The result is, the mark always being displayed on top of the first letter in the ligature.

renpytom:
Do you know if this is working right in harfbuzz?

TDCMC:
I'm pretty sure libreoffice, inkscape, firefox and chromium all use harfbuzz. And in libreoffice it does work correctly so yes, harfbuzz itself does support this.

TDCMC:
I read up on the harfbuzz documentation on this and it seems that the solution to this problem is a bit complicated. Upon reordering, the cluster value for a ligature is set just like a normal glyph, but the cluster value for marks are set every other number after that, in a way that the values for where the ligature glyphs are supposed to go are skipped. So a solution that I think of is checking if a value doesn't exist, then holding the last continuous value and applying the non continuous "marks" according to it.

khaledhosny:
You don't need to do anything special to get marks correctly placed over ligatures with HarfBuzz. If it is not working, then most likely HarfBuzz is not used correctly.

TDCMC:
I would assume harfbuzz itself would set the glyph advances and positions. In that case shouldn't the text be completely positioned by harfbuzz? This problem shouldn't be happening in the first place.

renpytom:
At least right now, Ren'Py's harfbuzz support is a mix of the old font renderer, and what harfbuzz produces. That's because Ren'Py gives the creator a lot of control over text layout, and we can't completely give that up.

TDCMC:
But from what I can see in hbfont.pyx, the positions are still calculated by harfbuzz, no? Are those not being used, or are they being adjusted incorrectly?

TDCMC:
Okay, I just realised something. If I put a mark after the first character in the ligature and another mark after the second character, the marks both appear on the correct spots. So the problem is that it's always prioritising the first anchor over every other one.

khaledhosny:
Do you have a sample text and a font to better understand the issue you are seeing?

khaledhosny:
I think this is because renpy calls fribidi’s log2vis then set text direction to LTR when calling HarfBuzz. HarfBuzz needs the text to be in logical order and direction to be set properly to work correctly. See discussion in #6188

[...]

PR #6280, opened 2025-04-16
Prevent shaping the text using fribidi when using harfbuzz

TDCMC:
This resolves (https://github.com/renpy/renpy/issues/6249) and removes the need to unmap Arabic glyphs when using harfbuzz. It also removes the `l.reverse()` line from `rtl_paragraph()` since it appears to do nothing.


TDCMC:
Closing this to for (https://github.com/renpy/renpy/pull/6286) to be merged into the master branch.

PR #6285, opened 2025-04-19
Fix for recent literals optimisation

mal:
Also use literal `NoneType` over `type(None)`.

**Using the console in launcher**

Before this change:
```
> sum(type(i) is tuple for i in renpy.python.py_compile_cache.values())
0
```

After this change:
```
> sum(type(i) is tuple for i in renpy.python.py_compile_cache.values())
50595
```


mal:
Resolved in 4bd4cbc0e5bdee7a370b2a95362830a2f7d103b9.

PR #6286, opened 2025-04-19
Prevent shaping the text using fribidi when using harfbuzz

TDCMC:
This resolves (https://github.com/renpy/renpy/issues/6249) and removes the need to unmap Arabic glyphs when using harfbuzz. It also removes the l.reverse() line from rtl_paragraph() since it appears to do nothing.


mal:
This appears to include unintentional changes from the `fix` branch.

TDCMC:
Oh oops, sorry about that. I will change those.

renpytom:
Thank you for the contribution.

PR #6289, opened 2025-04-19
Layeredimage 2.0

Gouvernathor:
Updates some aspects of the Layeredimage syntax and adds minor features. See https://github.com/Gouvernathor/layeredimage2 for a more detailed rationale.

The only incompatibility, which is so slight as to be virtually impossible, is documented in the incompat page with a workaround. All the rest, including the multiple group change, is fully backwards-compatible.

- Adds a new anonymous group syntax for multiple groups, intended to be the only documented one, which makes the behavior of auto multiple groups more consistent (if using this new syntax, otherwise the behavior is unchanged).
- The prefix and variant properties now take names without quotes, instead of requiring an expression.
- Attributes can take a variant, if they don't have a provided displayable and are not in a group with a variant.
- if_any, if_all and if_not are deprecated (but still supported), and replaced with a new if_attr with a new, more straightforward syntax which is both more readable and allows more expressions.
  - The Python equivalent still lists them as if_attr has no documented Python equivalent, but it's not a hard thing to do.
- Adds `at transform:` where `at` was, and `image:` where a displayable is expected, both taking ATL (adding `pass` everywhere was already done).
- Also some parts of the code have been reorganized, modernized, adapted to PY3...

- [x] Introduce the changes on the changelog and incompat pages
- [ ] Update the main layeredimage documentation page
- [ ] Add a IfAttr python equivalent and drop the if\_ properties for the python equivalents too


renpytom:
Question - do prefix and variant also take names with quotes? So that the old values will still  work?

Gouvernathor:
Yep, it is still supported, though not documented.

PR #6293, opened 2025-04-21
Enable console output in the Mac OS Launcher

JamieDerg:
I noticed that console output isnt available on Macos, so i decided to make my own fix.


renpytom:
Thanks for the feature, this will be in 8.4.

PR #6308, opened 2025-04-30
call `main_window.update` on `pump_window`

qTich:
Main window ignored on `renpy.display.presplash.pump_window`.
But `window` exists on `reload` or `load_language`.

### DIFF
--- a/renpy/display/presplash.py
+++ b/renpy/display/presplash.py
@@ -160,24 +160,25 @@ def pump_window():
     global last_pump_time
     global pump_count, pump_total
 
-    pump_count += 1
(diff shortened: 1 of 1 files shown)


mal:
Everything that can be after the `last_pump_time` check (such as the main window fetch) should be, as it exists to prevent this function impacting boot times.

What problem is this change seeking to address?

renpytom:
@qTich You need to write better justifications if you want to get things merged. We shouldn't have to figure out what problem you're trying to solve here. 

I don't get why there is a call to main_window.update here. That function doesn't do much with an opengl window when OpenGL isn't being run.

Issue #6284, opened 2025-04-18
Shape text in segments determined by the embedding levels


TDCMC:
Implementing this should ultimately fix every issue with bidirectional text. The problem right now is that every "line" of text is shaped as a whole. This causes several problems:
- Having any kind of LTR text in the middle of RTL text breaks the RTL text shaping
- Some special characters like U+200D (Zero-width joiner) need to precede a character to have the effect of succeeding that character.

This (https://github.com/renpy/renpy/pull/6286) pull request should take a step by reordering but not shaping the text when using "harfbuzz".
After this, the next step would be to not reorder the text, but split the text into segments with different directions and shape each of them with the correct direction based on the embedding levels. Currently, after reordering the text, renpy passes it to "harfbuzz" in LTR mode, regardless of which direction it is; and this causes the problems described above. After the shaping, the RTL text can then be reversed for applying LTR text algorithms that renpy has.

TDCMC:
As additional info for the actual implementation:
`fribidi_get_embedding_levels_ex()` takes a string and also the variable to store the embedding levels in. Simply put, every adjacent letter with the same direction has the same embedding level and when a segment in the string has a different direction, the embedding level for that segment differs. Also, according to the unicode consortium, the embedding levels for RTL text is an odd number, and for LTR text, an even number. So by using that info, we can then split the texts into segments, and then pass them to harfbuzz with the direction of the text. If `vertical` is set for a text, the odd number in the levels would represent BTT text, and the even number would represent TTB. The only problem with this is that renpy might need to change how it calculates the text positions because it would be broken into multiple segments with different starting positions.

PR #6313, opened 2025-04-30
Fix bidirectional paragraphs

TDCMC:
Addresses the issues mentioned in (https://github.com/renpy/renpy/issues/6284).
This pull request will eventually contain the following changes:
- Split paragraphs into segments based on the direction.
- Shape each segment (using harfbuzz) according to its direction
- Adjust any position advancement/offset issues that might occur

**After this pull request is finished, a full clean-up of the written code will be needed**

### DIFF
--- a/renpy/text/text.py
+++ b/renpy/text/text.py
@@ -1585,13 +1585,26 @@ class Layout(object):
         direction = ON
 
         l = [ ]
+        levels = [ ]
+        seg_dir = [ ]
+        seg_str = [ ]
+        curr_seg = 0
 
         for ts, s in p:
             ft_enabled = (getattr(ts, "shaper", "") != "harfbuzz")
 
-            s, direction = log2vis(str(s), ft_enabled, direction)
-            
-            l.append((ts, s))
+            s, direction, levels = log2vis(str(s), levels, ft_enabled, direction)
+            seg_dir.append(levels[0]%2)
+            seg_str.append(s[0])
+            for i in range(1, len(s)):
+                if seg_dir[curr_seg] != levels[i]%2:
+                    seg_dir.append(levels[i]%2)
+                    seg_str.append(s[i])
+                    curr_seg += 1
+                else:
+                    seg_str[curr_seg] += s[i]
+            for ss in seg_str:
+                l.append((ts, ss))
 
         rtl = (direction == RTL or direction == WRTL)
         
--- a/src/_renpybidi.pyx
+++ b/src/_renpybidi.pyx
@@ -30,7 +30,7 @@ cdef extern from "fribidi/fribidi.h":
 
 cdef extern from "renpybidicore.h":
     object renpybidi_log2vis(object, int *)
-    object renpybidi_reorder(object, int *)
+    object renpybidi_get_embedding_levels(object, int *, object)
 
 WLTR = FRIBIDI_TYPE_WL
 LTR = FRIBIDI_TYPE_LTR
@@ -39,10 +39,10 @@ RTL = FRIBIDI_TYPE_RTL
 WRTL = FRIBIDI_TYPE_WR
 
 
-def log2vis(s, shapetext=True, int direction=FRIBIDI_TYPE_ON):
+def log2vis(s, levels=[], shapetext=True, int direction=FRIBIDI_TYPE_ON):
 
     if shapetext:
         s = renpybidi_log2vis(s, &direction)
     else:
-        s = renpybidi_reorder(s, &direction)
-    return s, direction
+        s = renpybidi_get_embedding_levels(s, &direction, levels)
+    return s, direction, levels
--- a/src/renpybidicore.c
+++ b/src/renpybidicore.c
@@ -35,52 +35,36 @@ PyObject *renpybidi_log2vis(PyObject *s, int *direction) {
     return PyUnicode_FromKindAndData(PyUnicode_4BYTE_KIND, dstuni, size);
 }
 
-PyObject *renpybidi_reorder(PyObject *s, int *direction) {
+PyObject *renpybidi_get_embedding_levels(PyObject *s, int *direction, PyObject *seg_levels) {
     Py_ssize_t size;
     FriBidiChar *srcuni;
-    FriBidiLevel *levels;
     FriBidiCharType *types;
-    FriBidiLevel ret;
+    FriBidiLevel *levels;
 
     PyUnicode_READY(s);
     size = PyUnicode_GET_LENGTH(s);
 
     srcuni = (FriBidiChar *) alloca(size * 4);
-    levels = (FriBidiLevel *) alloca(size * 4);
     types = (FriBidiCharType *) alloca(size * 4);
+    levels = (FriBidiLevel *) alloca(size * 4);
 
     PyUnicode_AsUCS4(s, (Py_UCS4 *) srcuni, size, 0);
-
+    
     fribidi_get_bidi_types(
         srcuni,
         size,
         types);
-
-    ret = fribidi_get_par_embedding_levels_ex(
+    
+    fribidi_get_par_embedding_levels_ex(
         types,
         NULL,
         size,
         (FriBidiParType *) direction,
         levels);
-
-    if (ret == 0) {
-        return s;
-    }
-
-    ret = fribidi_reorder_line(
-        FRIBIDI_FLAG_REORDER_NSM,
-        types,
-        size,
-        0,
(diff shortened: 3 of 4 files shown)


TDCMC:
Currently, the `TextSegment` class doesn't have an `rtl` or`direction` property. Doing it the way I can think of, which is setting the direction of harfbuzz to that of a text segments' requires such a thing. @renpytom do you have any opinion on if this is even a good way of doing things?

renpytom:
TextSegment is meant for information from the style system and text tags - it currently doesn't store any information that's being computed during layout. 

Would it make sense for <whatever>Segment.glyphs to take the text direction as an argument, and return a new text direction in addition to the list of glyphs, which would then passed to the next segment's glyphs function?

TDCMC:
As far as I understand, that would be viable, but just in case, I'll explain what I intend to do.
Previously, `rtl_paragraph()` returned a segment list with one member. With the changes I made in the commits, it returns multiple a list with multiple segments and a direction for each one. I want the text shaping of each segment to be seperate, in a way that the harfbuzz buffer direction for rtl, ltr, ttb, and btt segments would be set to `HB_DIRECTION_RTL`, `HB_DIRECTION_LTR`, `HB_DIRECTION_TTB` and HB_DIRECTION_BTT` respectively. But right now, I don't have a way of accessing the direction of the text segment in HBFont.glyphs.

renpytom:
I think I'd be fine with adding a field to TextSegment to store this information. Just have __init__ initialize it.

TDCMC:
Will do, thanks.

TDCMC:
I have encountered a problem. In `rtl_paragraph()` a for loop iterates through each text segment and sets the direction attribute. However, it seems that the way I did it, which is using the same `ts` of each `p` (paragraph) for every segment in that same paragraph, results in every segment containing the same reference, and thus having the same text property including direction. Any ideas on how to fix this?

mal:
Is the only reason to add the information to the segments to set it on the glyphs later?

TDCMC:
No, glyphs get their direction set already. But (to my ability) those cannot be accessed to be used to set the buffer direction. If they can, that would also work.

TDCMC:
I did manage to fix the problem mentioned above, but the properties passed onto harfbuzz still remain the same for some reason. I might need to dig deeper into how renpy handles the text.

mal:
Curiosity has has me poking around this issue a little, but I'm not very familiar with RTL languages. Do you have any test strings you've been using, especially any that make use of the AN and/or EN embedding levels?

TDCMC:
Sure, I can give some:

"This is a نوشته in rtl."
image

"سلام player چطوری؟"

Screenshot From 2025-05-08 23-29-24

I can write longer sample texts that have even more embedding levels in them, but for the general idea, these do the trick.
I also wrote them with DejaVu fonts in the pictures so it would be a bit easier to see. Beyond that, since Arabic letters have different forms, it's not that simple to figure out what exactly is wrong with the displayed text.

TDCMC:
"من امروز 25 تا apple خریدم."
Screenshot From 2025-05-08 23-43-00

"I put ۲۵ apples in my bag."
image

These are also weird examples of text containing EN and AN in the middle of the text.

mal:
Thanks, those examples were a great help!

I've taken a pass at this, building on your Cython and C changes, but with a slightly different approach in `text.py`, I believe it behaves correctly for all the examples, including when using slow text. See what you think.

https://github.com/renpy/renpy/compare/master...mal:6313-attempt

<details><summary>And here's the script I was using to test it out.</summary>

```rpy
define config.rtl = True

label start:
    "This is a \u0646\u0648\u0634\u062A\u0647 in rtl."
    "This is a \u0646\u0648\u0634\u062A\u0647 in rtl." (show_shaper='freetype')
    "\u0633\u0644\u0627\u0645 player \u0686\u0637\u0648\u0631\u06CC\u061F"
    "\u0633\u0644\u0627\u0645 player \u0686\u0637\u0648\u0631\u06CC\u061F" (show_shaper='freetype')
    "\u0645\u0646 \u0627\u0645\u0631\u0648\u0632 25 \u062A\u0627 apple \u062E\u0631\u06CC\u062F\u0645."
    "\u0645\u0646 \u0627\u0645\u0631\u0648\u0632 25 \u062A\u0627 apple \u062E\u0631\u06CC\u062F\u0645." (show_shaper='freetype')
    "I put \u06F2\u06F5 apples in my bag." 
    "I put \u06F2\u06F5 apples in my bag." (show_shaper='freetype')
    "When I type \u0644\u0627\u064B, a mark should be on the first glyph, and when I type \u0644\u064B\u0627\u064B, a mark should be on both."
    "When I type \u0644\u0627\u064B, a mark should be on the first glyph, and when I type \u0644\u064B\u0627\u064B  a mark should be on both." (show_shaper='freetype')
    return

screen say(who, what, shaper='harfbuzz'):
    style_prefix "say"

    window:
        id "window"

        window:
            id "namebox"
            style "namebox"
            text "using: [shaper]"

        text what id "what" shaper shaper
```
</details>

TDCMC:
> Thanks, those examples were a great help!
> 
> I've taken a pass at this, building on your Cython and C changes, but with a slightly different approach in `text.py`, I believe it behaves correctly for all the examples, including when using slow text. See what you think.
> 
> [master...mal:bidi-support-with-tdcmc](https://github.com/renpy/renpy/compare/master...mal:bidi-support-with-tdcmc)
> 
> Main goals were to:
> 
> * Avoid needing to create duplicate/copy `TextSegment` objects.
> * Avoid holding directionality as part of a font's state.
> * Slice `str`s rather than rebuild them using `+=`.
> 
> Was unsure if `Glyph.rtl` really needs to be set as I couldn't see any visual impact, but if so it could now be done easily when first constructed in `HBFont.glyphs`.
> 
> Finally, here's the script I was using to test it out.

Thank you very much for your hard work. As far as I can tell, it works flawlessly! There are some more changes I wanted to propose. Adding a `force_direction` text property that forces the general flow of the text to be LTR/TTB when set to 0 (default) and reversed (RTL/BTT) when set to 1. This would be useful for text that only contain weak characters like CS or ES characters (for example "...!" or "--?!"). These text normally default to LTR, which isn't correct for RTL text. Another proposal I have is changing `config.rtl` to `config.bidi` and making `config.rtl` an alias since with these changes, only calling it rtl would be misleading.

mal:
Good to hear! If it's okay with you, I'll file it a separate pull request that includes your initial work. 🙂 

As for forcing RTL, my initial though was that prefixing the string with either `\u202b`, or `\u202e` might be enough, and while this does render correctly, it doesn't switch the paragraph itself into RTL mode, so slow text doesn't get the correct timings. I suggest we investigate options for that tackle it as a separate feature.

As it stands `config.rtl` is to be read as an "enable RTL support" flag due to the extra processing that it requires, and so while I agree `bidi` might be more correct in the strictest sense, I'm not sure it's worth revisiting for that alone.

TDCMC:
> Good to hear! If it's okay with you, I'll file it a separate pull request that includes your initial work. 🙂
> 
> As for forcing RTL, my initial though was that prefixing the string with either `\u202b`, or `\u202e` might be enough, and while this does render correctly, it doesn't switch the paragraph itself into RTL mode, so slow text doesn't get the correct timings. I suggest we investigate options for that tackle it as a separate feature.
> 
> As it stands `config.rtl` is (I believe) meant to be read as an "enable RTL support" flag due to the extra processing that it requires, and so while I agree `bidi` might be more correct in the strictest sense, I'm not sure it's worth revisiting for that alone.

That's okay, I will close this pr as soon as you open the other one. Thank you again.

TDCMC:
> Good to hear! If it's okay with you, I'll file it a separate pull request that includes your initial work. 🙂
> 
> As for forcing RTL, my initial though was that prefixing the string with either `\u202b`, or `\u202e` might be enough, and while this does render correctly, it doesn't switch the paragraph itself into RTL mode, so slow text doesn't get the correct timings. I suggest we investigate options for that tackle it as a separate feature.
> 
> As it stands `config.rtl` is (I believe) meant to be read as an "enable RTL support" flag due to the extra processing that it requires, and so while I agree `bidi` might be more correct in the strictest sense, I'm not sure it's worth revisiting for that alone.

As for the RTL forcing, what happens if the direction returned by `glyphs_paragraph()` is changed by the style?

mal:
I've opened a new issue to track the preferred paragraph direction feature, and put some thoughts on a possible solution together for it. You can find it over in #6333. 🙂

PR #6320, opened 2025-05-03
Do not emit carets if the whole statement is an error.

Andykl:
This PR addresses the problem that new tracebacks with carets can be too noizy by not emitting them in cases where frame line is simple function call, which is majority of such noize. E.g. in following cases carets are not rendered
1. `expr.call(...)`
2. `return call(...)`
3. `assign = call(...)`
4. `aug_assign += call(...)`

But _are_ rendered if it is part of a more complex expression, e.g. `call(...) + call(...)`


renpytom:
The unmerged branch associated with this pull request has been renamed to work/andy_kl/pep-657-less-verbose at commit 776a7756e4b9b8b1b96da5c90080ae2e60164db4.

PR #6304, opened 2025-04-29
Preserve some niceties of older traceback style

mal:
- Drop "Traceback (most recent call last):" lines.
   We provide our own more specific context to traces already, no need for two.
- Prevent double indentation of stack frames in text file output.
  Fewer characters is nice in some scenarios (such as pasting traces into Discord).
  Also makes each increase in indentation level consistent, and matches how it appeared in 8,3 and previous.
- Restore breathing room around tracebacks on exception screen.
- Include the initial "File " of frames in the location hyperlink.
  I noticed this when scanning the left side of traces, having each line start the same colour was distracting in a way that I hadn't previously remembered, and I believe this was why.
  Also matches how we dynamically highlight parse errors:
  https://github.com/renpy/renpy/blob/a9dd010dc87ff065987639cb1231f9225066c54b/renpy/common/_errorhandling.rpym#L579-L582
- Fixes sizing on strangely oversized final line.

**Comparisons**

<details><summary>GUI</summary>

<table><tr><th>8.3</th><th>8.4-nightly</th><th>8.4-this-patch</th></tr><tr><td>

</td><td>

</td><td>

new

</td></tr></table>

</details>
<details><summary>Console</summary>

<table><tr><th>8.3</th></tr><tr><td>

```
I'm sorry, but an uncaught exception occurred.

While running game code:
  File "game/script.rpy", line 1, in script
    init python:
  File "game/script.rpy", line 2, in <module>
    import problem3
ZeroDivisionError: division by zero

-- Full Traceback ------------------------------------------------------------

Full traceback:
  File "game/script.rpy", line 1, in script
    init python:
  File "renpy/ast.py", line 827, in execute
    renpy.python.py_exec_bytecode(self.code.bytecode, self.hide, store=self.store)
  File "renpy/python.py", line 1178, in py_exec_bytecode
    exec(bytecode, globals, locals)
  File "game/script.rpy", line 2, in <module>
    import problem3
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 664, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 627, in _load_backward_compatible
  File "renpy/loader.py", line 864, in load_module
    exec(code, mod.__dict__)
  File "problem3.py", line 1, in <module>
ZeroDivisionError: division by zero

[...]

PR #6322, opened 2025-05-04
Fix context lines of chained exceptions

### LINKED ISSUE #6321: Chained exception context lines can get mixed up
Context lines for chained errors can end up unintended places if the primary exception has a private-only stack.

## Recreation

**game/problem3.py**
```py
try:
    1 / 0
except:
    1 / 0
```

**game/script.rpy**
```rpy
init python:
    import problem3
```

## Ren'Py 8.4.0
**Problem 3**
```
ZeroDivisionError: division by zero

During handling of the above exception, another exception occurred:

While running game code:
  File "game/script.rpy", line 1, in script
    init python:
  File "game/script.rpy", line 2, in <module>
    import problem3
ZeroDivisionError: division by zero
```
_The "While running game code:" line appears before the chained exception, rather that at the top of the section._

mal:
Revisits #6304 to simplify the approach to context lines, and in the process resolve #6321.

Effectively reverts 7a95e5be21934792a21048d9a15ac1e6cc0d4465, replacing it with a simpler alternative, and leaves context lines to the calling code as was previously the case.

### DIFF
--- a/renpy/common/_errorhandling.rpym
+++ b/renpy/common/_errorhandling.rpym
@@ -496,14 +496,14 @@ init python:
         BRIGHT_RED = "#e00000"
         RED = "#800000"
 
-        def __init__(self, *, filter_private: bool, context_line=None):
+        def __init__(self, *, filter_private: bool, emit_context=True):
             import io
             super().__init__(
                 io.StringIO(),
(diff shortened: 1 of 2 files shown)


Andykl:
LGTM.

PR #6323, opened 2025-05-04
Adjustment in config.bottom_layers docs

Quetzalicious:
The explanation for the following config variables is identical, motivating this adjustment.
- https://www.renpy.org/doc/html/config.html#var-config.bottom_layers
- https://www.renpy.org/doc/html/config.html#var-config.top_layers

This adjustment is a guess on my part, since I'm not intimately familiar with the source code and have no idea how to look it up.


Andykl:
Good catch. Thanks!

PR #6324, opened 2025-05-05
Error suggestions

Andykl:
This PR adds error suggestions to Ren'Py error handling. This works by adding `config.error_suggestion_handlers` (is hook a better term here?) dictionary from exception type to function that takes raised exception and returns suggestion text.
Also it adds handlers for AttributeError, NameError and ImportError from Python, and Ren'Py specific handlers for unknown label names, screen names and image tags (but not attributes).

### DIFF
--- a/renpy/error.py
+++ b/renpy/error.py
@@ -1092,6 +1092,188 @@ def _safe_string(value, what, func: Callable[..., str] = str):
         return f'<{what} {func.__name__}() failed>'
 
 
+def _compute_best_suggestion(values: list[str], value: str) -> str | None:
+    MAX_CANDIDATE_ITEMS = 1200  # Python uses 750, but renpy.store can contain a lot more.
+    MAX_STRING_SIZE = 40
+    MOVE_COST = 2
+    CASE_COST = 1
+
+    def substitution_cost(ch_a: str, ch_b: str):
+        if ch_a == ch_b:
+            return 0
+
+        if ch_a.lower() == ch_b.lower():
+            return CASE_COST
+
+        return MOVE_COST
+
+
+    def levenshtein_distance(a, b, max_cost):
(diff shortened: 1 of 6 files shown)


renpy/script.py
```diff
@@ -1117,7 +1136,10 @@ def lookup(self, label):
             rv = self.namemap.get(label, None)
 
         if rv is None:
-            raise ScriptError("could not find label '%s'." % str(original))
+            if isinstance(original, str):
+                raise LabelNotFound(original)
+            else:
+                raise ScriptError(f"could not find label '{original}'.")
```
mal, line 1142: Feels pretty icky to have the exception type governed by whether we're looking up a string vs anything else. Perhaps this check should reside in `LabelNotFound.get_suggestion`?

renpy/error.py
```diff
@@ -1170,6 +1358,21 @@ def __init__(self, exception: BaseException, /, _seen=None):
             self.msg = exception.msg
             self._is_syntax_error = True
 
+        else:
+            suggestion = None
+            # Reversed so that the most specific handler is called first.
+            for typ, handler in reversed(renpy.config.error_suggestion_handlers.items()):
```
mal, line 1364:
```suggestion
        elif handlers := renpy.config.error_suggestion_handlers:
            suggestion = None
            # Reversed so that the most specific handler is called first.
            for typ, handler in reversed(handlers.items()):
```
Would make it clear that this branch is for the sole purpose of suggestion, while also short-circuiting it completely if there are none installed or the feature has been disabled.

mal (commented):
Couple of small things.

Andykl:
Good suggestions.

I still don't know how to correctly handle misspeled attributes, but this could be added later.

PR #6325, opened 2025-05-06
Added extra information because of dubious use of "filename" instead of "save slot"

Quetzalicious:
For https://www.renpy.org/doc/html/save_load_rollback.html#renpy.can_load
- Returns true if `filename` exists as a save slot, and False otherwise.
+ Returns true if `filename` (e.g. '3-2' for page 3, slot 2) exists as a save slot, and False otherwise.


renpytom:
Merged, thank you.

PR #6332, opened 2025-05-09
Support embeded RTL when using harfbuzz as shaper

mal:
Builds on and supersedes #6313, fixes #6284.

Credit for figuring out the problem, and how to solve it goes to @TDCMC, I mostly just wrapped it in a bit of Python.

Main goals were to:
- Have RTL text in the middle of a LTR paragraph display correctly when using `harfbuzz`.
- Avoid needing to create duplicate/copy `TextSegment` objects.
- Avoid holding directionality as part of a font's state.
- Slice `str`s rather than rebuild them using per-character concatenation (`+=`).

<details><summary>Finally, here's the script I've been using to test it.</summary>

```rpy
define config.rtl = True

label start:
    "This is a \u0646\u0648\u0634\u062A\u0647 in rtl."
    "This is a \u0646\u0648\u0634\u062A\u0647 in rtl." (show_shaper='freetype')
    "\u0633\u0644\u0627\u0645 player \u0686\u0637\u0648\u0631\u06CC\u061F"
    "\u0633\u0644\u0627\u0645 player \u0686\u0637\u0648\u0631\u06CC\u061F" (show_shaper='freetype')
    "\u0645\u0646 \u0627\u0645\u0631\u0648\u0632 25 \u062A\u0627 apple \u062E\u0631\u06CC\u062F\u0645."
    "\u0645\u0646 \u0627\u0645\u0631\u0648\u0632 25 \u062A\u0627 apple \u062E\u0631\u06CC\u062F\u0645." (show_shaper='freetype')
    "I put \u06F2\u06F5 apples in my bag." 
    "I put \u06F2\u06F5 apples in my bag." (show_shaper='freetype')
    "When I type \u0644\u0627\u064B, a mark should be on the first glyph, and when I type \u0644\u064B\u0627\u064B, a mark should be on both."
    "When I type \u0644\u0627\u064B, a mark should be on the first glyph, and when I type \u0644\u064B\u0627\u064B  a mark should be on both." (show_shaper='freetype')
    return

screen say(who, what, shaper='harfbuzz'):
    style_prefix "say"

    window:
        id "window"

        window:
            id "namebox"
            style "namebox"
            text "using: [shaper]"

        text what id "what" shaper shaper
```

Examples of what some of the various lines should look like. Again, credit to @TDCMC:

441867139-24e53398-a99c-4f24-b512-d31440b07045
441867765-e30459c6-3d53-4f9d-ba6c-f11b3d0553d4
[...]

PR #6334, opened 2025-05-10
Reading order style property

### LINKED ISSUE #6333: Ability to set prefered text directionality
This is a brain-dump stemming from a discussion that started in #6313 between myself and @TDCMC.

**Brief summary**

When writing a script predominantly in a RTL language with `config.rtl` enabled, some lines (examples below) are determined to be LTR rather than RTL, and so are displayed as such despite not being the authors intention and with no good way to address the issue.

**Example lines**

`"...!"`
`"--?!"`

**Ideas**

Initially the use of [Unicode Directional Formatting Codes](https://learn.microsoft.com/en-us/globalization/fonts-layout/text-directionality#explicit-formatting-characters) was considered. ~~As of the time of writing this fixes the rendering of such lines, but does not alter the determined directionality of the paragraph (line) itself, which results in slow text still appearing LTR rather than RTL.~~

Further experimentation reveals that while `U+202B`, `U+202E` are unsuitable for this purpose, both `U+061C` and `U+200F` do work. i.e.

`"\u200f...!"`
`"\u200f--?!"`

**Possible approach**

A potential approach to this might be a `config` flag. As any such flag would only have an effect when `config.rtl = True`, it may make sense to use the previously proposed `config.bidi` for this, and deprecate `config.rtl`, i.e.

`config.bidi` | `config.rtl`<br>equivalent | description
-|-|-
`None` (default) | `False` | Unconditionally assume all text is LTR, current default behaviour.
`neutral` | `True` | Use current method to determine directionality based on content.
`ltr` | N/A | Use `\u200e` prefix to prefer LTR directionality while still supporting embedded RTL runs.
`rtl` | N/A | Use `\u200f` prefix to prefer RTL directionality while still supporting embedded LTR runs.

This `config` could potentially be set in a `translate python` block in much the same way we currently expect `config.rtl` to be.

mal:
Fixes #6333.

Introduces `reading_order` style property to govern paragraph level directionality when `config.rtl = True`. Named `reading_order` as this is what Notepad calls it (right click in text area, see "Right to left Reading order" toggle).

This does not alter how LTR and RTL characters themselves are rendered, but rather operates on the paragraph/line level. Using the value of `"rtl"` would force lines to be treated as RTL, for example meaning that slow text would always render in from the right, even if the characters being rendered are LTR. There are five possible values available: `None`, `"ltr"`, `"rtl"`, `"wltr"`, and `"wrtl"`.

Realistically, most people using this feature will use either `None`, the default which decides directionality based on the characters in the text (i.e. defers to `fribidi`), or `"wrtl"` (weak right-to-left) to ensure that directionally neutral lines (such as `..?!`) are treated as RTL ensuring a visual consistency with other dialogue when using an RTL language. Other values are included for completeness, and to pre-emptively support potential edge-cases (i.e. needing to force RTL on a line that starts with Latin characters, or conversely, needing to force LTR on a line that starts with Arabic characters.).

Additionally there a small bit of clean up and optimisation of work done in #6332, as well as documentation for both that change and this one.


TDCMC:
Another thing I should say is that Arabic text uses another question mark character (؟). So it's best if the examples for neutral text direction forcing include "...!" and the likes since those are much more likely to be used in a way that would require such a feature.

mal:
I've taken a pass at incorporating your notes, please see what you think and let me know if you feel any other changes or clarifications are required. 🙂

TDCMC:
Line 743 of `style_properties.rst` has a typo ("unfluence" => "influence")
But other than that, it looks really good! The documentation and changelog is quite clear with the added features and changes.

I also tested the reading_order and bidirectional text with several other examples, which included weak rtl text right after strong ltr, such as Arabic punctuation marks and they work perfectly.

Issue #6328, opened 2025-05-07
Ren'Py `script` module overhaul


Andykl:
There are several ideas related to `renpy.script` that bubble in my mind for some time (#6150 already implements some of them). 
Each of them alone does not justify big refactor of one of the core classes of the engine, but together I think it does, so I will list all of them here, so we can discuss what would be a good addition to the engine and what is not (I am not sure for each of them).
1. `renpy.script.Script` is on the edge of god object. It not only responsible for holding AST but also loading/compiling/analyze/compress of files, making backups, holding reference to global translator, and managing bytecode cache. I am not sure what roles are excessive, if any, but every time I read the file, I feel I read extra code that could be abstracted out.
2. Add a way to store compiled files somewhere else, not only next to its source but somewhere else. In #6150 I use `gamedir/cache/compiled/branch-name/...` as a prefix, but it could be something else. In git environment especially, rpycs are visual junk.
3. Add [PEP 552](https://peps.python.org/pep-0552/) for compiled files. I.e. right now Ren'Py computes hashes of the source to see if file need to be recompiled, but Python used size and mtime for a while to invalidate compiled files, and Ren'Py can do so too. In 8.4 loading time was optimized pretty well, so now reading source file to just see that it has the same hash adds ~5% overhead when it could be just couple of sys calls.
4. RPYCv3 - new version of compiled code with a structure 
```
HEADER
int[4] sections_offset_array
pickled metadata dictionary
pickled diff info dict
ast before static transforms
ast after static transforms
```
Where metadata with current 3 values also store invalidation mode and hash/mtime of the rpy file, and name of the source file it was compiled from, so if user moved rpy but not rpyc we can catch it,
And diff info section is `{node.name: node.diff_info() for node in file_nodes}`. That way, even if there was incompatiable change in the user statement, that makes the rpyc unloadable, it is still possible to do `difflib.SequenceMatcher` and assign names, making it easier to do breaking changes in AST/changes that requere force recompile without breaking names/saves (sadly not in Ren'Py realm until v3 is wide-spread).

renpytom:
So, I'd say I wouldn't want anything in this direction to start until after 8.4 is released. And even there, we need to balance the effort spent refactoring portions of Ren'Py that work pretty well with fixing reported issues, and adding new features that will be visible to players. 

[...]

PR #6378, opened 2025-05-30
Modular counterpart of _ren.py files

valery-iwanofu:
As the title says, it simply adds a modular counterpart of _ren.py files.

Things to discuss:
- Suffix for modular ren_py files (currently `_rem.py`).
- Now there are places where the presence of a module is checked via loader. Maybe we should add a generic method for such a check https://github.com/renpy/renpy/blob/6983c6fc86172bf39be76d0007b73397bdf840f4/renpy/main.py#L406-L407


Andykl:
I like the idea and the name of `_rem.py` files, as it will simplify writing explicit load order of early python code.
But in my opinion it futher complicates script loading and shuld be postponded to after 8.4 release and be done as part of #6328.

valery-iwanofu:
On the one hand it's better to postpone any script loading changes until #6328 is complete, on the other hand the current PR isn't that big and won't complicate things too much.

renpy/script.py
```diff
@@ -793,7 +800,7 @@ def load_file(self, dir, fn): # @ReservedAssignment
 
         try:
 
-            if fn.endswith(".rpy") or fn.endswith(".rpym") or fn.endswith("_ren.py"):
+            if fn.endswith(".rpy") or fn.endswith(".rpym") or fn.endswith("_ren.py") or fn.endswith("_rem.py"):
```
Andykl, line 803:
```suggestion
            if fn.endswith((".rpy", ".rpym", "_ren.py", "_rem.py")):
```

renpy/translation/generation.py
```diff
@@ -448,7 +450,7 @@ def translate_list_files():
 
         filename = os.path.join(dirname, filename)
 
-        if not (filename.endswith(".rpy") or filename.endswith(".rpym") or filename.endswith("_ren.py")):
+        if not (filename.endswith(".rpy") or filename.endswith(".rpym") or filename.endswith("_ren.py") or filename.endswith("_rem.py")):
```
Andykl, line 453:
```suggestion
        if not filename.endswith((".rpy", ".rpym", "_ren.py", "_rem.py")):
```

sphinx/source/ren_py.rst
```diff
@@ -54,6 +54,10 @@ Ren'Py store (rather than in an isolated module), and the Python code is subject
 to the transformations that enable rollback and save/load to work - see
 :doc:`python` and :ref:`rollback`.
 
+Modular counterpart
+-------
+Replacing \_ren.py with \_rem.py turns the regular script file into a loadable module.
+
```
Andykl, line 60: Need an explanation that loading this module does not include `_rem` part, i.e. `modules/my_module_rem.py` -> `renpy.load_module("modules/my_module")`.

renpytom:
My thinking is we can have a function that loads a module in the _ren.py syntax, but we don't need a special extension. We could simply use: 

```renpy
renpy.load_module("utility.py")
```

valery-iwanofu:
The standard .py suffix could be used, but we need some way to distinguish a regular python module from a ren_py module, at least for translation extraction, right?

valery-iwanofu:
`_rpym.py` looks better than `_rem.py`, since you can misspell and write `_ren.py` (like i did in the description of this PR)

PR #6386, opened 2025-06-05
chore: modify quickstart.rst

logic-finder:
from 'reddish' to 'bluish' since `#c8c8ff` seems blue to me.


logic-finder:
Hi @renpytom, I've just pushed one commit 91eb0eb, but this PR seems wrong since it contains many other commits... So it would be greatly appreciated if you could tell me how to solve this problem.

What I did is...
1. forking the original `renpy` repository
2. making a branch, named chore/modify-quickstart-doc, from the master branch
3. switching to that branch and making this commit 91eb0eb on that branch
4. `git push -u origin chore/modify...`
5. making this PR to the renpy:fix branch

Thank you.

mal:
As you're targeting the `fix` branch, in step 2 you would want to have used `fix` instead of `master` when creating the `chore/...` branch.

Your most straight-forward path forward would probably be:
```sh
git checkout chore/modify-quickstart-doc

# This should shift your commit from being based on the master branch, to
# being based on the fix branch.
git rebase --onto origin/fix master

# Now check your git log, and make sure you're happy.
# Should see your commit atop the most recent commit made to the fix branch.

# Force-push your branch to Github to update this PR.
git push -f
```

logic-finder:
@mal Many thanks for your kind help.

PR #6403, opened 2025-06-15
Check constness in `*` imports

Elckarow:
this was discussed in #6111 as being a bad idea, but i'd argue that it's worth it for modules / frameworks that have a lot of exportable stuff

i've also made it work only for modules that define `__all__`.

### DIFF
--- a/renpy/pyanalysis.py
+++ b/renpy/pyanalysis.py
@@ -206,6 +206,19 @@ def import_from(from_module_name, in_module_name, *names):
                 else:
                     const(imported_subname)
 
+def import_all_from(from_module_name, in_module_name):
+    """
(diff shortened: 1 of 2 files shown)


Andykl:
I still think that `from mod import *` is a bad practice, and people should use `import mod` in such case, and we should not add code that handles bad practice. After all, auto-import feature is pretty standart in IDE. It's up for Tom to decide.

renpytom:
I don't think we'd want to encourage import * like this. It feels much more of a footgun than either importing the module or listing the names, and unlike regular Python where a module can be imported into multiple places, here you're generally only importing names into a single Ren'Py namespace. 

So I'm going to close this without merging it.

PR #6435, opened 2025-06-26
Always wrap child render when making mesh

### LINKED ISSUE #6426: 'Blur' Effect no longer applied to 'add' statements if applied through an ATL Block.
This is an error for 8.4.0 (Prerelease). I apologise if this isn't the exact place to post pre-release bug reports.

When supplying the 'blur' effect through an ATL animation to an image supplied through an 'add' statement, the 'blur' command is lost and not applied to the image. 

Example code:

Image

Through testing:
- if supplying the 'blur' command outside of an ATL block (i.e. directly in screen code), the blur is applied successfully.
- if wrapping the 'add' image in a hbox/vbox/container and supplying the ATL animation to that container, the blur is applied successfully
- providing a time for the blur to take effect (e.g. linear 0.0 blur 20), the error still persists.

Likely this issue will have been caused to updates regarding shader ordering & application (Seen at the bottom of [incompatabilities](https://www.renpy.org/dev-doc/html/incompatible.html) for 8.4.0)

mal:
Fixes #6426, and covers a few other things I noticed while investigating.

Recommend reviewing each commit individually as it may make sense to take only some (mostly unsure about 4).

1. Tidy - Mostly just whitespace and condensed syntax.
2. Readability - Found the double offset during mesh_pad very confusing until realising the second was effectively discarded.
3. Shaders - Seemed better to simply add the necessary shaders, rather then opt-in, then out of texture shader.
4. Mesh size - Not 100% sure on this one, but if I understand correctly this ought to be the pixel size of the incoming texture.
5. Fix - Resolves the issue outlined in 6426 by always wrapping the child render in another, see notes below.

---

Putting this in draft initially as, while the fix works, I can't reconcile _why_ the additional `Render` instance makes it work.

**Observations**
1. It does appear to be mipmap related, as the following works in the prerelease:
   ```rpy
       add Transform("gui/window_icon.png", gl_mipmap=True) blur 32
   ```
2. Using `mesh_pad` (which has the side effect of forcing an additional render) also works in the prerelease:
   ```rpy
       add "gui/window_icon.png" blur 32 mesh_pad (0, 0)
   ```

The latter lead me to the fix presented, but as I say, not sure why this is the case.

### DIFF
--- a/renpy/display/accelerator.pyx
+++ b/renpy/display/accelerator.pyx
@@ -202,7 +202,6 @@ cdef class RenderTransform:
         # The reverse transform.
         self.reverse = None # type: renpy.display.render.Matrix|None
 
-
     cdef make_mesh(self, cr):
         """
         Creates a mesh from the given render.
@@ -210,13 +209,10 @@ cdef class RenderTransform:
         Handles the mesh, mesh_pad, and blur properties.
         """
 
-
         mesh = self.state.mesh
         blur = self.state.blur
-        mesh_pad = self.state.mesh_pad
 
-        if self.state.mesh_pad:
-
+        if mesh_pad := self.state.mesh_pad:
             if len(mesh_pad) == 4:
                 pad_left, pad_top, pad_right, pad_bottom = mesh_pad
             else:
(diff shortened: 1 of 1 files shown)


mal:
Closing as we don't think this is the correct fix for the problem in #6426. The other general changes have been refiled as #6436.

PR #6442, opened 2025-06-27
Allow to load persistent if tokens dir does not exists, or security keys are empty.

Andykl:
After an hour of remote debuging, I figured out that somehow our users `security_keys.txt` was empty, which was making all persistent variables to reset to default values on each load.

In second commit I added the same checks that we use for regular saves for persistent file check, so this should not happen.

But I am not sure if we want that. Maybe it is better to create new tokens if for some reason user has keys file empty.

### DIFF
--- a/renpy/savetoken.py
+++ b/renpy/savetoken.py
@@ -184,6 +184,16 @@ def check_persistent(data: object, signatures: str) -> bool:
     This checks a persistent file to see if the token is valid.
     """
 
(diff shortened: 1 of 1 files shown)


renpytom:
I've fixed this in a different way - if the key is lost, we should write a new one. I don't think we should salvage the persistent data in this case. It's unlikely, and doing so feels insecure.

PR #6445, opened 2025-06-28
Store gpu name and driver version in GL2Draw.info.

### LINKED ISSUE #6335: Ability to fetch available GPU informations in Ren'Py
It would be helpful to have a built-in method in Ren'Py to query and retrieve GPU-specific information such as available VRAM, total VRAM, and potentially GPU model. 
This would be useful for adjusting performance settings dynamically based on hardware capabilities like modern 3D games do.

Personal use case:
I currently offer three selectable levels of preloading in my game: low, medium, and high VRAM usage. However, I do not have a reliable way to recommend or default to a setting that makes sense based on the user’s GPU. I have to rely on explaining it in the settings menu and hope users choose appropriately based on their system, which is not ideal for non-technical players.

Overall I think this would benefit everyone in the long run, especially as Ren'py start to have more advanced support for 3D in general and having a way to scale performance based on user's hardware becomes more important.

Andykl:
Adds `gpu_vendor`, `gpu_name` and `gpu_driver_version` in GL2Draw.info, so developer can read it at runtime.
Fixes #6335.

### DIFF
--- a/renpy/exports/displayexports.py
+++ b/renpy/exports/displayexports.py
@@ -1023,22 +1023,34 @@ def get_renderer_info() -> renpy.display.core.RendererInfo:
 
     Returns a dictionary, giving information about the renderer Ren'Py is
     currently using. Defined keys are:
+        ``"renderer"``
+            A string giving the name of the renderer that is in use.
 
-    ``"renderer"``
-        A string giving the name of the renderer that is in use.
+        ``"resizable"``
+            True if and only if the window is resizable.
 
-    ``"resizable"``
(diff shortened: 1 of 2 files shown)


renpytom:
Please document this change in get_renderer_info (in displayexports.py). While you're at it, indicate that this only becomes available during the first interaction, which is sometime after the game starts. Also, add it to the changelong, and I'll merge it for 8.4.

Andykl:
Done.

Andykl:
Yes, because it was added in 8.4. You need to update your SDK to get it.

PR #6447, opened 2025-06-28
Return `RenpyImporter.get_data` method

NyashniyVladya:
The deleted `get_data` method was returned, because without it, the function of the same name in `pkgutil` breaks.


NyashniyVladya:
upd: Set `spec.has_location` to `True` without conditions.

By the current condition, the modules in the `.rpa` archive are assigned `spec.has_location == False`, because of this, the module does not receive the `__file__` property. In turn, this breaks the loading of module files via `pkgutil.get_data`. I consider it necessary to set `spec.has_location` to `True` without conditions.

PR #6449, opened 2025-06-29
Update launcher.rpy

Veydzher:
Updated some strings :D

### DIFF
--- a/launcher/game/tl/ukrainian/launcher.rpy
+++ b/launcher/game/tl/ukrainian/launcher.rpy
@@ -827,15 +827,15 @@ translate ukrainian strings:
 
     # game/front_page.rpy:197
     old "Navigate Script"
-    new "Навігація за скриптом"
+    new "Навігація скриптів"
 
     # game/front_page.rpy:198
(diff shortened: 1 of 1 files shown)


Andykl:
"Типова тема" seems a bit off. My Ukrainian friend say that he would expect "Тема за замовчуванням" or "Стантандартна тема" here. Same for "типової мови".
And overall, I think it should be consistent through all the translations, and most of them uses "Стандарт..." for "default".
Other fixes seems correct.

Veydzher:
> "Типова тема" seems a bit off. My Ukrainian friend say that he would expect "Тема за замовчуванням" or "Стантандартна тема" here. Same for "типової мови". And overall, I think it should be consistent through all the translations, and most of them uses "Стандарт..." for "default". Other fixes seems correct.

Well spotted by your friend. I'll use "стандартний/стандартна" to keep the word translation consistent. Thanks.

Andykl:
Merged, thanks.
If you think "за замовчуванням" is better translation for "default", PR where all cases are changed would be welcome.

PR #6450, opened 2025-06-29
Update ukrainian

Veydzher:
Updated translations of "default" to keep the constistence, plus some fixes.


Andykl:
Merged, thanks.

PR #6461, opened 2025-07-02
Changes to if_attr

### LINKED ISSUE #6453: `if_attr` needs further discussion
The introduction of `if_attr` from #6289 should be omitted from the next release until it has had time for further discussion and improvements. I've had conversations about the change with several community members, where a major sentiment was that the current iteration of `if_attr` looks confusing and not like Ren'Py, and a preference for the older style. These are the issues that have been brought up and should be addressed before it is added to a full release:

- The `&`/`!`/`|` symbols should be switched for pythonic `and`/`not`/`or`. The symbols are used nowhere else in Ren'Py in this way, which will make it unintuitive for an average user to pick up (which, remember - the average Ren'Py user has genuinely no programming experience. Not even with `if x and y` style statements). If we can at least keep the formatting consistent with the Python and Pythonic style statements elsewhere, we can avoid new users needing to learn another format. It should not be an option to use `&`/`!`/`|` for this reason, even for advanced users.
- This feature gives way to a new possible problem: `if_attr` looks *too much* like a regular Pythonic `if` statement and will almost certainly result in users mistakenly adding regular conditions to it e.g. `if_attr (sunglasses and hat and location == "outside")` looks like a sensible extension of the feature. `if_all ["sunglasses", "hat"]` does not have this issue.
- `if_attr` (the name) doesn't clearly indicate what the property does. `if_not "hat"` tells me clearly "if not the hat attribute". `if_attr (not hat)` does not quite achieve this clarity, as `if_attr` is a short form for something that might be stated like "if it has this combination of attributes" or "if the attributes satisfy the provided condition". As mentioned, the similarity to regular `if` statements also opens the door for misinterpreting the property to be used with regular variables rather than just attributes. A better name could help solve this communication problem.
- The goal should primarily be to guide users towards the correct way to use the property and away from misconceptions of how it's used. Some possible additional avenues to better communicate the purpose and use of the property include:
  - a different start/end indicator than `()` to differentiate the property from regular `if` statements, as `if_attr [sunglasses and hat]` takes the property an additional step away from regular variable checks, though I don't think `[]` is a great choice due to confusion with lists and interpolation. 
  - add quotes around the attribute names - singling out the attributes in this way may help differentiate them from regular variables
- Even if `if_not`/`if_all`/`if_any` are being phased out in favour of `if_attr`, they should still be documented, at least for the next few versions. This could include there being a section at the bottom of the layered images page with "deprecated" if preferred.

I overall think this is a useful feature which expands upon the layered image attribute logic, but I think it's premature to commit `if_attr` to the engine release in place of `if_not`/`if_all`/`if_any` without considering the above points to ensure it's understandable to new and old users. Layered images are a major feature which lots of people very new to programming will be attempting to figure out. Users have been getting by fine with `if_any`/`if_all`/`if_not` prior to the introduction of this feature, so there is time to adjust and come up with a better version before committing it to the next engine version.

Gouvernathor:
- [x] rename the `if_attr` keyword to `when`
- [x] make parentheses optional
- [x] turn the operators into words
- [x] update documentation

fixes #6453

### DIFF
--- a/renpy/common/00layeredimage_ren.py
+++ b/renpy/common/00layeredimage_ren.py
@@ -331,8 +331,8 @@ class Attribute(Layer):
         A transform or list of transforms that are applied to the
         image.
 
-    `if_attr`
-        A string containing an if_attr expression, described in the :ref:`if_attr` section.
+    `when`
+        A string containing a ``when`` expression, described in the :ref:`when` section.
         The displayable is only shown when the expression is satisfied by
         the pool of attributes currently applied to the layeredimage.
 
@@ -404,8 +404,8 @@ class Condition(Layer):
         If not None, this should be a displayable that is displayed when
         the condition is true.
 
-    `if_attr`
-        A string containing an if_attr expression, described in the :ref:`if_attr` section.
+    `when`
+        A string containing a ``when`` expression, described in the :ref:`when` section.
         The condition is only evaluated when the expression is satisfied by
         the pool of attributes currently applied to the layeredimage.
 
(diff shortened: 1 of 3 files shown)


Gouvernathor:
tested

PR #6464, opened 2025-07-04
Fix caret blinking

lauwurence:
Remove blinking during position change and add a bigger delay before starting to blink.

Before:
https://github.com/user-attachments/assets/f4b20002-3c29-4382-876d-0fddaff4f251

After:
https://github.com/user-attachments/assets/6bd85758-ce9e-46a1-9d94-71f36a9153fb

### DIFF
--- a/renpy/display/behavior.py
+++ b/renpy/display/behavior.py
@@ -1376,6 +1376,7 @@ class CaretBlink(renpy.display.displayable.Displayable):
         self.caret = caret
         self.caret_blink = caret_blink
 
+        self.st = 0
         self.st_base = 0
 
     def get_placement(self):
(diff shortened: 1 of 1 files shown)


renpy/display/behavior.py
```diff
@@ -1385,14 +1386,16 @@ def visit(self):
...
 
         cr = renpy.display.render.render(self.caret, width, height, st, at)
         rv = renpy.display.render.Render(cr.width, height)
 
-        ttl = self.caret_blink - st % self.caret_blink
+        delay = self.caret_blink / 2.0
+        ttl = self.caret_blink - (delay + st) % self.caret_blink
 
-        if ttl > self.caret_blink / 2.0:
+        if st < delay or (ttl > self.caret_blink / 2.0):
```
mal, line 1398: I believe this is functionally equivalent.
```suggestion
        if st < delay < ttl:
```

mal (commented):
Haven't given any thought to the behaviour change, so this isn't a +1 or -1, just skimmed the code and spotted a potential improvement.

Andykl:
+1 to this. It is pretty annoying in console to move cursor around.

PR #6468, opened 2025-07-05
Several fixes for CDS init logic.

Andykl:
1. Store `init_priority` of the statement in registry instead of node itself, so changing it does not require recompilation.
2. If statement does not need to run its `execute_init`, do not include it into initlist.

### DIFF
--- a/renpy/ast.py
+++ b/renpy/ast.py
@@ -2273,14 +2273,27 @@ class UserStatement(Node):
     def execute_init(self):
         self.call("execute_init")
 
-        if renpy.statements.get("init", self.parsed):
-            self.init_priority = 1
-
         if renpy.statements.get("execute_default", self.parsed):
             default_statements.append(self)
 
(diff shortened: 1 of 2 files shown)


mal:
<details><summary>Snip: Original post</summary>
Not a fan of the amount of extra work being done for every `pause`, `play`, `stop`, etc statements, none of which require it. Seems disproportionate to the issue being solved, especially in light of the fact that it's expected that recompilation is required after changing custom `UserStatement`s, this just ends up being a half-and-half footgun, where it's more difficult to comprehend why certain aspects need recompilation and some won't.</details>

---

Concerns investigated and addressed. 🙂

Andykl:
Well, it is a cost of being generic. This reduces the amount of work needed to handle non-init statements tho, but not much.

Andykl:
I included fix from #6469 into this PR. It marginally speed ups time to run init code.

PR #6469, opened 2025-07-05
Avoid processing UserStatements with no init at init time

mal:
By returning `None` instead of `0` from initless `UserStatement.get_init` the statements are no longer included in `renpy.script.initcode`.
https://github.com/renpy/renpy/blob/0cdb3a576913f69c4706e59f01864129ecdd066f/renpy/script.py#L686-L687

This reduced the `initcode` list by around 80% in a test game (just under 4k down to 800 and change), I suspect mostly due to `pause` and various sound management statements.

### DIFF
--- a/renpy/ast.py
+++ b/renpy/ast.py
@@ -2201,7 +2201,7 @@ class UserStatement(Node):
     translation_relevant: bool = False # type: ignore
     rollback: RollbackType = "normal"
     subparses: list['renpy.lexer.SubParse'] = [ ]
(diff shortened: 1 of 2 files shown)


Andykl:
This is incorrect. Statement could be `init=False` but have `execute_init` which should be run regardless.
I believe I fixed the issue in #6468.

mal:
Updated the condition to account for `execute_init` as well as `init`.

Andykl:
Superseded by #6468.

PR #6472, opened 2025-07-06
Fix regression with init_priority of UserStatement parsed before 8.4

Andykl:
With this fix, UserStatement nodes parsed before 8.4 would call its init code with whatever priority was set on the node, and nodes parsed before 8.1 would call it with priority 0.

### DIFF
--- a/renpy/ast.py
+++ b/renpy/ast.py
@@ -2201,8 +2201,21 @@ class UserStatement(Node):
     translation_relevant: bool = False # type: ignore
     rollback: RollbackType = "normal"
     subparses: list['renpy.lexer.SubParse'] = [ ]
-    init_priority: SignedInt | None = 0
     atl: 'renpy.atl.RawBlock | None' = None
+
+    init_priority: SignedInt | None = None
+    """
+    Used to store statement init priority before 8.4.
+
+    All new instances should keep this set to None.
+    """
+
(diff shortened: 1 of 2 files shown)


renpy/ast.py
```diff
@@ -2201,9 +2201,22 @@ class UserStatement(Node):
...
     atl: 'renpy.atl.RawBlock | None' = None
 
+    init_priority: SignedInt | None = None
+    """
+    Used to store statement init priority before 8.4.
+
+    All new instances should keep this set to None.
+    """
+
+    init_offset: SignedInt | None = 0
```
mal, line 2213: Suggest defaulting this to `None` as it's the more common value, thus saving the most space in cslots.

renpy/ast.py
```diff
@@ -2201,9 +2201,22 @@ class UserStatement(Node):
     translation_relevant: bool = False # type: ignore
     rollback: RollbackType = "normal"
     subparses: list['renpy.lexer.SubParse'] = [ ]
-    init_priority: SignedInt | None = 0
     atl: 'renpy.atl.RawBlock | None' = None
 
+    init_priority: SignedInt | None = None
```
mal, line 2206: Redefining this default will cause any node parsed prior to this change with a value of `0` to now have a value of `None`.

PR #6473, opened 2025-07-07
Introduce a version attribute to UserStatement

mal:
A new approach to managing changes in the UserStatement object, seeking to provide a more understandable approach to compat management both now, and for any future iterations.

- Repurpose `UserStatement.init_priority` to hold `init_offset` to avoid a new attribute or rendering this one dead weight.
- Add a `UserStatement.version` attribute to manage this and any future alterations to the object.
- Support redefinition of the default of `UserStatement.init_priority`, via the new `version` attribute.

### DIFF
--- a/renpy/ast.py
+++ b/renpy/ast.py
@@ -2119,16 +2119,20 @@ class UserStatement(Node):
 
     init_priority: SignedInt | None = None
     """
-    Used to store statement init priority before 8.4.
+    If not None, this is the enclosing init_offset. The init_priority is
+    calculated by combining this value with init_priority from the statement.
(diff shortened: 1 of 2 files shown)


mal:
Discussed and came to the following conclusions:
- Not ideal that `init_priority` could represent one of two different things.
- The use of `cslots` makes disused attributes more of a cognitive overhead than a performance (speed or memory) one.
- That we're now forcing a recompile means the impact of the default value change is already mitigated.
- Adding a `version` attribute would be a constant overhead even if minimised with special handling (always serialising).

PR #6480, opened 2025-07-09
Translate the launcher to Persian

TDCMC:
These series of commits will translate every file created by Ren'Py's built-in launcher translation file generation function.


TDCMC:
The comments in gui.rpy almost never look right in a code editor/IDE. I can transcript it in the UN Romanisation scheme instead of mixing the Arabic script with the Latin script.

renpytom:
Can you change the language name to "persian", lowercase? That would make it match what we have for other languages. 

> The comments in gui.rpy almost never look right in a code editor/IDE. I can transcript it in the UN Romanisation scheme instead of mixing the Arabic script with the Latin script.

I don't 100% understand this, so please do what you think is right.

TDCMC:
Oh oops, I guess didn't change the casing in the actual files, I'll change it. I'll also transcript every comment in the script files to latin to make them actually readable. Another proposal I have is the use of the [Vazirmatn](https://github.com/rastikerdar/vazirmatn) font for the Arabic script. As far as I could tell, it supports everything dejavu supports with the added benefit of being much more readable and much better looking. It also fits very nicely with Roboto which I believe is used in the Ren'Py launcher.

renpytom:
Works for me - please put Vazirmatn into sdk-fonts. (Pick one weight, please.) 

And if you can get me this tomorrow, it'll be in 8.4. Otherwise, I'll be merged into 8.4.1.

TDCMC:
I'll try, but I think launcher.rpy and screens.rpy are way too big for me to translate today.  As for the font, I'll put the light UI with rounded dots and no latin characters in, is that alright?

renpytom:
Probably best to include the latin characters. My feeling is things like filenames and project names will be substituted in.

TDCMC:
That's fine, I just thought that a font group would be used and render latin glyphs redundant.

TDCMC:
The transcription system I used for romanisation is the [United Nations Romanization System for Persian](https://unstats.un.org/unsd/geoinfo/UNGEGN/docs/10th-uncsgn-docs/econf/E_CONF.101_118_Rev.1_Transcription%20symbols%20for%20Persian_Updated.pdf).

This is not a commonly used system by Persian speaking people, but it has much better clarity for how certain words are pronounced and thus a much more suitable system than what is known as "Finglish" and used for chatting.

TDCMC:
@renpytom The translation except a final proofread is done. I have two questions though. Is there a way to set `config.rtl`? And is there any other thing that I need to do?

mal:
On the `config.rtl` point, it should already be set to `True` in launcher:
https://github.com/renpy/renpy/blob/b3794e806575b731170d55fc20980a802a4a204c/launcher/game/options.rpy#L451-L452

renpytom:
Looks good to me. No need to set config.rtl - as of 8.4.1, config.rtl defaults to True. 

Could you retarget this to the fix branch? If that's too hard, don't worry about it - I can squash-merge into master and then cherry-pick the change. Since it's a translation, this can go into 8.4.2.

TDCMC:
I *can* but I don't think I'm confident enough to do so without causing a bunch of problems at first. Is it possible to change the target branch and only have these select commit on the pull requests, or do I need to recommit them onto a fork of the fix branch?

TDCMC:
I think I'll unfortunately need to leave the merging of the commits into the fix branch to you. I can push a fork of the fix branch onto my "master", cherry pick the relevant commits, and retarget the pull request, but I haven't done this so it might just cause more harm than good.

renpytom:
Yeah, no problem. I'll take care of it soon.

TDCMC:
Thank you very much. One concerning problem is that the translation doesn't actually work properly on the precompiled v8.3.7 sdk that I worked with to create. I see a second "English" entry in the launcher preferences that when selected, makes most of the buttons (I'm guessing every button that I translated) disappear. I'm also confused as to where the label of the translation gets defined, which is displayed as "English" right now.

Okay, I found the language name and font tag which I forgot to replace. But the button labels still don't get shown.

TDCMC:
@renpytom I compiled renpy's upstream repo and the non latin scripts still don't show up. Have I missed something?

Okay, so I took a look at the Arabic translation and there is an additional file called "style.rpy" which defines a bunch of font variables. Should I do the same and send you the relevant definition to commit to the repo, or is there a better way of dealing with this?

Another problem I found is that bidirectional text for buttons such as "Ren'Py sponsor information" and "Ren'Py website" buttons break for Arabic text. Is the text processing the same for them?

renpytom:
I've merged it in. I don't thionk the button processing text is in. Is it possible that "Ren'Py" being there is causing things to assume it's latin text?

TDCMC:
Okay, nevermind. This was a problem with v8.3.7 which didn't have the relevant changes to bidirectional text handling when using Harfbuzz (thanks again to mal for the implementation). I tested it with v8.4.1 and it worked perfectly. I just need the scaling of the font to be set to 0.9 because it looks a bit iffy right now.

PR #6487, opened 2025-07-12
Migrate build system to scikit-build-core.

Andykl:
This PR simillar to [pygame_sdl2](https://github.com/renpy/pygame_sdl2/pull/164), migrates build system of Ren'Py to use scikit-build-core and CMake.
This simplifies configuration needed on Windows (I was unable to configure my system to tell setuptools to use msys), makes build 2-3 times faster (at least on my machine in WSL), and pretty much make it possible to `pip install renpy`.

This is a draft before we decide on several things, most of which are dropped from the build system:
1. Do we still need RENPY_COVERAGE and if so, what the drawback to have it permanently be on? PEP 669 seems to make it less intrusive
2. Do we need cython annotations and if so, do we need it to be always on?
3. Do we need RENPY_STATIC and if so, should we switch to [inittab](https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#integrating-multiple-modules)?

Also, right now to build, you need to set environment variable `PYGAME_SDL2_HEADERS` equal to absolute path to pygame headers (absolute path of `.venv/.../site-packages/include` and to fix it we need to integrate pygame_sdl2 into Ren'Py directly.


renpytom:
I don't thinkl we need RENPY_COVERAGE anymore. 

Annotations are very useful - they make it clear how optimized the Cython is. 

We either need RENPY_STATIC, or some other way of generating the Cython in a way that's suitable for static linking through renpy-build. We also need to make sure that the init functions have unique names, which they don't by default.

PR #6498, opened 2025-07-14
Add easing customization to Scroll() screen action

DevilSpiderAkumo:
This pull request adds an optional `easing` argument into the `Scroll()` screen action.
The current implementation of this action only allows the `ease` warper, so I thought it would be nice to have more easing options.

### DIFF
--- a/renpy/common/00action_other.rpy
+++ b/renpy/common/00action_other.rpy
@@ -697,15 +697,21 @@ init -1500 python:
 
         `delay`
             If non-zero, the scroll will be animated for this many seconds.
+
+        `easing`
+            The easing function used for the scroll action. For built-in
+            functions, a string. For custom easing functions,
+            the function itself.
         """
 
         delay = 0.0
 
-        def __init__(self, id, direction, amount="step", delay=0.0):
+        def __init__(self, id, direction, amount="step", delay=0.0, easing="ease"):
             self.id = id
             self.direction = direction
             self.amount = amount
             self.delay = delay
+            self.easing = easing
 
 
         def get_adjustment_and_delta(self):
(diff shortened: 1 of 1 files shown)


Andykl:
I think it is better to 
```py
if isinstance(easing, str):
    try:
        easing = renpy.atl.warpers[easing]
    except KeyError:
        raise Exception(f"unknown warper name {easing!r}. Did you forget to define it with renpy.atl_warper?")
self.easing = easing
```

That way we catch unknown/misspeled names not at the time the action is called but eariler, maybe even in init time.
This should work because `atl_warper` is documented to be used in early python. Also, this allow to use user-defined named easing functions, and not only Ren'Py build in.

DevilSpiderAkumo:
Oh, true. Let me just do another commit with this in mind.

mal:
We have existing public APIs allowing warpers to be used already, such as in `ImageDissolve`, `MoveTransition` and `Drag`, i.e.
```rpy
define circirisout = ImageDissolve("circiris.png", 1.0, time_warp=_warper.easeout)
```
They all expect to be provided the warper function, not a string. I think for consistency it would be better to match that approach here, and also give it a similar name such that the default would be `warper=_warper.ease`. This would also give the flexibility present in those other APIs of being able (and indeed expected) to provide a function - be it inbuilt or user defined.

DevilSpiderAkumo:
Right, having it more consistent with displayables/functions that also have `time_warp` could be useful too. I'll wait for further comments before doing another commit.

Andykl:
I agree with mal that consistency is more important here. It feels weird that you need to reach 'private' _wraper' module for common ones, so maybe later we will fix all instances to take name or function, or to alias _warper to something different, but it would be separate issue.

PR #6530, opened 2025-07-25
Check for flatpak install when launching VSCode(System) on Linux

dynamiclinking:
Like many users my Visual Studio Code is installed using flatpak, I have added a check for visual studio installed via flatpak, it can get some improvements, but this version should allow flatpak users to enjoy Ren'py and save space and data.

The idea I have implemented is:
When the launcher cannot find ```code``` in the path, retry using flatpak, if flatpak is not found, raise the original exception.


renpytom:
Two changes at least: 

1. This code path should only run when self.system is True. We don't want it running when the user asked to use the Ren'Py installed vscode. 
2. It should be "if renpy.linux:". There is a typo, "renpu".

dynamiclinking:
I have fixed the typo and added a check for ```self.system```

PR #6535, opened 2025-07-26
Ensure _in_current_store call in transformed layer

### LINKED ISSUE #6534: Layer transforms can result in transitions being skipped
```rpy
image test = '[color]'
image r = Solid('f773', xysize=(320, 240))
image y = Solid('ff73', xysize=(320, 240))

label start:
    $ color = 'r'

    scene black
    show test as person
    'initial red'
    $ color = 'y'
    with {'master': dissolve}
    'dissolves to yellow'

    $ color = 'r'

    scene black
    show layer master at Transform()
    show test as person
    'initial red'
    $ color = 'y'
    with {'master': dissolve}
    'instantly yellow'
    return
```

mal:
Fixes #6534.

This is a very basic fix. I don't know it this path would benefit from either of the following:
- ~~A short circuit to return `self` if `not self.children`.~~
- Combining
   https://github.com/renpy/renpy/blob/99456f2f26e89799e86eaf93796bf9403e7e5562/renpy/display/layout.py#L678-L679
   into
   ```py
       if self.layer_name is not None and self.scene_list is not None:
   ```
   such that it can fallback to the else clause (passing `elif self.layers`) which already does the right thing:
   https://github.com/renpy/renpy/blob/99456f2f26e89799e86eaf93796bf9403e7e5562/renpy/display/layout.py#L731-L732
   <br/>Also checked other instances of `return self` and all appear to correctly check for changes beforehand.

### DIFF
--- a/renpy/display/layout.py
+++ b/renpy/display/layout.py
@@ -675,10 +675,7 @@ class MultiBox(Container):
         self.scene_list = None
 
     def _in_current_store(self):
-        if self.layer_name is not None:
-            if self.scene_list is None:
-                return self
-
+        if self.scene_list is not None:
             scene_list = []
 
             changed = False
@@ -729,7 +726,7 @@ class MultiBox(Container):
                 return self
 
         else:
-            return super(MultiBox, self)._in_current_store()
+            return super()._in_current_store()
 
         if self.offsets:
             rv.offsets = list(self.offsets)


renpytom:
I believe your second (now only) suggestion is a good idea.

mal:
After more thought, I don't think this is fit for merge in it's current form. The following things need to be addressed:

- ~~Does `layer_name` need to be retained on the `rv`?~~
- ~~Does `untransformed_layer` need to be retained, and if so, does it too need to have `_in_current_store` called on it?~~
- ~~Reasonable chance we need to retain `offsets`, `start_times`, and `anim_times`.~~

I'm actively working on the last item, feedback/guidance would be welcome on the first two, especially the second.

---

**Update**

Turns out the parent method is robust enough to handle all of those things "for free", so I'll update the condition and call it good.

mal:
I couldn't find a scenario in which `scene_list` would be set and `layer_name` would not be, so I further simplified the condition. Even in the event such a scenario does materialise, I believe the `else` is robust enough to handle that case without issue.

PR #6574, opened 2025-08-15
Add `CallableValue`.

NyashniyVladya:
Adding a bar value `CallableValue` that displays the result of a function call. It can be useful for displaying any dynamic value.

For example, a bar displaying seconds:

```renpy
init python:

    import time

    def get_seconds():
        t = time.localtime()
        return t.tm_sec
```

```renpy
screen test:
    bar value CallableValue(get_seconds, 60, .0)
```


renpytom:
This is a great feature. Merged.

PR #6623, opened 2025-09-16
Move dependencies into pyproject.toml

Andykl:
This simplifies install process by one step, and moving us one step closer to PyPI package.

Note, that `pip install` will use build isolation, making rebuilds slower, but when `setuptools` is installed in current venv (i.e. after `pip install -e .[dev]`, you could use `setup.py build_ext` as before. I believe with some changes in build process, where cython output cached outside of build directory and proper ccache handling, that would no longer be a case.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ ehthumbs_vista.db
 *~
 *.bak
 .gdb_history
+uv.lock
 
 # Compiled Ren'Py files.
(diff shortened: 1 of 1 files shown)


Andykl:
Simplify setup even futher by suggesting using `uv`, which will download correct version of python, setup venv and install dependencies in a single command.

PR #6644, opened 2025-09-27
Testcase System Revamp

arsym-dev:
## Overview
This is an overhaul of Ren'Py's testing framework and is an extension of the work done by @Gouvernathor in #4630. Key features are improved test execution, reporting, and more powerful commands.

To use it, run the following CLI command:
```bash
# Linux
./renpy.sh <basedir> test [testcase] [--no-skip] [--print-details] [--print-skipped]

# Windows
./lib/py3-windows-x86_64/python.exe <basedir> test [testcase] [--no-skip] [--print-details] [--print-skipped]
```

The following projects have been updated. Substitute `<basedir>` with any of these:
- gui
- launcher
- testcases
- the_question
- tutorial

## Test Execution and Test Suites
Previously, a testcase would be launched and it controlled the execution flow with `call` and `jump` commands to other testcases.

In this PR, testcases are grouped into testsuites and bundled with hooks such as `before_each_case` and `after`. Test execution is controlled by a state machine that runs the cases and hooks in a structured way.

In case an exception is thrown, the error is logged, the test is marked as a failure, and flow continues to the next test block.

Testing continues after an `utter_restart()`. This is useful for testing features like the director, trying different environments, or clearing persistent in a game and resetting to a fresh copy.

The following test hooks are available:
- `before` (ie. before any other hook or test is run within this suite)
- `before_each_suite`
- `before_each_case`
- `after_each_case`
- `after_each_suite`
- `after` (ie. the last block run after all subtests and hooks are executed)

<details>
<summary>Example: Script and Execution Flow</summary>

```python
testsuite global:
    before:
        pause until main_menu

    before_each_suite:
        if not screen main_menu:
            run MainMenu(confirm=False)
        click "Start"

    after_each_suite:
        $ print("Test suite complete")

    after:
        exit

    ## Subtests
    testsuite basic:
        testcase first_testcase:
            advance

    testsuite test_choices:
        before:
            run Jump("chapter1")

        before_each_case:
            advance until menu choice

        after_each_case:
            $ print("Finished a choice test.")

        after:
            $ print("Finished all choice tests.")

        ## Subtests
        testcase choice1:
            click "First Choice"

        testcase choice2(skip=True):
            click "Second Choice"

        testcase choice3:
            click "Third Choice"
```

would be executed as:

```
+ global
    global.before
    global.before_each_suite
  + basic
    - first_testcase
    global.after_each_suite
    global.before_each_suite
  + test_choices
      test_choices.before
      test_choices.before_each_case
    - choice1
      test_choices.after_each_case
      test_choices.before_each_case
    - choice2
      test_choices.after_each_case
      test_choices.before_each_case
    - choice3
      test_choices.after_each_case
      test_choices.after
    global.after_each_suite
    global.after
```
</details>

This was inspired by [QUnit](https://qunitjs.com/lifecycle/).

## Reporter
Currently a console reporter is implemented, and an interface exists for future extensions (eg. UI reporter). This tracks the execution order, exceptions thrown, duration, status, number of failures, etc.

The image below shows a successful test run of "The Question":

the_question

In the image below, an exception is thrown when trying to `assert False`. It reports tracebacks on three levels: one for the python code, the game script, and the test execution.

error_logging

## Nodes

There are too many changes here and it would be best to list all changes in documentation, but the following are now possible:

```python
advance until screen choice
skip until label chapter2

if not screen main_menu:
    run MainMenu(confirm=False)

click id "next_btn" until "End of menu"
keysym "K_BACKSPACE" repeat 30
keysym "game_menu" until id "main_menu_btn"
assert "This takes a second to display" timeout 3.0
assert screen phone and not "Contact 1"
exit
```

## Misc Changes
- `Launch Tests` button on the front page of the launcher. Launches the default test case.
- `renpy.exports.is_in_test()` helper function
- Mark tests skippable (either with `skip=True` or `only=True`)
- Exit codes make it possible to make this part of a CI/CD pipeline (if terminated with `exit` or similar)

## To do
- Update `drag` and `scroll`
- Update documentation
- Better gamepad emulation
- Bug: If you use `exit`, the case/hook is marked as not run

## Ideas for future PRs
- Sensitivity flag for selectors (eg. `pause until id "next_btn" sensitive`)
- Image selector (eg. `pause until image "mc happy hat"`)
- Screenshot testing
- Test speed is limited by the framerate. An option to unlock FPS would make tests run faster
- Allow cases to be called with parameters? (Allows reuse)
- List all tests in the Navigate Script menu (along defines, todos, labels, transforms...), with a button on each to launch it
- Make some testcases "publics" (meant to be triggered directly from the launcher) and others not

## Final remarks
I suggest running the five tests listed in the Overview. In particular, `tutorial` and `testcases` show how we can test the engine itself.

I've been using this on my own projects for a few months which is how it settled into this state. I'd really like to hear your thoughts if you feel something needs consideration or clarification.


renpytom:
I really like this. It solves one of the biggest problems with Ren'Py, and I really like the hooks system as a way to remove duplicate scripting between cases. I'd like to merge this ASAP, for the 8.5 prerelease - how does that look. 

Some minor comments: 

I don't love  ``before_each_suite`` and ``before_each_case``. I'm thinking the hooks should be: 

* ``before all`` (was ``before``)
* ``before testsuite`` (was ``before_each_suite``) 
* before ``testcase`` (was ``before_each_case``) 

And same with ``after``. 


I don't love description/skip/only/parent being arguments supplied to testcase. To me: 

```renpy
testsuite default(description="Default project testsuite"):
    ...
```

reads like a definition, and so the parenthesis reads as a parameter list, not an argument list. And so having it be meaningful like this feels bad. Perhaps: 

```renpy
testsuite default:
     description "Default project testsuite"
```

as more inline with how the rest of Ren'Py handles properties like this. 

These are pretty minor quibbles, though. I'm very happy with this, and would like to expedite the merge.

arsym-dev:
> I don't love `before_each_suite` and `before_each_case`. I'm thinking the hooks should be:
> * `before all` (was `before`)
> * `before testsuite` (was `before_each_suite`)
> * `before testcase` (was `before_each_case`)

Those are much nicer to read. I worry that `before all` would be interpreted as "before each testsuite and testcase". How about `setup` and `teardown`? That should remove the ambiguity.

> I don't love description/skip/only/parent being arguments supplied to testcase

I like that your suggestion frees up testcases so they can be supplied with parameters in the future.

However, it blurs the lines between boilerplate and actual test commands. For example, the `skip` flag means "activate skip mode" in a testcase, and I'm not sure how I feel about:

```renpy
testsuite default:
    description "Default project testsuite"
    only True

    testcase expensive_operation:
        skip True
        ...
```

How about using decorator-like syntax?
```renpy
@description("Default project testsuite")
@only
testsuite default:

    @skip(reason="We don't have all day")
    testcase expensive_operation:
        ...
```
This would give a way to use [marks similar to pytest](https://docs.pytest.org/en/stable/reference/reference.html#marks) such as 
- `@skipif(condition)`
- `@xfail` if a test is expected to fail
- `@parameterize` to [feed a testcase different inputs to test](https://docs.pytest.org/en/stable/how-to/parametrize.html#parametrize)

I feel this gives both clarity and room to extend the framework later. Maybe something like `@test.skip` is better if you feel we might want decorators elsewhere in the future.

> I really like this. It solves one of the biggest problems with Ren'Py, and I really like the hooks system as a way to remove duplicate scripting between cases. I'd like to merge this ASAP, for the 8.5 prerelease - how does that look.

Thank you! If you agree with me about decorators then I'd like to get that syntax is in first, otherwise I can wrap up the documentation and get it merged!

renpytom:
I like ``setup`` and ``teardown``, if you want to go that way. 

I don't like the decorator syntax - it puts minor information (like the description) before the name of the testcase. That feels weird to me. We also don't use it anywhere else in Ren'Py, and the semantics are somewhat different than the way decorators work in Python - so I'd really like to avoid decorators, if possible. 

```renpy
testsuite default:
    description "Default project testsuite"
    only True

    testcase expensive_operation:
        skip True
        ...
```

This is very similar to the way properties and statements in screen language work, so I think the basic structure is going to be familiar to the sort of advanced Ren'Py user than is likely to want to use the test system. The one restriction I'd put on it  is to disallow the test properties after the test statement - so that: 


```renpy
testsuite default:
    testcase expensive_operation:
        ...

    skip True
```

would be an error.

arsym-dev:
Hmmm, alright. My one contention is that `skip` has two different meanings:
- Ignore the testcase
- Activate dialogue skipping

I'll rename it so the first version becomes `enabled False`, and keep in mind what you said about test properties not coming after test statements.

renpytom:
That's fine, I like ``enabled``. 

I've also added renpy.config.gl_vsync. The testcase code can set it to False, and that speeds up the tutorial test from 46s to 17s, on my system. 

Are you using this in CI/CD? I'm curious how you're handling the OpenGL requirement there. (Or if I should look at a way of disabling the video and audio output for CI/CD use.)

arsym-dev:
Woah, thanks for the FPS unlock! Can't wait to try it!

For CI/CD, I set up a self-hosted runner for my GitHub Actions, so the requirement is already met by my development machine.

Here's what my `.github/workflows/renpy-test.yml` file looks like. It's really clunky and is intended to run out of my personal project:

```yaml
name: Ren'Py Testcases

on: [workflow_dispatch, push, pull_request]

permissions:
  contents: read

jobs:
  test:

    runs-on: self-hosted

    steps:
    - uses: actions/checkout@v4

    - name: Modify Project Appdata Directory
      shell: cmd
      run: |
        echo: >> game\options.rpy
        echo define config.save_directory = "IW-1510512572-dev" >> game\options.rpy

    - name: Delete old compiled files
      shell: cmd
      run: |
        del /s /q *.rpyc
        del /s /q *.bak

    - name: Delete saves and persistent data
      shell: cmd
      run: |
        del /s /q .\game\saves
        del /s /q %appdata%\RenPy\IW-1510512572-dev

    - name: Run Ren'Py Testcases
      run: C:\Data\Renpy\renpy\lib\py3-windows-x86_64\python.exe C:\Data\Renpy\renpy\renpy.py . test
```
This makes a lot of assumptions about my system (Windows, where the Ren'Py SDK is stored, etc) but it could be modified to run out of the renpy repo itself rather than my personal project.

I'm not sure what to do about OpenGL if you do something that's not self-hosted.

arsym-dev:
I've make the requested changes, updated the documentation (`testcases.rst`), and rebased to resolve conflicts. The only thing left from my "to do" section is the `exit` bug, but it's very minor.

A few points:

1) Should we standardize the `screen` selector so it uses a quoted string? Right now the selectors are:

```
screen choice
id "back_btn"
layer "master"
```
The alternative is to make it:
```
screen "choice"
```

Renpy scripts have a similar issue and the workaround there is to use `show screen expression expr`. We have the option now to sidestep it.

2) `renpy.config.gl_vsync = False` does nothing for me. I'm on Win10, GL2 renderer, and I got the `lib` directory from the Sept 29 2025 nightly build, but I'm capped at 60 fps. I've enabled the setting for tests, let me know if it works for you. Either way, I'm ready to merge it now and avoid another rebase.

3) Thanks to your suggestion, we can use expressions like `enabled renpy.linux` to conditionally skip tests. I really like it.

renpytom:
1. I'm fine with the screen selector taking a quoted screen name. 

2. I'll take a look and see if I can get gl_vsync working post-merge. 

I'll merge this in momentarily so it won't go out of sync. Thank you for bringing this up to date.

PR #6664, opened 2025-10-08
Testcase Patch 2

arsym-dev:
## NEW: xfail / xpass

Inspired by pytest, `xfail` means "expected to fail" and can be used to make sure a testcase or an assertion throws an error. For example, these following will be reported as having `XFAILED`:

```renpy
testcase example:
    assert 1 == 2 xfail True

testcase example2:
    xfail True

    click "button that doesn't exist"
```

If a case marked as xfail does not fail, it is reported as `XPASSED`, meaning "unexpectedly passed". This is considered to be an unsuccessful run.

## NEW: Screenshot 
Takes a screenshot of the game and saves it to the specified location. Locations are relative to `_test.screenshot_directory`, which is set to "tests/screenshots" by default. If the file already exists, it's compared to the new screenshot pixel-by-pixel.

This statement takes an optional `max_pixel_difference` which is either an int or float between 0.0 and 1.0. If the differences exceed this value, the test fails. It also takes an optional `crop (x1, y1, x2, y2)`

```renpy
screenshot "screens/main_menu.png" max_pixel_difference 0.01 crop (10, 10, 100, 50)
```

If comparison fails, it creates a `*.new.png` and `*.diff.png` to allow the dev to see what differs.

For the sake of tracking changes, the hash of the current commit is automatically appended to the filename if the project is in a git repository. For example, the above example might create `screens/main_menu@75e8d6d9870a8be6d7e77fbc4a927aa3782ffe6c.png`

### TODO
Right now the image comparison is done with a python loop across all pixels. Pygame has the [pygame.transform.threshold function](https://www.pygame.org/docs/ref/transform.html#pygame.transform.threshold) that might net us a speed boost, and  might be worth implementing in Ren'Py.

Something like this should hopefully do the trick:
```
diff_count = renpy.pygame.transform.threshold(  # type: ignore
    diff, img, (0, 0, 0), (0, 0, 0), (255, 255, 255), 1, old_img, True
)
```

## NEW: Parameterized Tests

This test would be run three times with different values of `x`:
```renpy
testcase example:
    parameter x = [1, 2, 3]
    assert eval (x > 0)
```

We can group parameters. This test is run thee times:
```renpy
testcase addition:
    parameter (x, y, z) = [ (1, 2, 3), (2, 3, 5), (3, 5, 8) ]

    assert eval (x + y == z)
```

If we use multiple `parameter` statements, the test will be run for every possible combination of values. The following test is run eight times and takes the values `(a, b, c)` = `(1, 3, 5)`, `(1, 3, 6)`, `(1, 4, 5)`, `(1, 4, 6)`, `(2, 3, 5)`, `(2, 3, 6)`, `(2, 4, 5)`, `(2, 4, 6)`:

```renpy
testcase combinations:
    parameter a = [1, 2]
    parameter b = [3, 4]
    parameter c = [5, 6]

    assert eval (a + b + c in [9, 10, 11, 12])
```

Finally, we can mix grouped and non-grouped parameters. The following test is run four times, and takes the values `(a, (b, c))` = `(1, (3, 5))`, `(1, (4, 6))`, `(2, (3, 5))`, `(2, (4, 6))`

```renpy
testcase mixed:
    parameter a = [1, 2]
    parameter (b, c) = [ (3, 5), (4, 6) ]

    assert eval (a + b + c in [9, 10, 11, 12])
```

Any statement that accepts an expression can use the test parameters. Scope is handled by state machine, and test statements make use of `scoped_eval()` and `scoped_exec()`. For example:

```renpy
testcase choice_test:
    parameter x = [0, 1, 2]
    
    assert eval (x < 2) xfail x == 2

testcase show_menu:
    parameter screen_name = ["preferences", "load"]

    $ print(f"Showing screen '{screen_name}'")
    run ShowMenu(screen_name)
    pause until screen screen_name
    run Return()
```

## NEW: untranslated and unsubstituted flags
The `untranslated` flag performs the text matching on untranslated strings. This makes it possible to reuse the same statements when testing different languages without having to write new tests.

The `unsubstituted` flag compares with strings that have not been interpolated.

```renpy
testsuite chapters:
    parameter language = ["japanese", None]

    run Language(language)
    run Start()
    advance until screen "choice"

    click "No" untranslated

    # Equivalent to the following:
    if eval language is None:
        click "No"
    elif eval langauge == "japanese":
        click "いいえ"

    assert "Welcome, [player_name]!" untranslated unsubstituted
```

## Other
- Bug fixes for crashes happening in setup and teardown
- Updated cli command (added more reporting options, renamed some to be more intuitive)
- Updated documentation, state machine, test reporter, testcases

Apologies for how arbitrarily these changes are grouped together, a lot of this is still in flux.

I'd like to hear your feedback and any criticism.

### DIFF
--- a/renpy/test/testast.py
+++ b/renpy/test/testast.py
@@ -1417,8 +1417,8 @@ class Screenshot(Node):
         self,
         loc: NodeLocation,
         filename: str,
-        max_pixel_difference: int | float = 0,
-        crop: tuple[int, int, int, int] | None = None,
+        max_pixel_difference: str | None = None,
+        crop: str | None = None,
     ):
         self.max_pixel_difference = max_pixel_difference
         self.crop = crop
@@ -1454,7 +1454,7 @@ class Screenshot(Node):
 
             if self.crop:
                 # img = renpy.display.scale.smoothscale(img, (renpy.config.screen_width, renpy.config.screen_height))
-                img = img.subsurface(self.crop)
+                img = img.subsurface(scoped_eval(self.crop))
 
             base_filename, ext = os.path.splitext(filename)
             ref_img_path = self.get_reference_image_path(filename)
@@ -1503,14 +1503,19 @@ class Screenshot(Node):
                     else:
                         diff.set_at(pos, color_black)
 
-            if isinstance(self.max_pixel_difference, float) and 0 < self.max_pixel_difference < 1:
-                self.max_pixel_difference = int(self.max_pixel_difference * img.get_width() * img.get_height())
+            max_pixel_difference = 0
+            if self.max_pixel_difference is not None:
+                max_pixel_difference = scoped_eval(self.max_pixel_difference)
+                if not isinstance(max_pixel_difference, (int, float)):
+                    raise ValueError("max_pixel_difference must be an int or float.")
+                if isinstance(max_pixel_difference, float) and 0 < max_pixel_difference < 1:
+                    max_pixel_difference = int(max_pixel_difference * img.get_width() * img.get_height())
 
-            if diff_count > self.max_pixel_difference:
+            if diff_count > max_pixel_difference:
                 self.save_image(img, new_fname)
                 self.save_image(diff, diff_fname)
                 raise RenpyTestScreenshotError(
-                    f"{filename} (pixel difference: {diff_count} > {self.max_pixel_difference})\n"
+                    f"{filename} (pixel difference: {diff_count} > {max_pixel_difference})\n"
                     f"Current image saved to {new_fname}\n"
                     f"Difference image saved to {diff_fname}"
                 )
--- a/renpy/test/testfocus.py
+++ b/renpy/test/testfocus.py
@@ -45,7 +45,7 @@ def find_focus(pattern: str, raw: bool) -> Focus | None:
         if pattern in text.casefold():
             candidates.append((len(text), focus))
 
-    return min(candidates, default=(None, None))[1]
+    return min(candidates, key=lambda x: x[0], default=(None, None))[1]
 
 
 def _get_default_focus_text(raw: bool) -> str:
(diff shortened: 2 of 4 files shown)


renpytom:
The one thing that I'm not liking about this is untranslated/unsubstituted - I don't like the idea of the test system digging into the Text displayable like that. I'd much rather give the _tts functions on Displayables an argument that determines if text is translated/substituted or not, and have match use that. 

Bypasssing the translation system for dialogue is also a bit ... well, it kind of scares me, but I see the point there, so we probably should let it happen. 

Do we really need "untranslated" and "unsubstituted" as separate items? I'm wondering if we should just have "raw" that does both, and then the normal mode.

arsym-dev:
I like the simplicity of "raw", I've removed "untranslated" and "unsubstituted" in favor of it. It's possible someone might want a translated but unsubstituted string, but admittedly it's rare and can be handled if requested.

I was worried that changing TTS to return raw text was very un-TTS-like and risks changing the meaning of the function. However, I've gone ahead with the change in the latest commit and it works well with displayables.

I made one change I'm uncomfortable with, and I don't know how to fix it. In `renpy\text\text.py` and `renpy\display\displayable.py`, I've replaced:
```python
if alt is not None:
    rv = renpy.substitutions.substitute(alt, scope={"text": rv})[0]
```
with:
```python
if alt is not None:
    if raw:
        rv = alt
    else:
        rv = renpy.substitutions.substitute(alt, scope={"text": rv})[0]
```

The problem is that `rv = alt` completely ignores all changes done to rv previously. Ideally there should be a way to substitute only the variables given in the scope and return the remaining tags unchanged.

----

Is an [implementation of pygame.transform.threshold](https://github.com/pygame/pygame/blob/85fda3f719d437cf27106afae8c890e6b88ba5f5/src_c/transform.c#L1776) possible, or should I continue with my python loops for screenshot comparison?

renpytom:
Let's go with python loops for now, and we can consider threshold or something similar to it, perhaps simpler. 

Let me think about how alt substitution should work... it's probably going to involve  treating {text} as something special.

PR #6687, opened 2025-10-22
Add support for custom actions in the launcher via project.json

sa20001:
Hi, I noticed that while the launcher supports adding options through `project.json`, there’s currently no way to customize the **Actions** tab.

This pull request introduces the ability to define custom actions directly in `project.json`. Example:

```json
"open_directory": {
  ...
},
"edit_file": {
  ...
},
"actions": {
  "Install dependencies": "./install.ps1"
}
```

These custom entries will appear in the **Actions** tab of the launcher.

image

---

**Notes:**

* Marked as draft because testing was done on version **8.4.1**, and the changes were then ported to the **master** branch.
* Currently tested only on **Windows**.

### DIFF
--- a/launcher/game/front_page.rpy
+++ b/launcher/game/front_page.rpy
@@ -206,6 +206,9 @@ screen front_page_project:
             frame style "l_indent":
                 has vbox
 
+                for button_name, path in p.get_renpy_launcher()["actions"].items():
+                    textbutton button_name action RunScript(os.path.join(p.path, path))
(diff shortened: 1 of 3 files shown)


renpytom:
I don't think this is appropriate for Ren'Py. The other things make sense, but by the time you're going to run general actions, you should just use your OS's launcher function. (Plus, there is little to no more room in the launcher.)

PR #6691, opened 2025-10-23
Update distribute.rpy

### LINKED ISSUE #6689: Incorrect destination folder when building macOS distribution
When building distribution for macOS, the folder gameXXXX-dist (where .zip file is created) is made in gameXXXX.app/Contents/Resources instead being created out of gameXXXX.app completely.

Once (haven't been able to recreate it completely) it even created folder c: in my Downloads and inside it created temp (so, ~/Downloads/c:/temp/).

Tested with:
- renpy-8.4.2.25101001+nightly-sdk

It worked correctly in renpy-8.3.7-sdk

the66F95:
this should fix #6689.
first i was using `parent = os.path.dirname(project.path.partition(project.name)[0])`, but this results in the wrong parent folder when the sub folder uses the package name.

### DIFF
--- a/launcher/game/distribute.rpy
+++ b/launcher/game/distribute.rpy
@@ -639,7 +639,8 @@ fix_dlc("renios", "renios")
             # The destination directory.
             if destination is None:
                 destination = build["destination"]
-                parent = os.path.dirname(project.path)
+                path, sep, package = project.path.rpartition(project.name)
+                parent = path or package
                 self.destination = os.path.join(parent, destination)
             else:
                 self.destination = destination


Andykl:
It took me a minute to realise that if there is no `project.name` in `project.path` partition would return `("", "", project.path)` and I think it does not handles path separation in partition so not sure how it will join later.

I would suggest use `Path` in here so it would be something like
```py
p = Path(project.path)
if project.name in p.parts:
    p = Path(*p.parts[:p.parts.index(project.name)])
else:
    p = p.parent
self.destination = str(p / destination)
```

the66F95:
and yet it does exactly what it's supposed to do.
the following `os.path.join()` already handles missing/unnecessary trailing slashes.
in my tests the distros are built in the expected locations, the project folder or the sub folder, for normal packages or .app packages.

PR #6695, opened 2025-10-26
gl2: Expose GPU max texture size

### LINKED ISSUE #6694: Feature to gauge GPU capability
Some way for creators to get an idea of what sort of hardware the game is played on. This would be useful for disabling GPU intensive shaders on low end, older, and budget devices.

mal:
Fixes #6694.

Typically indicative of overall GPU strength, making it a potentially useful metric for scaling more intensive shaders and VFX in order to avoid overloading weaker hardware found in older phones, and budget tablets and laptops.

Intentionally very low-level exposure/code impact for now while we consider its reliability/usefulness. Can always pick some thresholds and add it to the existing variant system down the line if it turns out to be super useful for that.

### DIFF
--- a/renpy/gl2/gl2draw.pyx
+++ b/renpy/gl2/gl2draw.pyx
@@ -774,7 +774,7 @@ cdef class GL2Draw:
         glGetIntegerv(GL_MAX_TEXTURE_SIZE, &max_texture_size)
         glGetIntegerv(GL_MAX_RENDERBUFFER_SIZE, &max_renderbuffer_size)
 
-        max_texture_size = max(max_texture_size, 1024)
+        max_texture_size = self.info["max_texture_size"] = max(max_texture_size, 1024)
         max_renderbuffer_size = max(max_renderbuffer_size, 1024)
 
         # The number of pixels of additional border, so we can load textures with


renpytom:
Would this provide a meaningful piece of information? Just about every device supports textures of size 1024, nowadays.

mal:
It's exposing the `max` value, so `1024` is the floor, not the ceiling. I have tested with a GPU that reports 16384, and some poking around on the internet suggests that modern GPUs support 32768, while older mobile and embedded GPUs are in the 1024-2048 range.

During my travels I did find this site ([GLES reports](https://opengles.gpuinfo.org/displaycapability.php?name=GL_MAX_TEXTURE_SIZE) | [GL reports](https://opengl.gpuinfo.org/displaycapability.php?name=GL_MAX_TEXTURE_SIZE)) which gives a rough distribution of reported values based on thousands of reports, and allows you to click into the different groups to get a feel for the sorts of devices reporting the various values. It's not foolproof, but seems like a good starting point. Part of the reason I'm only adding this to `renpy.display.draw.info` for now is because it would definitely benefit from validation before we choose whether to rely on it for a more friendly public-facing API.

This wasn't the first thing I looked at either, I spent some time trying to work out a reliable way to pull something like VRAM from hardware, but currently — except _maybe_ via Vulkan — there's not really a good way to do so. The `video_mem` value in `renpy.display.draw.display_info` reported via `pygame`/`SDL` is notoriously unreliable even in the best case of having official and fully functioning drivers; it reported a value of 256M to for a card with 2GB which, best I can tell is a fallback/placeholder value for OpenGL, behaviour which appears to be what prompted its removal from SDL3.

tl;dr: It seems like it would be a useful piece of information, but it's not certain, hence the very low-touch exposure for now.

PR #6697, opened 2025-10-26
Type and document `renpy.object` module

Andykl:
This PR adds types and some documentation for renpy.object.Object. I also added TypeError if one adds non-empty slots to a subclass, because right now slots could exist but are silently discarded.

I was also thinking that we could automatically extent `nosave` of current class with lists from its parents. Right now if you add a field to nosave in parent, you need to also check if all its children would need it.

### DIFF
--- a/renpy/object.py
+++ b/renpy/object.py
@@ -19,7 +19,7 @@
 # OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 # WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-from typing import Any, Callable, ClassVar, final
+from typing import Any, Callable, ClassVar, Self, final
 
 # Allow pickling NoneType.
 import builtins
@@ -42,7 +42,7 @@ class Object:
     you could use class variable with a sensible default value.
     """
 
-    nosave: list[str] = []
+    nosave: ClassVar[list[str]] = []
     """
     A list of field names that would be automatically removed from the
     pickle data. This is used to prevent storing unpickleable, cached
@@ -52,15 +52,14 @@ class Object:
     def __getstate__(self) -> dict[str, Any]:
         rv = vars(self).copy()
 
-        for f in self.nosave:
-            if f in rv:
-                del rv[f]
+        for f in type(self).nosave:
+            rv.pop(f, None)
 
         rv["__version__"] = self.__version__
 
         return rv
 
-    after_setstate: Callable[[], Any] | None = None
+    after_setstate: ClassVar[Callable[[Self], Any] | None] = None
     """
     A function that is called after the object is unpickled.
     This is mostly used to set values for `nosave` fields.


valery-iwanofu:
I just thought maybe Sentinel could be generic, something like this:
```python
class Sentinel[T: LiteralString]:
    name: str

    def __new__(cls, name: T) -> 'Sentinel[T]':
        ...
```

And then with some boilerplate it can become typesafe:
```python
type DefaultType = renpy.object.Sentinel[Literal["default"]]
Default: DefaultType = renpy.object.Sentinel("default")

def translate_string(s: str, language: str | DefaultType | None = Default) -> str:
    ...

translate_string('str', Default) # ok
translate_string('str', renpy.object.Sentinel('default')) # ok
translate_string('str', renpy.object.Sentinel('other sentinel')) # fail
```

Andykl:
Unfortunatelly, no. Without proper support from type checker there is no way to type that, but PEP 661 is deffered.

`Sentinel('other sentinel')` evaluates to `Sentinel[Unknown]` which is assignable to `Sentinel[Literal["anything"]]`, so your fail case does not work with pyright. But even if it works, we still need type checker to understand that `language is Default` narrows type to `str | None` after that, which what PEP 661 was trying to specify. Without it we will need ugly 
```py
    if isinstance(language, Sentinel):
        if language is Default:
            ...
        else:
            raise TypeError("Wrong sentinel value")
```
to make types work.

Any works good here practically speaking, and it is sound according to Python semantic. The only change I would like to do here, is to tweak sphinx to understand Sentinel and write `str | None = Default` in documentation, instead of [exposing Sentinel type](https://renpy.org/doc/html/translation.html#renpy.translate_string).

valery-iwanofu:
I was 100% sure that `language is Default` would work. Although this particular problem could be solved with `TypeGuard`, the more details there are, the more awkward it becomes.
I thought there was a clever solution, even with type checker support, but it didn't work out. It's definitely better to wait for built-in Sentinels.

lauwurence:
Since it was decided to document it, I'll try to share one issue that I encountered. The objects that are called in callbacks like `config.periodic_callback` or `config.scene_callbacks` are indeed updated with `after_upgrade`, but then are being rewritten with save files discarding all the new variables.

Here I tried to show that `var_2` does not appear in the World class after loading save file if the `world.add_time()` function is called in callbacks.

Video:
https://drive.google.com/file/d/12wRtCD7tpAl7mMRYwJjia_oM9sYJ-UUA/view?usp=sharing

<details>
  <summary>test.py</summary>
  
```renpy
################################################################################

init python:

    class Object(renpy.revertable.RevertableObject):

        __version__ = 0

        nosave = [ ]

        def __getstate__(self):
            rv = vars(self).copy()

            for f in self.nosave:

                if f in rv:
                    del rv[f]

            rv["__version__"] = self.__version__

            return rv

        # None, to prevent this from being called when unnecessary.
        after_setstate = None

        def __setstate__(self, new_dict):
            version = new_dict.pop("__version__", 0)

            self.__dict__.update(new_dict)

            if version != self.__version__:
                self.after_upgrade(version)  # type: ignore

            if self.after_setstate:
                self.after_setstate()  # E1102

    class World(Object):

        __version__ = 0
        nosave = []

        def after_upgrade(self, version):
            self.var_2 = True

        def __init__(self):
            self.time = 0
            self.var_1 = True

        @property
        def vars(self):
            return str(", ".join([str(v) for v in vars(self)]))

        def add_time(self):
            print(f"Time: {self.time}")
            self.time += 1


    ############################################################################
    # Callbacks

    def periodic_callback():

        if renpy.is_init_phase() or main_menu:
            return

        world.add_time()

    config.periodic_callback = periodic_callback

    def character_callback(event, **kwargs):

        if renpy.is_init_phase():
            return

        if event == 'begin':
            world.add_time()

    config.character_callback = character_callback

    def scene_callback(layer):

        if renpy.is_init_phase():
            return

        world.add_time()

    if scene_callback not in config.scene_callbacks:
        config.scene_callbacks.append(scene_callback)


################################################################################

default mc = Character("Character")
default world = World()

label start:
    scene black
    mc "..."
    mc "World Variables: [world.vars]"
    jump start
    
```
</details>

In my project I fixed it with a variables that is only `True` after `after_load` and used it in those callbacks to prevent the `world` variable from too early updating. But still it does not look like a good solution to me.

<details>
  <summary>Workaround</summary>
  
```renpy
define _is_game_phase  = False

init python hide:

    def start_callback():
        _is_game_phase = False

    config.start_callbacks.append(start_callback)

    def before_load_callback():
        _is_game_phase = False

    config.before_load_callbacks.append(before_load_callback)

    def after_load_callback():
        _is_game_phase = True

    config.after_load_callbacks.append(after_load_callback)
```

</details>

Andykl:
The problem here is buggy(?) combination of revertable protocol which call `World._rollback({'time': 176, 'var_1': True})` after unpickle called `World.after_upgrade(1)`.
Because `renpy.object.Object` is still internal API, I am not sure if it is somethig need to be fixed, but Ren'Py has combination of `Object` and `RevertableObject` in its public API - `renpy.Displayable`.

Andykl:
Could this be merged in fix, so [Sphinx wart](https://renpy.org/doc/html/translation.html#renpy.translate_string) would be fixed in next build? With new repr it will document function as `renpy.translate_string(s, language=renpy.object.Sentinel('Default'))`

Andykl:
Agree with @mal. Some of those changes are fixes of appearing typing errors, some are fixing unsound behavior that are not type error per se. It is knee-jerk reaction to fix typing error after I see it appearing, but that makes scope of PR bigger, which is bad.

I will close this PR and will open several others splitting changes on typing only, fixing typing errors, and other changes.

PR #4064, opened 2022-11-01
Add documentation for Ren'Py Extensions

Ayowel:
I hope this closes #3521

I have been fiddling with extensions recently and tried to document the most obvious pitfalls/limitations. The example is not great, as extensions should probably not be used for something so simple, but its aim is more to demonstrate the how than the what and I thought that re-using something already documented would give a comparison reference.

Shameless plug: I tried to create a small 'game' that explains how to create an extension and progressively shows more complex examples I'd love some feedback on it https://ayowel.itch.io/renpy-extensions-demo (the repository is here: https://github.com/Ayowel/renpy-extension-demo)

PR #6701, opened 2025-10-27
Document RPE files

### LINKED ISSUE #5623: Clarify in documentation what .rpe files are and how they can be used
Renpy has had `.rpe` files since version 6.1 however, there's no documentation that they even exist besides release notes.

Please clarify them in the documentation. Including who are the target audience for the feature and who is supposed to create such files.

arsym-dev:
Builds on #4064, fixes #5623

### DIFF
--- a/sphinx/source/developer_tools.rst
+++ b/sphinx/source/developer_tools.rst
@@ -182,3 +182,65 @@ that are available to click on. To do this, create project.json with the lines b
 
 If the file already exists, you'll want to edit in the renpy_launcher key and the lines below it.
 You can then edit the dictionaries to change the available files and directories.
+
+.. _rpe-extensions:
+
+Ren'Py Extensions
+-----------------
+
+Ren'Py Extensions (RPE) allow you to patch the Ren'Py engine before any other file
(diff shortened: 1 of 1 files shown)


renpytom:
As mentioned in the issue, I'm really skeptical about documenting .rpe files. I wouldn't have added them today, and it's basically encouraging people to play with internal engine details in a way I don't want to support.

arsym-dev:
Oh, that confuses me since #5623 is marked as an 8.4.2 milestone.

I can change the documentation to say it's development-only and shouldn't be released in project builds. Maybe a giant warning that any changes to the engine made this way are unsupported and it's on the script writer to debug their own issues.

I feel it has a legitimate role in IDE extensions, debugging tools, custom build stuff, etc.

Andykl:
Agreed. I think today we would advice people to tweak build process to include their patch, so at least we would be able to see it is unofficial distribution of Ren'Py, so we would ask what changes they did to the engine. With existence of RPE files we could only guess if the reported bug is not caused by bad RPE. I wonder should we mark installation unofficial once rpe file is loaded?

PR #6730, opened 2025-11-08
Correct wrong brightness -> contrast in ContrastMatrix documentation

RubenGarcia:


renpytom:
Thanks for the contribution, I appreciate it. 

In the future, consider making the change on the fix branch. The documentation on the website comes from the fix branch, so it goes live much faster.

PR #6734, opened 2025-11-09
use "input_up" and "input_down" to move the caret

### LINKED ISSUE #6732: up/down arrow in an input should move the caret up/down
the keysyms are defined, but they aren't used anywhere

Elckarow:
fixes #6732


renpytom:
Merged. This works well for monospaced text, less so for proprtional text. But that's fine.

PR #6753, opened 2025-11-20
Fixed priority on define, default and transform statements

bkats:
Fixed not always handling priority set on the statements define, default and transform.

Set priority was ignored when already in an init block:
```renpy
init:
    define -10 not_at_right_priority = -10
```

Also this priority option is undocumented for these statements. Is that on purpose or is it just missing?

### DIFF
--- a/renpy/parser.py
+++ b/renpy/parser.py
@@ -925,9 +925,11 @@ def image_statement(l, loc):
 
 @statement("define")
 def define_statement(l, loc):
+    create_init = not l.init
     priority = l.integer()
     if priority:
         priority = int(priority)
+        create_init = True
     else:
         priority = 0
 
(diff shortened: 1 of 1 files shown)


Andykl:
I don't think that option should be exposed. IMO it is harder to read than init offset or init block with priority, and when used together it's even harder. Without reading docs (if we document it) can you tell if it adds or overwrite outer priority? 

So I would leave it as is (I think it exists for backward compatibility) and would not fix its quirks. But final call is Tom's here, it is just my opinion.

bkats:
I'm not sure about adds or overwrite, I would say adds? The following code fails before the fix and works with the fix:
```renpy
init:
    default test = "one" + test_first
    default -10 test_first = "three"
```
Which would parsed into something similar as following:
```renpy
init:
    default test = "one" + test_first
    init -10 default test_first = "three"
```

About legacy/unclear code I had already some idea like that about it. So probably leave the documentation as is.
Fix would still make sense or maybe remove the priority option completely.

PR #6762, opened 2025-11-24
Update ukrainian (8.5.0)

Veydzher:
Updated translation files.


renpytom:
Just a suggestion - when we're just after a release like this, it makes sense to target the fix branch. I've merged this in now, but in general the fix branch will get released earlier and into people's hands.

Veydzher:
> Just a suggestion - when we're just after a release like this, it makes sense to target the fix branch. I've merged this in now, but in general the fix branch will get released earlier and into people's hands.

Sorry about that. I'll target the fix branch for future commits.

renpytom:
Nothing to be sorry about! I appreciate your translations, just wanted to let you know how to make sure they get out to people ASAP.

Issue #6389, opened 2025-06-06
Multiple image loading threads.


renpytom:
It may make sense to have multiple image loading threads, to load images faster on a multi-core computer. 

Suggested in the discussion of #6370.

PR #6770, opened 2025-11-28
Add multithreading option to image loading

ronchonbon:
Per #6389, with config option for how many threads to allocate. Defaults to 0, which sets behavior to automatically allocate up to 4 threads based on `os.cpu_count()`. Setting option to 1 reverts to previous, single-threaded behavior.

### DIFF
--- a/renpy/display/im.py
+++ b/renpy/display/im.py
@@ -514,6 +514,10 @@ class Cache:
         Creates it lazily if needed. Returns None if parallel decoding is disabled.
         """
 
+        # Multithreading is not supported on Web
+        if renpy.emscripten:
+            return None
+
         if self.decode_pool is not None:
             return self.decode_pool
 


bober-gith:
This is amazing so far, thank you. Tested it and it reduced the load time from 20sec to 8 sec in a preloading scenario I had on PC, so far didn't see any issues.
EDIT: Just a thought but maybe a small safeguard could be to make it consume 1 or 2 core less than the max supported by the system (but still capped at a max value), mainly thinking about those low end android phones. So that it always leaves at least some cores unused to give some breathing room if the system is weak.

Kassy2048:
On Web, multithreading is not supported, so `get_decode_pool()` should ignore the value of `config.preload_threads` and always return `None` in this case (when `renpy.emscripten` is true-ish).

renpytom:
This is looking good to me. Thank you for the contribution.

Andykl:
I am not sure it would work on Android (are threads are java green threads here?) or iOS (I have no idea how Obj-C implemet threads), so someone need to test it.

bober-gith:
I'll try test next week later for android, but if someone has done it already, please let us know.

Wanted to share some small searches...

- For Android, modern devices use a combination of high performances core and low performance cores and pushing too much can lead to worse performance and overheating. The value of 4 threads seem show up often as a reasonable default, so the setup here is well tuned for Android it seems.
- On PC there isn't a problem using all cores, though it seems to be considered good practice to do max cores - 1 core, just for extra safety measure, and it's a minimal give away of speed anyway.

Mixing Android and PC into one profile might leave a lot of performance on the table for PC, because for Android you have to be much more conservative, so the config might be hard to use if you want to scale on PC but keep it nice for Android. 
Right now it's setup for Android, which makes sense since it's all in one and the lowest denominator, but I tried 8 threads instead 4 threads on my 8 cores CPU and the gains were also great compared to 4 cores (in my test, 1 thread was ~20sec, 4 threads ~8sec, 8 threads ~4sec). It seems like it can keep scaling much more beyond 4.
However with current setup, you would have to leave it at 4 just because this might lead to issues on Android.
(I guess you can just change the config value before an android build or pc build, so this might be a non issue.)

ronchonbon:
I haven't had any issues testing so far on a Pixel 7 Pro - can't test on iOS myself.

I figured it made sense to default at something that would work for most devices, and then add this to specific projects:

`define config.preload_thread_autocap = 4 if renpy.android else 8`

For my projects, I set it up with a toggleable persistent variable.

bober-gith:
I did the same setting (but in the sdk). 😄
And I mean I agree, this is has good as it can get for all in one I think. I'm just thinking if someone try to raise it to up his performance, it might look good on PC but he might not realize he is creating problems for his Android build because everything would look fine on the PC. Just wanted to share this observation, at the end of the day, this isn't a problem for me or you.

Issue #4562, opened 2023-04-23
Default Layer/Camera Transforms / Full Screen Transforms


colinharvie:
I've implemented this in my own game, but I think many people would find it useful.
Being able to set a "default" transform for a layer is very useful, OR a transform that is applied to every layer, regardless of other transforms. As the game currently is, I would need to re-set the default transform every time the scene reset, or run the risk of accidentally replacing the transform on the layer. 

Why do I want to do this? I have a custom transform that makes uses of shaders that I want to be able to toggle on and off in the settings, meaning that I ALWAYS want the transform applied:

I've probably not implemented it the best way, because I was just futzing around until I got it to work, but here's how I did it:

    default_at_list = {
        "screens":[FullScreenTransform]
        }

    always_at_list = {
        "master":[FullScreenTransform]
    }

    def screenshader_set_layer_at_list(self, layer, at_list, reset=True, camera=False):
        global always_at_list
        layer_at_list = list(at_list)
        if (layer in always_at_list):
            layer_at_list.append(*list(always_at_list[layer]))
            
        if camera:
            self.camera_list[layer] = (None, list(layer_at_list))
        else:
            self.layer_at_list[layer] = (None, list(layer_at_list))

        if reset:
            self.layer_transform[layer] = None

    renpy.display.core.SceneLists.set_layer_at_list = screenshader_set_layer_at_list

    

    def screenshader_clear(self, layer, hide=False):
            global default_at_list
            if layer not in self.layers:
                return

            if not hide:
                self.layers[layer][:] = [ ]

            else:

                # Have to iterate in reverse order, since otherwise
                # the indexes might change.
                for i in range(len(self.layers[layer]) - 1, -1, -1):
                    self.hide_or_replace(layer, i, hide)

            self.at_list[layer].clear()
            self.sticky_tags = {k: v for k, v in self.sticky_tags.items() if v != layer}

            self.shown.predict_scene(layer)

            if renpy.config.scene_clears_layer_at_list:
                layer_at_list = []
                if layer in renpy.store.default_at_list:
                    layer_at_list = renpy.store.default_at_list[layer]
                self.layer_at_list[layer] = (None, layer_at_list)

    renpy.display.core.SceneLists.clear = screenshader_clear

Also for applying a full screen transform to everything:

    renpy.store.screen_transform = FullScreenShader

    def draw_screen(self, root_widget, fullscreen_video, draw):
        try:
            renpy.display.render.per_frame = True
[...]

PR #6793, opened 2025-12-10
Type `renpy.display.scenelists` module

arsym-dev:
Typing and documentation

The only non-typing change is a minor refactor of `SceneList.__init__()`. The line `self.focused = None` was common to both branches, so I pulled it out of the branches and annotated it.

The typing highlights logic issues in the module that should be resolved in a future PR.

### DIFF
--- a/renpy/config.py
+++ b/renpy/config.py
@@ -175,7 +175,7 @@ sticky_layers = ["master"]
 
 # Layers not automatically added to a scene and inherently sticky,
 # primarily for use with the Layer displayable.
-detached_layers = []
+detached_layers: list[str] = []
 
 # True if we want to show overlays during wait statements, or
 # false otherwise.
@@ -711,7 +711,7 @@ default_tag_layer = "master"
 tag_transform = {}
 
 # A map from the tag to the default zorder that's used for that tag.
-tag_zorder = {}
+tag_zorder: dict[str, int] = {}
 
 # The width of lines logged with renpy.log.
 log_width = 78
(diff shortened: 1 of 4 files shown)


renpytom:
Please hold off before producing typing PRs like these. These aren't terrible, but I'm considering if we want to move to Python 3.14, and that changes typing. 

(I would like to know about the logic issue.)

arsym-dev:
Sure thing, I won't mess with these until we look into python 3.14

Regarding logic, the main issue is that `self.config_layer_transform` doesn't know if it's `dict[str, Transform]`, or `dict[str, list[Transform]]`.

### Acting as list[Transform]
https://github.com/renpy/renpy/blob/2fea8043e69718ee8c7ad702a6bc4c940bddcd94/renpy/display/scenelists.py#L103-L104

### Acting as Transform

In both snippets, you want to focus on the first and last lines. Notice how `old_thing._target()` in `transform_state()` would not work on a list.
https://github.com/renpy/renpy/blob/2fea8043e69718ee8c7ad702a6bc4c940bddcd94/renpy/display/scenelists.py#L763-L779

https://github.com/renpy/renpy/blob/2fea8043e69718ee8c7ad702a6bc4c940bddcd94/renpy/display/scenelists.py#L274-L288

I think the correct behavior is `list[Transform]` since the comments explicitly state that it's meant to map onto `config.layer_transforms`. These changes were introduced to address #4562 but I have no idea if it resolved OP's issue.

PR #6796, opened 2025-12-13
Imrove Sentinel repr.

Andykl:
That way, when it is used as function parameters default, sphinx will produce reproducible signature, instead of `<Sentinel at 0x...>`.


Andykl:
Example of documentation that is changed each time docs are updated.
image

Issue #4614, opened 2023-05-08
Physics calculations not being performed or displayed on live2d models despite the physics3 file being present and defined


thynetruly:
The physics and animations are working normally in the live2d viewer, but despite the animation/motion/expression calls working normally, the physics are not appearing on the model during the movements, which makes it look stiff and plastic. Is there a way to verify that the code which does the physics calculations in the renpy engine using the live2d native sdk runtime is working normally?

renpytom:
Right now, that isn't implemented.

thynetruly:
Can this be turned into a request to have that implemented?

renpytom:
I'll have to see. Last time, I couldn't find documentation on how the format was meant to be interpreted.

thynetruly:
> Don't know if this can help: https://docs.live2d.com/en/cubism-sdk-manual/physics/

It can! Thank you!

stunnings:
Physics support for Live2D is absolutely necessary. The hair movement is completely gone, and the hair looks stiff and fixed as it moves. @renpytom, is there any progress on this issue?

When Ren’py did not officially support Live2D, there was a third-party module made by asfdfdfd ( https://github.com/asfdfdfd/renpy-live2d ) that is now deprecated since Ren’py has officially integrated Live2D. If you use this module, you can see in this video ( https://youtube.com/watch?v=mAe7IXWvZBo ) that the hair flows naturally. This suggests that the module supported physics. I hope this will help you to apply physics to Ren’py.

thynetruly:
> Physics support for Live2D is absolutely necessary. The hair movement is completely gone, and the hair looks stiff and fixed as it moves. @renpytom, is there any progress on this issue?
> 
> When Ren’py did not officially support Live2D, there was a third-party module made by asfdfdfd ( https://github.com/asfdfdfd/renpy-live2d ) that is now deprecated since Ren’py has officially integrated Live2D. If you use this module, you can see in this video ( https://youtube.com/watch?v=mAe7IXWvZBo ) that the hair flows naturally. This suggests that the module supported physics. I hope this will help you to apply physics to Ren’py.

I would also like an update on this since it's essential to displaying hair and fur accurately such as in https://twitter.com/KetKatto/status/1704308095986651581?t=2rlYeiNg1iV-ey3qB-awAw&s=19. If https://github.com/asfdfdfd/renpy-live2d can be implemented or merged so that physics would work as expected that would be a big help to our development.

thynetruly:
[...]

PR #6818, opened 2025-12-23
Add physics engine for GL2 models

ronchonbon:
Per https://github.com/renpy/renpy/issues/4614, re-implementation of Live2D physics support in Cython.

I use an alternate formulation of Live2DModel for my projects with a different flow, but I've placed the example call to `evaluate_physics` in I think a logical place within `Live2D.render` - as the last call after parameters have been updated by the `update_function` pipeline. 

This placement does have a downside: because Live2D physics operates on parameters, any modification of a physics-affected parameter by `update_function` or an expression will be overridden by the physics engine, leading to glitchy movement. The most obvious workaround is to only manually set non-physics parameters. There are relatively simple alternate workflows if you want a parameter to have both physics and be manually settable, which I'm happy to discuss with anyone who's interested.


Keinart:
Absolutely incredible implementation for all Live2D devs, almost obligatory since baking physics during the animation process usually lead to strange results. I really hope @renpytom checks it out and this goes live as soon as possible. Thank you so much for this.

raspberry-soft:
Wow! Waited so long for this! Any updates on this?

ronchonbon:
Probably not a priority while 8.5.3/8.6.0 are being worked up, but the changes are drop-in compatible with both `master` and `fix` so you can start using physics with your Live2D models as is if you wish.

renpytom:
I'd expect this to go in right after 8.5.3 is out.

ronchonbon:
I've been using this module for some other physics modeling, since it's really just an implementation of pendulum physics applied to a parameter set. It's had me mulling whether to rename to `gl2physics` or something - I'm leaning towards it on my fork, but I thought I'd offer to do it for this PR as well if @renpytom doesn't object.

ronchonbon:
The most recent commit shows how I'm doing it on my fork now. `live2dphysics` is now a small module that parses physics3.json files to the format `gl2physics` expects, which is now a little more lean as a module.

This makes sense to me and appeals to my aesthetics, but we can go back to the previous state if it's your preference.

Andykl:
I really like this general solution, because it can be later used for other things, e.g. Ren'Py particles class. 
My nits are: 
1. Why raw malloc and not python PyMem? 
2. Why have separate initialize method instead of using `__init__`? It's not like one would create a lot of physics objects in a hot loop.
3. Because of 2, live2d initialization looks clumsy. In my opinion physics object should live  in Live2D common, all parsing and creation live in live2dphysics, so live2dmodel does not care about physics (if I understand correctly, they are not directly wired). 
4. Formatting nits - no new lines at end of files and some lines are wrapped even on my wide monitor, we have 120 chars as wrap limit in ruff settings.

ronchonbon:
1. Just preference and habit to be honest, it's the pattern I saw most while poking around in /gl2/ and I adopted it. Used in `live2dmodel`, for example.

2 + 3. It's because `PendulumPhysics` expects raw C pointers for `parameter_values`/`parameter_minimum/maximum`/etc., blocking `__init__` and `__cinit__`. I don't disagree that it's kind of awkward as an API, but I wasn't bothered enough to try something more clever. Happy to brainstorm.

4. Sure, happy to format this (and https://github.com/renpy/renpy/pull/7152) to repo preference once the structure is agreed upon.

Andykl:
1. I think because it is a type managed by Python it is better to also allocate memory using python allocator. At some point I think it worth update other places that do not pass pointer to that alloc to SDL or other non-python-aware code.
2. Well, it expects raw C pointer because it takes e.g. `parameter_count` that is cdef. But it can be computed from `len(Live2DModel.parameters)`. I think of something like `PendulumPhysics(*parameters: ParameterLike, **rig: Unpack[RigData])` where `ParameterLike` and `RigData` are `TypedDict`s with well-defind structure.
3. After 2 is done, with some changes that allow to get correct `ParameterLike` from `Live2DModel` init of `PendulumPhysics` for live2d would look much more pythonic and can be done outside of cython.

This could be done later ofc, I have no strong preference here because this code is not public API.

ronchonbon:
Could do that, but it would require replumbing `Live2DModel` since e.g. `parameter_values` is a pointer read directly from the model through the Cubism SDK. It's way more convenient to just have the physics engine write in-place.

I don't feel strongly about the end result either, but I think since it would require more extensive rework of Live2D, it's probably not in the scope of this PR.

ronchonbon:
I want to spend some time thinking about @Andykl's suggestions, I was hesitant about tweaking too much, but maybe I can either keep it relatively contained or break off some concurrent PRs to change how `live2d` and `live2dmodel` interact with the parameter buffers.

ronchonbon:
@Andykl, what do you think of this clean up? It feels better to me, is it closer to what you were imagining?

Andykl:
Yes, this is much cleaner. I am not sure without deeper code read, but it feels some things could be more RAII (have fully initialazed object in cinit/init). Also some nits like depricated DEF which could be replaced with anonymous enum. 
Overall I think it could be merged, and cleaned later.

Andykl:
I see my confusion. I expected
```
cdef class PendulumPhysics:
    def __cinit__(self, ...):
        do all the allocations and C initialization and validation

    def __init__(self, ...):
        do validation of Python side

    def __dealloc__(self):
        do all free
```
Because `__cinit__` could not be called multiple times, it is idiomatic RAII. With current `cdef void _parse` it was not obvious.

Also ParameterBuffer seems to be used differently in 2 contexts, via `ParameterBuffer(...)` and `bind_parameters(...)`, and in first case owner is `tuple` and in other owner is `Live2DModel`, or does I miss something?

ronchonbon:
Ah yes I see how that's confusing. `_parse` is a remnant of the original Cubism structure that has persisted - I have no problem with rearranging initialization to be more intuitive.

I think what you're referring to is the crossing of Cython/Python boundaries that happens on `live2dphysics.py:34`? `bind_parameters` returns `ParameterBuffer`, and that gets passed directly to `PendulumPhysics` through the `def` intermediate `get_physics_parameters`. Am I missing an obvious way to make that interface more elegant or avoid it entirely?

Andykl:
Yes, for `ParameterBuffer`, if I understand the code correctly, I think it makes more sense to drop `bind_parameters` and instead add to live2dmodel
```py
@cython.final
cdef class _FloatView:
    """
    Buffer exporter over model-owned floats. Holds the owner, so the exporter
    (and anything pinning it) keeps the model object alive.
    """

    cdef object owner
    cdef const float* data
    cdef Py_ssize_t length
    cdef Py_ssize_t stride
    cdef bint readonly

    def __init__(self, *args, **kwargs):
        raise TypeError("_FloatView is created by Live2DModel.")

    @staticmethod
    cdef _FloatView create(object owner, const float* data, int count, bint readonly):
        cdef _FloatView view

        if owner is None or count < 0 or (count and data == NULL):
            raise ValueError("A float view requires an owner and a valid pointer.")

        view = _FloatView.__new__(_FloatView)
        view.owner = owner
        view.data = data
        view.length = count
        view.stride = sizeof(float)
        view.readonly = readonly
        return view

    def __getbuffer__(self, Py_buffer* info, int flags):
        if self.readonly and (flags & PyBUF_WRITABLE):
            raise BufferError("Buffer is read-only.")

        info.buf = <void*> self.data
        info.obj = self
        info.len = self.length * sizeof(float)
        info.itemsize = sizeof(float)
        info.readonly = self.readonly
        info.ndim = 1
        info.format = "f" if (flags & PyBUF_FORMAT) else NULL
        info.shape = &self.length
        info.strides = &self.stride
        info.suboffsets = NULL

def get_physics_parameters(self):
    cdef int n = self.parameter_count

    return ParameterBuffer(
        _FloatView.create(self, self.parameter_values, n, False),
        _FloatView.create(self, self.parameter_minimum_values, n, True),
        _FloatView.create(self, self.parameter_maximum_values, n, True),
        _FloatView.create(self, self.parameter_default_values, n, True),
        {name: parameter.index for name, parameter in self.parameters.items()},
    )
```

and
```py
cdef class ParameterBuffer:
    def __cinit__(self, values, minimum, maximum, defaults, dict indices not None):
        # body of current ParameterBuffer.__init__
        # body of current ParameterBuffer.validate

    # No __init__ at all, validate is done in __cinit__
```

So, there is only one initialization path, and it is impossible to have half-initializated or unvalidated `ParameterBuffer` object. Creation of ParameterBuffer should not be in a hot loop so small amount of buffer rewrapping I think is okay.

Not sure if that makre interface cleaner.

PR #6821, opened 2025-12-25
Don't translate Preference() names internally

SpikyCaterpillar:
Previously the Preference() function was checking to see if the preference name matched the *translation* of various constants, so if you had a translation that included a translation for a preference name then games would crash on start because they can't find the right pref. AFAICT nothing actually calls Preference() with a translated name.

### DIFF
--- a/renpy/common/00preferences.rpy
+++ b/renpy/common/00preferences.rpy
@@ -292,7 +292,7 @@ init -1500 python:
 
         def get():
 
-            if name == _("display"):
+            if name == "display":
                 if value == "fullscreen":
                     if renpy.can_fullscreen():
                         return SetField(_preferences, "fullscreen", True)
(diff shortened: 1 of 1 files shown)


mal:
The single underscore function (`_`) is a [no-op function](https://github.com/renpy/renpy/blob/f5fcd835377acded883a1a184e4979fb44bf9a6a/renpy/minstore.py#L90-L98) used to signify that a string should be made available for translation when generating translations. It does not perform any kind of runtime substitution, so the only result of this PR would be the removal of these strings from generated translations which is not desirable for the project and does not match the description of the issue being addressed.

If your analysis of the problem you're seeing is correct, it seems likely that something in your code base is overriding the `_` function, which is not supported as names and identifiers beginning with a single leading underscore are reserved for use by Ren'Py.

renpytom:
@mal  Is correct here - and the translated strings are used for TTS. So this change isn't necessary.

Andykl:
Yes, @mal is correct, that PR most likely fixes an issue with user code.

PR #6829, opened 2025-12-29
check for empty list in `include_module`

### LINKED ISSUE #6828: `renpy.include_module` from `python early` blocks

Elckarow:
fixes #6828


Elckarow:
update: this was happening because it was called from a `python early` block

PR #6840, opened 2026-01-01
Ensure can_rollback does not mutate values

mal:
- Removes mutation of `self.rollback_block` during the read-only `can_rollback` check.
- Removes double decrement of local `checkpoints` (see the if statement immediately following removed block).

This is mostly a speculative fix, I've done some cursory testing and believe it to be correct but I've not used it in anger.

---

Second commit addresses the short circuit check that wasn't accounting for `checkpoints` being a numeric value.

### DIFF
--- a/renpy/rollback.py
+++ b/renpy/rollback.py
@@ -878,8 +878,8 @@ class RollbackLog(renpy.object.Object):
         it returns False.
         """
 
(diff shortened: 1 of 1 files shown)


renpytom:
6dc4d05 was right, but 1e93148 seems wrong - even if it passes the check, we still want to check the rollback will succeed. 

I'll make the change directly.

PR #6842, opened 2026-01-01
Support explicit clear of roll-forward data

mal:
A small signature extension to make clearing of roll-forward data after rollback possible explicitly, rather than only as a side-effect of loading. Previously this was possible with a `finally` block, but the rollback changes in 8.5.1 have closed off this avenue.

Picked `forget` for the argument since it's effectively choosing to forget the forward action, but open to other names if people think that's easier to grok, perhaps `clear`.

### DIFF
--- a/renpy/rollback.py
+++ b/renpy/rollback.py
@@ -900,7 +900,7 @@ class RollbackLog(renpy.object.Object):
         return False
 
     def rollback(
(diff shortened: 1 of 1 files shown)


renpytom:
I'm leaning towards this, but what's the use case? It probably should be exposed in exports as well.

mal:
Following some discussion agreed that there are probably better ways to support target use case with existing functionality.

PR #6847, opened 2026-01-03
Improve lexer part 1

Andykl:
This PR is made on top of #6839.

This PR moves tokenization of strings and operators into `lexersupport` and splits `match_whitespace` from current `match_logical_word`.

Also, in this PR I added support for nested same quotes in f-strings, but `match_string` does not validate its content in any way, so there are still corner case that Python tokenizer handles better:
```
x = f"{"
...
```
Python will raise `SyntaxError: f-string: expecting '}'` at closing quote, my code will raise `SyntaxError: unterminated string literal` after reading whole file unless there are also unbalanced `}` somewhere.

Overall this PR improves speed of file tokenization by 2-3 times

<details>
  <summary>Benchmark code</summary>

```py
import re
import os
from pathlib import Path

import renpy

renpy.import_all()

basedir = Path(os.environ["BENCH_DIR"])

files = {
    str(p): p.read_text("utf-8")
    for p in [
        *basedir.rglob("*_ren.py"),
        *basedir.rglob("*.rpy"),
    ]
}

print(f"Found {len(files)} files with total of {sum(len(i) for i in files.values()):_} characters.")


def bench():
    for fn, data in files.items():
        renpy.lexer.list_logical_lines(fn, data)
```
and command `.\lib\py3-windows-x86_64\python -m timeit -n 1 -r 10 -s "from bench import bench" "bench()"`
</details>

All stats are compared with 8.6.0.26010201 nightly
My project with all comunity tl files from
```
1 loop, best of 10: 26.1 sec per loop
```
to
```
1 loop, best of 10: 10.7 sec per loop
```

Lanuncher from
```
1 loop, best of 10: 744 msec per loop
```
to
```
1 loop, best of 10: 206 msec per loop
```

The Question from
```
1 loop, best of 10: 107 msec per loop
```
to
```
1 loop, best of 10: 40.8 msec per loop
```

### DIFF
--- a/renpy/common/00keymap.rpy
+++ b/renpy/common/00keymap.rpy
@@ -544,13 +544,6 @@ label _load_reload_game:
     if not renpy.session.get("_reload_slot", None):
         return
 
-    if renpy.session.get("_did_load_reload", False):
-        python:
-            renpy.rename_save(renpy.session["_reload_slot"], "_reload-2")
-            del renpy.session["_reload_slot"]
-            del renpy.session["_did_load_reload"]
-        return
-
     if not renpy.can_load(renpy.session["_reload_slot"]):
         $ del renpy.session["_reload_slot"]
         return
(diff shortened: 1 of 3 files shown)


Andykl:
I mixed up my branches, so everything before I marked that PR as ready for review is related to cython and `list_logial_lines` changes.

Everything after that updates lexer class to use new cython functions for most common functions - `skip_whitespace` and `word`, reducing amount of calls to match_regexp by about 80%.

Also it now uses f-string aware cython function in `python_string`, so users could nest their quotes by the same rules as Python.

Altogether that PR reduces force recompile time of the launcher from ~5 secs to ~2.8, and my main project from ~100 secs to ~60.

Andykl:
I merged master, so now diff in `lexer.py` is much more clean.

PR #6858, opened 2026-01-11
Centralize python cache code

Andykl:
This PR decouples `renpy.script` and `renpy.python` referencing each other to cache result of `py_compile` storing it all in one dedicated class. This is related to #6328 and removes one of responsibilities of `Script` class.

### DIFF
--- a/renpy/python.py
+++ b/renpy/python.py
@@ -31,15 +31,19 @@ import ast
 import collections
 import contextlib
 import copyreg
+import importlib.util
 import marshal
+import os
 import sys
 import types
 import warnings
 import weakref
-from typing import Any
+import zlib
+from typing import Any, Final, Literal
 
 import renpy
 from renpy.astsupport import hash32
+from renpy.compat.pickle import dumps, loads
 from renpy.pydict import DictItems, find_changes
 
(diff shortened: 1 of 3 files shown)


renpy/python.py
```diff
@@ -1061,60 +1212,23 @@ def py_compile(source, mode, filename="<none>", lineno=1, ast_node=False, cache=
...
-        warnings_key = None
-        key = None
-
     source = str(source)
     source = source.replace("\r", "")
 
     if mode == "eval" and not ast_node:
         # If possible, compute the value of immutable literals.
-        try:
+        with contextlib.suppress(Exception):
```
mal: Suspect may be a fairly hot path, in which case it may matter that `contextlib.suppress` is markedly slower than `try ... except`, even in 3.12 (maybe beyond too, but at least in 3.12). Might be worth a benchmark.

renpytom:
I also don't like contextlib.suppress - it puts the error handling before what's actually being done. 

@copilot+claude-sonnet-4.6 Revert the two uses of contextlib.suppress(Exception): to try:/except Exception:/pass.

PR #6870, opened 2026-01-15
mark transform fully qualified name as const

### LINKED ISSUE #6869: Transform AST execution incorrectly registers namespaced transforms as global constants
## Issue:
When defining a transform inside a named store (`renpy.python.StoreModule`) (e.g., `transform my_module.t1():`), the AST execution logic calls `renpy.exports.pure()` using only the leaf variable name (`t1`), ignoring the store (`my_module`).

This causes the global name `t1` to be registered as pure (added to `renpy.pyanalysis.constants` and `renpy.pyanalysis.pure_functions`).

This is a false positive that pollutes the global namespace. If a developer has a global function or callable named `t1` that is not pure (has side effects), the screen language analyzer may incorrectly optimize screens that use it, believing it to be the pure transform.

## Reproduction:
```renpy
init:
    python:
        # Helpers to inspect the constants set

        def is_const(name):
            return name in renpy.pyanalysis.constants

        def is_pure(name):
            return name in renpy.pyanalysis.pure_functions

    # 1. Define a transform inside a module
    transform my_module.t1:
        pass

    # 2. Check constant analysis
    python:
        # BUG: These return True. They should be False.
        # The global string "t1" has been marked constant because of my_module.t1
        print("Is global t1 constant?", is_const("t1"))
        print("Is global t1 pure?", is_pure("t1"))

        # This is currently False (missing), but should be True
        print("Is my_module.t1 constant?", is_const("my_module.t1"))
        print("Is my_module.t1 pure?", is_pure("my_module.t1"))
```

## Analysis:
The issue lies in `renpy/ast.py` inside `Transform.execute`. It currently registers the leaf name regardless of the store:
```python
renpy.exports.pure(self.varname) # Registers "t1"
```
The screen language analyzer (`renpy/pyanalysis.py`) already supports scoped/dotted names.
1. `renpy.pyanalysis.Analysis._check_name` reconstructs dotted strings (e.g., `"my_module.t1"`) during analysis.
2. `renpy.pyanalysis.pure` handles string arguments containing dots correctly.
3. `renpy.pyanalysis.pure` also automatically strips the `"store."` prefix if present.

Therefore, the correct behavior is to always register the fully qualified name. This ensures namespaced transforms are correctly analyzed without polluting the global scope.

## Suggested Fix:
In `renpy/ast.py`, modify the `Transform` class `execute` method:

```python
# Old:
# renpy.exports.pure(self.varname)

# New:
renpy.exports.pure(self.store + "." + self.varname)
```

Elckarow:
fix #6869

### DIFF
--- a/renpy/ast.py
+++ b/renpy/ast.py
@@ -1318,8 +1318,6 @@ class Transform(Node):
         renpy.dump.transforms.append((self.varname, self.filename, self.linenumber))
 
         renpy.exports.pure(f"{self.store}.{self.varname}")
-        if not self.store or self.store == "store":
-            renpy.exports.pure(self.varname)
 
         ns, _special = get_namespace(self.store)
         ns.set(self.varname, trans)


juperMonkey:
The `pure()` function in  `renpy/pyanalysis.py`, already strips the `"store."` prefix if present.

We can safely simplify this change to just one line:
```diff
- renpy.exports.pure(self.varname)
+ renpy.exports.pure(self.store + "." + self.varname)
```

That covers both the namespaced case and the default store case without needing the extra if check.

Elckarow:
good catch

juperMonkey:
Looks good to me. Just a heads-up, your PR is currently targeting `master`. Since this looks like a bugfix, it probably belongs on `fix` (per the repo README) so it can ship in the next patch release.

PR #6872, opened 2026-01-16
Improve linter logs on im.ImageMatrix usage

Ayowel:
Consider the following code in a `game/images.rpy` file:

```py
define img1 = im.Grayscale("gui/namebox.png")
define img2 = im.MatrixColor("gui/namebox.png", [0]*20)
```

Currently, if an `im.ImageMatrix` proxy function is used (e.g.: `im.GrayScale` above), the linter logs an issue within Ren'Py's code, which makes the original issue way harder to identify for non-technical users:

```
Obsolete Image Manipulators:

/home/ayowel/data/development/sdk/renpy/renpy-8.5.2-sdk/renpy/display/im.py:
    * line  2137 im.MatrixColor

game/images.rpy:
    * line    2 im.MatrixColor
```

This improves linting by ensuring that the logged frame is out of im.py, which will usually more closely match user's expectations as a step toward permanently deleting this deprecated usage:

```
Obsolete Image Manipulators:

game/images.rpy:
    * line    1 im.MatrixColor
    * line    2 im.MatrixColor
```

### DIFF
--- a/renpy/display/im.py
+++ b/renpy/display/im.py
@@ -692,8 +692,9 @@ class ImageBase(renpy.display.displayable.Displayable):
         self.identity = (type(self).__name__,) + args + flat_properties
 
         if self.obsolete and renpy.game.context().init_phase:
-            frame = sys._getframe(2)
-            while frame.f_code.co_filename == __file__ and frame.f_back:
+            frame = sys._getframe()
+            current_filename = frame.f_code.co_filename
+            while frame.f_code.co_filename == current_filename and frame.f_back:
                 frame = frame.f_back
             filename = frame.f_code.co_filename
             line = frame.f_lineno


Andykl:
Instead of iterating through integers you could use `frame.f_back` with chech for none. 
The overall idea is good.

Ayowel:
> Instead of iterating through integers you could use `frame.f_back` with chech for none. The overall idea is good.

Yep, just noticed this. The changeset looks much better with it

Andykl:
I am not sure `__file__` is correct choice here, e.g. if pyc file is compiled on different machine co_filename would be local path to that machine so not equal to `__file__` if I am not mistaken. I don't remember who sets file on module object, but gut feeling it's import machinery. 
Probably it's better to get the filename from any function of the module code object? 

Worst case that change would be a noop.

Ayowel:
Sounds good to me, just made changes accordingly. I use the current file's frame to get a reference name and then just check against it while going back

Ayowel:
@Andykl let me know if you still see an issue with the changeset

Andykl:
LGTM.

Small nitpick is that you access `f_back` without check for `None`, but to trigger that one would need to run `im` module itself, which is pretty nonsensical.

Ayowel:
> Small nitpick is that you access f_back without check for None, but to trigger that one would need to run im module itself, which is pretty nonsensical.

Yeah, I dropped it as the module does not even have a `__main__`. Just put it back in as a security measure.

@renpytom This should be good to merge

PR #6877, opened 2026-01-18
Fix `@pure` decorator logic to support class methods via `__qualname__`

### LINKED ISSUE #6871: `renpy.pure` incorrectly registers class methods as global functions (ignores `__qualname__`)
## Issue:
The current implementation of `renpy.pure` relies on `fn.__name__`  to generate the registration key for the `renpy.pyanalysis.constants` and `renpy.pyanalysis.pure_functions` sets.. This works for global functions but fails for methods defined inside classes, as `__name__` does not include the class name (e.g., it returns `"my_func"` instead of `"MyClass.my_func"`).

This causes two specific failures:
1. Pollution: It incorrectly registers a global constant named `"my_func"`, even though no such global function exists.
2. Analysis Mismatch: The screen language analyzer (`renpy/pyanalysis.py`) correctly identifies calls as `"MyClass.my_func"`, but looks that up in the registry and fails to find it, causing the method to be treated as impure.

Additionally, `renpy.pure` does not handle standard decorators like `staticmethod` or `classmethod` if they wrap the function before it receives them, leading to potential introspection issues.

## Reproduction:
```renpy
init python:

    def is_pure(name):
        return name in renpy.pyanalysis.pure_functions

    # 1. Register a static method inside a class
    class MyClass:
        @renpy.pure
        @staticmethod
        def my_func():
            pass

    # 2. Check registration / analysis
    # BUG:
    # "MyClass.my_func" should be in constants.
    # Instead, "my_func" is in constants (pollution), and "MyClass.my_func" is missing.

    # Expected: True. Actual: False
    print("Is 'MyClass.my_func' pure?", is_pure("MyClass.my_func"))

    # Expected: False. Actual: True
    print("Is 'my_func' pure?", is_pure("my_func"))
```

## Analysis:
The screen analyzer (`renpy/pyanalysis.py`) reconstructs dotted paths for attribute access (e.g. `MyClass.my_func`).
To match this behavior, `renpy.pure` must register the function using its `__qualname__` to capture the class nesting.

## Suggested Fix:
In `renpy/pyanalysis.py`, modify the `pure` function to prefer `__qualname__` and unwrap common descriptors.

```python
def pure(fn):

    name = fn

    # Unwrap classmethod/staticmethod to get to the real function
    if isinstance(fn, (staticmethod, classmethod)):
        name = fn.__func__

    if not isinstance(name, str):
        # Use qualname if available to capture class nesting
        qual = getattr(name, "__qualname__", name.__name__)
        module = name.__module__
        name = module + "." + qual

    if name.startswith("store."):
        name = name[6:]

    if name not in not_constants:
        pure_functions.add(name)
        constants.add(name)

    return fn
```

juperMonkey:
Previously, using `fn.__name__` caused class methods to be registered incorrectly as top-level functions (e.g., `'module.method'` instead of `'module.Class.method'`).

Swapping to `__qualname__` ensures the fully qualified name is used, allowing the analyzer to correctly track pure methods nested inside classes.

**Technical Note**: This change relies on Python 3.10+ behavior where `staticmethod` and `classmethod` wrappers directly expose `__name__`, `__qualname__`, and `__module__`, removing the need to manually unwrap the underlying function.

Closes #6871

### DIFF
--- a/renpy/pyanalysis.py
+++ b/renpy/pyanalysis.py
@@ -240,7 +240,7 @@ def pure(fn):
     name = fn
 
     if not isinstance(name, str):
-        name = fn.__name__
+        name = fn.__qualname__
 
         module = fn.__module__
         name = module + "." + name


Andykl:
LGTM. I never saw static/classmethods/nested classes used with `renpy.pure` but I can imagine valid use case for it.

PR #6892, opened 2026-02-04
Correct capitalization and punctuation in config.rst

DipeshAggarwal:


DipeshAggarwal:
Converted this to draft, because apparently, I forgot to sync my branch with the repo and now, there are a ton of merge conflicts. I'm going to fix all of it and then mark it ready.

DipeshAggarwal:
@Andykl All merge conflicts are resolved.

One more change has been added. I used Rewrap Revived to correctly wrap lines after 80 columns to respect the column length limit, because line length were all over the place and it seems like PyTom intended to have 80 column as a hard limit, looking at the first few lines.

Andykl:
Now your PR mixes 2 things at once. Would you mind splitting it into 2 different PR, one with original changes that I can read one by one, and second with rewrap so I would check it didn't mess anything by comparing output html. I don't want to read throught 800 changes.

DipeshAggarwal:
Done.. I'm going through the entire documentation, should I just do away with wrapping lines?

Andykl:
I am not sure about it. I think that is more a @renpytom decision. IMO, 80 chars is too restrictive in modern world. I would go with 80 being a soft limit, and 120 a hard one.

Andykl:
Thank you for the fixes.

Gouvernathor:
Personally, I prefer not wrapping at all and enable auto-wrap in the IDE settings (Alt+Z in Code, and defaults are configurable by language)

DipeshAggarwal:
I think the same.

But right now, the wrapping is all over the place and one consistent style needs to be chosen in my opinion. Looking across documentation, 70% of it is set to around 80 column width.

renpytom:
I wrap to 120 columns for new writing. Do not re-wrap, as it makes the diffs bad.

Issue #3864, opened 2022-08-12
Thai font render issue


PhakphumV:
Hello,

I found that Thai fonts are not rendered correctly when there is a combination of tone mark and upper vowel.
Here are some examples between what ren'py render and notepad with the same font (Cordia New)

![2022-08-12 20_18_27-_Untitled - Notepad](https://user-images.githubusercontent.com/35300139/184362728-99dbf26e-5f1f-4172-a79a-fe65ae93b9a6.png)
![2022-08-12 20_19_00-_Untitled - Notepad](https://user-images.githubusercontent.com/35300139/184362737-cec9fcbb-d7cf-40f8-a755-d52d536d3363.png)

There was a script to make adjustment from this link https://gitlab.com/sahabandha/renpy-thai-font-adjuster
It's compatible only with renpy-7.4.10-sdk and some limitation. I tried code modification in renpy/text/text.py with the script but it doesn't work properly as I'm still new to renpy. So any guidance would be really appreciated.

In the above Gitlab link, there is also a reference to a very detailed documentation about encoding for Thai which could help on the font rendering script. I believe it could be useful for develop a fix for this issue.

renpytom:
Seems reasonable. I think I should be able to get this into the next major release.

PhakphumV:
Thanks, please feel free to reach out to me if you need some test samples for different combination of Thai characters.

Here is an example from W3.org that can be used as reference.
`แผ่นดินฮั่นเสื่อมโทรมแสนสังเวช  พระปกเกศกองบู๊กู้ขึ้นใหม่
  สิบสองกษัตริย์ก่อนหน้าแลถัดไป       สององค์ไซร้โง่เขลาเบาปัญญา
    ทรงนับถือขันทีเป็นที่พึ่ง           บ้านเมืองจึงวิปริตเป็นนักหนา
  โฮจิ๋นเรียกทัพทั่วหัวเมืองมา         หมายจะฆ่ามดชั่วตัวสำคัญ
    เหมือนขับไสไล่เสือจากเคหา      รับหมาป่าเข้ามาเลยอาสัญ
  ฝ่ายอ้องอุ้นยุแยกให้แตกกัน          ใช้สาวนั้นเป็นชนวนชื่นชวนใจ
    พลันลิฉุยกุยกีกลับก่อเหตุ          ช่างอาเพศจริงหนาฟ้าร้องไห้
[...]

Issue #3525, opened 2022-05-18
Hindi/Devanagari fonts not working


terimautchu:
The Hindi translation for a Ren'py game is getting all broken down. We have tried Noto Sans Regular/Mukta/Fira Go but all of them fail to properly render many of the words. Like हिन्दी is shown हनि्दी

Any suggestions on how to fix this

rajsingh010:
That's actually not a bug but the font which you are using itself has problem so you can just use other font's which support Hindi (Devnagari texts) like [https://www.typingguru.net/hindi-fonts/devanagari](https://www.typingguru.net/hindi-fonts/devanagari) download and use something like this one.

If the problem is still there then please share screenshot or error message.

renpytom:
I think it's possible that this might be a problem with (a lack of) support for indic text shaping, which (and I'm not an expert on this) would be required to render things correctly.

Any help in this area would be appreciated.

terimautchu:
> That's actually not a bug but the font which you are using itself has problem so you can just use other font's which support Hindi (Devnagari texts) like https://www.typingguru.net/hindi-fonts/devanagari download and use something like this one.
> 
> If the problem is still there then please share screenshot or error message.

We did try fonts like Lohit Devanagari/Kruti Dev 010 Regular but they were dropped because in order to enable a font switch, the font pack must contain English/Latin characters as well.  This is required as Ren'Py will try to switch back to english characters when reaching the end of the translated sections. If the font pack doesn't have any Latin characters then the game breaks and becomes unplayable. We therefore used Noto Sans/Mukta/Fira Go which contain BOTH Devnagari as well as Latin characters thus enabling a smooth switch.

This was an improvement as the earlier fonts containing only Devanagari were merely showing some boxes whereas the new fonts now show Devanagari. It's just that the Devanagari that is shown is broken which I believe is due to inability of the Ren'Py engine to properly render Indic characters. 

Is there some fix you guys could suggest that makes Ren'Py render these characters properly?

terimautchu:
> I think it's possible that this might be a problem with (a lack of) support for indic text shaping, which (and I'm not an expert on this) would be required to render things correctly.
> 
> Any help in this area would be appreciated.

^^^ This. We would REALLY appreciate it if someone who knows about this stuff can suggest how to make necessary tweaks or do whatever is required to enable smooth indic text shaping in Ren'Py

renpytom:
[...]

PR #6911, opened 2026-02-22
perf: Thai C90 further optimization for FreeType shaper

hiohlan:
Thank you for previously merging my Thai C90 implementation from issue #3864.

This pull request includes performance improvements, a definition fix, and removal of an invalid documentation link.

I ran a benchmark using `bench.py` (from my source repository), and the optimized version runs almost ×2 faster than the previous implementation. This should help improve in-game Thai text rendering performance.

(Prefer keeping it selectable via thaic90 instead of making it the default.)


hiohlan:
Since version 8.4, improvements in HarfBuzz have eliminated the need for ThaiC90 for Thai text rendering.

However, I believe it is still worth keeping this option for developers who rely on legacy Thai fonts that lack proper OpenType rules (e.g., GSUB/GPOS support).

Should we also update the documentation to clarify that Thai text should use `Unicode` by default, unless `thaic90` is explicitly required for legacy fonts?

renpytom:
Please update the documentation as you see best, and when it's updated I'll merge this.

hiohlan:
~~From my reading of the code, it seems that `thaic90` may no longer be necessary.
HarfBuzz already includes built-in support for Thai shaping (including C90-style handling), which makes the separate `thaic90` mode appear redundant.~~

~~If confirmed, we could consider deprecating and eventually removing it as an incompatible change.~~

Correction: I misread the code earlier. thaic90 must be kept, as it is still required when the FreeType shaper is used. HarfBuzz has built-in Thai shaping support, but FreeType does not.

mal:
Is it worth considering having use of `thaic90` config value imply/force the use of harfbuzz for shaping? That might fit well the dual goals of not breaking existing games while also allowing for the code reduction/simplification.

renpytom:
At this point, I'm wondering if there's a reason to have the freetypes shaper at all, now that harfbuzz has proven itself.

hiohlan:
> Is it worth considering having use of `thaic90` config value imply/force the use of harfbuzz for shaping? That might fit well the dual goals of not breaking existing games while also allowing for the code reduction/simplification.

Wouldn’t that make it redundant? `thaic90` exists only because the FreeType shaper does not support Thai shaping.
If selecting `thaic90` simply forces HarfBuzz, then it would be clearer to just use `unicode` instead and remove `thaic90` as a breaking change during an engine update, since those already introduce incompatible changes anyway.
Game developers should be able to adjust their language configuration accordingly.

> At this point, I'm wondering if there's a reason to have the freetypes shaper at all, now that harfbuzz has proven itself.

HarfBuzz still has issues when mixing multiple complex scripts in the same dialog.
For example, when Thai and Tai Tham are used together in a single dialogue line, the shaping can break (eg. wrong position for pre-vowel like #3525).
From my testing, I’m not sure whether this is an issue in HarfBuzz itself or in Ren’Py’s HarfBuzz integration.
(But FreeType shaper does not handle complex shaping at all, so it would not solve this problem either)

For now, I think we should keep the ThaiC90 code for game developers who still choose to use the FreeType shaper.
It can be removed once FreeType support is dropped.

renpytom:
Sounds good to me. Merged.

PR #6917, opened 2026-02-23
Optimize loader.py

Wasa16261627:
I tried to optimize. The first one runs 10 times faster, the second by 30%, as far as I can tell.


renpytom:
Do you have a a sense for how much absolute time this saves? This is on a slow code path that often hits disk, after all.

ghost:
No, I can't rate it. I tried, but it didn't work out quickly, so I put it off.

PR #6946, opened 2026-03-09
Extract colorspace from the video instead of assuming default

Galo223344:
Currently renpy assumes that all videos use the mpeg (limited) color range and the (very outdated) bt601 color-space for what appears to be no good reason, this made it so that videos looked different outside of renpy and made seamless transitions from starting frame to video file impossible unless you encoded your video to these standards.
This PR makes it so that both color range and color space are extracted from the metadata of the file, with the color-space falling back to the default if not specified.

### DIFF
--- a/src/ffmedia.c
+++ b/src/ffmedia.c
@@ -881,8 +881,15 @@ static SurfaceQueueEntry *decode_video_frame(MediaState *ms) {
 			return NULL;
 		}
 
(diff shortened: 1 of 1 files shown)


renpytom:
Thanks for the contribution. I'd say that the only reason we did this is I didn't know how colorspaces worked  I appreciate the fix.

PR #6949, opened 2026-03-09
Remove branch names from official builds

mal:
Official builds already have a well managed versioning system, adding `.master` and `.fix` to the end of them doesn't seem necessary.

### DIFF
--- a/renpy/versions.py
+++ b/renpy/versions.py
@@ -94,7 +94,7 @@ def _make_version_string(
     if dirty:
         suffixes.append("dirty")
 
-    if branch != "main" or dirty:
+    if not official and branch != "main" or dirty:
         suffixes.append(branch)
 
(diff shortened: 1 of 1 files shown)


Andykl:
I agree that such change would make version look better but probably better fix is to fix `main` to `master`. I don't know what I halucinated when used `main` here. Your current fix mean "any build made on official Ren'Py build server does not have branch name suffix", which is probably fine too.

mal:
Took this approach as the "blessed" patch versions (i.e. `8.5.3`) shouldn't have suffixes, and so aligning the nightlies with that `+nightly` felt less confusing than `+nightly.fix` which invariably leads to "Yes, I see `8.5.3` is out, but I wanted the fixed version so I kept using that" after the blessed version has been published.

PR #6951, opened 2026-03-11
allow .pyc files but prefer .py files in the renpy.importer

erazor83:
This allows the use of .pyc files in renpy to avoid source code distribution. It prefers the .py file if that exists too.

I've checked this with Android and Linux.

I'm doing something like this to bytecompile my library:
````bundle_common.py
import os
import shutil
import compileall
def create_pyc_files(start_dir,temp_base_dir):
	print("Running pre-build: Compiling modules...")
	#os.system("python -m compileall %s"%start_dir)
	compileall.compile_dir(start_dir, force=True, legacy=True, quiet=0)
	for root, _, _ in os.walk(start_dir):
		cache = os.path.join(root, '__pycache__')
		if os.path.exists(cache):
			for f in os.listdir(cache):
				temp_dir=os.path.join(temp_base_dir,root)
				temp_dir=temp_dir.replace('game/', '')
				os.makedirs(temp_dir, exist_ok=True)
				shutil.copy(
					os.path.join(cache, f),
					os.path.join(temp_dir, f.split('.')[0] + '.pyc')
				)
		
c_dir='game/common'
temp_dir='temp'
if os.path.isdir(c_dir):
	create_pyc_files(c_dir,temp_dir)
````

And:
````dist_common_py.sh
#!/bin/sh

echo "Cleanup... (removing __pychache__ and *.pyc)"
rm -r game/common/__pycache__ game/common/py/__pycache__
rm game/common/*.pyc game/common/py/*.pyc
rm -rf temp/*

~/RenPy/renpy-8.1.3-sdk/lib/py3-linux-x86_64/python -m compileall game/common
~/RenPy/renpy-8.1.3-sdk/lib/py3-linux-x86_64/python bundle_common.py

echo "Copy to game/python-packages/common"
[[ -d  game/python-packages/common ]] || mkdir -p game/python-packages/common

echo "Cleaning game/python-packages/common"
rm -rf  game/python-packages/common/*

echo "Copy to game/python-packages/common"
cp -vr temp/common/* game/python-packages/common/
````

It'd be great if these changes find their way into renpy.

Thanks for the engine and have a great day.

### DIFF
--- a/renpy/importer.py
+++ b/renpy/importer.py
@@ -20,6 +20,7 @@
 # WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 from typing import Iterable, Literal, NamedTuple, overload
+from importlib._bootstrap_external import _classify_pyc, _compile_bytecode
 
 import renpy
 import sys
@@ -172,43 +173,60 @@ class RenpyImporter(importlib.abc.MetaPathFinder, importlib.abc.InspectLoader):
 
     def _visit_dir(self, *path: str, files: Iterable[str]):
         dir_to_fn: dict[str, list[str]] = {}
-
         seen_init = False
         prefix = ""
         if path:
             prefix += "/".join(path) + "/"
+        mod_name_base = ".".join(path) if path else ""
 
         for fn in files:
             if "/" in fn:
-                top_directory, _, fn = fn.partition("/")
+                top_directory, _, sub_fn = fn.partition("/")
                 if top_directory not in dir_to_fn:
                     dir_to_fn[top_directory] = []
-
-                dir_to_fn[top_directory].append(fn)
+                dir_to_fn[top_directory].append(sub_fn)
                 continue
 
-            mod_name = ".".join(path)
-            if path and fn == "__init__.py":
+            if fn.endswith(".py"):
+                stem = fn[:-3]
+            elif fn.endswith(".pyc"):
+                stem = fn[:-4]
+
+            if stem is None:
+                continue
+
+            if stem == "__init__":
                 seen_init = True
+                mod_name = mod_name_base
                 is_package = True
             else:
-                if mod_name:
-                    mod_name += "." + fn[:-3]
-                else:
-                    mod_name = fn[:-3]
-
                 is_package = False
 
-            mod_info = RenpyImporter._ModuleInfo(prefix + fn, is_package, False)
+                mod_name = mod_name_base + ("." + stem if mod_name_base else stem)
 
+            py_file  = stem + ".py"
+            pyc_file = stem + ".pyc"
+
+            if py_file in files:
+                chosen_filename = prefix+py_file
+            elif pyc_file in files:
+                chosen_filename = prefix+pyc_file
+                # Optional: log when falling back
+                # print(f"Falling back to {chosen_filename} (no {py_full})")
+            else:
+                # Should not happen because fn came from files, but safety
+                continue
+            
+            #print('mod_name', mod_name, is_package, '->', chosen_filename)
+            mod_info = RenpyImporter._ModuleInfo(chosen_filename, is_package, False)
             yield mod_name, mod_info
 
         if path and not seen_init:
-            yield ".".join(path), RenpyImporter._ModuleInfo(prefix, True, True)
+            yield mod_name_base, RenpyImporter._ModuleInfo(prefix, True, True)
 
-        for add_dir, files in dir_to_fn.items():
-            yield from self._visit_dir(*path, add_dir, files=files)
-
+        for add_dir, subfiles in dir_to_fn.items():
+            yield from self._visit_dir(*path, add_dir, files=subfiles)
+            
     def _cache_entries(self) -> dict[str, _ModuleInfo]:
         if self._finder_cache is not None:
             return self._finder_cache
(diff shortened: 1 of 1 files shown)


renpytom:
I don't like this, from a game preservation perspective. This change would make it easier to tie a game to a very specific version of Python - which means it couldn't be run in future versions of Python. That's not something I want to encourage. 

It also uses a number of Python-internal functions that I don't believe are meant as part of a public API.

erazor83:
Hi Tom,

thanks for the feedback and for explaining your concerns — I completely understand wanting to protect games from becoming unplayable in the future.

The main motivation for this change is actually alignment with how Python itself works. In a standard Python environment, `.pyc` files are first-class citizens: the importer prefers source when both exist, falls back to bytecode when source is missing, and validates the bytecode header (magic number, timestamp/hash) to reject anything mismatched or corrupted. Ren'Py being the exception (silently ignoring `.pyc` entirely) felt very inconsistent to me when I first encountered it — especially since the engine already ships a fixed Python interpreter with every release (currently Python 3.12 in 8.5.x). So games are already tightly coupled to one specific Python version anyway; this PR doesn't introduce new version-lock risks, it just lets the importer behave like normal Python.

A few points addressing the preservation and API concerns directly:

- The code **prefers `.py` files** when both `.py` and `.pyc` are present (exactly like CPython). Only when the source is deliberately omitted (typical for distributed builds) does it use the `.pyc`. This avoids any risk from stale or invalid bytecode during development.

- It uses `_classify_pyc` which performs exactly the same validation the standard library does — so mismatched `.pyc` files (wrong Python version, corrupted, etc.) raise a clear `ImportError` instead of causing silent failures or mysterious crashes later. From a preservation perspective this is arguably safer than the current behavior.

- The two functions (`_classify_pyc` + `_compile_bytecode`) are the same ones used internally by the documented `importlib.machinery.SourcelessFileLoader` (public API since Python 3.3+). They've been stable since at least 3.7/3.8, and the importer already depends on several Ren'Py-private APIs anyway (`game_files`, `transpath`, `TreeEntry`, etc.), so this is consistent in style.

- The original importer had an explicit `TODO: add bytecode handling?` comment. This simply completes that TODO in the most straightforward, Python-idiomatic way.

As a small side benefit, it gives creators a zero-cost, optional layer of lightweight protection (harder to casually read/decompile than plain `.py` or `.rpy`), which some people already want.

Would you be open to reconsidering the PR with these points in mind? I'm happy to discuss any adjustments that would make it more acceptable for the project.

Thanks again for maintaining Ren'Py!

mal:
This is the crux of it IMO:

> As a small side benefit, it gives creators a zero-cost, optional layer of lightweight protection (harder to casually read/decompile than plain .py or .rpy), which some people already want.

It isn't zero-cost. The result of a change like this would be that creators would do this (believing, as you seem to) that it offers some kind of "protection" for their code, despite it still being trivial to decompile. Meanwhile, the shipped game becomes hard-locked to the version of Python for which it was compiled. That materially harms Ren'Py's game preservation goals because it means those games won't ever run on newer versions of Ren'Py (a lot of care has been taken over the last 20 or so years to ensure the vast majority of even very old games remain playable on newer SDKs which can run on modern systems) and so are at a much higher likelihood of becoming unable to be run and thus obsolete as time passes.

erazor83:
I completely get the worry that some creators might ship `.pyc`-only builds thinking it offers real protection (when it doesn't), and that this could quietly hurt preservation down the line. That part makes sense.

At the same time — and this is just my honest feeling — I find it a bit hard to follow the logic that we should prevent everyone from having `.pyc` support *at all* because a subset of users might not fully understand the consequences. Most Ren'Py creators are technically capable people who already make choices like shipping only `.rpyc` files, stripping debug info, or using other forms of obfuscation. They already live with the same trade-offs (fragile bytecode, version coupling). Treating `.pyc` differently feels a little like assuming users aren't able to weigh those risks themselves.

A clear warning in the documentation along the lines of  
“Do not distribute `.pyc`-only modules if you want your game to remain playable on future Ren'Py versions — always keep the `.py` sources for recompilation”  
would, in my opinion, be more than enough to inform anyone who cares about long-term compatibility. The vast majority of people would either keep their sources anyway (especially during development, where `.py` is preferred) or consciously accept the limitation if they choose bytecode-only.

I'm starting to run a bit low on fresh arguments here, and I do accept that preservation is the top priority for you. I just personally feel that `.pyc` support aligns better with normal Python expectations and doesn't meaningfully increase the existing risks when documented properly.

If after all this the answer is still "no, we prefer to keep `.pyc` unsupported entirely", that's totally fine — I'll happily keep using the change in my own projects/forks. Just wanted to make sure the other side was heard.

Thanks again for taking the time to discuss this — really value your perspective.

Andykl:
I can give you a different POW. 
I don't value games preservation that much as other Ren'Py devs, but I have different objection for this PR. 
In my opinion pure python packages are intended for PyPI packages - those essentially requiring them to be sourcefull, because python bytecode compatibility is basically nonexistent for major updates of python. 
The only reason to omit source files I can think of is to make build size smaller, but that is way weaker justification, given that couple of PNG images could be bigger than entire site-packages. 

On the other hand, _ren.py files compiled into rpyc can be sourceless because well Ren'Py will store source code into rpyc precisely to allow re compilation if cached bytecode can't load. 

I think _ren.py files is the way to add pyre python code for games and abuse of site-packages is the same missunderstanding as with "let's hack around RPA format to prevent decompilation of our assets", which just doesn't work, never(?). That happens because capable Ren'Py dev is not necessarily capable python dev. 

That said, I think it may make sense to add support for .pyc files as a fast path of getting parsed and compiled bytecode, that's why I left the comment there, but it should use public python api. I had no time to figure it out, and speed gain is so marginal compared to loading of Ren'Py script so I didn't bother.

renpytom:
Note that the .rpyc file format doesn't have this problem - it contains enough information to work with future versions of Python and Ren'Py. 

For the reasons mentioned above, I'm not going to merge this.

PR #6955, opened 2026-03-13
Introducing RenpyPath

### LINKED ISSUE #6898: Ren'PyPath: Integrate PurePath-like behavior into renpy
Since python 3.12 (current renpy version), [PosixPath](https://docs.python.org/3.12/library/pathlib.html#pathlib.PosixPath) from pathlib [is officially subclassible](https://docs.python.org/3.12/whatsnew/3.12.html#pathlib), I see Ren'Py has got a RenpyPath in the meantime: (https://github.com/renpy/renpy/blame/master/renpy/importer.py#L44).

I'd like to implement something more similar to RenpyPath I implemented in the past that can be found here: https://gitlab.com/brunoais/renpy-RenpyPath/-/blob/master/renpy_path/renpy_path.py (also made available at https://brunoais.itch.io/renpy-renpypath)

I believe that this library can be of major usefulness to libraries and to do advanced file handling.

I would like to have that but more extensive like how [renpy_path.py](https://gitlab.com/brunoais/renpy-RenpyPath/-/blob/master/renpy_path/renpy_path.py) is.

Would that be accepted to Ren'Py?

brunoais:
This is my first part in my 2-part effort to bring `RenpyPath` to Ren'Py.

This is about having a `PathLike` class, which is as compatible as viable with `Path`'s API as I reasonably could, while adding extra features which I see as advantageous for its users.

~~This draft PR already has all the core concepts I want to introduce and the core changes I have planned.~~

This PR has all the content planned for the first part I planned.

I've been testing and using this in Ren'Py VNs I have access to. Some with a few hundred files such as Tutorial and The Question and a few between 10K-140K file range (voice taking the majority of the count), some with `.py` files (Python libraries installed and present), some without. Assuming I didn't make any significant copy&paste mistakes between what I used and the master branch, it should work without any known issues at the moment.

This code (except metadata) was fully written by me, however, AI was used. This code was "peer-reviewed" (within its limitations) by AI, most type annotations, unit tests and performance tests (not included) were written by AI, all with my strict supervision. Initial function/method documentation (the one in the first commit here) was some written by me, some by AI.

- [x] To finish this first part, I still need to write proper documentation which can be picked up by Sphinx and
- [x]  expose RenpyPath to `renpy.exports` (i.e. making it available either as `RenpyPath`(preferred) or `renpy.RenpyPath`.
- [x] Improve this initial documentation. Make it less repetitive and easier to read.
- [ ] Update code for python 3.14.
- [ ] Move `RenpyPath` to a separate module.

For the 2nd part, which will be a separate PR and discussion, I intend to replace the majority (but not all) internal calls to `renpy.list_files()` to use `RenpyPath` instead. As part of testing if  `RenpyPath` appears to be working right, I already did that in my testing instance of Ren'Py and couldn't find issues.
Regardless, just because I don't find bugs, doesn't mean they aren't there. I may just be testing it wrong.

Resolves #6898

### DIFF
--- a/renpy/common/00build.rpy
+++ b/renpy/common/00build.rpy
@@ -27,6 +27,7 @@ init -1500 python in build:
     _constant = True
 
     from store import config, store
+    from renpypath import RenpyPath
 
     import sys, os
 
@@ -235,9 +236,21 @@ init -1500 python in build:
         Classifies files that match `pattern` into `file_list`, which can
         also be an archive name.
 
+        If the `pattern` is a :renpy_path:`RenpyPath`, either
+        the file it represents is added to the list or
+        the directory it represents and all its contents are added to the
+        file list
+
         If the name given as `file_list` doesn't exist as an archive or file
         list name, it is created and added to the set of valid file lists.
         """
+
+        if isinstance(pattern, RenpyPath):
+            path = pattern
+            pattern = str(path).replace('*', '[*]')
+            if path.is_dir():
+                pattern += '/**'
+
 
         base_patterns.append((pattern, make_file_lists(file_list)))
 
--- a/renpy/defaultstore.py
+++ b/renpy/defaultstore.py
@@ -368,6 +368,9 @@ def At(d, *args):
 Color = renpy.color.Color
 color = renpy.color.Color
 
+# The RenpyPath class.
+RenpyPath = renpy.loader.RenpyPath
+
 # The default menu functions.
 menu = renpy.exports.display_menu
 predict_menu = renpy.exports.predict_menu
--- a/renpy/defaultstore.py
+++ b/renpy/defaultstore.py
@@ -369,7 +369,7 @@ Color = renpy.color.Color
 color = renpy.color.Color
 
 # The RenpyPath class.
-RenpyPath = renpy.loader.RenpyPath
+RenpyPath = renpy.renpypath.RenpyPath
 
 # The default menu functions.
 menu = renpy.exports.display_menu
--- a/renpy/exports/__init__.py
+++ b/renpy/exports/__init__.py
@@ -509,6 +509,12 @@ from renpy.exports.platformexports import (
     open_url,
 )
 
+from renpy.exports.renpypathexports import (
+    RenpyPath,
+    rpy_path,
+    rpyc_path,
+)
+
 from renpy.exports.predictexports import (
     cache_pin,
     cache_unpin,
@@ -653,6 +659,7 @@ if typing.TYPE_CHECKING:
     from . import persistentexports as persistentexports
     from . import platformexports as platformexports
     from . import predictexports as predictexports
+    from . import renpypathexports as renpypathexports
     from . import restartexports as restartexports
     from . import rollbackexports as rollbackexports
     from . import sayexports as sayexports
--- a/renpy/exports/loaderexports.py
+++ b/renpy/exports/loaderexports.py
@@ -140,6 +140,12 @@ def list_files(common=False):
     `common`
         If true, files in the common directory are included in the
         listing.
+
+    For targeted path operations (listing a directory, listing many
+    directories, globbing a subtree) :class:`RenpyPath` is usually more
+    performant, convenient, and easier to use. ``list_files`` is better
+    when you need all files and not just a subset, or if your only filter
+    is a file name or extension but you do not know where it is located.
     """
 
     rv = []
--- a/renpy/exports/renpypathexports.py
+++ b/renpy/exports/renpypathexports.py
@@ -1,0 +1,24 @@
+# Copyright 2004-2026 Tom Rothamel <pytom@bishoujo.us>
+#
+# Permission is hereby granted, free of charge, to any person
+# obtaining a copy of this software and associated documentation files
+# (the "Software"), to deal in the Software without restriction,
+# including without limitation the rights to use, copy, modify, merge,
+# publish, distribute, sublicense, and/or sell copies of the Software,
+# and to permit persons to whom the Software is furnished to do so,
+# subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+# LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+# OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+# WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+from __future__ import division, absolute_import, with_statement, print_function, unicode_literals  # type: ignore
+
+from renpy.renpypath import RenpyPath, rpy_path, rpyc_path  # *
--- a/renpy/importer.py
+++ b/renpy/importer.py
@@ -77,18 +77,18 @@ class RenpyImporter(importlib.abc.MetaPathFinder, importlib.abc.InspectLoader):
     """
 
     def __init__(self):
-        self.prefixes: list[str] = []
+        self.prefixes: list[RenpyPath] = []
         "List of prefixes where modules can be found."
 
-        self._finder_cache: dict[str, RenpyImporter._ModuleInfo] | None = None
+        self._finder_cache: dict[str, RenpyImporter._ModuleInfo] = {}
         """
         A dict from module name to module info of that module for all paths that
         can be loaded with this importer.
         """
 
-    def add_prefix(self, prefix: str):
-        if prefix and not prefix.endswith("/"):
-            prefix = prefix + "/"
+    def add_prefix(self, prefix: str | RenpyPath):
+        if isinstance(prefix, str):
+            prefix = RenpyPath(prefix)
 
         if prefix in self.prefixes:
             return
@@ -101,70 +101,42 @@ class RenpyImporter(importlib.abc.MetaPathFinder, importlib.abc.InspectLoader):
         is_package: bool
         is_namespace: bool
 
-    def _visit_dir(self, *path: str, files: Iterable[str]):
-        dir_to_fn: dict[str, list[str]] = {}
+    def _get_module_info(self, fullname: str) -> _ModuleInfo | None:
 
-        seen_init = False
-        prefix = ""
-        if path:
-            prefix += "/".join(path) + "/"
-
-        for fn in files:
-            if "/" in fn:
-                top_directory, _, fn = fn.partition("/")
-                if top_directory not in dir_to_fn:
-                    dir_to_fn[top_directory] = []
-
-                dir_to_fn[top_directory].append(fn)
-                continue
-
-            mod_name = ".".join(path)
-            if path and fn == "__init__.py":
-                seen_init = True
-                is_package = True
-            else:
-                if mod_name:
-                    mod_name += "." + fn[:-3]
-                else:
-                    mod_name = fn[:-3]
-
-                is_package = False
-
-            mod_info = RenpyImporter._ModuleInfo(prefix + fn, is_package, False)
-
-            yield mod_name, mod_info
-
-        if path and not seen_init:
-            yield ".".join(path), RenpyImporter._ModuleInfo(prefix, True, True)
-
-        for add_dir, files in dir_to_fn.items():
-            yield from self._visit_dir(*path, add_dir, files=files)
-
-    def _cache_entries(self) -> dict[str, _ModuleInfo]:
-        if self._finder_cache is not None:
-            return self._finder_cache
-
-        if not renpy.loader.game_files:
+        if not self._finder_cache and not renpy.loader.game_files:
             renpy.loader.scandirfiles()
 
-        self._finder_cache = dict(
-            self._visit_dir(files=(fn for _, fn in renpy.loader.game_files if fn.endswith(".py") if not fn.endswith("_ren.py")))
-        )
-        return self._finder_cache
+        try:
+            return self._finder_cache[fullname]
+        except KeyError:
+            pass
 
-    def _get_module_info(self, fullname: str) -> _ModuleInfo | None:
-        cache_entries = self._cache_entries()
+        fullname_base = fullname.replace(".", "/")
 
+        info = None
         for prefix in self.prefixes:
-            prefix = prefix.replace("/", ".")
-            if rv := cache_entries.get(prefix + fullname):
-                return rv
+            dir: pathlib.Path = prefix / fullname_base
 
-        return None
+            if dir.is_dir():
+                init = dir / '__init__.py'
+                if init.exists():
+                    info = RenpyImporter._ModuleInfo(str(init), True, False)
+                else:
+                    info = RenpyImporter._ModuleInfo(str(dir), True, True)
+                break
+            else:
+                file = dir.with_name(f'{dir.name}.py')
+                if file.exists() and file.is_file():
+                    if not file.name.endswith('_ren.py'):
+                        info = RenpyImporter._ModuleInfo(str(file), False, False)
+                    break
+
+        self._finder_cache[fullname] = info
+        return info
 
     # MetaPathFinder interface
     def invalidate_caches(self):
-        self._finder_cache = None
+        self._finder_cache.clear()
 
     def find_spec(self, fullname, path, target=None):
         if module_info := self._get_module_info(fullname):
@@ -268,7 +240,7 @@ class RenpyImporter(importlib.abc.MetaPathFinder, importlib.abc.InspectLoader):
 meta_backup = []
 
 
-def add_python_directory(path: str):
+def add_python_directory(path: str | RenpyPath):
     """
     :doc: other
 
--- a/renpy/importer.py
+++ b/renpy/importer.py
@@ -31,7 +31,7 @@ import importlib.abc
 import importlib.util
 import importlib.resources.abc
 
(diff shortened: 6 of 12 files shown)


brunoais:
@renpytom It's ready for review.
Note I was unable to place `RenpyPathDirEntry` in the documentation. I'd like to document it but, at the same time, not export it. I couldn't find how.

Andykl:
I do like the idea of RenpyPath. I have a few questions about the implementation:
1. Why does it inherit PurePosixPath instead of PurePath. I don't undesrtand what are implications of that, but it just a feels wrong.
2. Ren'Py already normalizes paths to lower-case NFC Unicode. Why do we need that flag for methods?
3. Do we really need `rpy_path`, `rpyc_path` and `build_classify`? Such utility functions IMO should live outside of class as functions that takes path because it bloat class interface. 
4. Why does it have _FILE_SEPARATOR and _DIR_SEPARATOR? Doesn't it change load order for script?
5. I think the class should be in separate file that is loaded as early as possible, and it could bring utility functions that it needs with it (import it in loader for backward compatibility).

brunoais:
> 1. Why does it inherit PurePosixPath instead of PurePath. I don't undesrtand what are implications of that, but it just a feels wrong.

`PurePosixPath` is `PurePath` with Unix flavour. There is one or another optimization which checks for `PurePosixPath` but, frankly, the biggest optimizations are for the `os.posixpath` done in the builtin `PurePath` in the "borrowed" functionality of `PurePath` which shortcuts the execution if the flavour is `os.posixpath`.
In its essence, `PurePosixPath` is `PurePath` with flavour of `os.posixpath`
In Python 3.12: https://github.com/python/cpython/blob/v3.12.4/Lib/pathlib.py#L801-L808
In Python 3.15.0a7: https://github.com/python/cpython/blob/v3.15.0a7/Lib/pathlib/__init__.py#L594-L601


> 2. Ren'Py already normalizes paths to lower-case NFC Unicode. Why do we need that flag for methods?

I decided, for methods which already exist in `Path` (directly or indirectly) to copy or extend the call signature whenever possible. My main intent is to make it compatible with as many libraries that exist for `pathlib` as viable. I also considered keeping enough familiarity with people already used to `pathlib`.
For me, that's more valuable than keeping a tight signature to the exact behavior (except for the additional methods, not present in `Path`, of course)
 
> 3. Do we really need `rpy_path`, `rpyc_path` and `build_classify`? Such utility functions IMO should live outside of class as functions that takes path because it bloat class interface.

I don't mind that. Where do you suggest me placing them?
Also, if doing like that, I'd suggest changing build.classify() to support a `RenpyPath` instead of a new function.
That leaves us with the other 2 to place.

> 4. Why does it have _FILE_SEPARATOR and _DIR_SEPARATOR? Doesn't it change load order for script?

I need a better clarification of what you mean related to load order.
I'll try to answer closest to what I understand, if it doesn't answer, please clarify.
`_FILE_SEPARATOR` and `_DIR_SEPARATOR` are present as a trick to force total sorting order of files and directories. It enforces strict order where, for each directory, files appear first, then directories and they are all sorted alphabetically. Methods such as `_find_ranges` and `_iter_children` rely on that behavior to function.
Using `/` instead would prevent that because files or directories with any character less than `/` would break that order. Also, files at the corresponding level would appear in between directories, instead of before directories.
Is this what you wanted to know?

> 5. I think the class should be in separate file that is loaded as early as possible, and it could bring utility functions that it needs with it (import it in loader for backward compatibility).

It's not an advantage if this class is loaded earlier than this. Actually, it's only useful when used after `scandirfiles()` runs because it's `scandirfiles()` which finds the files to populate it. 
It may live in a different file, however, I can't see how it should be done so it's available earlier than if it's in `loader.py` that, given it's useless before `scandirfiles()` runs.
I also don't understand what you mean with "backward compatibility". `RenpyPath` is being introduced here the first time.
If it's in a separate file, something must provide the result of `scandirfiles()` to `RenpyPath` or something needs to "trigger" `RenpyPath`'s separate file to load from `scandirfiles()`'s result.

brunoais:
@Andykl Please respond.

Andykl:
I gave it a good thought and I will try to convey my reasoning.

You propose to add RenpyPath as public API, so any decisions you made are gonna stuck for a good amount of years. It is much easier to iterate with functions because you can leave old one with bad interface or something as is and make a new function. But with class, it is more like "should be good first try". That thought was an implicit premise in my questions.

I agree that "make it compatible with as many libraries that exist for pathlib as viable" is a good goal, so making signatues compatible is correct, but that reinforce my objection to inherit from PosixPath, because if library does `isinstance(path, PurePosixPath)` it would be correct from library POV to assume path represents path on real posix filesystem, which Ren'Py is not.

At the same time, I think RenpyPath should be 100% compatiable with any other Ren'Py API that currently uses `str` to represent paths, so case-sensitivity and unicode order should be preserved. So `renpy.loadable(RenpyPath("test.txt"))` and `renpy.loadable(RenpyPath("TEST.TXT"))` should do the same as currently with strings, and any (Ren'Py or user) function that does any sorting of str paths should sort RenpyPath the same. Because Ren'Py currently does not has an API to list directories your approach (with _DIR_SEPARATOR that looks crypitc, but I can live with it) works, but I am not sure if adding such API is a good thing, but that's a thing to Tom to decide. Another thing that I think should be given explicit thought is are `RenpyAllPath("game/file.txt")` and `RenpyPath("file.txt")` equal? (given that gamedir is "game").

So, to sum this up, I think that RenpyPath should be inherited from `PurePath` with a custom flavor/parser, so libraries don't make incorrect assumptions, even if that flavor is indistinguishable from posixpath. I am sure there are libraries that test for `path._flavour == posixpath` to see if they can do some operations.

About the class being in separate file, my reasoning is that your PR doubles the size of `renpy.loader`. You can do something like
```
renpy/renpypath.py
class RenpyPath:
    def _create_path(...): ...
    def _create_too_early(...):
        raise Exception("RenpyPath can't be instantiated before loader system ready.")
    _create = _create_too_early

renpy/loader.py
from renpy.renpypath import RenpyPath
def scandirfiles():
    ...
    RenpyPath._create = _create_path
```
or something like that. I don't mind if you leave it as is, that's also something that Tom may decide one way or the other.

brunoais:
> You propose to add RenpyPath as public API, so any decisions you made are gonna stuck for a good amount of years. It is much easier to iterate with functions because you can leave old one with bad interface or something as is and make a new function. But with class, it is more like "should be good first try". That thought was an implicit premise in my questions.

With a class, you can leave the old method that had a bad interface as is and make a new method. I don't see how a class makes that worse, at that level. I do agree that methods create an explicit inter-dependency between them. However, this `RenpyPath` didn't come out of nowhere as the first thing done.

1. The one I'm producing here is my 7th iteration
1. This one is my 2nd iteration: https://gitlab.com/brunoais/renpy-RenpyPath/
1. I have a blueprint and behavior base to work on top of which is Python's native `PurePath`.

However, that point is valid for the new methods I added. As a middle ground, I can start with separate functions (and using your new module idea, they'd be in the `renpypath` module) and then, after they show themselves as useful and working as intended, migrate the fixed version or the intended version to become a `RenpyPath` method.

And on that 3rd one, I made extensive testing (which I didn't include in this PR) to ensure as-perfect behavior parity with `PurePath` as I could. I ran all [tests python has for their PurePath made for `posixpath` paths](https://github.com/python/cpython/blob/v3.12.13/Lib/test/test_pathlib.py#L723) and I had AI helping me select which ones to run out of all `Path` tests. 

> I agree that "make it compatible with as many libraries that exist for pathlib as viable" is a good goal, so making signatues compatible is correct, but that reinforce my objection to inherit from PosixPath, because if library does `isinstance(path, PurePosixPath)` it would be correct from library POV to assume path represents path on real posix filesystem, which Ren'Py is not.

I think you meant "my objection to inherit from `PurePosixPath`". Assuming that:
`PurePath`s  do not represent filesystem. They are memory-level path arithmetic. For those pure string arithmetic methods, `PurePosixPath` and `RenpyPath` are interchangeable. If a program does `isinstance(path, PurePosixPath)` and trusts that result, the outcome is the expected, according to python's own battery of tests.

 
> At the same time, I think RenpyPath should be 100% compatible with any other Ren'Py API that currently uses `str` to represent paths, so case-sensitivity and unicode order should be preserved. So `renpy.loadable(RenpyPath("test.txt"))` and `renpy.loadable(RenpyPath("TEST.TXT"))` should do the same as currently with strings, and any (Ren'Py or user) function that does any sorting of str paths should sort RenpyPath the same.

If we provide support to execute, I'll use the `renpy.loadable(RenpyPath("TEST.TXT"))` as example, `loadable` would simply get the path by `filename = str(filename)` and then run normally without further change. Same for `open_file`, same for `exists`. The other functions in `loaderexports.py` are clearly for strings and providing a `RenpyPath` makes no sense but we can also support it by calling `str` with the object instance.

If there's a need to normalize input from user into lower case and unicode folding for `RenpyPath`, then that's something I'm ok with and can be done at the user APIs for `RenpyPath`.

> Because Ren'Py currently does not has an API to list directories your approach (with _DIR_SEPARATOR that looks crypitc, but I can live with it) works, but I am not sure if adding such API is a good thing, but that's a thing to Tom to decide. 

I wonder if that cryptic can be mitigated with documentation. It is not API and it's only used for the internal representation. `PurePath`  has no file order, `Path` explicitly claims file sort order to be implementation-dependent, `RenpyPath` makes no claims about sort order of its output. `RenpyPath` files being sorted in its internal structure (due to implementation detail) and the sorting order of output are both not API. The way `list_files` sorts files is not documented anywhere either, so I assume that not being API either because Ren'Py's own code that uses it doesn't rely on the fact files are lexicographically sorted.
I don't know what API you refer to in that paragraph excerpt because that comment is about something that is there but is not API.

> Another thing that I think should be given explicit thought is are `RenpyAllPath("game/file.txt")` and `RenpyPath("file.txt")` equal? (given that gamedir is "game").

I believe you mean them being the same file, as in, calling `.samefile()`. Because equality in pathlib just means that [they have the same pure content.](https://github.com/python/cpython/blob/v3.12.13/Lib/pathlib.py#L519), so by `PurePath`'s API, and mimicking their behavior, either `RenpyAllPath(...) == RenpyPath(...)` must always be `False` or they are `True` when they encapsulate the same text. I.e. these 2 are True `RenpyAllPath("file.txt") == RenpyPath("file.txt")` & `RenpyAllPath("game/file.txt") == RenpyPath("game", "file.txt")`

Assuming you meant `.samefile()`, you make a valid point. In the current implementation, a `RenpyPath` and a `RenpyAllPath` `.samefile()` check isn't correctly implemented and I forgot to test that. I may need to fix that before submitting. In this first iteration, I'm only exposing `RenpyPath` as public API, so I forgot to mix both classes together.
 
> So, to sum this up, I think that RenpyPath should be inherited from `PurePath` with a custom flavor/parser, so libraries don't make incorrect assumptions, even if that flavor is indistinguishable from posixpath. I am sure there are libraries that test for `path._flavour == posixpath` to see if they can do some operations.

Given my explanation above, libraries can assume a `RenpyPath` to be a fully fledged `PurePosixPath` and behave as one. `_flavour` is not public API. But even then, it's true libraries break that. But, for everything `PurePosixPath`-wize, they can use the `path._flavour` and it will work but only for the subset of `PurePath`.

However, if you are still uncertain, I'm ok with making a flavour like I did with my [2nd iteration](https://gitlab.com/brunoais/renpy-RenpyPath/) and make it extend `PurePath` instead.

> About the class being in separate file, my reasoning is that your PR doubles the size of `renpy.loader`. You can do something like
> 
> ```
> renpy/renpypath.py
> class RenpyPath:
>     def _create_path(...): ...
>     def _create_too_early(...):
>         raise Exception("RenpyPath can't be instantiated before loader system ready.")
>     _create = _create_too_early
> 
> renpy/loader.py
> from renpy.renpypath import RenpyPath
> def scandirfiles():
>     ...
>     RenpyPath._create = _create_path
> ```
> 
> or something like that. I don't mind if you leave it as is, that's also something that Tom may decide one way or the other.

I think we can move to a different file with RenpyPath as a module, no issues.
However, I think a guard like that is unnecessary. Code can only instantiate `RenpyPath` too early if it's done by someone editing and customizing Ren'Py's source code or by `rpe` code (archived or not). By the time `python early` blocks run, `scandirfiles` had already been called.
It's done when `renpy.loader.index_files()` is executed, which is right after the multiple `load_rpe(path + "/" + fn)` calls.


About 
> `_create = _create_too_early`

I think sphinx doesn't like that. Complains both methods are the same. [That's why I did the `partial` workaround](https://github.com/renpy/renpy/pull/6955/changes#diff-677c6d58052bbd68c5b100d60e97740837ec2149adc2443915a1c90686177101R1410).


Thank you for the attention.

Andykl:
Classes are not only methods. Yes, for extra methods we can always add a new one, but the devil in the details. Things that are tricky to change after class is exposed as public are - public attributes (if any), `__new__` and `__init__` signatures, and especially pickle methods.

I will take a deeper look at python implementation of pathlib and your library and how it changed over time to see your iterations. It is great that you have a solid ground on the API. 

I want to tell that my comments are not deal breaker for your implementation, just my input, based on working with renpy and python overall, so that we can make better decision on what end user will get. 

Also I think it's important to note that I am currently working at free time on updating renpy to 3.14. I don't see any blockers for it to be ready for next renpy release, but can't be certain when it will be ready, so if I understand correctly your code will need to change after that.

brunoais:
> Classes are not only methods. Yes, for extra methods we can always add a new one, but the devil in the details. Things that are tricky to change after class is exposed as public are - public attributes (if any), `__new__` and `__init__` signatures, and especially pickle methods.

In the case of `RenpyPath` (and `PurePath`) there are no public attributes, so I'm covered there. `__new__` and `__init__` are just the `PurePath` ones, so covered there.
About pickling, I went to do a quick test and [`PurePath`](https://github.com/python/cpython/blob/v3.12.13/Lib/pathlib.py#L356)'s implementation looks correct to me. At least, it seems to work as intended. 

> I will take a deeper look at python implementation of pathlib and your library and how it changed over time to see your iterations. It is great that you have a solid ground on the API.

I don't have a public git history of most of my iterations. But I can try to find what I have in my backups if you need it.
 
> I want to tell that my comments are not deal breaker for your implementation, just my input, based on working with renpy and python overall, so that we can make better decision on what end user will get.

Thank you for clarifying 🙂. I agree. It's productive to discuss and make sure I didn't forget anything. I did extensive work on it but it's not rare for me to fo 99%, believing I did 100% and then someone calls me out of that 1% which I missed. Rarely it's more than a small thing to fix or some tweak but it happens more often at work than I wish it did.

> Also I think it's important to note that I am currently working at free time on updating renpy to 3.14. I don't see any blockers for it to be ready for next renpy release, but can't be certain when it will be ready, so if I understand correctly your code will need to change after that.

There are optimizations I can do for 3.14. Some parts of it, such as `glob` and `rglob` can be optimized better, for certain use-cases, for example. With python 3.14, I can remove the path restrictions of `.files_matching()` and also use [`glob.translate`](https://docs.python.org/3/library/glob.html#glob.translate).

Then I'll need to test again, to make sure all works as expected. But I'm not expecting any regressions.

brunoais:
Moved back to draft waiting for python 3.14 and will implement the optimizations I already predicted I could do for 3.13 or higher.

brunoais:
FYI: I have the update. I'm waiting for Ren'Py python update to 3.14 so I can test if my code works.

brunoais:
@Andykl Any estimate of when renpy using python 3.14 will happen? I already have the code update ready, I just need it to happen so I can download the nightly and test the changes before I undraft this PR.

Andykl:
I was thinking that Python 3.14 would go into 8.6, but we decided to speed the release so SDL3 changes are avaliable sooner than later. Python 3.14 was working before merge of SDL3 but now I need to fix things after rebase, you can see it [here](https://github.com/renpy/renpy-build/pull/189).

So, given that I am not sure what is correct course of actions here. I would not object this going in 8.6 as an experimental/interal class for 8.6 with no guarantee that interface would not change, but I am certan that if it is merged as public class, there will be breaking changes once Python 3.14 is merged, as pathlib was heavyily refactored.

Let's get @renpytom input on the issue.

renpytom:
Can we go into what sort of breaking changes we can expect from 3.14? I haven't been following it too close - but are there really changes that would break user code?

Andykl:
My main concern is pickle. In pathlib it is reduce to the defining class with path parts as arguments, and no extra fields. Not sure how different kinds of renpy paths will interact with it. Plus there were some deprecations removed and new deprecations added. 

There were also refactoring of internals, but it is for class attributes, so not a big deal.

brunoais:
Thank you for the link. I'll track it.
@Andykl 
I don't mind waiting for the merge of python 3.14.
I'd be expecting to not have issues with pickle also because it's a class not designed to be stored in save files or anywhere permanent just like other few Renpy classes that are not meant to be stored permanently. 

At this point I also have already spent the time and effort to code and do basic testing with fake data to make it work with 3.14. I rather wait anyway, even if it will only end up in renpy 8.7 due to any delays merging https://github.com/renpy/renpy-build/pull/189.

Actually, I just checked:
`TypeError: a class that defines __slots__ without defining __getstate__ cannot be pickled` <- This one sets `__slots__` but not `__getstate__`, so it won't ever be pickled.

renpytom:
We should be writing either a getstate or a reduce method, anyway, so we're not dependent on Python details.

renpytom:
I think my main issue is that I don't think this should live in loader. It should be pretty isolated, and so it should live in its own file.

brunoais:
> I think my main issue is that I don't think this should live in loader. It should be pretty isolated, and so it should live in its own file.

I already did that. I'll update this code despite the conflicts so you can see.

brunoais:
@renpytom I pushed with whatever code I have right now without cleaning it up nor turning into commits, if you want to see now.

In loader.py, the only part that matters for version 3.14 is:
```
    from operator import itemgetter
    def rpy_path_normalize_file_list(fn):
        return tuple(sorted(map(normalize_renpy_path, map(itemgetter(1), fn))))

    RenpyPath._paths = rpy_path_normalize_file_list(game_files)
    RenpyCommonPath._paths = rpy_path_normalize_file_list(common_files)
    pos = bisect_left(RenpyPath._paths, RenpyCommonPath._paths[0])
    RenpyAllPath._paths = RenpyPath._paths[:pos] + RenpyCommonPath._paths + RenpyPath._paths[pos:]
```
And deleting the current `RenpyPath`.

I can implement `__setstate__` and `__reduce__`.

PR #6978, opened 2026-03-28
Fix write_target looking for wrong file locations when launcher creates new project

UlyssesZh:


renpytom:
This PR doesn't actually do anything, as far as I can tell. It doesn't change the files that are written from what the current version writes.

UlyssesZh:
Could you try creating a new project using Ren'Py 8.5.2? It fails, and my PR fixes that.

UlyssesZh:
Without this patch, `copy_script` calls `write_target` with full path, but `write_target` expects relative path to the prefix. In this patch, the behavior of `write_target` is changed to expect full paths.

renpytom:
Write_target can take either full or relative paths, but I think it makes sense to make everything in the file take relative paths. That's what I did in ff325d39423b .

UlyssesZh:
I'm okay as long as it fixes the bug that happens when the user creates new projects.

Issue #6981, opened 2026-03-29
Bug on compiling


loviuz:
I am trying to compiling renpy with `./run.sh` because I am preparing a PR for italian translation of the tutorial, but I get this error at the end:

```
﻿I'm sorry, but an uncaught exception occurred.

While running game code:
  File "launcher/game/front_page.rpy", line 260, in script call
    call editor_check
  File "launcher/game/front_page.rpy", line 270, in script
    call screen front_page
  File "renpy/display/render.pyx", line 494, in renpy.display.render.render_screen
    rv = render(root, width, height, st, st)
  File "renpy/display/render.pyx", line 267, in renpy.display.render.render
    rv = d.render(widtho, heighto, st, at)
  File "renpy/display/render.pyx", line 170, in renpy.display.render.render
    cpdef render(d, object widtho, object heighto, double st, double at):
  File "renpy/display/render.pyx", line 267, in renpy.display.render.render
    rv = d.render(widtho, heighto, st, at)
  File "renpy/display/render.pyx", line 170, in renpy.display.render.render
    cpdef render(d, object widtho, object heighto, double st, double at):
  File "renpy/display/render.pyx", line 267, in renpy.display.render.render
    rv = d.render(widtho, heighto, st, at)
  File "renpy/display/render.pyx", line 170, in renpy.display.render.render
    cpdef render(d, object widtho, object heighto, double st, double at):
  File "renpy/display/render.pyx", line 267, in renpy.display.render.render
    rv = d.render(widtho, heighto, st, at)
  File "renpy/display/render.pyx", line 170, in renpy.display.render.render
    cpdef render(d, object widtho, object heighto, double st, double at):
  File "renpy/display/render.pyx", line 267, in renpy.display.render.render
    rv = d.render(widtho, heighto, st, at)
  File "renpy/display/render.pyx", line 170, in renpy.display.render.render
    cpdef render(d, object widtho, object heighto, double st, double at):
  File "renpy/display/render.pyx", line 267, in renpy.display.render.render
    rv = d.render(widtho, heighto, st, at)
  File "renpy/display/render.pyx", line 170, in renpy.display.render.render
    cpdef render(d, object widtho, object heighto, double st, double at):
  File "renpy/display/render.pyx", line 267, in renpy.display.render.render
    rv = d.render(widtho, heighto, st, at)
  File "renpy/display/render.pyx", line 170, in renpy.display.render.render
    cpdef render(d, object widtho, object heighto, double st, double at):
[...]

PR #6985, opened 2026-03-29
Added italian translation of the tutorial

loviuz:
Hi,

I have added italian translation of the tutorial. I cannot fully test it because I have some problem on compiling the game: https://github.com/renpy/renpy/issues/6981

I have worked almost only on new files.


renpytom:
Thanks for the translation. 

I merged this - but it sure looks like a lot of files were missing, and also that there were some constructions that didn't make sense in Ren'Py - like a translation. 

I'll need you to translate the translations added in 2ae939d1fa1fc970891bba5392be886d55c66141 before I can release this - I'll back it out if we get too close to 8.6, but we have time. 

Thank you!

PR #6986, opened 2026-03-30
Restore `renpy.loadsave.dumps` removed in af9679f

UlyssesZh:
The API `renpy.loadsave.dumps` was removed in af9679f, which is a breaking change introduced in 7.5.0 and 8.5.2.

https://github.com/renpy/renpy/commit/af9679f3ab3becd5516e949427505bd25279cb50#diff-9c341eebc74dbd209775af1ea6d4072346feb1bdbbe0af5a35ce9bf0a7238d4fL46-L76

### DIFF
--- a/renpy/loadsave.py
+++ b/renpy/loadsave.py
@@ -32,7 +32,7 @@ import time
 import renpy
 from json import dumps as json_dumps
 
-from renpy.compat.pickle import dump, loads, dump_paths, find_bad_reduction
+from renpy.compat.pickle import dump, dumps, loads, dump_paths, find_bad_reduction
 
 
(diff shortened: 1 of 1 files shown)


renpytom:
This isn't a breaking change. 

renpy.loadsave isn't a public API, it's an internal detail, and so changes to it aren't public. 

How did you detect this? Do you have a game that was accessing renpy.loadsave.dumps?

UlyssesZh:
Yes. I encountered this problem when trying to play Summertime Saga with latest Ren'Py 7.

renpytom:
That's a bug in Summertime Saga - you'll need to contact them and ask them to fix it. This isn't a public API.

PR #6989, opened 2026-03-31
Fix compatibility with older games (backport a455860)

UlyssesZh:
The compatibility fix was not backported to v7.


renpytom:
I merged it, but I don't plan to do any more work on v7 - Ren'Py 8 is the only supported version.

UlyssesZh:
It's okay.

PR #6996, opened 2026-04-03
Add config variable to determine if controller input hides the mouse

jsfehler:
Currently, controller input will hide the mouse. I'm using a controller to move the mouse so having controller input hide the mouse is not ideal. This PR makes this functionality configurable, with the default being the current behaviour.

### DIFF
--- a/renpy/defaultstore.py
+++ b/renpy/defaultstore.py
@@ -77,6 +77,8 @@ _autosave = True
 # Should live2d fading happen?
 _live2d_fade = True
 
+# Should the mouse be hidden when a controller event happens?
+_controller_hides_mouse = True
 
 class _Config(object):
     def __getstate__(self):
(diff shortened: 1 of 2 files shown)


mal:
Wouldn't this want to be a player decision, rather than a developer one?

renpytom:
This should be a store variable, `_controller_hides_mouse`, set in defaultstore.py and accessed through renpy.store. 

I can see sections of the game where controller-moves-mouse is necessary for a minigame, and a store variable would let you show the mouse for those sections and hide it elsewhere.

jsfehler:
@mal I don't mind if it's a preference, it just depends on how Ren'Py wants to promote controller usage. The current assumption is that the mouse should vanish when using a controller. A config variable lets the developer change that if they disagree. A preference says "This decision belongs to the player" (even if the developer hides the preference).

@renpytom I can do that.

renpytom:
Thinking about this, and I could be wrong here...

Is what we want the controller not hiding the mouse, or that the mouse is always shown. I'm thinking for a mouse dependent minigame, we wouldn't want (eg) the keyboard hiding the mouse, either.

renpytom:
2ebc2e667994c11ddcdecaa3f7200bcdb24e3daf adds a new value, "always", to the mouse_visible variable in the main store. I think this accomplishes your goal better than treating controller input specially would.

renpytom:
I'm going to close this, as I think 2ebc2e6 does the same thing.

PR #6997, opened 2026-04-04
sphinx: use renpy.config.basedir to look for incdir

UlyssesZh:


renpytom:
I believe there are problems with this PR (especially with renpy_json.py not being updater).  But more importantly, I don't see the problem you're trying to solve with this PR - why this change is worth making in the first place..

I'm administratively closing this PR, as it's unmotivated.

UlyssesZh:
I just felt it weird to sometimes refer to the Ren'Py base dir and sometimes refer to the cwd to look for incdir. This way one can use the doc generation game without cd and without having it inside Ren'Py.

PR #6999, opened 2026-04-09
Update Reserved Names Docs

HB38-RPY:
This minor update adds Python naming restrictions to the Reserved Names page to assist newer coders with what is and isn't allowed.  The page already lists other Python specific exceptions/rules; these should be here as well.

### DIFF
--- a/sphinx/source/reserved.rst
+++ b/sphinx/source/reserved.rst
@@ -11,6 +11,12 @@ Ren'Py reserves all names beginning with a single underscore (\_). Do not
 use names beginning with an underscore, as that may cause your game to break
 in future versions of Ren'Py.
 
+In addition, Python has rules for what names are valid variables:
+
+* A variable name must start with a letter
+* A variable name cannot start with a number
+* A variable name can only contain alpha-numeric characters and underscores (A-z, 0-9, and _ )
+
 The following is a list of names that are used by Python. Re-using these
 names can lead to obscure problems.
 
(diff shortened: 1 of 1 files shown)


Andykl:
I am not sure I agree with wording here. Ren'Py as well as Python has much more lax requerements for identifiers, most notably, you can have non-ascii chars in names - `ﬁⁿₐˡᵢᶻₐᵗᵢᵒₙ = 3`. I am not a big fan of non-ascii names in Python code, but I can see a value for Ren'Py, so that non-English native devs could write dialogues completly on their native language.

I think we should remove `* A name can only contain alpha-numeric characters and underscores (A-z, 0-9, and \_ )` and add something like "Basically, Ren'Py follow [Python rules for identifiers](https://docs.python.org/3/reference/lexical_analysis.html#names-identifiers-and-keywords)".

KiloOscarSix:
* Must start with a Unicode letter (`a–z`, `A–Z`, or many non-English letters) or an underscore `_`
  * Cannot start with a digit (`0–9`)
  * After the first character, may contain Unicode letters, Digits (`0–9`), Underscores `_`, and Certain Unicode combining characters
  * Cannot contain Spaces, Most punctuation (`! @ # $ % ^ & * ( ) - + =`, etc.), Emojis or symbol characters (e.g. `🔥`, `①`)

michaelkstahlsr:
Python doesn't strictly follow convention either.  There's an entire list of non-normative XID_start and XID_continue Unicode characters that work in Python.  The important convention is non-normative characters have to convert  to Normalization Form KC (NFKC) | Compatibility Decomposition, followed by Canonical Composition.  There's a list.

How specific should the documentation be?  It should be Ren'Py specific, not Python specific, correct?

https://www.unicode.org/Public/16.0.0/ucd/DerivedCoreProperties.txt

renpytom:
I've just taken it right from Python's explanation, in 8a7b6c097e4188579bb751a1f7c7dd011a826cf3

PR #7007, opened 2026-04-13
C++ Optimized RPA Archive Reading Module

NetizenNo1:
Newly Created Files:
src/rpa_archive.h | Note: C++ header file - Defines RPA archive data structures src/rpa_archive.c | Note: C++ implementation - Core reading logic (~500 lines) renpy/loader/rpa_archive.pyx | Note: Cython wrapper - Python interface renpy/loader/rpa_cxx.py | Note: Python caching layer renpy/loader/init.py | Note: Package initialization file Modified Files:
renpy/loader.py | Note: Import the C++ module and modify the load_from_archive() function to prioritize the C++ reader setup.py | Note: Add build configuration for renpy.loader.rpa_archive Optimization Points:
C++ native zlib decompression - Stream decompression via C-level zlib.h Memory optimization - Use malloc / realloc instead of Python's pickle deserialization Binary index parsing - Directly parse binary format and avoid Python's pickle.loads() Caching mechanism - CachedRPAFile class caches opened archives Backward compatibility - Automatically fall back to the Python implementation if the C++ module fails to load

### DIFF
--- a/launcher/game/archiver.rpy
+++ b/launcher/game/archiver.rpy
@@ -32,13 +32,31 @@ init python in archiver:
 
     from pickle import dumps, HIGHEST_PROTOCOL
 
+    try:
+        from renpy.archiver.packer import RPAPacker, derive_key, RPA_ENCRYPTION_KEY_SIZE
+        _HAS_CXX_PACKER = True
+    except ImportError:
+        _HAS_CXX_PACKER = False
+
 
     class Archive(object):
         """
         Adds files from disk to a rpa archive.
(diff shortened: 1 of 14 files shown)


renpytom:
I don't see any motivation for this pull request, or any discussion of of what you were trying to accomplish here. 

If you'd like to open a ticket to discuss this problem, that would make sense. But a major PR like this that contains a lot of software, with no discussion of why it should go into Ren'Py, isn't going to get merged.

PR #7019, opened 2026-04-19
Don't try to kill the image Cache if renpy.display.draw isn't available

nekopy:
In some cases, such as linting, by the time display.im.cache.quit is called, display.draw is no longer available, causing a key error in im.cache.kill

Fix:
Moved im.cache.kill into the if-block in bootstrap containing the display.draw.quit() function


renpytom:
Thanks for the contribution.

Issue #5184, opened 2023-12-16
Renpy Web 7.6.2: Loading save from imported save causes error


heeeeeeeeeeh:
When loading a save I get this error. Loading saves that come from running the game in the browser don't get this error.
```Full traceback:
   File "renpy/common/_layout/screen_preferences.rpym", line 29, in script
     $ ui.interact()
   File "./renpy/ast.py", line 1138, in execute
     renpy.python.py_exec_bytecode(self.code.bytecode, self.hide, store=self.store)
   File "./renpy/python.py", line 1122, in py_exec_bytecode
     exec(bytecode, globals, locals)
   File "renpy/common/_layout/screen_preferences.rpym", line 29, in <module>
     $ ui.interact()
   File "./renpy/ui.py", line 299, in interact
     rv = renpy.game.interface.interact(roll_forward=roll_forward, **kwargs)
   File "./renpy/display/core.py", line 3587, in interact
     repeat, rv = self.interact_core(preloads=preloads, trans_pause=trans_pause, pause=pause, pause_start=pause_start, pause_modal=pause_modal, **kwargs) # type: ignore
   File "./renpy/display/core.py", line 4548, in interact_core
     rv = root_widget.event(ev, x, y, 0)
   File "./renpy/display/layout.py", line 1179, in event
     rv = i.event(ev, x - xo, y - yo, cst)
   File "./renpy/display/layout.py", line 1179, in event
     rv = i.event(ev, x - xo, y - yo, cst)
   File "./renpy/display/layout.py", line 1179, in event
     rv = i.event(ev, x - xo, y - yo, cst)
   File "./renpy/display/screen.py", line 770, in event
     rv = self.child.event(ev, x, y, st)
   File "./renpy/display/layout.py", line 1179, in event
     rv = i.event(ev, x - xo, y - yo, cst)
   File "./renpy/display/layout.py", line 1403, in event
     rv = super(Window, self).event(ev, x, y, st)
   File "./renpy/display/layout.py", line 281, in event
     rv = d.event(ev, x - xo, y - yo, st)
   File "./renpy/display/layout.py", line 1179, in event
     rv = i.event(ev, x - xo, y - yo, cst)
   File "./renpy/display/layout.py", line 1179, in event
     rv = i.event(ev, x - xo, y - yo, cst)
   File "./renpy/display/behavior.py", line 1142, in event
     return handle_click(self.clicked)
   File "./renpy/display/behavior.py", line 1075, in handle_click
[...]

Issue #4256, opened 2023-01-16
Error when save in web version


AdamSmasher1:
when trying to save in the web version of the game, 
not an isolated case

![Error_when_save](https://user-images.githubusercontent.com/104849745/212666566-a2e9169d-b768-47d7-a8c5-7e6188d1d9d4.png)

renpytom:
Closing this, unless new information comes to light.

PR #7021, opened 2026-04-19
Add portable save export/import API (save_to_string/load_from_string)

josephrobertlopez:
## Summary

Adds `renpy.save_to_string()` and `renpy.load_from_string()` functions to enable cross-platform save sharing via copy-paste strings.

## Problem

Mobile browsers often fail to download files from iframes (common issue on itch.io), leaving players unable to backup saves. Additionally, itch.io's shared-origin storage means one game's data can evict saves from other games.

## Solution

String-based save export/import bypasses file download restrictions and gives players direct control over their save data.

## API

```python
# Export save as base64 string
code = renpy.save_to_string("slot-1")  # Returns "RPYS1-..."

# Import save from string  
renpy.load_from_string(code, "imported")  # Loads and continues game
```

## Implementation

- **renpy/loadsave.py**: Core string export/import functions (+219 lines)
- **renpy/config.py**: Configuration variables (+19 lines)  
- **renpy/exports/__init__.py**: API exports (+5 lines)
- **testcases/game/tests/test_savestring.rpy**: In-game test coverage

## Features

- Preserves existing save format and signature validation
- Web builds require explicit opt-in via `config.allow_save_string_web = True`
- Configurable size limits prevent zip bomb attacks
- Comprehensive error handling with specific exception types

## Testing

- 32 automated tests covering validation, error cases, round-trip behavior
- In-game test suite added to `testcases/game/tests/test_savestring.rpy`
- Verified via Playwright with mobile viewport emulation on simulated iframe hosting
- Live hardware testing (Android Chrome, iOS Safari on real itch.io) is pending and will be included as follow-up evidence

## Related Issues

Related: #4256, #5184 (uses string transport that avoids the failure class in #5184)

## Files Changed

- `renpy/loadsave.py` - Core string export/import functions
- `renpy/config.py` - Configuration variables
- `renpy/exports/__init__.py` - API exports
- `testcases/game/tests/test_savestring.rpy` - Test coverage

## Development Note

Developed with Claude Code assistance; functionality verified via Playwright testing. All code follows existing Ren'Py patterns and compilation verified.

Fixes mobile save export issues reported in forum threads and addresses itch.io storage limitations.


mal:
None of the listed "related issues" are actually related to the stated problem or this proposal.

mal:
IMO if there are verifiable problems with web saves that are as problematic as asserted then they should be addressed, but I doubt this is the way to do it. The commit message specifically mentions that this is in response to "reported issues" so it would be good to present those reports as currently there's no visibility on them.

The Itch eviction problem especially seems like something well outside the scope of Ren'Py, any issue reports in that area probably ought to be sent to the team at Itch.

josephrobertlopez:
You're right. I linked unrelated issues and cited 2023 forum threads without current verification.

The actual capability: expose existing JS save export/import as Python API so developers can integrate save transport into custom UI flows.

My real question: How can I actually help indie developers with Android/mobile save challenges? Is this narrow API useful, or should I focus elsewhere?

renpytom:
> How can I actually help indie developers with Android/mobile save challenges? 

So, we already have the Ren'Py Sync system to address this issue, and this issue doesn't really address that. I'd like to start by understanding the problems that you currently are experiencing with the save system, and then we can talk about the right way to solve them. You're the one asserting there is a current problem. 

Starting with a pull request (especially a claude generated one) isn't usually a good idea when the problem isn't clearly established.

josephrobertlopez:
Fair point - should have communicated my experiences first and asked where effort would be most useful. What should I learn about real developer workflows before identifying where to help?

From playing VNs: on some titles, saves work when manually copied between updates for said titles but don't transfer well automatically. I'd wonderer if this was a developer experience issue.

Also curious about broader developer pain points - Android/Linux builds, itch.io distribution, post-signing support.

renpytom:
I'd suggest if you're having problems, please create an issue so we can discuss it directly, rather than a PR.

I'm currently working on an AI policy for Ren'Py, but it's almost certainly going to disallow AI generated PRs at this time. If you'd like to work on an existing project and get it down to root cause - that's reasonable. But changes require full understanding before they go into Ren'Py, and this didn't exhibit that.

PR #7035, opened 2026-04-29
Add 'while' statement to testing suite

A-Voski:
Of course there exists an repeat statement, but that takes a single statement.

A while loop makes the testing suite more flexible. In my usecase, I have a list of scenes that (after choosing in several menus) arrive at the same label, so it's convenient to be able to say "While you are at a menu, pick a random choice and advance until you either reach another menu or label X. Rinse and repeat until you reach X".

I feel like even though the while statement might not be used by a lot of people, it'll help out a minority of developers enough to justify the 34 lines of additional code.


arsym-dev:
A while loop is very handy, I've found myself wanting one a few times.

If possible, it should also be added to the documentation, right under "Until".
https://github.com/renpy/renpy/blob/master/sphinx/source/testcases.rst#until

A-Voski:
Glad to hear someone else also has an use case for it!

As suggested, I've added the while statement to the documentation as well now.

renpytom:
Merged, thank you.

PR #7036, opened 2026-04-30
Update 000statements.rpy

the66F95:


renpytom:
Can you target this to the fix branch, please? It feels useful enough to go there.

renpytom:
I think I'll merge this one, rather than the retargeted one, and cherry-pick - since the other PR is too big.

PR #7040, opened 2026-05-01
Update 000statements.rpy

the66F95:


the66F95:
`config.audio_filename_callback` itself might need a bit of clarification in the docs.
not only audio files are affected by the linked function but also videos. so you have to be precise with the code, you are running there.

Issue #7045, opened 2026-05-07
Dragging stays active without mouse press


kihamala:
Hi All,

This is a follow-up for
https://github.com/renpy/renpy/issues/6980 since I'm not allowed to reopen it.

I used the 8.5.3 pre-release to test the fix for https://github.com/renpy/renpy/issues/6980 but unfortunately it could be even more broken now. However, it requires a different test project to reproduce.  This time there are two similar issues.

[MissingViewportDragEnd_v2.zip](https://github.com/user-attachments/files/27474868/MissingViewportDragEnd_v2.zip)

Steps to reproduce the first case:
1) open the attached project
2) start the game
3) at the draggable image's first line of text, left click
=> now the dragging is active on the second line of text even though you're not pressing the mouse button

(Again, this may require more than one try to reproduce.)

Steps to reproduce the second case:
1) open the attached project
2) start the game
3) keep clicking until the brain image
4) now mouse wheel back to the draggable image
=> now the dragging is active even though mouse button is up

(Seems to happen always?)

I tested with 8.5.2 and these same issues seem to exist there too.

Thanks for RenPy and your help!

renpytom:
Pretty sure this fix solves the problem.

kihamala:
Hi,

After I added those two lines to 8.5.3, I'm getting this exception by just dragging the image. Doesn't happen every time though.
<img width="2578" height="1617" alt="Image" src="https://github.com/user-attachments/assets/a8c8039a-9741-4ff5-873c-c4e142b8d5a8" />

kihamala:
the stacktrace continues like this:
<img width="1159" height="907" alt="Image" src="https://github.com/user-attachments/assets/48e848a4-6b3d-4094-a385-c1114d112cc0" />

renpytom:
Added a fix, I think.

the66F95:
doing it this way, you can still dismiss the say screen when releasing a grabbed viewport while in motion.
d492003ae0cfdb2e6f85218bca9c862d3e8bd69f fixed the issue by ensuring to grab only if there's actually button 1 pressed.
otherwise, even a simple click to dismiss the say screen would have grabbed a focused viewport.

kihamala:
I tested version 68ac025 of viewport.py on top of 8.5.3 and it seems to work beautifully now. I didn't find any more bugs. Great work!

PR #7046, opened 2026-05-08
Update viewport.py

the66F95:
this should fix the initial #7045 and the issues due to e93c86670813e56d5448f09465d1ead363823987. however it needs further testing of nested viewports or viewport as child of a drag.


renpytom:
89d7901 handles this a different way.

PR #7055, opened 2026-05-11
Fix/ruff bare except hygiene

dmitriy-belkin:
Replaces bare except: handlers with explicit exception types (Exception, OSError, etc.) so failures are not silently masked by catching BaseException, and runs Ruff’s safe auto-fixes in `renpy/` (notably removing unused imports and other low-risk cleanups).
Bare except makes debugging harder and is flagged by Ruff (E722). Narrowing the caught exceptions preserves intended behaviour while avoiding overly broad handlers.

**Changes**
- error.py - _safe_string: catch Exception when formatting values for tracebacks.
- update.py - OSError for rmdir / chmod; create empty .new.rpu via open() and .close() to avoid an unused context variable.
- savelocation.py / savetoken.py / pygame/sysfont.py - explicit Exception where the intent is “log and continue” or “fallback and skip”.
- launcher (Visual Studio Code editor) - Exception on the flatpak fallback path.
- Ruff - ruff check renpy --fix for automated fixes across renpy/.


renpytom:
This seems to contain tons of unrelated changes (basically, merging master into fix). 

Let's wait for 8.5.3 to come out, and target this at master.

renpytom:
I'm going to reject this, as it includes a number of changes I don't think benefit Ren'Py. (like turning "not a in b" to  "a not in b"), for minimal benefit.

PR #7075, opened 2026-05-19
Wavedash sdk integration

cloud9c:
Adds native support for the [Wavedash](https://wavedash.com) SDK.

The integration registers a `WavedashBackend` so `achievement.grant(...)` flows to the platform, syncs the savedir as a zip on every save, and exposes a small `wavedash` module for stats / username / fullscreen / overlay calls.

I added `config.save_callbacks` for the cloud-save push. Wavedash hooks it to upload the savedir, but any mod could use it for backups/analytics. The matching loop in `loadsave.py` wraps each callback in try/except so a flaky upload doesn't fail the local save.

Everything else lives in `00wavedash.rpy` and is gated on `renpy.emscripten` plus a `window.Wavedash` probe.

### DIFF
--- a/renpy/common/00start.rpy
+++ b/renpy/common/00start.rpy
@@ -59,6 +59,10 @@ init -1600 python hide:
     # Callback to run before load.
     config.before_load_callbacks = [ ]
 
+    # Callbacks to run after a save slot has been written. Each callback
+    # is called with the slotname that was just saved.
+    config.save_callbacks = [ ]
+
     # Should we suppress overlay during the splashscreen?
     config.splashscreen_suppress_overlay = True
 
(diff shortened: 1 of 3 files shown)


mal:
Friendly reminder that when submitting integration code, it's considered polite to disclose if you are directly affiliated with, or employed by, the organisation or product in question.

renpytom:
@mal They contacted me on twitter about this.

cloud9c:
> Friendly reminder that when submitting integration code, it's considered polite to disclose if you are directly affiliated with, or employed by, the organisation or product in question.

@mal sorry yes! My PR here is affiliated with Wavedash.

renpytom:
Looks good to me. My only concern is with the fullscreen handling not working - let me run some tests and see if it can be integrated.

PR #7078, opened 2026-05-22
Sync gamecontrollerdb.txt

arsym-dev:
6 years worth of updates. Current Ren'Py SDL version is 2.0.20, so we should be good.


renpytom:
This is going to be replaced again in the SDL3 branch, but it's good to have this in now.

PR #7101, opened 2026-06-19
Add `config.failed_save_dump`

Andykl:
This adds a save_dump option that only triggers when pickle failes. Sometimes it is hard to tell how unpickable object got into save paths by a `find_bad_reduction` path, and because the problem is with save itself there is no tracesave to show the full picture, hence this variable that I could enable in my project and ask people to send the dump file.

### DIFF
--- a/renpy/config.py
+++ b/renpy/config.py
@@ -524,6 +524,9 @@ python_callbacks = []
 # If true, we dump information about a save upon save.
 save_dump = False
 
+# Same as save_dump, but only triggers when the save fails.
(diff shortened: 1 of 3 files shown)


renpytom:
I like this. Any reason why this shouldn't be enabled by default? It seems like a valuable debugging tool - just make sure the dump doesn't kill the game if it throws an exception.

Andykl:
Yeah, enabled by default does make more sense. Also added try-except for it, as it could fail if e.g. game can't open dump file.

Issue #6396, opened 2025-06-10
AttributeError when node.statement_start is None


ronchonbon:
Maybe a bit of a convoluted use case and totally possible I'm doing something highly inadvisable, but the following test game:

```renpy
label start:
    show screen dummy_screen()

    while True:
        menu:
            "Does nothing":
                pass

    return

label dummy:
    "Oh, hi there."
    
    return

screen dummy_screen():
    hbox:
        textbutton "Hey":
            action Call("dummy", from_current = True)
```

results in the following error when the "Hey" button is pressed.

```
Traceback (most recent call last):
  File "renpy/bootstrap.py", line 365, in bootstrap
    renpy.main.main()
    ~~~~~~~~~~~~~~~^^
  File "renpy/main.py", line 612, in main
    run(restart)
    ~~~^^^^^^^^^
  File "renpy/main.py", line 137, in run
    renpy.execution.run_context(True)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "renpy/execution.py", line 1029, in run_context
    context.run()
    ~~~~~~~~~~~^^
  File "renpy/execution.py", line 692, in run
    return_site = getattr(node, "statement_start", node).name
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'name'
```

I'm testing on the latest [nightly](https://nightly.renpy.org/8.4.0.25060903+nightly/).

Seems like the easiest fix would be:

```renpy
            except renpy.game.CallException as e:

                if e.from_current:
                    if statement_start := getattr(node, "statement_start", None) is None:
                        statement_start = node

                    return_site = statement_start.name
```

For context, my use case is that I have a bunch of persistent screens that are shown at the same time as the main Ren'Py script. I'm open to ideas about better ways to do this.

renpytom:
statement_start should have always been set, but there is a case where it was missing. I added that and will also use your 
version as backup.

ronchonbon:
Thanks!!

Issue #4755, opened 2023-06-18
BUG: Adding 'mousedown_6' to config.keymap['skip'] causes crash.


henvu50:
BUG: 'mousedown_6' no longer works when added to config.keymap['skip']
Ren'py Version: 8.1.1.23060707

**How to reproduce:**
1. Create a brand new Ren'py Project.
2. Add these two lines of code to gui.rpy
```
init python:
    gui.init(1920, 1080)
    #NEW LINE
    config.keymap['skip'] = [ 'K_LCTRL', 'K_RCTRL', 'mousedown_6']
    #NEW LINE
    renpy.clear_keymap_cache()
```
The above two lines will cause a crash with this error:
```
  ..................
  File "C:\renpy-sdk\renpy\display\behavior.py", line 59, in compile_event
    return compile_event(renpy.config.keymap[key], keydown)
  File "C:\renpy-sdk\renpy\display\behavior.py", line 55, in compile_event
    return "(" + " or ".join([compile_event(i, keydown) for i in key]) + ")"
TypeError: sequence item 2: expected str instance, NoneType found
```

NOTES:
- 'mousedown_6' still works when used for other key functionality outside of 'skip'.
- 'mousedown_6' used to work with config.keymap['skip'] in much older versions of Ren'py. I remember it used to work fine, but I've updated to the newest version of ren'py, and it no longer works.
- 'mousedown_6' is the bottom of the two buttons on the left side of a mouse. 'mousedown_7' is the top one.

The following code works, binding the o hotkey will successfully skip dialogue when pressing o:
```
init python:
    gui.init(1920, 1080)
    #NEW LINE
    config.keymap['skip'] = [ 'K_LCTRL', 'K_RCTRL', 'K_o']
    #NEW LINE
    renpy.clear_keymap_cache()
```

PR #7126, opened 2026-07-03
Fix #6581, #6396, #4755 – loadable empty string, node None crash, keymap int

### LINKED ISSUE #6581: renpy.loadable edge case
Up until Ren'Py 8.3.7, `renpy.loadable("")` returned `False`.
As of Ren'Py 8.4.1, it now returns `True`.

This seems like a mistake as an empty string is not actually loadable and attempting to display it as an image will result in an error.

Mubashir78:
This PR addresses three bugs:

- **#6581:** `renpy.loadable('')` incorrectly returned `True`. An early guard now returns `False` for falsy filenames.
- **#6396:** Accessing `node.statement_start` when `node` is `None` caused an `AttributeError`. Added proper `None` checks before accessing.
- **#4755:** An integer value in a keymap (e.g., `'mousedown_6'`) caused a `TypeError` because it was not iterable. The code now wraps a bare integer in a tuple before iterating.

### DIFF
--- a/renpy/exports.py
+++ b/renpy/exports.py
@@ -1,0 +1,6 @@
+    if not fn:
+        return False
+    if not fn:
+        return False
+    if not fn:
+        return False


Mubashir78:
Closing per request

PR #7127, opened 2026-07-03
Fix renpy.loadable('') returning True incorrectly

Mubashir78:
Adds a guard clause to renpy.loadable to return False immediately if the filename is empty or falsy.


Andykl:
This looks like bot behavior.

PR #7137, opened 2026-07-07
Sort shader names before writing to shaders.txt

### LINKED ISSUE #7134: Consider sorting shaders.txt + line end inconsistency
shaders.txt isn't currently sorted, so the order is always random when it gets written.

I find value in preserving shaders.txt tracked on a git repository (and I deal with the consequences of this decision such as sometimes cleaning it when I add/remove shaders) so that all shaders can be cached at the start. Some shaders are potentially only used in certain scenes so to fully populate shaders.txt with all of the shaders used on the game the game has to be fully run and it's valuable for me to preserve that kind of information in a repository and share it with a team working on the same VN.

Because of this I wanted to check if a change could be made so shaders.txt could be sorted in some way that makes its output consistent so that, for example, the git diff doesn't trigger every time because the lines dance around.

Also, now that I'm talking about shaders.txt, I wanted to point out this:
https://github.com/renpy/renpy/blob/b3486265660dea5d350c1e2027403f83e6ec12d5/renpy/gl2/gl2shadercache.py#L411-L415
`io.open` is called with just `"w"`, but then `"\r\n"` is used.

This is technically incorrect because this defaults to text mode io (as opposed to binary mode with e.g. `"wb"`). On text mode, written newlines are by default translated to the system default line separator `os.linesep`. On Windows this means that lines end on \r\r\n, because the \r is just passed as is, and \n is translated into \r\n.

You could either simply remove \r, or to make it consistent you can either specify the `newline` on `open` or use binary mode `"wb"`

Andykl:
This fixes #7134

### DIFF
--- a/renpy/gl2/gl2shadercache.py
+++ b/renpy/gl2/gl2shadercache.py
@@ -19,9 +19,7 @@
 # OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 # WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-from __future__ import print_function
 import re
-import io
 import os
 
 import renpy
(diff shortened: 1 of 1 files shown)


renpytom:
Ah, already was committing when you did.

Issue #7151, opened 2026-07-11
[Feature] Add a GL state cache to bypass redundant state changes


ronchonbon:
I've had this implemented on my build for a bit, figured it was worth proposing before submitting the full PR.

Most (every?) model draws currently set the full GL state unconditionally. In draw-heavy scenes where many of the draws share states, this leads to a bunch of redundant calls. These calls aren't crazy expensive (though it seems like ANGLE is more affected), but it is wasted work.

My solution was to add a small state cache that tracks the GL state and only issues the diff from current. It resets at the start of `draw_render` passes, so it doesn't blindly trust state between passes or in different contexts, and it restores some default states set by Ren'Py at the end of passes so that external readers still see the same state.

I can set up a PR with benchmarking (it's been a while since I measured myself so would be a good sanity check anyways) if this is of interest.

renpytom:
I like the idea of this, but I would want to see how you're doing things - specifically, we probably want to do something in Cython rather than Python. And it's important to make it easy to add new kinds of state. But fundamentally, I like this. 

Thank you for the other PRs, the ones I just merged.

ronchonbon:
Past me and you were in agreement about setting it up in Cython. I pulled the changes from my fork and drafted the PR. Will set up the benchmark as a sanity check, hopefully tonight or tomorrow.

PR #7152, opened 2026-07-11
Add a GL state cache to skip redundant state changes

ronchonbon:
Companion for https://github.com/renpy/renpy/issues/7151. Going to set up a clean benchmark since it's been a while.


ronchonbon:
Here's a synthetic worst-case benchmark for both this and https://github.com/renpy/renpy/pull/7153.

```renpy
define config.gl_vsync = False

init python:

    import os
    import time

    SPRITES = 2000
    MAX_SPRITES = 24000

    SETTLE = 0.5
    PROBE = 1.5
    MEASURE = 10.0

    VALID_FRACTION = 0.85

    def make_sprites(n):
        solids = []
        texts = []

        for i in range(n // 2):
            solids.append((Solid("#%02x%02x40" % (50 + (i * 13) % 200, 50 + (i * 29) % 200), xysize=(16, 16)), (i * 37) % (config.screen_width - 40), (i * 53) % (config.screen_height - 40)))

        for i in range(n - n // 2):
            texts.append((Text("%03d" % i, size=14), (i * 41) % (config.screen_width - 40), (i * 59) % (config.screen_height - 40)))

        return solids + texts

    bench_sprites = []

    def measure_fps(duration):
        renpy.pause(SETTLE, hard = True)

        start_frames = config.frames
        start_time = time.time()

        renpy.pause(duration, hard = True)

        return (config.frames - start_frames) / (time.time() - start_time)

    def set_sprites(n):
        store.bench_sprites = make_sprites(n)

        renpy.hide_screen("bench_screen")
        renpy.show_screen("bench_screen")

transform bench_spin:
    subpixel True
    rotate 0.0
    linear 1.0 rotate 360.0

    repeat

screen bench_screen():
    for d, x, y in bench_sprites:
        add d pos (x, y)

    # to drive redrawing
    add Solid("#ffffff", xysize=(2, 2)) pos (4, 4) at bench_spin

label start:
    python:
        refresh = renpy.display.get_info().refresh_rate or 60.0
        target = refresh * VALID_FRACTION

        sprites = SPRITES
        set_sprites(sprites)

        # if vsync-capped, scale the load up until the cap can't bind
        while sprites < MAX_SPRITES and measure_fps(PROBE) > target:
            sprites = min(sprites * 2, MAX_SPRITES)

            set_sprites(sprites)

        fps = measure_fps(MEASURE)

        capped = fps > target

        msg = "renderer=%s sprites=%d refresh=%.0f fps=%.1f frame=%.2fms%s" % (
            renpy.get_renderer_info().get("renderer", "?"),
            sprites,
            refresh,
            fps,
            1000.0 / fps,
            " VSYNC-CAPPED-INVALID" if capped else "",
        )

        print(msg)

        with open(os.path.join(config.basedir, "bench_result.txt"), "a") as f:
            f.write(msg + "\n")

        renpy.quit()
```

On my machine:

```
vanilla: renderer=gl2 sprites=2000 refresh=144 fps=80.0 frame=12.51ms

gl2-draw-tweaks: renderer=gl2 sprites=2000 refresh=144 fps=85.5 frame=11.69ms

gl2-state-cache: renderer=gl2 sprites=2000 refresh=144 fps=105.3 frame=9.50ms
```

Andykl:
I wonder is it possible to not use module level global variable and instead pass the object around. With current code it is hard to track where cache use starts and where ends but maybe I am just unfamiliar with this code.

ronchonbon:
Definitely possible, and I think you're right - better. Let me look into it today!

Andykl:
Last two commits make the data flow much better, now I see how things comes together. I still not quite understand when I would change GLCache, and when GLDrawingContext if I ever need to touch this code, but it's me-problem now for sure. 

Overall LGTM.

ronchonbon:
Thanks a lot for the improvement suggestions!

renpytom:
> I'm still not quite understand when I would change GLCache, and when GLDrawingContext if I ever need to touch this code, but it's me-problem now for sure.

I think I understand. GLDrawingContext is about figuring out "what" Ren'Py wants to draw. The state cache is about drawing it faster - intercepting logical calls to OpenGL and avoiding them if possible. 

In any case, I like this and am about to merge it. Thank you, @ronchonbon

PR #7153, opened 2026-07-11
Handful of micro-optimizations for gl2draw

ronchonbon:
Some performance tweaks to gl2draw that I've accumulated. Going to set up benchmark for this alongside https://github.com/renpy/renpy/pull/7152.


renpytom:
This also looks a bit nicer, so it's going in on those grounds.

PR #7154, opened 2026-07-12
Add support for custom cursor styles to Drag objects

altskop:
This is a simple change adding `mouse` style element to Drag objects. It defaults to `drag` (to differentiate it from the regular `button`), and can be overridden with any other user-defined value. It also works with the `pressed_` modifier. 

I've updated the documentation accordingly and added some examples on how this can be used.
Also, slightly reformatted the examples section to be more readable and added some comments for clarity.

### DIFF
--- a/renpy/common/00style.rpy
+++ b/renpy/common/00style.rpy
@@ -372,6 +372,7 @@ init -1800:
 
     style drag:
         focus_mask None
+        mouse "drag"
 
     # Out-of-game menu root windows
 
--- a/renpy/sl2/sldisplayables.py
+++ b/renpy/sl2/sldisplayables.py
@@ -506,6 +506,7 @@ Keyword("mouse_drop")
 Keyword("alternate")
 Style("child")
 Style("sound")
+Style("mouse")
 
 DisplayableParser("draggroup", renpy.display.dragdrop.DragGroup, None, many, replaces=True)
 Keyword("min_overlap")
--- a/sphinx/source/config.rst
+++ b/sphinx/source/config.rst
@@ -905,7 +905,7 @@ Mouse
     Otherwise, this should be a dictionary giving the
     mouse animations for various mouse types. Keys used by the default
     library include ``default``, ``say``, ``with``, ``menu``, ``prompt``,
-    ``imagemap``, ``button``, ``pause``, ``mainmenu``, and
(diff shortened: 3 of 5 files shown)


Andykl:
LGTM in general.
One question, if developer defined "button" in current version of Ren'Py would this continue to work or they need to do `config.mouse['drag'] = config.mouse['button']`? 
If that the case I think it needs mention in "incomatible changes" and redefine in `common/00compat.rpy`

renpytom:
I like this in general. This should be targeted at master, and not fix. What I'd like is that if drag/pressed_drag are not in config.mouse, the button/pressed_button variants can be use. @altskop , if you can retarget this to master and update the documention to mention that, I'll go ahead and handle the backwards compatibility.

altskop:
> One question, if developer defined "button" in current version of Ren'Py would this continue to work or they need to do `config.mouse['drag'] = config.mouse['button']`?

Regular button styles would continue to work. This change functionally only affects the Drag elements, which previously did not even accept `mouse` style property as a keyword, so it shouldn't break any existing projects.

> @altskop , if you can retarget this to master and update the documention to mention that, I'll go ahead and handle the backwards compatibility.

Sure thing, will do.

altskop:
Changed the target to `master` and updated the documentation to specify that `button` will be used as a fallback if `drag` is not defined.

That said, that part needs to be implemented. I'm thinking we could do it by adding additional checks in the `get_mouse()` method in `focus.py`, something along these lines:
```py
def get_mouse():
    """
    Gets the mouse associated with the currently focused displayable.
    """

    focused = get_focused()
    if focused is None:
        return None
    else:
        if isinstance(focused, renpy.display.behavior.Button): # this affects Button and all its subclasses (like Imagebutton)
            return focused.style.mouse or "button" # prioritize button style over default keyword

        # Additional checks for Drag cursors
        elif isinstance(focused, renpy.display.dragdrop.Drag):
            if focused.style.mouse:
                return focused.style.mouse
            elif renpy.config.mouse.get('drag'):
                return "drag"
            else:
                return "button"
        return focused.style.mouse
```
which would also mean removing the default style for it from 00style.rpy

You might have a better suggestion though, this was the first thing that came to mind for me.

Andykl:
I created #7161 based on your comment because I am not sure if another special case is the way to go. Let's see what Tom think of it.

PR #7166, opened 2026-07-16
Correctly oversample Live2D mask renders

### LINKED ISSUE #7140: Big performance drop in 8.6 (SDL3 / Nightly) with Live2D
I've been testing out the 8.6 nightly to check out SDL3, and I've been running into some big performance drops with Live2D. (On Windows)

I'm getting 2 to 3 times more GPU usage, and CPU usage going from 1% to 10%, as well as FPS drops (from max/60fps, down to around 27fps).
This only seems to happen when the Live2D model is updated or rendered, either via `update_function` or by playing a Live2D motion.

I've tested this on several different computers (all of which are higher end PCs, Windows 10, with NVidia cards), and all of which show the performance drop.

Here is a sample game showcasing the performance drop, using sample Live2D data.

[Live2D_PerfTest.zip](https://github.com/user-attachments/files/29820842/Live2D_PerfTest.zip)

ronchonbon:
Introduced by https://github.com/renpy/renpy/commit/825c4c7d767873434abdd8a579ed40d1d54b0ea9, the issue is that Live2D mask renders are oversampled by the new default factor of 8.0 when flattened. This PR passes the default `mesh_oversample` value as a render property in that path.

Fixes https://github.com/renpy/renpy/issues/7140.

### DIFF
--- a/renpy/gl2/gl2draw.pyx
+++ b/renpy/gl2/gl2draw.pyx
@@ -1130,12 +1130,10 @@ cdef class GL2Draw:
             tx, ty = reverse.transform(1, 1)
             oversample = math.hypot(tx, ty) / math.hypot(1, 1)
 
-            cap = renpy.config.mesh_oversample
-            
             if r.properties is not None:
-                cap = r.properties.get("mesh_oversample", cap)
+                oversample *= r.properties.get("mesh_oversample_scale", 1.0)
 
-            oversample = max(1.0, min(oversample, cap))
+            oversample = max(1.0, min(oversample, renpy.config.mesh_oversample))
 
             for i, c in enumerate(r.children):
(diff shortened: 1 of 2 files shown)


renpytom:
It should be up to 8.0, not just 8.0. Isn't the right thing to make sure it gets oversampled by the correct amount?

ronchonbon:
I was favoring keeping it simple, since the masks were hitting the 8.0 wall almost certainly because their meshes live in canvas space. It seemed like the best solution was a per-render override whenever the `gl2draw` side math wasn't appropriate for the context.

But I take your point that this means masks don't get oversampling even when they should. Let me think...

ronchonbon:
I think that's better, see if you agree.

renpytom:
Looks good to me.

PR #7170, opened 2026-07-18
Make sure text input is active when needed.

### LINKED ISSUE #7138: Game loses input focus after switching apps
Most notably in console, after Alt+Tab, all non control keys are ignored unless console is reopened.

Andykl:
This fixes #7138.


Andykl:
The fix was a bit more involved than I though. Tested on android, system keyboard is shown correctly even after minimizing the game.

Issue #7086, opened 2026-05-31
Prevent ATL Animations and movies from halting prediction


colinharvie:
Currently, ongoing ATL Animations and movies prevent images from being predicted, even when using `renpy.start_predict`

I realize that this is to prevent stuttering and lag, HOWEVER, this caues a huge problem for any game that uses an animated CTC indicator (or animated menu screens, etc). 

I've only just now realized that the reason my game seems to always fail on prediction is because my CTC screen contains a image that has a repeating ATL animation that pulses the opacity. So, instead of predicting, a line would finish displaying and then nothing would happen, because the CTC would display, halting any prediction.

I've temporarily found a work around by adding a function to the CTC element's ATL that calls `renpy.game.context().idle_frame(False)`, although this feels really hacky and I have no idea what problems it might cause.

renpytom:
The issue is that idle_frame can be expensive, which can lead to stutter. It's a hard problem.

colinharvie:
I've done some further work and testing, and I've extracted the portion of the code from `idle_frame` that deals with prediction (`interface.prediction_coroutine.send(False)`), and currently just trigger that.

In my specific case with an animated CTC indicator (and in most cases with animated CTCs, I imagine) any stuttering has been completely unnoticeable (and certainly FAR less noticeable than the 0.5s to 1s stutter I currently get because nothing has been predicted due to the animated CTC). 

Perhaps there could be an ATL property of sorts (like `animation`) called something like `allow_prediction` that prevents that particular ATL animation from blocking the animation? (And a similar solution for `Movie()`) 

Additionally, I've noticed that as long as the assets being predicted are small in size, I can have any ATL animation running (a smooth, 5 second linear camera move)  and don't experience any jitter or lag. It's only when the files being predicted are a bit larger (like Live2D Atlases) that I get any jittering.

Perhaps if there were a way to limit the size of elements that can be predicted during animations?

PR #7200, opened 2026-07-29
Reframe idle_frame work around configurable budget

ronchonbon:
Proposal for an extension of PR https://github.com/renpy/renpy/pull/7149, meant to address https://github.com/renpy/renpy/issues/7086.

Rather than waiting for a non-expensive pass, this would provide a configurable time budget within which (some) idle_frame work can be performed.

I've done my best to contain the scope of this change to texture uploads and predictions.

### DIFF
--- a/renpy/config.py
+++ b/renpy/config.py
@@ -947,8 +947,8 @@ gc_print_unreachable = "RENPY_GC_PRINT_UNREACHABLE" in os.environ
 # prediction.
 idle_frame = 4
 
-# The time budget, in seconds, for one inexpensive idle pass.
-idle_frame_budget = 0.0005
+# The time budget, in milliseconds, for one inexpensive idle pass.
+idle_frame_budget = 0.5
 
 # Does taking the transform state go through image reference targets?
 take_state_from_target = False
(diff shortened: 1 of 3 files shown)


Andykl:
This looks reasonable. 
1. I am not sure about idle frame budget being seconds. Could this be reframed to amount of frames or to integer number of ms or ns?
2. I am not sure python time.time has a good resolution, especially on windows. Maybe some pygame/SDL clock can fit better here. This would also address 3.
3. That even more tightly couple renpy.display.core and renpy.display.im. It's not a big concern, but probably at some point prediction and decoding could be factored out to its own module. Maybe after python 3.14 because I want to try free threading and doing everything related to prediction and decoding in its own native thread without gil make more sense.

ronchonbon:
1. For sure, I'll switch it to ms (ns is probably overkill and number of frames requires too much mental math to be useful imo).
2. A good point as well.
3. I wholeheartedly endorse this as a future move :)

ronchonbon:
Okay, see what you think now. I've done my best to parse through what should continue to use the warpable `renpy.display.core.get_time` (which continues to measure from the same reference point, just with higher resolution), e.g. displayables, and what really should be an absolute time measurement. Let me know if I took it too far anywhere.

Andykl:
LGTM.

PR #7206, opened 2026-08-03
doc: Clarify button style customization

### LINKED ISSUE #3649: Document some button style gripes
You can globally set button properties like this at the top of screens.rpy:

```
style button:
    activate_sound "click.ogg"
    mouse "hand_cursor"
```
1) I found some issues. The first and minor one is quick buttons are the only ones to take properties from "default" instead of "button". So it took time to figure out why they weren't affected by the change. If there is a reason for this, it should be noted on the gui docs.

2) A second one is imagebuttons need to be set apart like this. I didn't find it documented in the gui docs. So I guessed until I got it, although it can be inferred.

```
style image_button:
    activate_sound "click.ogg"
    mouse "hand_cursor"
```

3) In order to pass these properties to gallery buttons you need to use:

`g.make_button(name, unlocked, mouse="hand_cursor", activate_sound="click.ogg", tooltip="CG #01")`

I know it says it right there in the docs but an example would lead clueless devs like me to the right path. I didn't realized it until I was wondering why my gallery buttons didn't make a sound.

4) This could be my mistake but I can't figure out how to set those parameters on hyperlinks. Is this supported? I read on Ren'Py Gripes it was but I can't find the documentation. This feels intuitive:

```
style hyperlink_text is gui_text:
    activate_sound "click.ogg"
    mouse "hand_cursor"
```

*This came up due to the need of having any button (including hyperlinks) behaving the same way while being easy to mantain.

kyui-azusa:
Fixes #3649.

Document the button styles that need separate customization in projects created from the default template:

- Quick menu buttons use `quick_button`, which inherits from `default`.
- Imagebuttons use `image_button`.
- Gallery buttons created by `Gallery.make_button` default to `empty` and can receive either a shared style or individual style properties.

The added examples show how to apply an activation sound and mouse cursor in each case.

Validation:

- Built the HTML documentation with Sphinx.
- Confirmed the build has the same 147 pre-existing warnings as an unmodified baseline build and no new warning messages.
- Inspected the generated `gui.html` and `rooms.html` examples.

### DIFF
--- a/sphinx/source/gui.rst
+++ b/sphinx/source/gui.rst
@@ -1340,6 +1340,24 @@ add (to the end of :file:`screens.rpy`, or somewhere in :file:`options.rpy`) the
 
 A huge number of customizations are possible using style statements.
 
+Not every button inherits from the ``button`` style. In projects created from
+the default template, the quick menu uses the ``quick_button`` style, which
+inherits from ``default``. Imagebuttons use the ``image_button`` style. To give
+these buttons the same sound and mouse cursor as other buttons, style them
+separately::
+
+    style button:
+        activate_sound "click.ogg"
+        mouse "hand_cursor"
+
+    style quick_button:
+        activate_sound "click.ogg"
+        mouse "hand_cursor"
+
+    style image_button:
+        activate_sound "click.ogg"
+        mouse "hand_cursor"
+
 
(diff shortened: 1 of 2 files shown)


kyui-azusa:

PR #7207, opened 2026-08-03
Preserve scrollbar thumb grab position

### LINKED ISSUE #6258: [enhancement idea] Make grabbing viewport scrollbar thumb with mouse not so jolting
Currently the viewport scrollbar thumb will have its center point snap to the cursor position when you click on it to drag it. This causes the content within the viewport to jolt up or down depending on how far away from the center of the thumb you grabbed. This can make it confusing when you just want to move the content up or down a little by dragging the thumb, as you loose track of where you were looking.
In most applications I've used, the thumb stays in place where it is, no matter where you click on it and starts dragging from that position. It would be nice if that could be implemented with viewport scrollbars in Ren'Py as well.

Below is an example .gif of the current behaviour:
Image

kyui-azusa:
## Summary

- keep the scrollbar thumb under the point where it was grabbed instead of snapping its center to the pointer
- preserve the grab offset when a screen rebuild replaces the `Bar` displayable
- retain track-click behavior and support horizontal, vertical, and inverted bars

## Why

`Bar.event` previously calculated every mouse position relative to the thumb center. That is appropriate for clicks on the track, but it caused a thumb grabbed away from its center to jump before dragging began.

The bar now records the rendered thumb bounds and pointer offset on activation. Mouse motion subtracts that offset, while activations outside the thumb continue to use the existing center-based calculation.

Fixes #6258

## Validation

- `selectors::scrollbar_thumb_drag` (5 assertions covering click-without-jump, horizontal dragging, track clicking, and vertical/inverted dragging)
- `python3.12 -m compileall -q renpy/display/behavior.py`
- `git diff --check`

### DIFF
--- a/renpy/display/behavior.py
+++ b/renpy/display/behavior.py
@@ -2258,7 +2258,7 @@ class Bar(renpy.display.displayable.Displayable):
     def _draggable(self):
         return self.focusable
 
-    __version__ = 2
+    __version__ = 3
 
     def after_upgrade(self, version):
         if version < 1:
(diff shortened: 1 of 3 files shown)


kyui-azusa:

PR #7208, opened 2026-08-03
Support partial playback in MusicRoom

### LINKED ISSUE #2354: Partial Playback in MusicRoom
I'm trying to create moveable progress bar in MusicRoom. AudioPositionValue is "read-only", so I have chosen FieldValue with 'action' making something like this:
```
renpy.music.play("<from " + str(new_pos) + ">" + file)
mr.queue_if_playing()
```
It works perfect - bar changes current track position as intended.
But there is 1 big problem: after moving a bar function renpy.music.get_playing inside MusicRoom code start to return "\<from X>file.ogg" instead of simple "file.ogg". As a result Music Room doesn't recognise current track as one from playlist. Because of this Next/Previous actions works incorrectly and button for current track doesn't count as selected.
I'm not sure is it possible to change the behaviour of get_playing function itself, but it would be great if at least in MusicRoom code get_playing output will be parsed and unnessesary part ("\<from X>") be removed.
Something like this should work:
```
m = renpy.music.get_playing(channel=channel)
if m is not None:
    m = re.sub('<.*?>', '', m)
```

kyui-azusa:
## Summary

- recognize a MusicRoom track when `renpy.music.get_playing()` includes an audio specifier such as `<from 10>`
- use the normalized track identity for button selection, track actions, queue rebuilding, and next/previous navigation
- keep exact playlist entries unchanged, including entries that intentionally contain their own specifier

## Why

`renpy.music.get_playing()` correctly returns the name passed to the audio system, including audio specifiers. MusicRoom compared that value directly with its registered filenames, so seeking by replaying `<from X>file.ogg` made the current track look unknown. This broke selected buttons and caused next/previous navigation to restart from the beginning of the playlist.

MusicRoom now strips a leading specifier only when the remaining asset name is registered in that room. The public `get_playing()` behavior remains unchanged.

Fixes #2354

## Validation

- `selectors.musicroom::partial_playback` (3 assertions covering track identity, selected state, and next-track navigation)
- `git diff --check`

### DIFF
--- a/renpy/common/00musicroom.rpy
+++ b/renpy/common/00musicroom.rpy
@@ -44,7 +44,7 @@ init -1500 python:
 
             renpy.restart_interaction()
 
-            if renpy.music.get_playing(self.mr.channel) == self.filename:
+            if self.mr._get_playing() == self.filename:
                 if renpy.music.get_pause(self.mr.channel):
                     renpy.music.set_pause(False, self.mr.channel)
                     return
@@ -55,7 +55,7 @@ init -1500 python:
             return self.mr.is_unlocked(self.filename)
 
         def get_selected(self):
-            return renpy.music.get_playing(self.mr.channel) == self.filename
+            return self.mr._get_playing() == self.filename
 
         def periodic(self, st):
             if self.selected != self.get_selected():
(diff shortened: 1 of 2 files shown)


kyui-azusa:

PR #7209, opened 2026-08-03
Refresh Movie paths when they become loadable

### LINKED ISSUE #7105: `Movie()` never plays (black, no audio) when `config.search_prefixes` is set after the image is defined
A `Movie()` whose `play=` path resolves through `config.search_prefixes` plays as a **black, silent screen** if `search_prefixes` is configured *after* the `image NAME = Movie(...)` statement runs.
`Movie.__init__` decides **once, at construction time**, whether the play path is loadable and never re-checks, so if it isn't loadable yet, playback never starts.

Reproduced on **8.5.3 and 8.6.0-nightly**, on **desktop (Linux) and Android**. (Came up porting a game to Android, where an asset pipeline redirected movie paths via `search_prefixes`.)

**Minimal repro** : https://github.com/Quetzalicious/renpy-movie-searchprefixes-bug

**Root cause** (`renpy/display/video.py`):

```python
# Movie.__init__
if (play is not None) and self.any_loadable(play):
    self._play = play        # set ONLY if loadable at construction time
# Movie.play()
if (self._play is None) and (old_play is None):
    return                   # _play is None -> playback never starts
```

`any_loadable()` honors `search_prefixes`, but the image statement constructs the `Movie` during init before a later `search_prefixes` is in place, and `_play` is never re-evaluated before the movie is shown. (Not an FFmpeg issue: `init_input()` with a custom `AVIOContext` set never opens the filename; playback just never reaches the decoder.)

**Suggested fix:** re-check loadability at play time (as `after_setstate()` already does), or set `_play = play` unconditionally and decide playability when the movie is shown.

Repo (MVE, runtime-trace harness, prebuilt APKs for both versions, screenshots): <REPO_URL>

*Disclosure: I used AI to help track this down. I've kept it to a verifiable minimal reproduction + trace rather than speculation. It's my sincere hope that this contributes to solving this issue instead of adding to "AI noise".*

kyui-azusa:
## Summary

- retain unresolved movie `play` and filename-based `loop` paths
- retry their loadability from `Movie.per_interact()` before playback tracking
- add a regression testcase for paths that become loadable after construction

## Why

`Movie.__init__` only assigned `_play` when the path was loadable at construction time. If a later init block changed `config.search_prefixes`, the original path could become loadable, but the normal display lifecycle never checked it again and playback never started.

The refresh stops once each path resolves, and boolean `loop` behavior is unchanged.

Fixes #7105.

## Validation

- `./run.sh testcases test 'selectors.movie::late_search_prefix' --report-detailed` (1 testcase, 3 assertions)
- `python3.12 -m compileall -q renpy/display/video.py`
- `git diff --check`

### DIFF
--- a/renpy/display/video.py
+++ b/renpy/display/video.py
@@ -565,6 +565,13 @@ class Movie(renpy.display.displayable.Displayable):
             if self.mask_channel is not None:
                 self.mask_channel = "movie_mask"
 
+    def _refresh_loadable(self):
+        if (self._play is None) and (self._original_play is not None) and self.any_loadable(self._original_play):
+            self._play = self._original_play
+
+        if (self._queue is None) and (self._original_queue is not None) and self.any_loadable(self._original_queue):
+            self._queue = self._original_queue
+
     def ensure_channel(self, name):
         if name is None:
             return
@@ -648,6 +655,7 @@ class Movie(renpy.display.displayable.Displayable):
             self._play = play
 
         if not isinstance(loop, bool):
+            self._original_queue = loop
 
             if self.any_loadable(loop):
                 self._queue = loop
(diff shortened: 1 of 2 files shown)


kyui-azusa:

PR #7210, opened 2026-08-03
Preserve channel volume during audio emphasis

### LINKED ISSUE #6579: config.emphasize_audio_channels deemphasize resets channel_volume to 1.0 instead of restoring original value
In story scenes, I need to manually control the volume of BGM or ambient music. However, when this audio ducking effect is enabled, the volume during and after the ducking is adjusted to a fixed value, no longer matching the volume I originally set. I suggest saving the volume before emphasizing, changing the volume to the original volume multiplied by the configured ducking volume during ducking, and restoring it to the original volume during deemphasizing. The version used is 8.1.3, and after verification, newer versions retain this behavior.

kyui-azusa:
## Summary

- track audio emphasis as a separate per-channel volume multiplier
- combine that multiplier with the channel's configured secondary volume when updating the audio backend
- document the multiplicative behavior and add a regression testcase

## Why

Audio emphasis previously wrote `config.emphasize_audio_volume` directly into each non-emphasized channel's secondary volume. When emphasis ended, it wrote `1.0`, discarding any volume set with `renpy.music.set_volume()`.

Keeping emphasis separate preserves the configured channel volume. It also means changing the channel volume while emphasis is active immediately uses the new value, and removing emphasis restores that latest value rather than a stale snapshot.

Fixes #6579.

## Validation

- `./run.sh testcases test 'audio::emphasize_volume' --report-detailed` (1 testcase, 5 assertions)
- `python3.12 -m compileall -q renpy/audio/audio.py`
- `ruff format --check renpy/audio/audio.py`
- Sphinx dummy parse (no new diagnostics in `config.rst`; the checkout lacks generated `sphinx/source/inc/*` files)
- `git diff --check`

### DIFF
--- a/renpy/audio/audio.py
+++ b/renpy/audio/audio.py
@@ -298,6 +298,9 @@ class Channel(object):
         # The time the secondary volume of this channel was last set.
         self.secondary_volume_time = None
 
+        # The volume adjustment applied while other channels are emphasized.
+        self.emphasis_volume = 1.0
+
         # Should we stop playing on mute?
         self.stop_on_mute = stop_on_mute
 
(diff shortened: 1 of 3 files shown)


kyui-azusa:

PR #7211, opened 2026-08-03
Allow bars to render behind their thumbs

### LINKED ISSUE #5840: give bars a property to not break the bar underneath the thumb
Currently, thumbs are drawn over a break between the bar, and you are to use thumb_offset in order to make the bar continue unbroken.

I propose that bars have a boolean style property to just not have a break in the bar at all so that it's much simpler to achieve the effect of the bar continuing unbroken. One of the rationales being that thumb_offset is not ideal if your thumb can vary in size (such as the thumb in the bar used to scroll the history screen). Nor is thumb_shadow ideal either since it's sized and attached to the thumb and not the bar.

kyui-azusa:
## Summary

- add a `bar_unbroken` boolean style property, defaulting to `False`
- extend both sides of horizontal and vertical bars underneath the thumb when enabled
- support both resized and cropped bar rendering paths
- document the property and add focused rendering regression tests

## Why

Bars currently leave a gap for the thumb. `thumb_offset` can make a fixed-size thumb overlap that gap, but it cannot reliably handle thumbs whose rendered size changes, such as viewport scrollbar thumbs.

With `bar_unbroken True`, the fore and aft bar surfaces meet underneath the thumb while the thumb geometry and draggable range remain unchanged. Existing bars retain their current rendering by default.

Fixes #5840.

## Validation

- `./run.sh testcases test 'bar' --report-detailed` (2 testcases, 5 assertions)
- `.venv/bin/python setup.py build_ext --inplace`
- `python3.12 scripts/generate_styles.py`
- `python3.12 -m compileall -q scripts/generate_styles.py renpy/display/behavior.py renpy/sl2/slproperties.py`
- `ruff format --check scripts/generate_styles.py renpy/display/behavior.py renpy/sl2/slproperties.py`
- `git diff --check`

### DIFF
--- a/renpy/common/00style.rpy
+++ b/renpy/common/00style.rpy
@@ -238,6 +238,7 @@ init -1800:
         unscrollable None
         bar_invert False
         bar_resizing False
+        bar_unbroken False
         bar_vertical False
 
         # Viewport properties
--- a/renpy/common/_errorhandling.rpym
+++ b/renpy/common/_errorhandling.rpym
@@ -158,6 +158,7 @@ init label _errorhandling:
         thumb_offset 0
         unscrollable None
         bar_invert False
+        bar_unbroken False
         bar_vertical False
 
         # Viewport properties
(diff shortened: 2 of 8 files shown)


kyui-azusa:

PR #7212, opened 2026-08-03
doc: Explain ATL event handling

### LINKED ISSUE #6410: Improve Event Statement Documentation
https://www.renpy.org/doc/html/transforms.html#event-statement

I have no clue if I'm using `event` correctly, writing it correctly, what's supposed to happen, etc (more succinctly, I've never gotten it to work). I've also seen other people mention they couldn't figure out `event` either. It would be nice if there was an example snippet of code provided using `event` as a reference to learn from.

kyui-azusa:
## Summary

- clarify that an ATL `event` is handled only by an enclosing `on` statement
- explain that events do not search forward or cross into separate transforms
- add a working example that transfers control from a `start` handler to a `pulse` handler

## Why

The existing documentation described upward event propagation but did not show the structure required for an `event` statement to reach a handler. This made a later sibling `on` statement or a handler in another transform look like they should work, even though neither encloses the event.

Fixes #6410.

## Validation

- linted a minimal Ren'Py project containing the documented transform
- Sphinx dummy parse introduced no new diagnostics in the edited event section
- `git diff --check`

### DIFF
--- a/sphinx/source/transforms.rst
+++ b/sphinx/source/transforms.rst
@@ -1298,6 +1298,25 @@ The ``event`` statement causes an event with the given name to be produced.
     to the event handler. Otherwise, the event propagates to any containing event
     handler.
 
+    The event is only handled by an enclosing ``on`` statement. It does not search
+    forward for a later ``on`` statement or propagate to a separate transform.
+
+**Example**
+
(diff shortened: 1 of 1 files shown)


kyui-azusa:

PR #7213, opened 2026-08-03
Document Ren'Py version scheme

### LINKED ISSUE #6769: Document Ren'Py versioning scheme
[PEP 440](https://peps.python.org/pep-0440/) specifies the version schemes that other tools in Python ecosystem must recognize.

Currently Ren'Py uses versioning scheme that could be described as `MAJOR.MINOR.PATCH.YYMMDDCC[+(nightly|unofficial)]` where CC is amount of commits on a given day on the branch the version was built from. There are caveats that I think we need to evaluate and/or change and document.
1. There is no infromation from which branch the version was build from, so it is possible to confuse master and fix branch builds.
2. Current code has a special path for dirty tree which I think is error prone if automatic build somehow gets dirty tree.
3. There is no information if version is a prerelease.

Because current scheme conforms with the PEP, I think we should document it as is with following changes:
1. Add `.branch_name` after `+` part unless in master branch.
2. Remove path for dirty tree and add `.dirty` after `+` part.
3. Add `+prerelease` for prerelease versions.

So main release version and nightly is kept the same. Nightly fix version would look like `8.5.1.YYMMDDCC+nightly.fix`, and build from any other branch like `8.6.0.YYMMDDCC+unofficial.branch`.

This would help today for uv to distringuish versions better, because right now it thinks that Ren'Py is always version 8.99.99.

kyui-azusa:
## Summary

- document the `major.minor.patch.YYMMDDCC` version format and each local-version component
- update the public version variable examples to reflect the current format
- omit the local-version separator when an official release has no suffixes
- add focused tests for release, nightly, and unofficial branch versions

## Why

The version implementation gained explicit `unofficial`, `nightly`, `dirty`, and branch markers, but the public documentation still only showed the older unsuffixed shape. The formatter also always appended `+`, producing an invalid trailing separator for a clean official non-nightly build.

## Validation

- `./.venv/bin/python -m unittest unittests.testversions -v`
- `ruff check renpy/versions.py unittests/testversions.py`
- `ruff format --check renpy/versions.py unittests/testversions.py`
- `./.venv/bin/sphinx-build -b dummy -D master_doc=other sphinx/source /tmp/renpy-sphinx-version`
- `git diff --check`

The Sphinx build succeeds; the checkout reports its existing warnings for missing generated `inc/*` files.

Fixes #6769

### DIFF
--- a/renpy/versions.py
+++ b/renpy/versions.py
@@ -98,7 +98,12 @@ def _make_version_string(
     if not official and branch not in {"main", "master"}:
         suffixes.append(branch)
 
-    return f"{major}.{minor}.{patch}.{commit:08d}+{'.'.join(suffixes)}"
+    version = f"{major}.{minor}.{patch}.{commit:08d}"
+
+    if suffixes:
+        version += "+" + ".".join(suffixes)
+
+    return version
 
 
 def get_vc_version() -> VersionDict | None:
--- a/sphinx/source/other.rst
+++ b/sphinx/source/other.rst
@@ -11,18 +11,39 @@ Ren'Py Version
 
 .. include:: inc/renpy_version
 
+Ren'Py version numbers have the form ``major.minor.patch.YYMMDDCC``. The first
+three components identify the release series. In the final component, ``YY``,
+``MM``, and ``DD`` give the date of the latest commit, and ``CC`` gives the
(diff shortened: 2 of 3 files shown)


kyui-azusa:

PR #7214, opened 2026-08-03
Require button press and release on same target

### LINKED ISSUE #6613: Buttons respond to mouse release even if you didn't click on the button
Due to a player complaining about misinputs, I've discovered some odd behavior with buttons: if you click one button, but slide your cursor to a different button before releasing, the second button will be triggered instead of the first one. In fact, if you click somewhere with no buttons, but drag the cursor over to a button before releasing it, it will trigger the button. Typically buttons should only trigger if you click and release on the same button.

Maybe that level of input precision isn't really necessary for most visual novels, but it can be pretty annoying when you have a lot of custom menus, like my game does. If performance is a concern, it could maybe be a config setting?

https://github.com/user-attachments/assets/b60cebcd-eb6c-44d5-9e66-de095bce0c50

kyui-azusa:
## Summary

- remember which focus target each mouse button was pressed on
- activate a button on mouse release only when it matches the press target
- clear pending presses when the pointer leaves the window
- add a testcase covering cross-button release, blank-to-button release, and a normal click

## Why

Buttons currently activate whichever button is focused when the mouse button is released. This lets a press that began on another button, or on empty space, trigger a different button after the pointer moves.

A mismatched release is consumed without running the action, so it also does not fall through to unrelated dismiss behavior. Keyboard and controller activation continue to use the existing path.

## Validation

- `CPPFLAGS="-I$(brew --prefix fribidi)/include" LDFLAGS="-L$(brew --prefix fribidi)/lib" ./run.sh testcases test "selectors::button_release_origin"`
  - 1 testcase passed
  - 3 assertions passed
- `./.venv/bin/python -m compileall -q renpy/display/focus.py renpy/display/behavior.py`
- `git diff --check`

Fixes #6613


kyui-azusa:

arsym-dev:
Confirmed this solution works and is probably as simple as it can be.

The one thing I'd add to the test is this:
```renpy
        $ button_release_result = None
        drag id "button_release_first" to pos (0, 0)
        assert eval button_release_result is None

        drag pos (0, 0) to id "button_release_first"
        assert eval button_release_result is None
```

This PR handles it fine, but it's a good safety check in case someone tries another implementation in the future.

PR #7215, opened 2026-08-03
Support idle imagebutton focus masks

### LINKED ISSUE #5862: feature request: focus_mask 'idle'

kyui-azusa:
## Summary

- allow `focus_mask "idle"` to use an imagebutton's idle image as its focus mask
- use the selected idle image when the imagebutton is selected
- document the new `focus_mask` value
- add testcase coverage for normal and selected imagebuttons

## Why

With `focus_mask True`, an imagebutton uses its currently rendered child as the mask. If the hover image makes an area transparent that was opaque in the idle image, moving over that area can repeatedly switch the button between idle and hover.

The `"idle"` value keeps focus testing tied to a stable image while leaving all existing `focus_mask` values unchanged.

## Validation

- `CPPFLAGS="-I$(brew --prefix fribidi)/include" LDFLAGS="-L$(brew --prefix fribidi)/lib" ./run.sh testcases test "selectors::imagebutton_idle_focus_mask"`
  - 1 testcase passed
  - 2 assertions passed
- `./.venv/bin/sphinx-build -b dummy -D master_doc=style_properties sphinx/source /tmp/renpy-sphinx-idle-focus-mask`
- `./.venv/bin/python -m compileall -q renpy/display/behavior.py renpy/styledata/styleutil.py`
- `git diff --check`

The Sphinx build succeeds with the checkout's existing warnings for missing generated includes and unresolved references.

Fixes #5862

### DIFF
--- a/renpy/display/behavior.py
+++ b/renpy/display/behavior.py
@@ -998,7 +998,7 @@ class Button(renpy.display.layout.Window):
                 fw = rv.width - self.style.right_margin
                 fh = rv.height - self.style.bottom_margin
 
-            mask = self.style.focus_mask
+            mask = self.get_focus_mask()
 
             if mask is True:
                 mask = rv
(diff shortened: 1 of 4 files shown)


kyui-azusa:

PR #7216, opened 2026-08-03
Expose the current rollback identifier

### LINKED ISSUE #7111: Add a way of getting the current statement's rollback identifier
A creator on the Ren'Py Discord server wanted to make a "rollback to last choice menu" button. Jumping to the `menu`'s label instead could be somewhat of an option, but I think that comes with its own complications.

My mind goes to the History screen and its `HistoryEntry` objects that hold a rollback identifier, that can easily be used with the `RollbackToIdentifier()` action.
This _feels_ like the right approach, but as far as I can tell, there is no documented way of getting the rollback identifier for the `menu` statement - which would've made this as easy as storing the ID for later use.

Could you consider adding/documenting a way of getting the current statement's rollback ID (perhaps to be used in a statement callback), or something to that effect?
I'd be open to other solutions too, but this strikes me as something that'd be nice to have in general. Rollback identifiers somehow seem a bit underutilized, only ever mentioned in the `HistoryEntry` objects.

kyui-azusa:
## Summary

- add `renpy.get_rollback_identifier()` for the currently executing statement
- return `None` when the current context has no rollback entry
- document its use from `config.statement_callbacks`
- add a testcase that captures a menu identifier and resolves it through the rollback history

## Why

Rollback identifiers were only exposed through dialogue `HistoryEntry` objects. Other statements, such as menus, already execute within rollback entries but had no public way to expose the corresponding identifier.

This makes it possible for a statement callback to remember a menu location and later pass it to `RollbackToIdentifier()`.

## Validation

- `CPPFLAGS="-I$(brew --prefix fribidi)/include" LDFLAGS="-L$(brew --prefix fribidi)/lib" ./run.sh testcases test "rollback_identifier::current_statement"`
  - 1 testcase passed
  - 2 assertions passed
  - 5 hooks passed
- `./.venv/bin/sphinx-build -b dummy -D master_doc=config sphinx/source /tmp/renpy-sphinx-rollback-identifier`
- `./.venv/bin/python -m compileall -q renpy/exports/rollbackexports.py renpy/exports/__init__.py`
- `git diff --check`

The Sphinx build succeeds with the checkout's existing warnings for missing generated includes and unresolved references.

Fixes #7111

### DIFF
--- a/renpy/exports/__init__.py
+++ b/renpy/exports/__init__.py
@@ -553,6 +553,7 @@ from renpy.exports.rollbackexports import (
     checkpoint as checkpoint,
     fix_rollback as fix_rollback,
     get_identifier_checkpoints as get_identifier_checkpoints,
+    get_rollback_identifier as get_rollback_identifier,
     get_roll_forward as get_roll_forward,
     in_fixed_rollback as in_fixed_rollback,
     in_rollback as in_rollback,
--- a/renpy/exports/rollbackexports.py
+++ b/renpy/exports/rollbackexports.py
@@ -250,3 +250,25 @@ def get_identifier_checkpoints(identifier):
     """
 
     return renpy.game.log.get_identifier_checkpoints(identifier)
+
+
+def get_rollback_identifier():
+    """
+    :doc: rollback
+
+    Returns the rollback identifier for the currently executing statement. This
+    can be passed to :func:`RollbackToIdentifier` or
+    :func:`renpy.get_identifier_checkpoints`.
+
+    This returns None when the current context does not support rollback or a
+    rollback entry is not available.
+    """
+
+    if not renpy.game.context().rollback:
+        return None
+
+    current = renpy.game.log.current
+    if current is None:
+        return None
+
+    return current.identifier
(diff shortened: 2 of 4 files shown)


DevilSpiderAkumo:
How are you managing to make 12 pull requests in such a short period such as 8 hours?
Usually pull requests here appear at most a couple times per week, but 12 PRs in one day, all done by you?
I'm curious on your approach/workflow that makes these speedy pull requests.

kyui-azusa:
You're right, and I'm sorry — I didn't think this through.

Two things I found after going back and reading my own PRs properly instead of defending them:

1. This PR has a real defect. The new `testsuite rollback_identifier` block is inserted in the middle of `testsuite context`, so `context::replay_python` silently became `rollback_identifier::replay_python`. It's been renamed, and it now runs inside my setup/teardown, which appends to `config.statement_callbacks`. My Validation section only ran `rollback_identifier::current_statement`, so it never executed the test I broke. That is exactly the kind of thing volume hides.

2. Ten of my PRs touch `testcases/game/tests/test_testcase.rpy`. I merged them onto master in sequence locally: six of the ten conflict. So even in the best case, merging them costs you a rebase round-trip per PR. I made that for you.

So I've closed all of them, including this one — I'd rather withdraw it than ask you to review a fix for a defect I shipped. The only one I've left open is #7221. The rest I'll resubmit one at a time.

When I do, they'll come in a form that's actually reviewable: smaller, self-contained, and not all landing in the same file.

PR #7218, opened 2026-08-03
Document flattening layered images

### LINKED ISSUE #6379: LayeredImage object as one displayable image
Hello, I use the LayeredImage system and in the process of testing I realized something: if an object has joints that do not have defects in the normal scale, then with a different one than indicated in gui.init (), then there will be transparent lines. 

And there is also a problem that if I use the Alpha channel, then there will be problems with the fact that what is hidden by another image becomes visible and confusing.

I want to have a solution that allows you to simply get one image at the output of the call as with the im.composite manipulator, since there are no problems with it, but this method is simply inconvenient with the advantages that LayeredImage gives. 

Other solutions personally do not see the benefit due to the fact that if I use the completion of joints, there will be problems with alpha, and if I use exactly as it should, there will be problems with lines at the joints.

Example with two bugs:

Image

Example without alpha but windowed mode game:

Image

Example in normal scale and without alpha:

Image

kyui-azusa:
## Summary

- explain why scaling separately rendered layered-image parts can reveal transparent seams
- explain why applying alpha can expose covered portions of overlapping layers
- document `mesh True` with `gl_drawable_resolution False` as the flattening solution
- note the rendering cost and recommend enabling it only where needed

## Why

Layered images normally render their parts independently for performance. Creators encountering scaling seams or overlap artifacts had no documented path to flatten those parts into one texture, even though the engine already supports it through transform properties.

## Validation

- `./.venv/bin/sphinx-build -b dummy -D master_doc=layeredimage sphinx/source /tmp/renpy-sphinx-layeredimage-flattening`
- `CPPFLAGS="-I$(brew --prefix fribidi)/include" LDFLAGS="-L$(brew --prefix fribidi)/lib" ./run.sh /tmp/renpy-layeredimage-doc lint`
- `git diff --check`

The Sphinx build succeeds with the checkout's existing warnings for missing generated includes and unresolved references. The standalone layeredimage example parses successfully; its temporary placeholder project only reports an unrelated image-name lint warning.

Fixes #6379

### DIFF
--- a/sphinx/source/layeredimage.rst
+++ b/sphinx/source/layeredimage.rst
@@ -569,6 +569,31 @@ the images are being predicted properly, it generally won't improve
 performance or image size much to crop the images yourself.
 
 
+**Flatten layers when scaling reveals seams.**
+
+When adjacent parts of a layered image are rendered separately, scaling the
+image can reveal transparent lines at their edges. Applying alpha to the whole
+layered image can also expose parts of overlapping layers that were intended to
+be hidden.
+
+The layered image can be flattened into a single texture before it is scaled or
+given alpha by enabling model-based rendering. Rendering the texture at the
+layered image's virtual resolution keeps the result independent of the window
+resolution::
+
+    layeredimage eileen:
+        mesh True
+        gl_drawable_resolution False
+
+        always "eileen_base"
+        group outfit auto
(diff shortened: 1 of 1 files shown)


kyui-azusa:

Issue #7223, opened 2026-08-04
[Video] Desktop Movie playback is CPU-bound at 4K/60


yorkyang2333:
## Related issue

Related to [#4817](https://github.com/renpy/renpy/issues/4817), which discusses desktop Movie performance, `yuv444` costs, and the concern that Ren'Py should not require external codecs, drivers, or a platform media player.

## Problem

Ren'Py's desktop `Movie` displayable currently relies on a CPU-backed playback path in the normal configuration. At 4K/60, composited movies can stutter and consume substantial CPU because decoding and frame preparation compete with the rest of the render loop.

The cost is not limited to codec decode time. The engine path can include software decode, CPU YUV/RGB conversion, CPU RGBA surface allocation, CPU-to-GPU texture upload, composition, render-thread synchronization, and present/vsync lateness.

Smooth playback in Finder or a standalone media player does not prove that the Ren'Py `Movie` displayable is using hardware decoding, because those applications may use a different platform presentation path.

## Reproduction

The affected use case is a composited title-screen Movie on desktop macOS. The test asset is `game/images/main_menu_bg_2160p.webm` (original file ![here](https://github.com/yorkyang2333/OurBriefEternity/blob/main/game/images/main_menu_bg_2160p.webm)).

| Property | Value |
| --- | --- |
| Codec | VP9 Profile 0 |
| Resolution | 3840x2160 |
| Frame rate | 60 fps |
| Pixel format | `yuv420p` |
| Color range | `tv` |
| Bitrate | approximately 11.3 Mbps |

The Movie is transformed, scaled, blended with other displayables, and redrawn as part of the normal Ren'Py render loop.

## Current path

The current/default path is approximately:

```text
FFmpeg software decode
  -> CPU YUV/RGB conversion
  -> CPU RGBA surface
  -> GL texture upload
  -> Ren'Py composition and present
```

Even when a platform hardware decoder is selected, copying the decoded surface to the CPU and running a CPU color conversion still leaves a high-cost path:

```text
Hardware decoder surface
  -> CPU surface transfer
  -> CPU YUV/RGB conversion
  -> CPU RGBA surface
  -> GL texture upload
  -> Ren'Py composition and present
```

## Expected behavior

[...]

Issue #4817, opened 2023-07-04
Very poor software video playback performance on Windows


RadarNyan:
There were some closed issues related to video playback performance, but I'm still experiencing serious performance problems with the latest versions (both 7.6 and 8.1 branches).

The CPU usage is very high, and the video doesn't play smoothly at all. When I try to play the same video file with MPC-BE + LAV Filters it plays smoothly with no frames dropped and reasonable CPU usage.

If I understand correctly, Ren'Py uses FFmpeg for video playback, which is what LAV Filters is based on as well. Why is the performance difference so great?

(I wrote Windows in the title because it's the only platform I've tested on. The sample video provided in now-closed issue #4070 doesn't play nice for me either, so I don't think my video files are the cause.)

renpytom:
Do you have an example of a movie file that is having problems? Ren'Py's movie playback is optimize for some formats, and not others - specifically, yuv420 is fast, and yuv444 is very slow.

RadarNyan:
It turns out the GPU in my machine is too old (well, kinda hard to get a new GPU in recent years) to support hardware decoding of VP9, and the software decoding performance of Ren'Py is just very, very bad.

I can watch 1080P YouTube (VP9) videos just fine, and if I push the limit, the CPU can do software VP9 playback (using LAV filters) of 2560x1440 ~48fps (at ~20% CPU usage). However, when I try to play the same video file in Ren'Py despite very high CPU usage (reaching 100%), the playback experience is terrible (like no more than 10 fps to my eyes). Even if I lower the resolution of the video down to 1080p, Ren'Py still struggles to play it smoothly.

I understand that for legal reasons, Ren'Py couldn't include H.264 support (which is a shame since almost any device that somewhat works today supports hardware decoding of that). WebM's royalty-free nature seems to be a good choice, but with this kind of software decoding performance, it's unusable for my (old) machine. It would be nice if the software decoding performance could be improved to be on par with others, or maybe include platform-specific workarounds to do video playback natively, like using DirectShow Filters on Windows? Even without stepping into the gray area of using H.264, at least that would provide decent software decoding performance for VP9.

renpytom:
Ren'Py doesn't use hardware decoding, in any case - it's a visual novel engine, not a media player. My feeling is that there's likely something wrong here. Have you checked if your file is using yuv444p? That's slow to turn to into RGB.

RadarNyan:
[...]

PR #7224, opened 2026-08-04
[Video] Add macOS hardware-surface import for Movie displayables

yorkyang2333:
Related to [#7223](https://github.com/renpy/renpy/issues/7223) and [#4817](https://github.com/renpy/renpy/issues/4817).

### Summary

This PR adds an experimental macOS VideoToolbox-to-IOSurface path for composited `Movie` displayables. It covers macOS only; issue #7223 describes the broader cross-platform problem of CPU-bound desktop Movie playback at 4K/60.

The default `Movie` API and software fallback remain intact. Hardware decoding is optional and does not require an external codec, driver, or platform media player.

### Before and after

Before, the desktop path is approximately:

```text
FFmpeg software decode
  -> CPU YUV/RGB conversion
  -> CPU RGBA surface
  -> GL texture upload
  -> Ren'Py composition and present
```

On macOS, this PR enables the following path for compatible movies and renderer capabilities:

```text
FFmpeg VideoToolbox decode
  -> retained NV12 IOSurface
  -> IOSurface plane import
  -> GPU YUV conversion
  -> Ren'Py composition and present
```

This removes the CPU surface transfer, CPU color conversion, CPU RGBA allocation, and CPU texture upload for the supported path. It does not make playback zero-cost: GPU conversion, composition, synchronization, and present/vsync remain.

### Changes

- Add FFmpeg VideoToolbox capability detection and hardware-surface retention.
- Import compatible NV12 IOSurface planes into the macOS OpenGL renderer.
- Convert NV12 to RGBA in GLSL, including video/full-range and color-matrix metadata.
- Fall back to software decoding if hardware setup, transfer, or surface handling fails.

### DIFF
--- a/renpy/audio/audio.py
+++ b/renpy/audio/audio.py
@@ -894,11 +894,23 @@ class Channel(object):
 
         return renpysound.read_video(self.number)
 
+    def read_video_hardware(self):
+        if not pcm_ok:
+            return None
+
+        return renpysound.read_video_hardware(self.number)
+
     def video_ready(self):
         if not pcm_ok:
             return 1
 
         return renpysound.video_ready(self.number)
+
+    def video_stats(self):
+        if not pcm_ok:
+            return {"state": "audio-disabled"}
+
+        return renpysound.video_stats(self.number)
 
 
 # A list of channels we know about.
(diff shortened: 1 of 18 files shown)


renpytom:
Per #7223, this isn't the right approach, and so this isn't going into Ren'Py.

PR #7228, opened 2026-08-05
Use YUV plane textures for desktop video playback

yorkyang2333:
Re-implements #7223

This changes desktop playback of supported YUV420 video frames to use separate luma and packed chroma textures, with YUV-to-RGB conversion in the movie shader.

The existing path converts every supported frame to RGBA with swscale before uploading it to OpenGL. Keeping the decoded YUV representation reduces CPU-side conversion and RGBA upload cost for high-resolution movies.

This does not add hardware decoding or zero-copy decoding. FFmpeg still performs decoding, frames are copied into tightly packed YUV buffers, and the planes are uploaded to OpenGL before composition.

### DIFF
--- a/renpy/audio/audio.py
+++ b/renpy/audio/audio.py
@@ -894,6 +894,12 @@ class Channel(object):
 
         return renpysound.read_video(self.number)
 
+    def read_video_yuv(self):
+        if not pcm_ok:
+            return None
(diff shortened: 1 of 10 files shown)


renpytom:
I like this. Thank you.

PR #7231, opened 2026-08-05
Allow statement prediction to track store-variable values and skip impossible branches

ronchonbon:
Statement prediction currently walks every branch of every `if`, predicting images the game cannot show from its current state. These changes make prediction follow what it can prove about store variables so that `If.predict` can evaluate its conditions and return only the reachable branches.

I tried to set this up conservatively so prediction defaults to not pruning rather than pruning too aggressively on statements that could change game state in unexpected ways. Let me know if I missed any situations you can think of and what you think of the overall approach!

### DIFF
--- a/renpy/ast.py
+++ b/renpy/ast.py
@@ -319,6 +319,15 @@ class Node(Object):
     can't see.
     """
 
+    predict_trust_store_after: ClassVar[bool] = False
+    """
+    True if, when this node is the statement prediction starts from, the
+    statements after it may keep evaluating against the current values of
+    store variables. This requires that executing the node can't change a
+    store variable, and defaults to False as most statements can run
+    creator-supplied code.
+    """
+
     @property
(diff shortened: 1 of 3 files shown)


Andykl:
I think it is too convoluted and have a vague idea how to achieve the same simpler. I need some time to think it through.

ronchonbon:
I'm definitely interested in a way to approach this more simply! The bind I kept running into is how to handle all the times that user authored Python code can be run and therefore change game state in a way that we can't predict. I feel like a better solution to that problem would crack this into a much cleaner flow, but I wasn't able to come up with one.

ronchonbon:
Hmm, actually you've sparked an idea, I'll see if it works...

ronchonbon:
I agree with @Andykl, I think there's a better solution. I'll work on a new PR in the next week or so.

Issue #7094, opened 2026-06-11
Allow file-like objects in `renpy.pygame.image.save`


Andykl:
I used `pygame.image.save` to save jpeg images and because currently it is impossible to pass e.g. BytesIO, I used `tempfile`, but if the result path can't be opened by the SDL it would error with `renpy.pygame.error.error: SaveJPEG: could not open C:\Users\[BAD USERNAME]\AppData\Local\Temp\tmpjyc4r58j\screenshot_surface_1781193303.07166.jpg`.

Pygame allow signature `save(Surface, fileobj, namehint="")` so it will also make ours more in sync with upstream.

renpytom:
Sure. Do you want to handle this, or should I?

Andykl:
I will do it. After SDL3 is merged I will check if there are other other code paths that can't handle bad chars in paths, but it is a separate issue.

Andykl:
On second thought, I will do it after SDL3 is merged in main because likely it will be easier to open file in this function and write to its buffer or passed one, instead of different paths for filename and file-like.

PR #7234, opened 2026-08-06
Refactor iostream

Andykl:
This PR refactors `renpy.pygame.iostream` so it has more clean semantic in both Python and C sides. This also fixes several crashes that unusual but valid, or slightly innacurate file-like implementations could cause, e.g. buffer overflow if `filelike.read` returns more bytes than requested.

I think it would simplify passing border between Python and SDL, makes fixing #7094 trivial, and if I read docs correctly should simplify/make faster IO on mobile and web.

New interface looks like
```py
class IOStream(io.RawIOBase):
    def __init__(self):
        raise TypeError("IOStream is abstract, use one of its subclasses.")

    cdef SDL_IOStream *borrow(self) except NULL:
        """
        Borrow the reference to the underlying SDL_IOStream.

        Caller must make sure this object is kept alive while the stream
        is used by SDL. To transfer ownership use IOStream.take().
        """

    cdef SDL_IOStream *take(self) except NULL:
        """
        Transfers ownership of the underlying SDL_IOStream to the caller, and
        caller is responsible to call SDL_CloseIO on it.

        This object is closed from Python's point of view after the call.
        """

class IOPath(IOStream):
    """
    An IOStream backed by a file on disk, opened by SDL. All operations are
    GIL-free.

    `path`
        The file to open, as a str or os.PathLike[str].

    `mode`
        A string, "rb" to open file for reading, "wb" to open file for writing.

    `name`
        If given, a string with a file name this stream represents.
    """

    def __init__(IOPath self not None, object path, /, str mode="rb", str name=None):
        ...

class IOSubFile(IOStream):
    """
    A read-only window into a file on disk, of `length` bytes starting at
    `base`. Used for files stored inside Ren'Py archives and Android assets.

    The window opens and exclusively owns its own handle on the underlying
    file, so distinct IOSubFile objects over the same path are independent and
    each read is a single SDL_ReadIO with no redundant seek.

    `path`
        The file to open, as a str, or os.PathLike[str]. An IOStream may
        also be given, in which case its SDL stream is taken and this
        object assumes exclusive ownership of it.

    `base`
        Non-negative offset into `path` where sub-file starts.

    `length`
        Non-negative length of the sub-file.

    `name`
        If given, a string with a file name this stream represents.
    """

    def __init__(IOSubFile self not None, object path, /, Sint64 base, Sint64 length, str name=None):
        ...

cdef class IOBuffer(IOStream):
    """
    An IOStream over any object supporting the buffer protocol. The buffer view
    is held for as long as the underlying SDL_IOStream is open, independent of
    this Python object's lifetime.

    `buffer`
        An object supporting the buffer protocol that provides contiguous memory.
        Set to None when this stream is closed.

    `writable`
        If True, `buffer` must allow requesting writable buffer.

    `name`
        If given, a string with a file name this stream represents.
    """

    def __init__(IOBuffer self not None, object buffer, /, bint writable=False, str name=None):
        ...

class IOFileLike(IOStream):
    """
    An IOStream wrapping an arbitrary Python file-like object. Every operation
    reacquires the GIL, prefer IOPath, IOSubFile, or IOBuffer when the data is
    available another way.

    If the wrapped object implements readinto(), reads are zero-copy.

    `filelike`
        A Python file-like object.

    `close`
        If true, closing this object also closes the wrapped object.

    `name`
        If given, a string with a file name this stream represents.
    """

    def __init__(IOFileLike self not None, object filelike, /, bint close=False, str name=None):
        ...

cdef inline SDL_IOStream *SDL_IOStreamFromPython(object obj, str name=None) except NULL:
    """
    This accepts, in order:

    * An IOStream object, which is closed and the underlying SDL_IOStream
      object is returned.

    * A str or path-like filename, which is opened.

    * An object with a name field. The name is interpreted as a filename.
      and opened. The object will be closed.

    * An object that supports the buffer protocol.

    * A file-like object.

    It returns an SDL_IOStream object, or NULL on error.

    Calling this function transfers exclusive ownership of `obj` to the returned
    stream, including responsibility for closing it.
    """
```

### DIFF
--- a/renpy/audio/renpysound.pyx
+++ b/renpy/audio/renpysound.pyx
@@ -52,7 +52,7 @@ from __future__ import print_function
 
 from libc.stdint cimport uintptr_t
 
-from renpy.pygame.iostream cimport SDLIOStreamFromPython
+from renpy.pygame.iostream cimport SDL_IOStreamFromPython
 from renpy.pygame.sdl cimport SDL_IOStream
 
 cdef extern from "renpysound_core.h":
@@ -148,7 +148,7 @@ def play(channel, file, name, synchro_start=False, fadein=0, tight=False, start=
     if audio_filter is not None:
         audio_filter.prepare(get_sample_rate())
 
-    rw = SDLIOStreamFromPython(file)
+    rw = SDL_IOStreamFromPython(file)
 
     if rw == NULL:
         raise Exception("Could not create IOStream.")
@@ -176,7 +176,7 @@ def queue(channel, file, name, synchro_start=False, fadein=0, tight=False, start
     if audio_filter is not None:
         audio_filter.prepare(get_sample_rate())
 
-    rw = SDLIOStreamFromPython(file)
+    rw = SDL_IOStreamFromPython(file)
 
     if rw == NULL:
         raise Exception("Could not create IOStream.")
--- a/renpy/gl2/assimp.pyx
+++ b/renpy/gl2/assimp.pyx
@@ -25,7 +25,7 @@ import threading
 
 from typing import Iterable, Literal
 
-from renpy.pygame.iostream cimport SDLIOStreamFromPython
+from renpy.pygame.iostream cimport SDL_IOStreamFromPython
 from renpy.pygame.sdl cimport SDL_IOStream
 
 from assimpapi cimport (
@@ -968,7 +968,7 @@ cdef public SDL_IOStream *assimp_load(const char *filename) nogil:
 
         try:
             f = renpy.loader.load(fn)
-            rv = SDLIOStreamFromPython(f)
+            rv = SDL_IOStreamFromPython(f)
         except Exception as e:
             pass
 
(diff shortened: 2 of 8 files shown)


Andykl:
Backward compatibility note:
1. IOStream base class no longer can be created. If someone used it they would be broken. It had weird signature, and did some duck typing that no call used. I could add it back as cpdef function and name base class BaseIOStream if needed.
2. `renpy.loader.archives` now has different signature. `start` prefix used in `IOStream.from_split` was empty bytes for almost 2 decades in rpa3, so I moved it into compatibility path that joins it with other chunks and uses IOBuffer.

the66F95:
you also have to update `runtime/android/apk.py l.79` with your current implementation.

Andykl:
Complete missed that one. I think I will make pr that adds emscripten and android runtimes with fixes to main repo so it would be harder for them to desync. They look more like renpy proper than consuming renpy as api.

the66F95:
reading RPA file content is still not possible. it errors out with an unknown path.

ronchonbon:
Working on a fix, will submit a PR this morning!

PR #7237, opened 2026-08-07
Pass text_input to mesh renders

### LINKED ISSUE #7177: Unstable input field when parent displayable has `mesh True`
Input field does not work stably when it is in another displayable with mesh True.

<details>
  <summary>Code</summary>

```renpy
screen sc_test_input():

    default show = False
    default value = ""

    add "#333"

    button:

        # Input is not stable when mesh True
        at transform:
            mesh True

        background "#000"
        padding (0, 0)
        align (.5, .5)
        xsize 400
        ysize 80

        action ToggleScreenVariable('show')

        if show:
            input:
                xsize 1.0
                ysize 1.0
                align (.5, .5)
                size 24

                value ScreenVariableInputValue('value')

        else:
            text "Click to input":
                align (.5, .5)
                size 24

label start:
    call screen sc_test_input()

```
</details>

Used to work on renpy-8.5.4.26061901+nightly

arsym-dev:
Fixes #7177

The `text_input` flag was not being propagated correctly to the mesh render. The mesh render made by `make_mesh()` has `operation = FLATTEN`, so the draw function doesn't traverse its children and `text_input` was never set.

The end result is that `pygame.key.start_text_input()` was never called. When SDL processes keyboard events, it would create KEYDOWN events with unicode="" which gets ignored by `Input.event()` and no TEXTINPUT events would be created.

### DIFF
--- a/renpy/display/render.pyx
+++ b/renpy/display/render.pyx
@@ -797,6 +797,9 @@ cdef class Render:
             self.depends_on_list.append(source)
             source.parents.add(self)
 
+            if source.text_input:
+                self.text_input = True
+
         return 0
 
     cpdef int subpixel_blit(Render self, source, tuple pos, object focus=True, object main=True, object index=None):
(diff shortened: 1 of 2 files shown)


Andykl:
LGTM

PR #7238, opened 2026-08-07
Remove unit tests

Andykl:
Agents often trip over this folder thinking that it has real tests and spend tokens on exploring it. I think @arsym-dev work on Ren'Py aware test infra is a way to go, and with typing it will make us never write python style unit tests.


arsym-dev:
There's a place for python unit tests (e.g. to test environment variables, CLI commands, anything where a crash on startup is actually the desired result), but the unittests folder has not been touched in 13 years and is currently a distraction.

renpytom:
Ayup.

PR #7243, opened 2026-08-07
textshader: Respect {cps} tags when preferences.text_cps = 0

### LINKED ISSUE #7242: Text Shaders with a text_cps set to 0 will prevent the cps tag from being used
It appears that if you have textshaders enabled alongside text_cps = 0 (meaning you want, by default text to instantly appear) then when you override the global cps with a cps_tag it won't work.

`default preferences.text_cps = 0
define config.default_textshader = "typewriter"
label start:
    "Test" "{cps=40}This appears one letter at a time at a speed of 40 characters per second{/cps}"`

The above code will have the dialogue instantly appear, even though cps is set to 40.

`default preferences.text_cps = 0
label start:
    "Test" "{cps=40}This appears one letter at a time at a speed of 40 characters per second{/cps}"`

With textshaders disabled, however, the dialogue will properly appear at a speed of 40 characters per second.

arsym-dev:
Fixes #7242

`layout.cps` is the default cps for the layout (zero in this case). I changed the condition check to `layout.max_time` instead, which reflects the actual per-segment timings. You can do `layout.max_time > 0` if you want to be extra safe. Test with the new `textshaders` suite.

This line was last modified in 376c520b7509db062af0ccafb8650d94282bde20

I'm still not comfortable doing direct commits to the main branches outside of the tests folder, so apologies if this feels minor.

### DIFF
--- a/renpy/text/text.py
+++ b/renpy/text/text.py
@@ -2952,7 +2952,7 @@ class Text(renpy.display.displayable.Displayable):
     def render_textshader(self, render, layout, st, at):
         slow_time = layout.max_time
 
-        if self.slow and layout.cps:
+        if self.slow and layout.max_time:
             slow_time = min(slow_time, st)
 
         render.add_uniform("u_text_slow_time", slow_time)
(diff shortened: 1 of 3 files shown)


Grizzlebubs:
Any news on when a build with this fix integrated will be released?

PR #7244, opened 2026-08-08
Add renpy.reentrant_statement

ronchonbon:
This enables rollback and saving in scenes with Python-driven logic and looped screens. I set it up because I frequently invoke situations like below in my projects and players got frustrated when these functions didn't work as expected.

```renpy
default coins = 0

screen coin_pile():
    frame align (0.5, 0.5):
        vbox:
            text "Coins: [coins]"
            textbutton "Take a coin" action Function(take_coin)
            textbutton "Walk away" action Hide("coin_pile")
            textbutton "Rollback" action Rollback()

init python:

    def take_coin():
        store.coins += 1

        renpy.checkpoint()

    def coin_pile():
        renpy.reentrant_statement() # this is the change, commenting forces rollback to the start

        # makes re-execution harmless!
        if not renpy.get_screen("coin_pile"):
            renpy.show_screen("coin_pile")

        while renpy.get_screen("coin_pile"):
            renpy.pause()

label start:
    "You find a pile of coins."

    $ coin_pile()

    "You leave with [coins] coins."

    return
```

I recognize that this may be a niche use case. I also think it's fairly low-risk? See what you think, both about the approach and how I documented its usage.

### DIFF
--- a/renpy/exports/__init__.py
+++ b/renpy/exports/__init__.py
@@ -556,6 +556,7 @@ from renpy.exports.rollbackexports import (
     get_roll_forward as get_roll_forward,
     in_fixed_rollback as in_fixed_rollback,
     in_rollback as in_rollback,
+    reentrant_statement as reentrant_statement,
     retain_after_load as retain_after_load,
     roll_forward_core as roll_forward_core,
     roll_forward_info as roll_forward_info,
--- a/renpy/exports/rollbackexports.py
+++ b/renpy/exports/rollbackexports.py
@@ -178,6 +178,23 @@ def retain_after_load():
     renpy.game.log.retain_after_load()
 
 
+def reentrant_statement():
+    """
+    :doc: rollback
+
+    Declares that the currently executing statement is re-entrant: it may be
+    re-executed from its start, at any point, without applying its side
+    effects twice. This causes each interaction that occurs inside the
+    statement to get its own rollback record, so rollback is granular even
+    if the statement runs many interactions.
+
+    This should be called each time the statement executes, before its second
+    interaction begins. The declaration ends when the next statement begins.
+    """
+
+    renpy.game.log.statement_reentrant = True
+
+
 def rollback(force=False, checkpoints=1, defer=False, greedy=True, label=None, abnormal=True, current_label=None):
     """
     :doc: rollback
--- a/renpy/rollback.py
+++ b/renpy/rollback.py
@@ -474,9 +474,13 @@ class RollbackLog(renpy.object.Object):
 
     __version__ = 7
 
-    nosave = ["old_store", "mutated", "identifier_cache"]
+    nosave = ["old_store", "mutated", "identifier_cache", "statement_reentrant"]
     identifier_cache = None
     force_checkpoint = False
+
+    # True if the currently executing statement has declared itself
+    # re-entrant. Cleared when a new statement begins.
+    statement_reentrant = False
 
     def __init__(self):
         super(RollbackLog, self).__init__()
@@ -552,6 +556,8 @@ class RollbackLog(renpy.object.Object):
         if not context.rollback:
             return
 
+        self.statement_reentrant = False
+
         # We only begin a Rollback if the previous statement reached a checkpoint,
         # or an interaction took place. (Or we're forced.)
         ignore = True
@@ -581,13 +587,7 @@ class RollbackLog(renpy.object.Object):
         else:
             renpy.python.begin_stores()
 
-        # If the log is too long, prune it.
-        while len(self.log) > renpy.config.rollback_length:
-            if self.log.pop(0).hard_checkpoint:
-                if self.rollback_block:
-                    self.rollback_block -= 1
-                else:
-                    self.rollback_limit -= 1
+        self.prune_log()
 
         self.current = Rollback()
         self.current.retain_after_load = self.retain_after_load_flag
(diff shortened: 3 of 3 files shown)


renpytom:
I think I get the idea - but my one  question is, what advantage does this have over: 

```renpy
show screen coin_pile

while True:
    $ coin_pile()
    
hide_screen coin_pile
```

At least in my mind, this would only be useful if the Python function does something Ren'Py couldn't do. I can be convinced that's the case, but I need a less trivial example. 

We should probably make a list of statements that cause checkpoints to happen, and include that in the documentation - rather than saying every statement that does an interaction. But I can do that once I'm convinced this is going in.

ronchonbon:
Just so I understand your suggestion, the full code would be:

```renpy
default coins = 0

screen coin_pile():
    frame align (0.5, 0.5):
        vbox:
            text "Coins: [coins]"
            textbutton "Take a coin" action Function(take_coin)
            textbutton "Walk away" action Hide("coin_pile")
            textbutton "Rollback" action Rollback()

init python:

    def take_coin():
        store.coins += 1

        renpy.checkpoint()

    def coin_pile():
        # renpy.reentrant_statement() # this is the change, commenting forces rollback to the start

        # makes re-execution harmless!
        if not renpy.get_screen("coin_pile"):
            renpy.show_screen("coin_pile")

        while renpy.get_screen("coin_pile"):
            renpy.pause()

label start:
    "You find a pile of coins."

    show screen coin_pile()

    while True:
        $ coin_pile()
        
    hide screen coin_pile

    "You leave with [coins] coins."

    return
```

Right? Rollback on this proposed code rolls all the way to the start unfortunately. But I could totally be missing something about your suggestion.

mal:
The example in the OP seems potentially flawed if I'm reading it right. Since you can take 10 coins, leave, roll all the way back, roll all the way forward and you will have 11 coins because the Hide action in the screen won't get re-run so you'll end up with an accidental trailing addition, even though the changes in the value of coins have been pushed into the rollback log. 

This is one way to safely achieve the behaviour desired in the example with without changes to the engine, it functions reliably allowing you to take an arbitrary number of coins, leave, roll all the way back the start and then all the way forward again with consistent behaviour:
```rpy
default coins = 0

screen coin_pile():
    roll_forward True

    frame align (0.5, 0.5):
        vbox:
            text "Coins: [coins]"
            textbutton "Take a coin" action Jump('take_coin')
            textbutton "Walk away" action Return()
            textbutton "Rollback" action Rollback()

label take_coin:
    $ store.coins += 1
    jump coin_pile

label start:
    "You find a pile of coins."

    label coin_pile:
    call screen coin_pile()

    "You leave with [coins] coins."

    return
```

ronchonbon:
I think I might have shot myself in the foot by using such a trivial example - it was just meant to demonstrate the flow that I'm hoping to achieve when `take_coin` is replaced by some potentially complex Python logic. I'm specifically trying to avoid having to handle all the control flow within labels: it would force me to put every checkpoint-worthy mutating step into its own label with every screen action as a `Jump` or `Return` as opposed to a `Function`. Nested loops, code shared with non-interacting contexts - it would all have to be unrolled into a bunch of hard-coded label jumps.

I can work on whipping up a less trivial example if that's the block here, I initially didn't want to make something too contrived and was pretty sure pulling something out of one of my projects would mostly invite criticism that I'm not using the engine as intended (hehe).

ronchonbon:
Maybe this ported example is helpful?

```renpy
default state = State()

init python:

    def run_scene():
        renpy.reentrant_statement()

        for phase in state.phases:
            if phase.done:
                continue

            for event in phase.pending:
                present(event) # e.g. show a screen, do some pauses, checkpoints, etc.

                if event.aborts:
                    break

            choice = prompt(phase) # some sort of call-screen-like interaction, with checkpoints
            resolve(phase, choice) # maybe prompt again, with checkpoints

            phase.done = True

            renpy.checkpoint()

label start:
    "Something is about to happen."

    $ run_scene()

    "It happened."
```

Doing the same with labels to achieve granular rollback seems like it would really be a lot of duplicating work creating labels and defining store variables, wiring in the jumps, etc.

mal:
<details><summary>FWIW you can cut down on boilerplate in my previous example by genericising it.</summary>

```rpy
default coins = 0

screen coin_pile():
    roll_forward True

    frame align (0.5, 0.5):
        vbox:
            text "Coins: [coins]"
            textbutton "Take a coin" action Call('wrapper', take_coin, from_current=True)
            textbutton "Walk away" action Return()
            textbutton "Rollback" action Rollback()

label wrapper(fn):
    return fn()

init python:
    def take_coin():
        store.coins += 1

label start:
    "You find a pile of coins."

    call screen coin_pile()

    "You leave with [coins] coins."

    return
```
</details>

While I get what you want to do there are existing patterns that work for this kind of thing in Ren'Py (such as queuing events and using an rpy while loop to iterate over each—quick pseudo code below) where you the programmer may have to make some of your more complex code re-entrant but Ren'Py doesn't need to know or care about it.

That's not to say Ren'Py wouldn't benefit from being able to have something like the functionality you describe, but rollback stability is already fairly complex for people to wrap their heads around and the proposed decorator which makes things that won't work, kinda look like they _might work, almost, sometimes_, seems more harmful than useful, at least in it's current form.

```rpy
default coins = 0
default queue = ['coin']

screen coin_pile():
    roll_forward True

    frame align (0.5, 0.5):
        vbox:
            text "Coins: [coins]"
            textbutton "Take a coin" action Return(1)
            textbutton "Take more coins" action Return(5)
            textbutton "Walk away" action Return('done')
            textbutton "Break it down" action Return(dance)
            textbutton "Rollback" action Rollback()

label  start:
    "You find a pile of coins."

    label processor:
    while queue:
        $ process(queue)

    "You leave with [coins] coins."
    return

label woah:
    "Eh, that's probably enough."
    jump processor

init python:
    def process(q):
        # you'd probably dispatch to listeners or something in here, but for the sake of example:
        current = q[0]

        if current == 'coin':
            rv = renpy.call_screen('coin_pile')
            if rv == 'done':
                q.pop(0) # clear coin event
            elif isinstance(rv, int):
                store.coins += rv
            else:
                q.insert(0, dance)

            if store.coins > 10:
                q.append('woah')
                q.pop(0) # clear coin event

            return

        elif current == 'woah':
            q.pop(0) # clear event
            renpy.jump('woah')

        elif callable(current):
            current()
            q.pop(0) # clear event

    def dance():
        renpy.notify("TODO: Insert epic cool dance number here!")
        return
```

ronchonbon:
I hear you! I personally don't like working like the example above, but I appreciate the reasoning. Even if this doesn't belong on main, anyone trying to achieve the same thing can reference how I did it from this PR. We can see how Tom feels.

Issue #7245, opened 2026-08-08
Migrate to GLSL ES 3.00 / GL 3.3


ronchonbon:
Per discussion on Discord a few weeks ago. I thought I'd take a stab at starting some of the migration work and have opened a draft [PR](https://github.com/renpy/renpy/pull/7246). This issue can be used to track discussion.

The main idea is to get Ren'Py to the newest versions of GLSL ES and GL that maintain platform parity across desktop/mobile/web. That puts us at GLSL ES 3.00 and the desktop equivalent GL 3.3. This should be the easiest target versions for maintainers at the moment, since they both share a shader dialect. Maybe one day there should be consideration into supporting higher platform-specific versions, but I don't really think it's worth the headache at the moment.

We're currently on GLSL ES 1.00 and GLSL 1.20, so there's a decent set of features we gain that IMO make this well worth the effort.

The main decision point right now in my mind is how to maintain support for custom shaders written before the eventual migration. In the PR, I proposed that we modify `register_shader` to take a keyword `glsl` that gates how the shader is emitted: a developer can choose to write the shader in either the "modern" format of the newer versions or the "legacy" format that Ren'Py has supported to-date, with the latter case translated to the modern format at compile time. It would default to `glsl = 100` (the legacy format), and we could gently encourage new shaders to be written with the `glsl = 300` kwarg.

To-dos:
- [x] Add shading language and 3.3 context
- [x] Default `glsl` to `config.glsl_version`
- [x] Error on detection of obsolete glsl
- [x] Set up a basic test suite
- [x] Draft doc/changelog updates
- [x] Add GL 3.3 Core support

renpytom:
I'd suggest the glsl argument should default to None, where None means "the value of config.glsl_version". That could be a modern version for new games, and games that declare compatibility would get the old version - though we should probably have a Ren'Py-level error if the more obvious bits of of older glsl are present, telling them they need to use the keyword argument or change config.glsl_version.

ronchonbon:
I think that would work well. I was trying to be conservative about making people modify existing code, but it's a good point that someone trying to update a game with a newer version of Ren'Py is probably capable of adding a config line to opt-out?

Agreed about the error.

ronchonbon:
@renpytom I think I've reached a good checkpoint on the PR and it's ready for a sanity check.

The next items I was thinking of tackling, since they're both on the path to migrating Core:

[...]

PR #7246, opened 2026-08-08
Migrate to GLSL ES 3.00 / GL 3.3

ronchonbon:
Companion PR for https://github.com/renpy/renpy/issues/7245.


renpytom:
I think this is going in the right direction. I'd like to merge things in sensible pieces - merge this before some of the other work you've proposed. (This is good, but would not want it to get much larger before merging.) 

I don't love the name ``renpy_FragementColor``. It might sense to call it fragment_color,  or something like that.

ronchonbon:
Totally agree, I like contained PRs. Let me make the name change and then mark as ready for review.

ronchonbon:
Tried to sneak in a bit more: some better erroring, updates to docs that I noticed were out-of-date now, and translating existing shaders to the new syntax without adding any features that would break hardware support anywhere. Let me know if you want to peel off any of these new commits and save for the next PR - I thought they felt like strict improvements but maybe you disagree.

renpytom:
FYI, this is mostly reviewed, though I have a couple changes I'll make before merging it. (Simple enough I'll just make them rather than writing them out.)

renpytom:
And sorry I took that long, but I needed a night when  I could  really dig down into this to make sure I understood athe changes. They deal with four minor issue, and I'm on the fence about the last commit - not downgrading for a double-failure feels like the right behavior.

In any case, about to push the merge button. Thank you for your contribution - this is a really nice piece of work.

PR #7255, opened 2026-08-10
Fix RPA loading

ronchonbon:
Fixes an issue loading RPAs introduced by https://github.com/renpy/renpy/pull/7234. I tried to infer the intent of the original PR, but the first commit here solves the problem just by reverting to the old behavior.

`webaudio` is touched to prevent a silent sound drop if handed a raw stream directly.

### DIFF
--- a/renpy/audio/webaudio.py
+++ b/renpy/audio/webaudio.py
@@ -175,7 +175,7 @@ def play(
 
     try:
         if not isinstance(file, str):
-            file = file.raw.name
+            file = getattr(file, "raw", file).name
     except Exception:
         if renpy.config.debug_sound:
             raise
@@ -212,7 +212,7 @@ def queue(
 
     try:
         if not isinstance(file, str):
-            file = file.raw.name
+            file = getattr(file, "raw", file).name
     except Exception:
         if renpy.config.debug_sound:
             raise
--- a/renpy/audio/webaudio.py
+++ b/renpy/audio/webaudio.py
@@ -175,7 +175,7 @@ def play(
 
     try:
         if not isinstance(file, str):
-            file = getattr(file, "raw", file).name
+            file = file.name
     except Exception:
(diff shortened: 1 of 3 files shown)


Andykl:
https://github.com/renpy/renpy/blob/0db554ebef792811ff31e5531108f08fdbcd57c5/renpy/loader.py#L574-L611
Looking at the code from before the change I see that `from_split` path was dangling since 13cfd68491a38ca1d864c6b668b8c527e58923e2.

In any case, both branches used `BufferedReader`. Without it `renpy.open_file` would regress in performance in some cases.

For audio case, I believe it could be just `file = file.name` because BufferedReader.name is a property to the underlying `raw`.

Andykl:
Fix for the `read_index` is correct.

ronchonbon:
Ah, thanks for the catch. Seems like the right move is to add a branch to `SDL_IOStreamFromPython` then, e.g.

```python
    cdef IOStream stream
    if isinstance(obj, IOStream):
        stream = <IOStream> obj

...

    elif isinstance(getattr(obj, "raw", None), IOStream):
        stream = IOFileLike(obj, name=name)

...

    else:
        raise TypeError(f"{obj!r} is not a filename or file-like object.")
```

Andykl:
I don't like `SDL_IOStreamFromPython` in general. It did and does very optimistic dispatch for general file-likes, but yeah, we can add back
```
while hasattr(obj, "raw"):
    obj = obj.raw

cdef IOStream stream
if isinstance(obj, IOStream):
    stream = <IOStream> obj
...
```

Andykl:
Lgtm. I will test it locally to confirm it fixes the problem

Andykl:
I made the same change for web audio in #7257, without some change in JS code it was still not playing for me.

Also tested that files are loaded from the archives using The Question, with this PR it works, so it could be merged.

PR #7259, opened 2026-08-12
feat(audio): allow runtime seeking on channels

yorkyang2333:
Inspiration from [this commit](https://github.com/potato-flower-alt/renpy/commit/6354f6a600872463eca45d870ec9ad6e7ae25697).

This adds a public `renpy.music.seek(position, channel="music")` API for seeking the media currently playing on an audio channel, while `AudioPositionValue` can opt into an adjustable progress bar:

```renpy
bar value AudioPositionValue("music", adjustable=True)
```

The existing default remains read-only, so existing uses retain their current behavior. Seeking is routed through the channel API to native and web audio backends. The native FFmpeg backend processes requests on its decoder thread, flushes its decoder and resampler state, and clears queued decoded audio and video data. The WebAudio backend seeks buffered audio and video sources while preserving the current stream and channel queue, rather than restarting playback with an audio `<from ...>` specifier.

### DIFF
--- a/renpy/audio/audio.py
+++ b/renpy/audio/audio.py
@@ -857,6 +857,14 @@ class Channel(object):
 
         return renpysound.get_duration(self.number)
 
+    def seek(self, position):
+        """Seeks this channel without reopening its media stream."""
+
+        if not pcm_ok or self._number is None:
+            return
+
+        renpysound.seek(self.number, position)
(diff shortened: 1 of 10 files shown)


renpytom:
Looks good to me, thanks for the contribution.

PR #7265, opened 2026-08-14
Fixes doc label file typos

LucileGB:
This fixes a few typos in label.rst.

Sorry in advance if I've done anything wrong, it's my first PR!

Thank you for your hard work.

**Edit:** Oh no. This is showing that I'm trying to merge 717 commits into fix instead of just one. I _definitively_ did something wrong.


arsym-dev:
Your change was made on top of the `master` branch (which will eventually become renpy 8.6.0), but you tried to merge into `fix` (which is is the much older 8.5.4). That's why it's trying to bring so many other changes with it.

Make a new branch based off `fix` and re-apply your changes, then try another PR. Thanks for the contribution!

Issue #7267, opened 2026-08-14
Support language overrides by resource category


yorkyang2333:
Inspired by [this commit](https://github.com/potato-flower-alt/renpy/commit/6354f6a600872463eca45d870ec9ad6e7ae25697).

Ren'Py currently uses one language preference for translated text, images, and voice resources. This makes some common localization setups difficult to express without changing the resource layout or adding game-specific workarounds.

For example, a game may want:

- Text in Simplified Chinese.
- Voice in Japanese.
- Localized UI images that follow the text language.

The engine should support independent language selection for text, images, and voice while preserving the existing global language behavior for games that do not opt into the new controls.

The proposed behavior is:

- The global language remains the fallback for all resource categories.
- Text, image, and voice languages can each be overridden independently.
- A category can be fixed to a translation language or to its original resources.
- A category can be switched back to following the global language.
- Image and audio resource categories are selected from the existing loader directory context, without inferring categories from file extensions.

This should be available through engine APIs and persistent preferences.

yorkyang2333:
My implementation is at #7268, please take a look. Thanks!

renpytom:
I like the idea of this, but don't love that your implementation hardcodes the implementation for those three categories. 

My thinking is that this belongs at the path prefix level rather than the asset type level. So there would be _preferences.translate_paths = { "backgrounds" : "tl/english/background", "voice/eileen" : "tl/chinese/voice/eileen", "voice/lucy" : "tl/japanese/voice/lucy", "gui" : "tl/klingon/gui" }

This could be set up by a creator how they want, and would be used to translate all files except for text. (With the current behavior as a fallback for paths that aren't matched.) 

What do you  think?

yorkyang2333:
> I like the idea of this, but don't love that your implementation hardcodes the implementation for those three categories.
> 
> My thinking is that this belongs at the path prefix level rather than the asset type level. So there would be _preferences.translate_paths = { "backgrounds" : "tl/english/background", "voice/eileen" : "tl/chinese/voice/eileen", "voice/lucy" : "tl/japanese/voice/lucy", "gui" : "tl/klingon/gui" }
> 
> This could be set up by a creator how they want, and would be used to translate all files except for text. (With the current behavior as a fallback for paths that aren't matched.)
> 
> What do you think?

Yes, that simplified the logic. Reimplemented in #7270.

yorkyang2333:
Closing as #7270 has been merged.

PR #7268, opened 2026-08-14
Title: Add resource-specific language overrides

yorkyang2333:
Implements #7267.

This change adds persistent language overrides for text, image, and voice resources.

The global language remains the fallback, so existing games retain their current behavior. Each category can now be selected independently:

- `renpy.change_text_language(language)` controls text translation and translated strings.
- `renpy.change_image_language(language)` controls translated images loaded with `directory="images"`.
- `renpy.change_voice_language(language)` controls translated audio loaded with `directory="audio"`.

For each category, `None` selects the original resources, a language string selects a translation, and `renpy.GLOBAL_LANGUAGE` restores following the global language. The corresponding `renpy.get_*_language()` functions report the effective language.

### DIFF
--- a/renpy/common/00gui.rpy
+++ b/renpy/common/00gui.rpy
@@ -134,7 +134,7 @@ init -1150 python in gui:
         Note: This is a very slow function.
         """
 
-        renpy.translation.change_language(_preferences.language, force=True, rebuild=True)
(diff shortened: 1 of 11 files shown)


yorkyang2333:
According to https://github.com/renpy/renpy/issues/7267#issuecomment-5295943190, reimplemented in #7270.

PR #7270, opened 2026-08-15
Add language mappings for resource path translations

yorkyang2333:
Reimplements #7267, according to the idea of @renpytom (https://github.com/renpy/renpy/issues/7267#issuecomment-5295943190).

### DIFF
--- a/renpy/common/00action_other.rpy
+++ b/renpy/common/00action_other.rpy
@@ -431,6 +431,32 @@ init -1500 python:
             return self.language in renpy.known_languages()
 
 
+    @renpy.pure
+    class ChangeResourcePathTranslation(Action, DictEquality):
+        """
+        :doc: language_action
+
+        Changes the translated path used for resources beneath `path`.
+        Passing None as `translation` removes the mapping.
+        """
+
+        def __init__(self, path, translation):
+            self.path = path
+            self.translation = translation
+
+        def __call__(self):
+            if self.translation is None:
+                _preferences.resource_path_translations.pop(self.path, None)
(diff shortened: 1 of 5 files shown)


renpytom:
I like where this is going. The only thing is this is useful for things other than resource paths - for example, it can be used to toggle sprites. So it should be _preferences.resource_path_translations, and there needs to be a ChangeResourcePathTranslation action.

This also needs to be invoked at after the directory argument that is currently handled by get_prefixes is applied - that should be considered part of the path. (If that's not obvious, let me know and I can do that part.) Other than that, if you can draft a changelog entry - I'd appreciate it.

yorkyang2333:
@renpytom Thanks for your comment! 

> I like where this is going. The only thing is this is useful for things other than resource paths - for example, it can be used to toggle sprites. So it should be _preferences.resource_path_translations, and there needs to be a ChangeResourcePathTranslation action.
> 
> This also needs to be invoked at after the directory argument that is currently handled by get_prefixes is applied - that should be considered part of the path. (If that's not obvious, let me know and I can do that part.) Other than that, if you can draft a changelog entry - I'd appreciate it.

Done. Let me know if there's something needed from me.

renpytom:
Looks good to me. I'll merge it in a moment. 

Only suggestion is that I don't love the commit messages - I'd suggest something like: 

translation: Add resource path mappings. 
translation: Simply resource path mapping documentation.
translation: Honor directory in resource path mappings. 
translation: Document resource path mapping arguments.

But that's a minor thing.

Issue #7057, opened 2026-05-11
Allow expressions in test statements


arsym-dev:
A big oversight with the current test framework is that many test command expect literals even though expressions would be useful. This was seen in #6957.

I propose modifying test statements to use expressions natively. Many commands already support expressions, but not all.

## Behavior
Currently you're forced to use literals in many commands:
```renpy
keysym "K_BACKSPACE" repeat 10
```

To support the use of parameters and loops, I suggest allowing the following:
```renpy
$ keyval = "K_BACKSPACE"
$ num = 10
keysym keyval repeat num
```

## Affects
Commands:
- keysym [str]
- type [str]
- drag button [int] steps [int]
- skip fast [bool]
- repeat [int]

Selectors:
- screen [str]
- id [str]
- layer [str]
- label [str]

## Incompatible change: Label selector
Labels selectors would be changed to string literals rather than the current "naked" form.

The currently valid statement `if label chapter_1:` would become `if label "chapter_1":`

This change allows the use of:
```renpy
$ label_name = "chapter_1"
if label label_name:
```

## Incompatible change: Expression -> Text
The commit 3635af24c53e085fe9d8737f108f51fd3ca22db1 was used to fix #6957. It allowed the use of `click expression <expr>`, but this masks the fact that this is a text selector, not a general expression. For example, `if expression <expr>` is valid, but it's looking for text rather than evaluating a general expression.

I suggest renaming it to `text <expr>`.

The previous statements become `click text <expr>` and `if text <expr>`. This also makes it so that `click "Start"` and `click text "Start"` are equivalent.

The commit in question is on master and hasn't been released yet. @ChummyChonka, you might be interested in this discussion.

## Discussion
If these two incompatible changes are okay, I can create a PR to implement this proposal. Otherwise I'm open to suggestions on how to minimize the impact.

arsym-dev:
Any opinions on the breaking changes outlined here? Namely the label selector since it's been out since 8.5.0.

I can leave the second one (renaming `expression` to `text`) for another time, it's mostly a semantic thing, but should be clarified sooner rather than later to avoid entrenching it in 8.6.0.

Having to use literals is becoming a bottleneck in my projects so I'd like to get clarity on this soon if possible.

renpytom:
[...]

PR #7277, opened 2026-08-17
Test: Expressions allowed in all statements

arsym-dev:
Addresses #7057

- All numeric/string properties in test statements accept expressions now. These are evaluated when the statement is first executed in the given test.
- Renamed `expression` to `text` as per discussion (e.g. `click expression button_name` is now `click text button_name`)
- Label selector accepts expressions with backwards compatibility (valid: `label chapter1`, `label "chapter1"`, `label label_name`)
- Added test coverage for the above
- Slight reorganization of Tutorial tests

### DIFF
--- a/renpy/test/testast.py
+++ b/renpy/test/testast.py
@@ -1141,7 +1141,7 @@ class RepeatCounter(Condition):
 
     def ready(self):
         self.value -= 1
-        return self.value == 0
+        return self.value < 0
 
 
 class Pass(Node):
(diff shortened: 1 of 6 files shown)


renpytom:
Can you expand the changelog entry? I'm finding it a bit hard to understand on its own. And given that we've had quite a few changes to tesing in 8.6, maybe add a Testing section between Features and Other Changes, so it's easier to scan.

arsym-dev:
Updated

PR #7289, opened 2026-08-21
Fix SDL3 high-DPI window sizing on startup

yorkyang2333:
User attachment

Screenshot 2026-08-21 at 15 55 03

On macOS Retina displays, the initial Ren'Py window could open at roughly twice the intended size, and moving it between displays could resize it again. The regression was probably introduced by a816da259 ("Set up dpi_scale in gl2draw"), which applied the display content scale to SDL3 window dimensions even though SDL3 expects logical window pixels and reports drawable pixels separately.

This keeps initialization, resize, and saved window dimensions in logical pixels while retaining the drawable size for high-density rendering. It also avoids resizing the logical window when its display scale changes.

Existing Launcher users may need to reset the Launcher window preference manually once by deleting both ~/Library/renpy/launcher-4/persistent and /path/to/renpy-sdk/game/saves, because an old high-DPI-sized value can still be stored in launcher-4/persistent. After that reset, the corrected default size will be used.

Please help me test whether this PR would cause regressions on other OSes. Thanks!

### DIFF
--- a/renpy/gl2/gl2draw.pyx
+++ b/renpy/gl2/gl2draw.pyx
@@ -177,16 +177,15 @@ cdef class GL2Draw:
         vwidth, vheight = self.virtual_size
         virtual_ar = 1.0 * vwidth / vheight
 
+        # SDL3 window sizes are logical pixels, even when the window has a
+        # high-density drawable. The drawable size is handled separately
+        # below by draw_per_phys.
         # The requested size.
         pwidth, pheight = physical_size
 
         if pwidth is None:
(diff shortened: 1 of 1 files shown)


renpytom:
This is wrong - the behavior differs between Windows and macOS. I'll look at an improved fix.

yorkyang2333:
> This is wrong - the behavior differs between Windows and macOS. I'll look at an improved fix.

Thanks! Turned this into issue #7292. Closing for now.

PR #7221, opened 2026-08-03
Add statement information API

kyui-azusa:
## Summary

- add `renpy.get_statement_info()` and `renpy.StatementInfo`
- expose the filename, line number, and translation identifier for executing and predicted statements
- track the active prediction node with exception-safe restoration
- document the API and cover execution, prediction, tlid propagation, and prediction failure cleanup

## Motivation

Creator-defined statement prediction callbacks currently have no supported way to determine which statement Ren'Py is predicting. Existing helpers only report the statement being executed by the context, so they can return unrelated information during prediction.

The new API uses the current execution node normally and the active prediction worklist node while `node.predict()` is running. It also uses the translation identifiers already carried by prediction state.

## Validation

- `./run.sh testcases test 'statement_info'` (1 testcase, 12 assertions passed)
- `ruff check --select E9,F63,F7,F82` on changed Python modules
- `ruff format --check` on changed Python modules
- `python -m compileall` on changed Python modules
- Sphinx dummy build (succeeded; 4 pre-existing warnings)
- `git diff --check`

Fixes #7064

PR #7294, opened 2026-08-21
Preserve compiled ATL blocks when copying transforms

### LINKED ISSUE #7283: Spurious exception "NameError: name 'parameter' is not defined" when setting image property inside label with call parameter value then `show`ing again outside call
**Version**
Ren'Py 8.5.3.26051504

**Repro script.rpy**

```rpy
image white = Solid("#ffffff")

define e = Character("Eileen")

label start:
    call .initial_state(_alpha=0.0)

    show white

    e "Above should cause error: _alpha is undefined"

    return


label .initial_state(_alpha):
    show white:
        alpha _alpha
    return
```

**Error**

```
I'm sorry, but an uncaught exception occurred.

Compiling ATL code at game/script.rpy:18
  File "game/script.rpy", line 8, in script
    show white:
  File "game/script.rpy", line 17, in <module>
    alpha _alpha
          ^^^^^^
NameError: name '_alpha' is not defined. Did you mean: 'Alpha'?

-- Full Traceback ------------------------------------------------------------

  File "game/script.rpy", line 8, in script
    show white:
  File "renpy/ast.py", line 1364, in execute
    show_imspec(self.imspec, atl=getattr(self, "atl", None))
    ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  ...
  File "game/script.rpy", line 17, in <module>
    alpha _alpha
          ^^^^^^
NameError: name '_alpha' is not defined. Did you mean: 'Alpha'?

Windows-11-10.0.26200-SP0 AMD64
Ren'Py 8.5.3.26051504
MRP_ATL_interpolation_outside_call_undefined_var 1.0
```

**Expected**

I'm not using `_alpha` at all in the outer `show` statement so there shouldn't be an issue with it being undefined in outer scope.

**Note**

At first I got the error with:

```
    show white:
        alpha 1.0
```

on second `show`, but I noticed that even when removing `alpha`, just `show` would try to access the inner `_alpha` variable!

There is no issue if I'm using some local variable `_alpha`, error only occurs when using a parameter value. Somewhat it magically refers to the `show` statement inside the label and doesn't know `_alpha` at that point.

**Workaround**

I can define a dummy ` $ _alpha = 0` just to avoid the error, but really the value doesn't matter here.

pederbe:
Fixes #7283

## Summary

- preserve the compiled ATL block when making an exact transform copy
- avoid re-evaluating inline ATL after its label-local parameters leave scope
- add regression coverage that checks both successful replacement and preserved transform state

## Testing

- `renpy.py testcases test 'atl::inline_parameter_scope' --hide-execution all`
- `renpy.py testcases test camera_transform --hide-execution all`
- `renpy.py testcases test menus_and_branching --hide-execution all`
- `renpy.py testcases test menus_and_branching_variables --hide-execution all`
- `renpy.py testcases test label_control_flow --hide-execution all`
- `renpy.py testcases test textshaders --hide-execution all`
- `renpy.py testcases lint` (completed with existing testcase-project warnings)

### DIFF
--- a/renpy/display/transform.py
+++ b/renpy/display/transform.py
@@ -913,6 +913,11 @@ class Transform(Container):
         d = self()
         d.kwargs = {}
         d.take_state(self)
+
+        if isinstance(self, ATLTransform):
+            assert isinstance(d, ATLTransform)
+            d.block = self.block
+
         d.take_execution_state(self)
         d.st = self.st
         d.at = self.at
--- a/testcases/game/tests/test_atl.rpy
+++ b/testcases/game/tests/test_atl.rpy
@@ -1,0 +1,45 @@
+# This file tests ATL transforms and replacement behavior.
+
+# ==============
+# == Fixtures ==
+# ==============
+
+image atl__inline_parameter_scope__solid = Solid("#fff")
+
+label atl__inline_parameter_scope__start:
+    call .show(_alpha=0.0)
+
+    show atl__inline_parameter_scope__solid
+
+    "The image was replaced without re-evaluating the previous ATL block."
+    return
+
+label .show(_alpha):
+    show atl__inline_parameter_scope__solid:
+        alpha _alpha
+    return
+
+
+# ==============
+# === Tests ====
+# ==============
+
+testsuite atl:
+    after testcase:
+        if not screen "main_menu":
+            run MainMenu(confirm=False, save=False)
+
+    testcase inline_parameter_scope:
+        description "Does not re-evaluate replaced inline ATL outside its parameter scope"
+        # GH-7283: Replacing an image after returning from a parameterized label must not
+        # evaluate the old inline ATL block after the label parameter has left scope.
+        # https://github.com/renpy/renpy/issues/7283
+        run Start("atl__inline_parameter_scope__start")
+        pause 0.1
+
+        python:
+            entry = renpy.game.context().scene_lists.get_displayable_by_tag(
+                "master", "atl__inline_parameter_scope__solid"
+            )
+            assert entry is not None
+            assert entry.state.alpha == 0.0


Andykl:
Hi. Your last submissions looks like LLM generated. Could you please tell if you use one and if so does you, as a human, actually tested your submission? I don't want to alienate LLM usage, but knowing how much human effort was put into the change is essential to its evaluation. 
See #7221 for similar case.

pederbe:
Hey. Yes, I used Codex to help on this issue. I spent around 3 hours myself being assisted by Codex. I'm currently learning Renpy and trying to help where I can. In this case I tested everything myself and had to go back several times to get this to work and pass the tests as intended. 

I started with the nightly branch and did everything then realized I had to switch to the fix branch and that's different in several ways. So my testing/debugging took most of the time.

Is that kind of usage of LLMs acceptable for Renpy? I don't want to step on anyone's toes.

Andykl:
We have a section in readme about AI-assisted submitions, tldr it's ok as long as you fully understand the changes and basically responsible for it is if you wrote it yourself. 

I personally don't like when comments/PR description is written by LLM. I prefer broken English or message written in native language that I translate myself, because LLM could hallucinate during translation, and original intent is lost on the other side, but if it happens on my side, I can use different translator or ask for clarification. 

Also the volume could be an issue. A lot of submissions in short period of time overwhelm and disincentive to respond, because it feels like review take more effort than a submission. 

You could join renpy discord server where we have a channel dedicated for engine development where it's easier to coordinate.

mal:
Unfortunately I don't have time to dig into this, but a very similar change last year (https://github.com/renpy/renpy/commit/0e71cc4af58cc503853012f34f919bc904ff8110) ended up necessitating a slew of follow-up fixes (https://github.com/renpy/renpy/commit/c822255ae3a24f1afd2ddb1bcb00ad3947f2f27f, https://github.com/renpy/renpy/commit/d16f25602bfec0201a18bb22884bcbc66bd453d4, https://github.com/renpy/renpy/commit/92b5a2b35b5e985b03de7e7066f789db44396306). It's hard to say if the same problems would arise here, but I'd suggest it probably should be looked at in detail before taking it forward.

PR #7297, opened 2026-08-22
Fix doubled words in the documentation

zaveshaa:
Fixes 23 doubled-word typos across 19 files in `sphinx/source/`, found with a mechanical scan and confirmed by hand in context. Code samples (e.g. `size size`, `pos pos`, `delay delay`) and changelog entries were deliberately left untouched.

Includes one adjacent typo on the same line: "prevents thart behavior" → "prevents that behavior" (`transforms.rst`, `unique()` method docs).


Andykl:
Thank you

PR #7305, opened 2026-08-24
Fix ECDSA signature encoding and EC private key structure in save token signing

Andykl:
This PR fixes the spurious rejection of legitimate saves.
The exact fixes are:
1. In ECSign: non-canonical DER to raw conversion. Now it uses a real TLV walk (GetNumberFromDER)
2. Opposite non-minimal INTEGER encoding in AddNumberToDER. Now it correctly strips leading zeros.
3. EC private keys: SEC1 vs PKCS#8: `OSSL_ENCODER_CTX_new_for_pkey(pkey, EVP_PKEY_KEYPAIR, "DER", NULL, NULL)` leaves the output structure unspecified, OpenSSL's default for EC is the type-specific SEC1 ECPrivateKey. `crypto.subtle.importKey` has no SEC1 format — the web build requests "pkcs8" and throws DataError with an empty message. The fix now names "PrivateKeyInfo" explicitly and ToPkcs8 converts old SEC1 to PKCS#8.

This leaves saves signed by the old ECSign incorrect, so ~0.43% of users have permanently unloadable saves. I don't know how to handle this correctly. Probably pass saves upgrade when that's the case.

The analysis and the fix were made by Claude. I see the bug is fixed, but have limited knowledge in cryptography so could miss something.
@renpytom @bkats.

Also less related question. Is there a special reason to keep this code C? It was hard to wrap my head around. I was thinking to embed it into `renpy/ecsign.pyx` and `renpy/common/_ecsign.js` so it is easer to read.

### DIFF
--- a/src/ec_sign_core.c
+++ b/src/ec_sign_core.c
@@ -40,6 +40,7 @@
 
 static EVP_PKEY *GetKeyFromDer(int public, const unsigned char *key_der, size_t key_len);
 static int AddNumberToDER(char *der, int offset, const char *number);
+static int GetNumberFromDER(const unsigned char *der, size_t der_len, size_t *offset, char *number);
 static int Sign(EVP_PKEY *priv_key, const void *data, size_t data_len, void **sign, size_t *sign_len);
 static int Verify(EVP_PKEY *pub_key, const void *data, size_t data_len, const void *sign, size_t sign_len);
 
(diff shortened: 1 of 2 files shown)


renpytom:
The answer is no, not really. 

I'll be honest, this isn't really performance critical, and it's not even that security critical in the way that matters - it's used to protect people from being idiots who download hacked saves from f95, and not for traditional security.  I'm wonderijng if it makes sense to just go back to ecdsa.py (which we used in 8.5), and just embed it into Ren'Py so we don't trigger the CVE detector.

Andykl:
I completely forgot that there were no public release with new signing code, going back to ecdsa.py I think make sense. Maybe we could distill only needed parts out of it as pure python and stash them into savetoken, it's not that much code and is fairly fast for amount of signing we do.

bkats:
I apparently missed the leading zero issue (I only human, sorry). But at quick look the new code seems okay.

I didn't bother to make the pkcs8 conversion for web, don't expect import/export save file between a desktop version and a web version of the same game to be done that often and I think that actually worked. The problem was I think it  would create a second key once and then everything just works. Since the web version doesn't verify there was no real problem there. I think I mentioned that back when I finished the code. But now the conversion is made that might help.

The reason for the separate c-file was I develop the code is a small test project in pure c. Was quicker to tackle the gruesome openssl problems (documentation isn't great). Once the code worked I just only wrote than a pyx wrapper to prevent introducing a change for stupid mistakes.

I typed this quickly so I hope I didn't make to many grammar mistakes.

The decision to use this code or python ecdsa lib is not mine to make. Both have pros and cons. I offered a solution that can replace it, but is it really needed? I don't really know. My opinion about the security risk of original ecdsa code being low for use in renpy  isn't changed.

renpytom:
Obsoleted by #7327 .

Issue #7311, opened 2026-08-25
lint: option to write the report as JSON


wgu9:
Right now the only way to get lint results out of Ren'Py is the text report. That works fine for reading in an editor, but anything that wants to act on the results (the VS Code extension, a CI job, a pre-commit hook, a script that runs after `translate`) has to scrape lines that look like

```
game/script.rpy:42 The image 'eileen happpy' is not defined.
```

and hope the format never changes. The VS Code extension currently runs `lint lint.txt` and just opens the file, because there isn't anything better to do with it.

I'd like a `--format json` option on the lint command that writes the same information in a structured form (`--json` on its own is out, since argparse sees it as an ambiguous abbreviation of `--json-dump`):

```
./renpy.sh myproject lint lint.json --format json
```

giving something like

```json
{
  "version": "8.6.0",
  "error": true,
  "problems": [
    {"file": "game/script.rpy", "line": 42, "message": "The image 'eileen happpy' is not defined."}
  ],
  "notes": ["..."],
  "statistics": {"languages": [...], "menus": 2, "images": 0, "screens": 23}
}
```

`report()` in `renpy/lint.py` already knows the file and line for every problem, so this is mostly collecting what's currently printed instead of printing it. The text report stays the default, and `--error-code` keeps working the same way in both modes.

I've got this working locally and will put up a PR.

Andykl:
There are several issues with current lint. I don't want to lock-in json format before we settle on a new, documented interface of the module.

wgu9:
Fair enough. Two thoughts.

The CLI page already says the CLI isn't a stable interface, so I'd be fine with the JSON being documented the same way (keys may be added, existing ones kept where practical) rather than as a contract. The alternative is that tools keep scraping the text report, which is already a de-facto interface, just a worse one.

#7314 only touches the four places that print problems (`report`, `add`, `problem_listing`, the Python warnings), so it should carry over to a reworked module without much trouble. If you'd rather settle the interface first, what shape are you thinking of: a report object that formats itself, hooks getting handed a reporter, something else? I'm happy to rework #7314 on top of that once there's a direction, or park it until it lands.

wgu9:
Closing this proposal for now. The existing text output is workable, and I do not want to push a JSON contract before the lint interface direction is settled.

PR #7314, opened 2026-08-25
Add a --format json option to the lint command

wgu9:
Draft status: parked pending the lint interface direction in #7311. The maintainer does not want to lock a JSON schema before that interface is settled, so this is not requesting review.

The prototype adds --format json while preserving the text report. Once the repository settles on a documented report or module interface, this branch can be adapted to that design instead of making the current lint internals a contract.

Related: #7311.

### DIFF
--- a/renpy/lint.py
+++ b/renpy/lint.py
@@ -21,7 +21,9 @@
 
 import time
 import re
+import io
 import sys
+import json
 import collections
 import textwrap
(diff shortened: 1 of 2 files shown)


wgu9:
I am closing this draft rather than asking for review while the lint interface is still being reconsidered. Thanks for the direction on the issue.

PR #7315, opened 2026-08-25
Explain the error when flow falls through into a label with parameters

### LINKED ISSUE #5166: Feature request: Debug: improve error message when entering label with parameters due to missing return in previous label

wgu9:
Fixes #5166.

When a label with required parameters is reached by falling off the end of the label above it, the only error you got was:

```
TypeError: missing a required argument: 'duration'
```

pointing at the label that was entered. The label that's actually wrong is the one above it, which is missing a `return` or `jump`, and nothing in the traceback says so. hsandt's example in the issue is the typical case: a few utility labels in a row, one of them without a `return`, and the error lands on the next one.

With this change, that case now reads:

```
TypeError: missing a required argument: 'duration' - label 'ft_helper' takes parameters, but was reached by falling through from the statement before it, not by a call or jump. Check that label 'ft_start', above it, ends with a return or jump.
```

How it decides: in `Label.execute`, if applying the arguments raises `TypeError`, and `_args`/`_kwargs` are both `None`, and `context.last_abnormal` is false (nothing jumped or called into this label), it's a fall-through. `jump`, `call`, and `renpy.call()` all set `abnormal`, so a `jump` to a parameterised label without arguments keeps the original message. It's still a `TypeError`, so anything catching that is unaffected.

"The label above it" is the closest `Label` node earlier in the same file, found from `script.all_stmts` only on the error path. That's a heuristic, but since every file ends in an implicit `return`, fall-through can't cross files, so it's right in practice. It handles local labels (`ft_local.part_two`). If no label precedes it, the message just says to check the statements above.

Tested by running a small project with `RENPY_SIMPLE_EXCEPTIONS=1` through five scenarios: fall-through (new message), fall-through from a local label (new message naming `outer.local`), `jump` with no args (old message), `renpy.call()` with no args (old message), and a normal `call` with args (works). Nothing changes when `config.developer` is off, since argument errors are ignored there anyway.

### DIFF
--- a/renpy/ast.py
+++ b/renpy/ast.py
@@ -1172,7 +1172,13 @@ class Label(Node):
 
         renpy.game.context().mark_seen()
 
-        values = apply_arguments(self.parameters, renpy.store._args, renpy.store._kwargs)
+        try:
+            values = apply_arguments(self.parameters, renpy.store._args, renpy.store._kwargs)
+        except TypeError as e:
+            if self.reached_by_fall_through():
+                raise TypeError(self.fall_through_message(e)) from None
+
+            raise
 
         renpy.exports.dynamic(**values)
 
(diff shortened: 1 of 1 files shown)


wgu9:
Replaced by #7322. The original PR targeted `master`, but this is a bug fix and belongs on `fix`. The replacement also limits the enhanced error to labels with required parameters, so a TypeError from an optional default expression is not misreported as fall-through.

Issue #6676, opened 2025-10-17
Interactive visual flowchart


jaimebarrancos:
Hey, I would like to discuss a cool feature. Here goes:

**Description:**
A interactive visual flowchart tool (as a VSCode extension) that automatically maps out all the branching paths in your game and lets you better manage and organize the story structure visually.

**Core features**
- Automatically map choices of existing game
- add scenes/choices to the diagram, even if not implemented (useful for organizing and planning)
- jump to code of scene

**Discussion**
I would like to discuss the feasibility, problems and added value of something like this.
How difficult do you think it would be? What would be the main problems? And finally, would this help?

I know there are some existing tools for drawing static diagrams, but those require extra setup, and are very limited, since they are view only.
I also believe that this would serve as an easier entry to new users as it would give some structure and guidance to the game making process.

**Implementation**
The only way I thought of making this would be as a VSCode extension, since the diagram needs to be interactible. What alternatives are there? Would it be possible to put it in renpy itself?

bober-gith:
I think it's a case where it might be easy to do for simpler games, but those games won't need it because they are simple enough you don't need such a tool; and extremely difficult to have it reliable for complex games, the ones that would need this type of tools. I think it would require the tool to have a sort of 'intelligence', to understand all the jump labels and output this as something readable and useful. Might be possible but a lot of work for sure.

renpytom:
@jaimebarrancos Are you looking at being the one to develop this yourself, and add it to Ren'Py? Or is this something you want us to add? 

My feeling is it's at least possible, for simple to medium games, and it's potentially valuable for certain developers. It might make sense to have as a second VSCode extension you maintain, with Ren'Py exporting information about the static flow graph. (Say, what labels are sure to jump each other, and which might.) 

I don't know about how to actually maker the extension - I'm poor at VSCode, though I hope to learn soon. 

I agree with @bober-gith  - it is going to be a lot of work, but if it exists it could be a nice addition. (That being said, there are several other tools that compete in the graphical frontend to Ren'Py space.)

jaimebarrancos:
> Are you looking at being the one to develop this yourself, and add it to Ren'Py? Or is this something you want us to add? 

I don't know yet, I wanted to discuss it first to see if it's worth it or if there's some other idea I hadn't had.

What other tools compete with renpy in the graphical interface?
I would like to have a look.

Thanks.

bober-gith:
I think Twine is the most popular one for visual novels.

[...]

PR #7318, opened 2026-08-25
Include the label flow graph in the --json-dump output

### LINKED ISSUE #7312: Include label flow (jump/call/menu targets) in the --json-dump output
#6676 and a few external tools (the flowchart extension in that thread, the graphviz generator, the IDE @ghirpara mentioned) all end up re-parsing `.rpy` files themselves to work out which labels jump or call which. That's fragile. Each of them has to reimplement enough of the parser to deal with nested menus, `call ... from`, `jump expression`, labels inside `if` blocks, `menu` captions, creator-defined statements, and so on, and each of them silently gets a different answer on the things it doesn't know about.

Ren'Py already has all of this once the script is loaded. I'd like `--json-dump` (or a separate command, if that's cleaner) to include a flow section, roughly:

```json
"flow": {
  "start": {
    "file": "game/script.rpy",
    "line": 10,
    "jumps": ["chapter_1"],
    "calls": ["intro"],
    "menus": [
      {"line": 14, "choices": [{"label": "Go left", "jumps": ["left"], "calls": []}]}
    ],
    "falls_through": "chapter_1",
    "dynamic": true
  }
}
```

`dynamic` would mark labels that contain `jump expression` / `call expression` / `renpy.jump()` in Python, so a tool knows the static edges are incomplete there and can draw it differently instead of pretending it's complete.

Tom suggested in #6676 that Ren'Py could export the static flow graph and let editor tools consume it. This is the concrete ask for that. It doesn't have to be perfect; it just has to be the same data Ren'Py itself uses, so the tools stop guessing. Together with #7311 it would give editor tooling everything it needs from the engine without touching the parser.

wgu9:
Fixes #7312. The editor-tooling background is in #6676.

This adds a `flow` section to `--json-dump`, keyed by label name. Each record contains the label location, static jump and call targets, menu choices, the next label reached by fall-through, and a `dynamic` flag when the destination cannot be resolved without running the game.

The collector follows Ren'Py's parsed statements through `if`, `while`, translation blocks, menus, and creator-defined statement sub-blocks. Python blocks are parsed with Python's AST. Literal `renpy.jump("label")` and `renpy.call("label")` calls become static edges; comments, strings, and calls inside a function declared by the block do not. Non-literal targets and context-changing flow calls set `dynamic=true`.

The existing `--json-dump-private` and `--json-dump-common` filters also apply to flow records.

Validation with the 8.6 nightly:

- `the_question`, `testcases`, nested menus, local labels, conditional choices, dynamic jumps, Python calls, and creator-defined statements were exercised
- a focused probe confirmed that a real literal jump is exported while the same text in a comment, string, or nested function is ignored
- non-literal Python targets set `dynamic=true`
- the generated dump remains valid JSON
- Ruff and formatting checks passed for the changed code, excluding the file's existing compatibility-import warnings

Draft while I review the output semantics and final diff.

### DIFF
--- a/renpy/dump.py
+++ b/renpy/dump.py
@@ -27,9 +27,9 @@ from __future__ import division, absolute_import, with_statement, print_function
 from renpy.compat import PY2, basestring, bchr, bord, chr, open, pystr, range, round, str, tobytes, unicode  # *
 
 
+import ast as python_ast
 import inspect
 import json
-import re
 import sys
 import os
 
@@ -57,12 +57,13 @@ def file_exists(fn):
     return rv
 
 
-# Matches renpy.jump("label") and renpy.call("label") in Python code, so the
-# targets can be reported as static edges.
-python_flow_re = re.compile(r"""renpy\.(jump|call)\(\s*(["'])([^"'\n]+)\2""")
-
-# Matches uses of the flow functions that couldn't be resolved statically.
-python_dynamic_re = re.compile(r"renpy\.(jump|call|jump_out_of_context|call_in_new_context|invoke_in_new_context)\b")
+python_flow_functions = {
+    "jump",
+    "call",
+    "jump_out_of_context",
+    "call_in_new_context",
+    "invoke_in_new_context",
+}
 
 
 def new_flow():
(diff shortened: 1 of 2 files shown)


wgu9:

Issue #6593, opened 2025-08-28
Feature Request / Discussion: Add web interface to emit dialogue events


danielhammerl:
**TL;DR**

Add an optional HTTP/TCP interface that emits dialogue events whenever say is called, so external clients (e.g. TTS engines) can react to in-game text.

**Feature Request**
I’d like to propose adding an optional interface in Ren’Py that allows dialogue events to be exposed over HTTP or a TCP socket. Specifically, whenever the say function is called, Ren’Py could send the relevant data (speaker, text, etc.) to a configured endpoint.
Motivation / Use Case

My personal motivation is to connect Ren’Py games with external Text-to-Speech clients. I would love to see a voiced version of some games which probably never will get some, because voice over takes a lot of effort and budget, but thanks to technical advances in AI / TTS, this is no longer necessary. Such clients need to know who is speaking and what they are saying in real-time in order to generate and play audio. I tried achieving this with some kind of OCR tool, but this is not really practical due to various reasons.

While TTS is my own use case, this feature could benefit a variety of external integrations. Theoretically you could make this interface bidirectional for live translation, but this is out of scope here.

**Implementation Thoughts**
- Provide an optional configuration to enable the interface, port should also be configurable. Default setting is off, because of data privacy reasons, this is a niche feature addressed to tech savvy users.
- On every say, emit a structured JSON object like: `{ "speaker": "Alice", "text": "Hello world." }`
- Keep it non-blocking, should not interfere with normal game flow

I would love to try implementing this, but before I’d mainly like to open a discussion here about whether such an extension makes sense for Ren’Py, and if so, what the best design approach might be. Please let me know what you think about.

renpytom:
So, first off, just to check - you do know about the built in self-voicing that you can get to via the accessibility (shift+A) menu right? That includes clipboard voicing, that you can get now. 

But I'll admit, this is better. I think we'd want streaming JSON through a TCP socket, and probably it would be something like:

```json
[
{ "id": 1234567, "kind": "text", "text": "Eileen", "role": "who" },
{ "id": 1234568, "kind": "text",  "text": "Hi! My name is Eileen, and I'd like to welcome you to the Ren'Py tutorial!", "role": "what" 
},
{"id": 1234569, "kind": "text", "text": "Back", "role": "quick_menu" },
# ...
]
``` 

[...]

PR #7319, opened 2026-08-25
Add a text event stream for external tools (TTS, live translation)

wgu9:
This is a first pass at #6593, along the lines Tom sketched there: a TCP daemon streaming JSON, a gathering pass like the self-voicing one, and roles on the text. Two things differ from the sketch, explained below.

**Enabling it.** `define config.text_events_port = 4444` for a creator who wants it on, or `RENPY_TEXT_EVENTS_PORT=4444` in the environment for a player or tool author who wants it on for a game that doesn't set it. Off by default. Ren'Py listens on 127.0.0.1 only, and only writes; clients never send anything.

**What comes out.** One JSON object per line:

```
{"kind": "interaction", "image_tag": null, "id": 20, "interaction": 4}
{"kind": "text", "role": "quick_menu", "style": "quick_button", "text": "Back", "id": 21, "interaction": 4}
{"kind": "text", "role": "who", "style": "say_label", "text": "Sylvie", "id": 29, "interaction": 4}
{"kind": "text", "role": "what", "style": "say_dialogue", "text": "Hi there! How was class?", "id": 30, "interaction": 4}
{"kind": "text", "role": "choice", "style": "choice_button", "text": "To ask her right away.", "id": 127, "interaction": 14}
{"kind": "focus", "role": "choice", "style": "choice_button", "text": "To ask her later.", "id": 130, "interaction": 14}
```

`interaction` goes up when the set of text on the screen changes (interaction restarts that don't change the text send nothing). `focus` records go out when the focus moves. Text is what self-voicing would say: tags stripped, `{alt}` honoured, `alt` style property honoured.

**Where roles come from.** The sketch had a `role` style property. Adding a style property means touching `generate_styles.py` and rebuilding the Cython, which I can't verify in this environment, and it also means every screen has to be annotated before a tool sees anything. So for now the role is derived from the style: `config.text_events_roles` maps style names to roles, and the lookup walks up the style's parents. The default map covers the default screens (`say_label` -> `who`, `say_dialogue` -> `what`, `choice_button` -> `choice`, `quick_button` -> `quick_menu`, `navigation_button`, `input`, `notify`, and the nvl/bubble equivalents). Because it walks parents, `the_question`'s `say_thought` (which inherits `say_dialogue`) reports `what` with no configuration. The raw style name is sent alongside for tools that need more. A `role` property could be added later and take precedence; the map would then be the fallback.

**The pass.** `renpy/display/textevents.py` walks from `renpy.display.tts.root` mirroring `_tts_common`: front-to-back layer order (`config.tts_front_to_back`), screens in OLD/HIDE phase skipped, modal screens drop what's behind them, `NearRect` without a rect skipped, `alt` replaces children, `TTSDone` stops the walk. The one deliberate difference from self-voicing: buttons and bars are included in the screen pass (via `_tts_all`, so `alt` and "selected" work) rather than only when focused, because a TTS or translation tool needs to see the choices as part of the screen. It's hooked in `focus.py` right after `tts.displayable()`, so it runs whenever self-voicing would.

**Not blocking the game.** Records go on a queue and a thread does the `sendall`; sockets get a 5 s timeout and a client that stops reading is dropped.

Tested by running `the_question`'s `bad_end` and `good_end` testcases with a small client attached (708 records on `good_end`; main menu navigation, quick menu, narrator/`say_thought`, `Sylvie`/`Me` dialogue, both choices, and focus changes all come through with the expected roles; nothing logged to `log.txt`). Docs: a "Text Events" section at the end of `self_voicing.rst`, and the two config variables in `config.rst`.

### DIFF
--- a/renpy/__init__.py
+++ b/renpy/__init__.py
@@ -495,6 +495,7 @@ def import_all():
     import renpy.display.predict
     import renpy.display.emulator
     import renpy.display.tts
+    import renpy.display.textevents
     import renpy.display.gesture
     import renpy.display.model
     import renpy.display.quaternion
--- a/renpy/config.py
+++ b/renpy/config.py
@@ -1402,6 +1402,31 @@ extra_savedirs = []
 # The text-to-speech dictionary. A list of [ (RegEx|String, String) ] pairs.
 tts_substitutions = []
 
+# If not None, the port on 127.0.0.1 that the text events server listens on.
+# The RENPY_TEXT_EVENTS_PORT environment variable overrides this.
+text_events_port = None
+
+# A map from style name to the role reported for text using that style (or a
+# style descended from it) in the text event stream.
+text_events_roles = {
+    "say_label": "who",
+    "say_dialogue": "what",
+    "nvl_label": "who",
+    "nvl_dialogue": "what",
+    "nvl_thought": "what",
+    "bubble_who": "who",
+    "bubble_what": "what",
+    "choice_button": "choice",
+    "choice_button_text": "choice",
+    "quick_button": "quick_menu",
+    "quick_button_text": "quick_menu",
+    "navigation_button": "navigation",
+    "navigation_button_text": "navigation",
+    "input": "input",
+    "input_prompt": "input_prompt",
+    "notify_text": "notify",
+}
+
 # The base URL where unpacked web videos can be found.
 web_video_base = "./game"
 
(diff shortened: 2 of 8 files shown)


renpytom:
@wgu9 How much review did you do of this and your other pull requests before submitting them? One of our big limitations is human reviewer time - and with so many pull requests being emitted in such a short time, I'd like to know how much of this is your work, versus how much of  this is AI. 

It's important that we have a human in the loop that understands Ren'Py well enough to say "I believe this is good code."

wgu9:
Thanks for the feedback. I am closing this and stepping back from the related drafts so I do not add to the review burden.

PR #7320, opened 2026-08-25
Add dev_setup.py to set up a checkout from a nightly build

### LINKED ISSUE #7227: Improve developer setup workflow for new users and Windows users
## Summary
Setting up a development environment on Windows is unintuitive. Users have to manually download nightly libs, unpack them, and know about internal build steps. Switching between the fix and master branches with different SDK versions adds to the friction.

The reason for opening this issue is that I was running master branch from a nightly SDK on Windows, but I got `ModuleNotFoundError: No module named 'renpy.styledata.stylesets'`. The solution is run `uv run setup.py generate` which is already handled by `run.sh` for Linux developers, but a Windows dev does not have an equivalent.

## Proposed solution
Since uv gives us Python, the actual logic could be a Python script with .bat/.sh wrappers. The .bat/.sh are just so new devs on a new machine can double click/execute them and it would initiate the setup. It would be the first stop that handles everything.

Create a `dev_setup.sh`/`dev_setup.bat` that:

1. Checks if uv is installed; if not, prints instructions OR confirms with the user if they want the script to install uv for them.
2. Run `dev_setup.py` (or similar)

Create a `dev_setup.py` that:

1. Determines the current branch (master or fix) to pick the correct nightly.
2. If on a non-standard branch, asks the user which branch they're targeting.
3. Downloads and unpacks the latest or cached nightly SDK.
4. Runs `uv run setup.py generate`.
5. Passes remaining arguments to Ren'Py.

## Example flow
After running the `.sh`/`.bat` the user sees the following:

```
uv not detected on this machine. 

More information:
https://docs.astral.sh/uv/getting-started/installation/

Install uv? [y/N]
```

After installing uv:
```
Detected master branch.
```

```
Not on main branch (master or fix). Which would you like to develop for?

[*] Master branch (8.6.0, default)
[ ] Fix branch (8.5.4)
```

After detecting the branch:
```
Choose SDK source

[*] Download latest nightly
[ ] Build from source [Wouldn't start with this, but maybe long term]
[ ] Use cached SDK

Cached:
[*] Return to SDK selection
[ ] 8.6.0.26080208+nightly (2026-08-02) [Detected from a cache, correct for the target branch]
[ ] 8.6.0.26080301+nightly (2026-08-03)
```

After selecting SDK
```
Success! You can run Ren'py using the following commands:

./renpy.exe launcher
./renpy.sh launcher
```

## Related
@Andykl is working on devcontainers for renpy and [renpy_build](https://github.com/renpy/renpy-build/), but for the vast majority of devs, they only need to edit Pyth/Ren'Py code and don't need the full Cython build. Downloading the latest nightly, unpacking it, and running a few setup commands should be sufficient for most users.

wgu9:
Fixes #7227.

This adds `dev_setup.py`, with `dev_setup.sh` and `dev_setup.bat` wrappers, so a new contributor can get a runnable checkout with one command and no toolchain:

```
./dev_setup.sh          # Linux, macOS
dev_setup.bat           # Windows
```

What it does, in order:

1. Reads the branch-to-version table out of `renpy/versions.py` and the current git branch, so `master` looks for 8.6 nightlies and `fix` for 8.5 ones (matching on major.minor, since the `fix` nightlies are numbered 8.5.4 while the table says 8.5.4 too, but that won't always line up). Unknown branches fall back to main's version; `--version 8.5` or `--nightly <name>` override it, and `--list` shows the candidates.
2. Fetches the index at nightly.renpy.org, picks the newest matching nightly (preferring ones without a `.dirty`/branch suffix), and downloads and unpacks it into `nightly/` (gitignored, reused on later runs). `--sdk DIR` skips this for an SDK you already have.
3. Links `lib` and `renpy.app` (symlink; junction, then copy, as fallbacks on Windows) and **copies** `renpy.sh` and `renpy.exe`.
4. Runs `scripts/generate_styles.py`, which is pure Python, so `stylesets.py` exists. `vc_version.py` isn't needed; `versions.py` already handles its absence.
5. Runs the SDK's own Python with `import renpy` from the checkout and checks the package came from the checkout, then prints how to run.

Point 3 is the part I'd draw attention to. The current `after_checkout.sh` symlinks `renpy.sh`, but the nightly's `renpy.sh` starts by resolving symlinks to find `ROOT`, so a symlinked `renpy.sh` runs the nightly's own `renpy/` package, not the checkout's. Nothing tells you. I lost a round of testing to this before noticing my changes had no effect, and it's probably part of why #7227 exists. `after_checkout.sh` now copies `renpy.sh` and `renpy.exe` instead, and README explains why for people doing it by hand.

Only the standard library is used, on purpose: this has to work before `uv sync` or a venv exists, on a machine that only has a system Python. It stays out of the way of `run.sh`, which is still the route for compiling modules.

Tested on macOS: `--list` against the live index for both 8.6 and 8.5; a fresh `git clone` plus `dev_setup.py --sdk <unpacked nightly>`; and a fresh clone with a full download (it picked `8.6.0.26082407+nightly`, ~160 MB), after which `./renpy.sh the_question lint` runs from the checkout. I don't have a Windows machine to try the `.bat` and junction path on; the code paths are small but they're untested.

Opened as a draft so it can be reviewed properly first.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,6 +61,9 @@ cache/
 # Libraries.
 /lib
 
+# Nightly builds downloaded by dev_setup.py.
+/nightly
+
 # Platforms.
 /rapt
 /rapt2
--- a/README.rst
+++ b/README.rst
@@ -56,7 +56,20 @@ like MSYS.
 Nightly Build
 -------------
 
-Nightly builds can be downloaded from:
+The quickest way to set up a checkout is to let ``dev_setup.py`` do it. From
+this repository, run::
+
+    ./dev_setup.sh        # Linux and macOS
+    dev_setup.bat         # Windows
+
+This works out which nightly matches the branch you're on, downloads and
+unpacks it into ``nightly/``, links its compiled libraries into the checkout,
+generates the style data, and checks that Ren'Py imports from the checkout.
+It only needs a Python 3 on the machine. Run ``python3 dev_setup.py --help``
+for the options, including ``--sdk`` to use a nightly you've already
+unpacked and ``--version`` to pick a different release series.
+
+To do the same by hand, download a nightly from:
 
    https://nightly.renpy.org
 
@@ -64,9 +77,14 @@ Note that the latest nightly build is at the bottom of the list. Once you've
 unpacked the nightly, change into this repository, and run::
 
     ./after_checkout.sh <path-to-nightly>
+    python3 scripts/generate_styles.py
 
-Once this script completes, you should be able to run Ren'Py using renpy.sh,
-renpy.app, or renpy.exe, as appropriate for your platform.
+Once this completes, you should be able to run Ren'Py using renpy.sh,
+renpy.app, or renpy.exe, as appropriate for your platform. Note that
+``renpy.sh`` and ``renpy.exe`` must be copies rather than symlinks, as
+``renpy.sh`` resolves symlinks to find the directory to run from, and a
+symlinked one would run the nightly's copy of Ren'Py instead of the
+checkout. ``after_checkout.sh`` takes care of this.
 
 If the current nightly build doesn't work, please wait 24 hours for a new
 build to occur. If that build still doesn't work, contact Tom (`pytom at bishoujo.us`,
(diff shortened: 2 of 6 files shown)


wgu9:

PR #7321, opened 2026-08-25
Let the launcher add and remove projects outside the projects directory

### LINKED ISSUE #3668: [Feature Req] Allow specify multiple project directories
Basically what VSC (and other programs) lets you do.
Having one folder for all projects isn't suitable if you have a structure like
```
GitHub/Project_1/src
GitHub/Project_2/src
```
I think it's fine to leave the current setting as a default project dir, but when creating a new project, prompt to select a custom dir. Also would need to show all projects in the launcher even if they are not in the default dir.

wgu9:
Fixes #3668.

`projects.txt` has done the job since 2022, but only for people who know it exists and are comfortable editing it, which is exactly the group that didn't need help. This puts it in the launcher.

**Preferences, General tab.** Under the projects directory there's now an "Additional Projects" list showing each path in `projects.txt`, a Remove button beside each, and an "Add a project from another directory" button that opens the usual directory chooser. Adding checks the chosen directory actually contains a project (same `find_basedir` check the scanner uses) and tells you if it doesn't, or if it's already listed. The list rescans immediately, so the project shows up on the front page without a restart.

**Command line.** `renpy.sh launcher add_project <path>` and `renpy.sh launcher remove_project <path>`, alongside the existing `set_projects_directory` / `set_project`. Both print what they did and exit non-zero on failure (not a project, not listed, no projects directory yet). This was the "teammate who isn't technical" case from the thread: one command in a setup script instead of instructions to edit a file.

**Keeping hand edits intact.** The thread had a long back-and-forth about whether the launcher should own the file. It doesn't have to. Adding appends one line (creating the file with its explanatory comment if needed), and removing rewrites the file keeping every other line, blank lines and comments included, in their original order. Manual edits survive either way. Duplicate detection normalises paths (`abspath`, forward slashes) so `C:\foo\` and `C:/foo` count as the same entry. The file is now read and written as UTF-8 explicitly, since paths aren't always ASCII.

**Implementation.** `ProjectManager` gains `projects_txt()`, `ensure_projects_txt()` (which the existing `EnsureProjectsTxt` action now calls instead of carrying its own copy of the header text), `extra_projects()`, `add_extra_project()`, and `remove_extra_project()`. A `RemoveExtraProject` action sits beside `Select`. The chooser flow is a label in `preferences.rpy` like `projects_directory_preference`. Docs: the projects.txt section of `launcher.rst`, and two new command sections in `cli.rst`.

**Tested** with a copy of the launcher pointed at a scratch projects directory: `add_project` (added, already listed, not a project), `remove_project` (removed, not listed), exit codes, and that the front page's project list picks the project up and drops it again. Then a testcase through the actual preferences screen: open General, assert the Remove button is there, click it, assert it's gone, and check `projects.txt` on disk only lost that one line. The directory chooser itself can't be driven from a testcase, so the add button was only tried by hand.

New launcher strings are untranslated for now, as usual.

Opened as a draft so it can be reviewed properly first.

### DIFF
--- a/launcher/game/preferences.rpy
+++ b/launcher/game/preferences.rpy
@@ -59,17 +59,7 @@ init python:
         """
 
         def __call__(self):
-            fn = os.path.join(project.manager.projects_directory, "projects.txt")
-
-            if os.path.exists(fn):
-                return
-
-            with open(fn, "w") as f:
-                f.write("""\
-# This file can be used to add projects not in the projects directory
-# by listing the full path to each project, one per line.
-
-""")
+            project.manager.ensure_projects_txt()
 
 
 default persistent.force_new_tutorial = False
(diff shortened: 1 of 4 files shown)


wgu9:

PR #7322, opened 2026-08-25
Explain label parameter errors caused by fall-through

### LINKED ISSUE #5166: Feature request: Debug: improve error message when entering label with parameters due to missing return in previous label

wgu9:
Fixes #5166.

When execution falls through into a label with required parameters, the existing error points at the second label even though the usual mistake is a missing `return` or `jump` in the label above it. This adds that context to the `TypeError` and names the preceding label when one can be identified.

The extra message is deliberately limited to the fall-through case:

- no call arguments are present
- the previous transfer was not a call or jump
- the target label has at least one required parameter

Errors raised while evaluating optional default values are left unchanged, as are missing-argument errors after an explicit jump.

Validation with the 8.5.4 fix nightly:

- fall-through into `label needs_argument(value)` includes the preceding label and missing `return`/`jump` guidance
- an explicit jump with no argument keeps the original missing-argument error
- a `TypeError` inside an optional default expression keeps the original exception

This replaces #7315, which was based on `master`. Bug fixes target `fix` in this repository.

Draft while I review the final diff against the fix branch.

### DIFF
--- a/renpy/ast.py
+++ b/renpy/ast.py
@@ -1141,7 +1141,13 @@ class Label(Node):
 
         renpy.game.context().mark_seen()
 
-        values = apply_arguments(self.parameters, renpy.store._args, renpy.store._kwargs)
+        try:
+            values = apply_arguments(self.parameters, renpy.store._args, renpy.store._kwargs)
+        except TypeError as e:
+            if self.missing_required_parameters_after_fall_through():
+                raise TypeError(self.fall_through_message(e)) from None
+
+            raise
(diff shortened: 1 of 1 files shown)


wgu9:
I am closing this draft so it does not add to the review queue. Thanks.

Issue #1397, opened 2018-03-04
Wrap `python hide` blocks in functions?


eevee:
`python hide` blocks are executed as though they were top-level code, which has some interesting effects on scoping.  For example:

```
python hide:
    import re
    print ' '.join(re.escape(s) for s in ['abc', 'def', 'ghi'])
```

This produces a `NameError`.  Disassembly shows the `import` is writing into locals, whereas the generator expression (which has its own scope) is trying to read from globals.  A similar problem occurs when trying to create a closure; when it's called later, nothing it closed over exists any more.

I think this is because top-level declarations _don't actually close over_ outer variables — the outer scope is module scope anyway, which is expected to always exist.  But assignments write into locals no matter where they happen; it just so happens that in any top-level scope in normal Python, locals and globals are the same namespace.

So how about implicitly wrapping every `python hide` block in a function that's immediately executed?  i.e., treat the above as:

```
python hide:
    def _scope():
        import re
        print ' '.join(re.escape(s) for s in ['abc', 'def', 'ghi'])
    _scope()
```

The local scope is initially empty and discarded at the end of the block anyway, so offhand I don't _think_ this should make any visible difference, except that it'd make closures work.

----

Incidentally, since the global scope is the store, the `global` statement works perfectly well for reading and writing store values within a `hide` block.  Might be worth documenting.

renpytom:
Did you test your example? 

I just ran:

```
label start:
    python hide:
        import re
        print ' '.join(re.escape(s) for s in ['abc', 'def', 'ghi'])
```

And it did not cause a NameError. Is there something else special about doing this? Do you have a traceback when it comes to the name error?

Python should handle this properly, and from what I'm seeing, it does.

eevee:
Well that's interesting.  I _do_ get a NameError from exactly the same script, on master:

```
I'm sorry, but an uncaught exception occurred.

While running game code:
  File "game/script.rpy", line 2, in script
    python hide:
  File "game/script.rpy", line 4, in <module>
    print ' '.join(re.escape(s) for s in ['abc', 'def', 'ghi'])
  File "game/script.rpy", line 4, in <genexpr>
    print ' '.join(re.escape(s) for s in ['abc', 'def', 'ghi'])
NameError: global name 're' is not defined

-- Full Traceback ------------------------------------------------------------

Full traceback:
  File "game/script.rpy", line 2, in script
    python hide:
[...]

PR #7328, opened 2026-08-28
Fix semantic of `python hide`

Andykl:
## Description
This PR removes `wrap_hide` that was initially added to address #1397, which is now handled by more correct `wrap_generator`, but removing it adds a benefit that `python` and `python hide` has matching semantic, i.e.
1. `return` in python block does not get recognized as early exit from it and now errors regardles of hide with `'return' outside function`
2. Same for `'yield' outside function`
3. There is no longer spurious frame in traceback

## Human Oversight

- [ ] Other:

### DIFF
--- a/renpy/python.py
+++ b/renpy/python.py
@@ -753,26 +753,6 @@ class WrapNode(ast.NodeTransformer):
 
 wrap_node = WrapNode()
 
-
-def wrap_hide(tree):
-    """
-    Wraps code inside a python hide or python early hide block inside a
-    function, so it gets its own scope that works the way Python expects
-    it to.
-    """
-
-    hide = ast.parse("""\
-def _execute_python_hide(): pass;
-_execute_python_hide()
-""")
-
-    for i in ast.walk(hide):
-        ast.copy_location(i, hide.body[0])
-
-    hide.body[0].body = tree.body  # type: ignore
-    tree.body = hide.body
-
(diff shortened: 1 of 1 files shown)


mal:
Being able to early exit from `python hide` blocks was particularly useful and is already used "in the wild". Is there any kind of replacement for this functionality, and if not could we consider supporting early returns in all `python` blocks as a feature, ideally prior to the next milestone release.

Andykl:
I don't like "return" being available only in hide, but not regular python, plus it looks strange. I guess we could add a function that ends execution of current python statement by raising control exception. 
Most straightforward would be `renpy.call("_renpy_return")`

renpytom:
What is a return from a python block supposed to do. That's not really something I was expecting.

mal:
Nothing crazy, just to abort the block without processing the rest of it, exactly like an early return in a function.

I found it to be clear concise way to have boot time conditionals that short circuit around code only intended for certain users or tasks (developers, testers, lint tasks, etc), while still being able to keep them organisationally close to other related Ren'Py elements (screens, styles, transforms, etc) in an `rpy` and without stacking indents. Being able to early return out of a block not needed in the current context felt natural and the intent was immediately clear to anyone reading the code without scrolling sometimes quite far down to see if there was going to be an `else` or any follow up logic.

PR #7335, opened 2026-08-29
fix: `AudioPositionValue` stuttering

qTich:
## Description

`renpy.music.seek` is called for every `.changed`,
which leads to stuttering due to frequent locks.

## Human Oversight

- [ ] Other:

### DIFF
--- a/renpy/common/00barvalues.rpy
+++ b/renpy/common/00barvalues.rpy
@@ -606,6 +606,8 @@ init -1500 python:
             If true, the player can seek by dragging the bar.
         """
 
+        _position = -1.0
+
         def __init__(self, channel='music', update_interval=0.1, adjustable=False):
             self.channel = channel
             self.update_interval = update_interval
@@ -613,8 +615,12 @@ init -1500 python:
 
             self.adjustment = None
 
+            self._position = -1.0
+
         def changed(self, position):
-            renpy.music.seek(position, self.channel)
+            if abs(position - self._position) > 5e-2:
+                renpy.music.seek(position, self.channel)
+                self._position = position
 
         def get_pos_duration(self):
             pos = renpy.music.get_pos(self.channel) or 0.0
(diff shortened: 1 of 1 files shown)


Andykl:
Could you set `_position = None` on the class. Not sure if `AudioPositionValue` could be pickled in saves but better be safe.
Also, are you sure its enough to check for equality? I don't know what is the resolution of the bar is, so probably it is better to set `_position = -1.0` and check for small delta.

qTich:
I don't quite understand about the delta.

The boundaries are `0.0` and `duration`.
There is no conversion except for checking for `range` inside `.change`,
therefore the value will not differ in any way.

https://github.com/renpy/renpy/blob/3f9187f6e14741711029ac5dfe7e2e57f3045abb/renpy/display/behavior.py#L2134-L2165

Andykl:
I mean that if bar value changed slightly, it could still stutter. I don't think it make much sense to seek when position changed from 1.0 to 1.0005.

renpytom:
Does it make sense to have .changed queue it up, and then have the actual change happen in periodic, which happens every 0.05 seconds?

qTich:
It seems to me that the solution using `.periodic` would be incorrect,
because the user can set a high value for `update_interval`.

renpytom:
I don't understand why a high value for update interval would make queueing incorrect - the two seem unrelated to me. update_interval is how often the bar updates itself, while periodic is how often the bar can accept updates.

qTich:
Okay.

The main problem is that the `.seek` occurs for every `.changed`
because it is impossible to determine where the value came from.

That's why I set the value before calling `.change`,
because it doesn't change anywhere inside the call chain.

This solution avoids calling `.seek` if the value is in the range of the received value from `.get_pos_duration'.

So initially, I thought it was about `AudioPositionValue.periodic`,
because the change was specifically about `AudioPositionValue'.

If you are talking about `renpy.audio.audio.periodic` + `renpy.audio.audio.Channel.periodic`.

Then I don't understand how `renpy.audio.audio.periodic` is supposed to solve this problem,
because as far as I understand, you suggest always calling `.seek` if there is a value in the queue.
If not, then lock the mutex again to get a position?

What if `update_interval` is equal to `0.0`, meaning the value will always be in the queue?
In other words, the update will occur even more frequently.

PR #7339, opened 2026-08-31
fix: `get_sdl_dll` on `ios`

qTich:
## Description

Fixing getting `get_sdl_dll` on `ios`, as everything is statically linked in `sys.executable`.

## Human Oversight

- [ ] Other:

### DIFF
--- a/renpy/exports/platformexports.py
+++ b/renpy/exports/platformexports.py
@@ -179,7 +179,7 @@ def get_sdl_dll():
     :doc: sdl
 
     Returns a ctypes.cdll object that refers to the library that contains
-    the instance of SDL2 that Ren'Py is using. If this fails, None is returned.
+    the instance of SDL that Ren'Py is using. If this fails, None is returned.
     """
 
     global sdl_dll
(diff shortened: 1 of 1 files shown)


renpytom:
This seems wrong, as this case is covered by the first clause, with gets it out of the main executable. 

Can you explain what you did to verify the correctness of this?

renpytom:
I've updated this in 276f5ea29b0b7d2f9d8f4bc828c8ff46d6c818a7 to match SDL3.

qTich:
sh

Okay.

Andykl:
You need to use `renpy.get_sdl_dll(3)` for it to work. This is done so users that used this function to get SDL2 not segfault, but get a Python exception.

qTich:
ph

renpytom:
Hm... did you test your original version on iOS? I'm a bit confused as to how it would work, as it doesn't have access to the appropriate symbols, either. 

(My thinking is we probably need to adjust the iOS linker settings so dlopen(NULL) works.)

renpytom:
Thinking a bit more about this, I think the right answer is to remove the function. It's asking for breakage (everything that used it just broke). And now's the time to do it - as nobody has written code that uses SDL3.

renpytom:
Decided to keep it after all, with your change, somewhat reordered. 045e11abea54bb2ce6bed34b2072b92aebb1d67e should have everything.

PR #7340, opened 2026-08-31
fix: changing of `ShownImageInfo` inside `prediction`

qTich:
## Description

`RuntimeError: Set changed size during iteration`

## Human Oversight

- [ ] Other:

### DIFF
--- a/renpy/display/image.py
+++ b/renpy/display/image.py
@@ -865,13 +865,14 @@ class ShownImageInfo(renpy.object.Object):
         Returns true if name is the prefix of an image that is showing
         on layer, or false otherwise.
         """
+        _shown = tuple(self.shown)
 
         tag = name[0]
         rest = name[1:]
 
         layer = renpy.exports.default_layer(layer, tag)
 
-        if (layer, tag) not in self.shown:
+        if (layer, tag) not in _shown:
             return None
 
         shown = self.attributes[layer, tag]
@@ -892,21 +893,24 @@ class ShownImageInfo(renpy.object.Object):
         """
         Returns the set of tags being shown on `layer`.
         """
+        _shown = tuple(self.shown)
 
-        return {t for l, t in self.shown if l == layer}
+        return {t for l, t in _shown if l == layer}
 
     def get_hidden_tags(self, layer):
         """
         Returns the set of tags on layer that have attributes,
         but aren't being shown.
         """
+        _shown = tuple(self.shown)
 
-        return {t for l, t in self.attributes if l == layer if (l, t) not in self.shown}
+        return {t for l, t in self.attributes if l == layer if (l, t) not in _shown}
 
     def predict_scene(self, layer):
         """
         Predicts the scene statement being called on layer.
         """
+        _shown = tuple(self.shown)
 
         if layer is None:
             layer = "master"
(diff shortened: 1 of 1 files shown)


renpy/display/image.py
```diff
@@ -865,13 +865,14 @@ def showing(self, layer, name, exact=False):
...
         """
+        _shown = tuple(self.shown)
 
         tag = name[0]
         rest = name[1:]
 
         layer = renpy.exports.default_layer(layer, tag)
 
-        if (layer, tag) not in self.shown:
+        if (layer, tag) not in _shown:
```
mal, line 875: The changes to `showing` don't appear to be warranted as it as an atomic membership test, but now it's a tuple creation and search.

renpy/display/image.py
```diff
@@ -892,21 +893,24 @@ def get_showing_tags(self, layer):
...
 
     def get_hidden_tags(self, layer):
         """
         Returns the set of tags on layer that have attributes,
         but aren't being shown.
         """
+        _shown = tuple(self.shown)
 
-        return {t for l, t in self.attributes if l == layer if (l, t) not in self.shown}
+        return {t for l, t in self.attributes if l == layer if (l, t) not in _shown}
```
mal, line 907: Similarly, the changes to `get_hidden_tags` do the same thing changing an atomic membership test, into a tuple creation and search.

Andykl:
Could you clarify your changes in a bit more details? Especially changes for `showing`, `predict_show` and `predict_hide`. 
I assume it is to prevent stale data when doing concurrent checks?

mal (commented):
Reading this I ended up a little confused over some of the choices made. While I can understand the tuple creation to avoid the loop changing size (I assume via the prediction thread?), other changes (set membership ops, and add/discard ops) don't seem to fit that goal or become redundant after accounting for the tuple creation already done.

Please could you clarify?

renpy/display/image.py
```diff
@@ -934,7 +938,9 @@ def predict_show(self, layer, name, show=True):
         self.attributes[layer, tag] = rest
 
         if show:
-            self.shown.add((layer, tag))
+            _shown = self.shown.copy()
+            _shown.add((layer, tag))
+            self.shown = _shown
```
mal, line 943: Given the iteration is being protected by tuple creation, I'm not sure why it's necessary to copy the set here.

renpy/display/image.py
```diff
@@ -944,7 +950,9 @@ def predict_hide(self, layer, name):
         if (layer, tag) in self.attributes:
             del self.attributes[layer, tag]
 
-        self.shown.discard((layer, tag))
+        _shown = self.shown.copy()
+        _shown.discard((layer, tag))
+        self.shown = _shown
```
mal, line 955: And same thing here.

Andykl:
I reverted the merge. 
Given the traceback
```
File "renpy/display/core.py", in interact_core
File "renpy/display/scenelists.py", in clear
File "renpy/display/image.py", in predict_scene
File "renpy/display/image.py", in <genexpr>
RuntimeError: Set changed size during iteration
```
I see that it is only possible if you have a concurrent thread that changes content of `config.overlay_layers`.
If that's the case we can make a more focused fix for that that is less controversial.
Feel free to open it if you find the offender. I assume only changes to methods that iterate over `self.shown` need to change.

PR #7341, opened 2026-08-31
fix: `fetch_emscripten` incorrect timeout of status

qTich:
## Description

`status = ...` and `message = ...` resets on next cycle

## Human Oversight

- [ ] Other:


mal:
Wouldn't a `break` after meeting the timeout condition have been much more readable here?

Andykl:
You are right, I reverted and did it with a break as you suggested.

PR #7342, opened 2026-08-31
fix: missing `renpy.update` modules during `renpy.reload_all`

qTich:
## Description

Since modules are not imported during `import_all`, after `reload_all`
there may be an exception about the absence of, for example, `Update`

## Human Oversight

- [ ] Other:

### DIFF
--- a/renpy/__init__.py
+++ b/renpy/__init__.py
@@ -561,7 +561,9 @@ def import_all():
     import renpy.test.testcli
 
     import renpy.update
+    import renpy.update.common
     import renpy.update.deferred
+    import renpy.update.update
 
     try:
         import renpy.tfd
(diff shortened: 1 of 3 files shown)


Andykl:
Could you add a comment explaining why requests imported conditionally. I assume it's because of emscripten?

renpytom:
I don;t understand what problem this is trying to solve. Can you explain the motivation behind this PR?

qTich:
I'm using the `Update` subclass and redefining the `download_block_file`.

I remember for sure that the trace occurred after `utter_restart`,
saying that the class (`renpy.update.update.Update`) could not be found.

renpytom:
That's not a public API.

The current version is fairly carefully done - it import renpy.update.update only when the updater is in use, so that requests doesn't get imported on a normal run of Ren'Py without updating. 

renpy.update.update.Update is not a public API. I'd suggest you just run "import renpy.update.update" before doing your patching, if you need to. It doesn't require an engine change for this obscure case.

PR #7344, opened 2026-09-01
Add support for the GL 3.3 core profile

ronchonbon:
## Description

This adds a GL 3.3 core context between 3.3 and 2.0 compatibilities, such that 3.3 compatibility is tried first, and such that Mac doesn't get forced onto GLSL 1.20 with all GLSL 3.30 features translated away.

Some things that got caught up in these changes for compatibility with the core profile:

1. Had to add scratch buffers to `gl2statecache` as core profile substitutes for the client-side vertex and index arrays Ren'Py is used to.
2. Added an explicit default VAO.
3. `glGetString` and `GL_GENERATE_MIPMAP_HINT` are removed in the core profile, so added some workarounds.
4. Some tweaks due to changes in luminance formats.

I also did a thing I was told not to do and hand-edited `gl2functions`. Sorry!

## Human Oversight


mal:
> I also did a thing I was told not to do and hand-edited `gl2functions`. Sorry!

Will there need to be changes made to `scripts/uguu/generate_required_functions.py` to ensure these changes are not accidentally wiped out by someone updating `gl2functions` the "right" way in future?

ronchonbon:
Yes! Updating `generate_required_functions` just required a little more thinking, but maybe it is reasonable to fix it up in this PR. I'll dig in a bit.

ronchonbon:
Never mind it didn't need much at all!

PR #7349, opened 2026-09-04
Fix multiplied matrixcolor snapping to the target instead of interpolating

### LINKED ISSUE #7303: regression in ATL interpolation
as per https://github.com/renpy/renpy/commit/936fb03d241b4a3d7399185b489d4a6a959be247#commitcomment-197350044

zaveshaa:
## Description

Fixes #7303.

Since #7230, `interpolate()` treats any callable with a `.value` attribute as a value-based matrix. `_MultiplyMatrix` sets `self.value = 1.0`, so a product like `BrightnessMatrix(0.0) * TintMatrix("#fff")` matches that check even though it interpolates through its left/right factors rather than a scalar value. It then takes the new path, where `__call__` receives a plain `Matrix` and returns the end value on every frame — the animation snaps to the target instead of interpolating.

The fix removes the unused `value` attribute from `_MultiplyMatrix` so products take the original path again. Nothing reads `_MultiplyMatrix.value`.

Added a regression test next to the existing matrixcolor tests.

## Human Oversight

### DIFF
--- a/renpy/common/00matrixcolor.rpy
+++ b/renpy/common/00matrixcolor.rpy
@@ -133,7 +133,6 @@ init -1500 python:
         def __init__(self, left, right):
             self.left = left
             self.right = right
-            self.value = 1.0
 
         def __call__(self, other, done):
             if type(other) is not type(self):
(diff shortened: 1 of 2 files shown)


renpytom:
Nice fix.

PR #7354, opened 2026-09-07
Remove MOUSEMOTION from redraw-blocking events

ronchonbon:
## Description

Discards mouse motion from the set of pygame events that blocks redraw. This gets rid of the classic dramatic framerate drop that accompanies player mouse motion. I am unable to see a downside to this other than some imperceptible input -> display latency.

## Human Oversight


renpytom:
Hm... by default, Ren'Py drops down to a 5 fps framerate. I'd want to update sooner than that, because 200ms is quite noticable.

ronchonbon:
This shouldn't make mouse updates wait 200 ms even from idle, everything still wakes when input arrives and draws immediately with `needs_redraw`. Unless I'm missing something you see?

PR #7365, opened 2026-09-11
Only copy/render uniforms a transform actually sets

ronchonbon:
## Description

Avoids scanning all registered uniforms for every transform, such that states don't get populated with unused defaults.

## Human Oversight

### DIFF
--- a/renpy/display/transform.py
+++ b/renpy/display/transform.py
@@ -149,9 +149,18 @@ class TransformState(renpy.object.Object):
 
     def take_state(self, ts):
         d = self.__dict__
+        src = ts.__dict__
 
-        for k in all_properties:
+        for k in non_uniform_properties:
             d[k] = getattr(ts, k)
+
+        # Uniforms only live in __dict__ when explicitly set.
+        for k in uniforms.intersection(d):
+            if k not in src:
+                del d[k]
+
+        for k in uniforms.intersection(src):
+            d[k] = src[k]
 
         self.last_angle = ts.last_angle
         self.radius_sign = ts.radius_sign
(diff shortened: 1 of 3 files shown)


renpy/display/transform.py
```diff
@@ -149,10 +149,19 @@ def __init__(self):
...
+        for k in non_uniform_properties:
             d[k] = getattr(ts, k)
 
+        # Uniforms only live in __dict__ when explicitly set.
+        for k in d.keys() & uniforms:
+            if k not in src:
+                del d[k]
+
+        for k in src.keys() & uniforms:
+            d[k] = src[k]
```
mal, line 163:
```suggestion
        sk = src.keys()
        dk = d.keys()
        
        for k in uniforms:
            if k in sk:
                d[k] = src[k]
            elif k in dk:
                del d[k]
```

mal:
```suggestion
        sk = src.keys()
        dk = d.keys()
        
        for k in uniforms & dk - sk:
            del d[k]

        for k in uniforms & sk:
            d[k] = src[k]
```

mal (commented):
Had two thoughts about potential optimisations/readability improvements while reading, but nothing crazy. The first avoids a double loop, and the latter avoids the conditional inside the first loop.

renpy/display/transform.py
ronchonbon: Good points! I can fold into the next PR I submit that touches these files!

ronchonbon: I did some really quick timings and I think the single loop actually underperforms as we register more uniforms. The set modification is more benign but wasn't a clear winner to me. I think can defer for now?

mal: Sure, they weren't blockers, just things I spotted whilst reviewing but by the time I was done it was already merged. :stuck_out_tongue:

